Issue #204, opened 2013-02-07
Rendering a nil object with a custom serializer fails


LeeMallabone:
Hi there,

Attempting to `render :json` with a nil value and a custom serializer raises an exception.

For example, if I try this:

``` ruby
foo = some_function_that_may_return_nil
render :json => foo, :serializer => CustomSerializer
```

then the serialization will raise an exception whenever `foo` is nil. The message is:
_NoMethodError: undefined method `read_attribute_for_serialization' for nil:NilClass_

I've added a test case to demonstrate this behavior to my fork: https://github.com/fblee/active_model_serializers/commit/d35999a7eec185f6b334bd3a4e3e3c4dd8464f15

Best,
Lee.

PR #213, opened 2013-02-26
Fix rendering nil with custom serializer

michihuber:
Fixes https://github.com/rails-api/active_model_serializers/issues/204

If the root object is nil, it doesn't try to convert it into a hash but renders it as is.
Custom roots are respected.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -267,6 +267,7 @@ module ActiveModel
     # Returns a hash representation of the serializable
     # object without the root.
     def serializable_hash
+      return nil if @object.nil?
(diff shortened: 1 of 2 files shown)


k-rudy:
Guys, current master doesn't handle the `nil` objects like described in this PR. It tries to serialize it and instead of returning `nil` it returns `{ first_attribute: nil, second_attribute: nil }`. Was that changed intentionally, or should be considered as an issue? If that was done by reason, could you please explain why? Thanks

beauby:
@k-rudy Could you open a separate issue stating the version, the adapter, the expected behavior, etc.?

PR #300, opened 2013-05-11
ActiveModel Serializers support for serialization of a model embedded in a hash

AndyObtiva:
Covered the ActiveRecord::SerializerOverride module with a test, and added documentation for when it is useful as I encountered a problem recently with a client that has a lot of legacy code triggering json serialization for a model nested within a hash directly from controller actions, and we could not simply refactor the model to rely on active model serializers without going and changing all the controller actions because the actions were rendering the json of a hash that contained the model within instead of the model directly. 

Here is the documentation describing the problem included with the code change:
## Serialization of a hash containing a nested (embedded) model

There is a mixin that overrides ActiveRecord::Base#to_json and #as_json to enable
serialization via ActiveModel Serializers. We do not recommend that you use ActiveModel
Serializers in this way in general, but there is one exception. That is when a legacy
app's controller actions for example are rendering json for a model nested within a hash
(e.g. {param1 => value1, param2 => model}), which will not trigger ActiveModel Serializers
for the nested model by default, yet instead rely on normal to_json serialization.
To get around this problem, you can require the file 'active_record/serializer_override'
and include the module ActiveRecord::SerializerOverride, which will override the to_json
and as_json methods of the model to rely on ActiveModel Serializers instead.

Example:

In controller actions of a legacy app, there is a lot of this:

``` ruby
render :json => {param1 => value1, params2 => value2, data => model}
```

Add an initializer file: "config/initializers/active_model_serializers.rb"

with this content:

``` ruby
require 'active_record/serializer_override'
ActiveRecord::Base.send(:include, ActiveRecord::SerializerOverride)
```

or to override to_json and as_json for one model class only:

``` ruby
require 'active_record/serializer_override'
ActiveRecord::Base.send(:include, ActiveRecordModel)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,56 @@ serializer when you render the object:
 render json: @post, serializer: FancyPostSerializer
 ```
 
+To render an object embedded in a hash, you can simply render the hash, and
+Rails will search for a serializer for the embedded object and use it if available.
+
+```ruby
+render json: {data: @post, platform: 'mobile', format: 'json'}
+```
+
+Given the example above, the output is as follows
+
+```json
+{
+  "data":{"id":1,"title":"post1","body":"body1"},
+  "platform": "mobile",
+  "format": "json"
+}
+```
+
+This will work even if the hash contains multiple objects with serializers.
+
+```ruby
+render json: {post: @post, related_post: @related_post}
+```
+
+This yields the following output
+
+```json
+{
+  "post":{"id":1,"title":"post1","body":"body1"},
+  "related_post":{"id":2,"title":"post2","body":"body2"}
+}
+```
+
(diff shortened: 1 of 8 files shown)


sebastianvera:
+1 to @jonchay

jasonswett:
+1

adnguyen1989:
+1

bf4:
@adnguyen1989 and others, this PR has been closed for over a year.  If you'd like to bring it up as an issue or new PR, the best place is an issue or new PR.  Please see https://github.com/rails-api/active_model_serializers/blob/master/CONTRIBUTING.md  Thanks ! :rainbow:

PR #448, opened 2013-11-07
Added XML support

jokklan:
Hallo

I know that you have said before that you are not focusing on XML support at the time being, but I''ll need it for one of my own projects, so i have build it anyway. Even though it's not done, was it really simple to implement, as a hash easily can be transformed into XML. The only obstacle i came across was that XML needs a root, and json does not, but that doesn't seem to bad. Rabl eg can also render json without root with the same template as it will render xml. 

I don't know if you are at all interested, but i can't see a reason not to? 

Best regards
/Johan

### DIFF
--- a/test/integration/action_controller/xml_serialization_test.rb
+++ b/test/integration/action_controller/xml_serialization_test.rb
@@ -251,7 +251,7 @@ module ActionController
 
         get :render_xml_array_embeding_in_root
         assert_equal 'application/xml', @response.content_type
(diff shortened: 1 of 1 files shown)


reidmorrison:
Now that v0.10 is available, is it worth re-opening the discussion around XML support?

Is it as simple as just adding a new adapter that derives from the JSON adapter and/or the JSON API adapter, and then just calls `to_xml` instead of `to_json` on the final Hash to render the output as XML?

The challenge this would address is being able to render the "equivalent" JSON or XML output on demand. Over time clients can move away from the XML interface to the better JSON layout.

bf4:
@reidmorrison I would help you write an add-on that adds an XML adapter, but, at present, it's not a good fit for AMS.  Additionally, active model xml serializers have been extracted from Rails 5, so it would be problematic to make that a dependency.

Please open a new issue if you'd like to discuss more, esp with an example of what your use-case is.

PR #498, opened 2014-01-27
Nested and listed predicates

christhekeele:
Looking ahead to the 0.9 update I realized I'd have to rewrite some monkey patches I'm using to keep a simple feature I've hacked into my serializers. I thought I'd take the opportunity of the alpha/refactor to see if perhaps they belong in AMS proper.

This pull request does not rebase cleanly into the refactor branch, something I can tend to if the general concept is vetted.

It allows virtual attribute and inclusion/filter methods to behave differently depending on the context a serializer is invoked in, `nested?` from `ActiveModel::Serializer::Association` and `listed?` from `ActiveModel::ArraySerializer`. For example:

``` ruby
class PostSerializer < ApplicaitonSerializer
  attribute :title, :body, :published_at
  has_one :author
  has_many :comments

  def include_author?
    not nested?
  end

  def include_comments?
    not listed?
  end
end
```

This will not display the author of the post if serialized in a nested context, for example as a serialized user's `favorite_post`. It will also not display the post's comments if serialized as a group of posts, for example as a user's `publications`. `nested?` and `listed?` should only both be true when serializing `has_many` associations.

Since my API gets more verbose the more you drill down on a resource, this keeps me from needing up to 4 different classes in combination to handle the possible permutations, and could be a useful pattern to incorporate into AMS (YMMV). It also will play very nicely with the centralized `filter` method in the upcoming release, since attributes in these contexts tend to be included/excluded in batches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-[![Build Status](https://api.travis-ci.org/rails-api/active_model_serializers.png)](https://travis-ci.org/rails-api/active_model_serializers)
-[![Code Climate](https://codeclimate.com/github/rails-api/active_model_serializers.png)](https://codeclimate.com/github/rails-api/active_model_serializers)
+[![Build Status](https://api.travis-ci.org/rails-api/active_model_serializers.png)](https://travis-ci.org/rails-api/active_model_serializers) 
+[![Code Climate](https://codeclimate.com/github/rails-api/active_model_serializers.png)](https://codeclimate.com/github/rails-api/active_model_serializers) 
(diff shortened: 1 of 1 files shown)


christhekeele:
Have you had any further thoughts on this? I can work towards a mergeable PR.

kurko:
@christhekeele can I suggest you open an issue with a proposal for the new DSL before going for a PR? We'd need to discuss how it'd work, its interface and API. Sounds good?

Issue #398, opened 2013-10-09
CamelCase JSON format


wojtha:
Is there a recommended way how to produce camelized keys by default during serialization by using AMS?

In small APIs one can just override the attributes hash, but is not a feasible way for bigger APIs.

I guess that probably the best way would be to fork AMS and camelize the keys somewhere in the overriden `serialize` method? Or do someone has better idea?

Btw some interesting links about snake_case vs. CamelCase in JSON:
- https://blog.apigee.com/detail/restful_api_design_what_about_attribute_names
- http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#snake-vs-camel

ratbeard:
Has anyone written a camelcase adapter for 0.10?

joaomdmoura:
@ratbeard no, we haven't build a specific CamelCase adapter yet.

semikolon:
Some models have trailing questionmarks in method names like model.was_good? and I wanted it to translate to wasGood in JSON. Here's my solution. Some repetition for has_many/has_one/belongs_to  - how would you simplify it?

```
class ApplicationSerializer < ActiveModel::Serializer
  def attributes(options={})
    super(options).map { |k, v| [ k.to_s.camelize(:lower).chomp('?').to_sym, v ] }.to_h
  end

  def self.has_many(*args)
    options = args.extract_options!
    options.reverse_merge!({ key: args.first.to_s.camelize(:lower).to_sym })
    args << options
    super(*args)
  end
  def self.has_one(*args)
    options = args.extract_options!
    options.reverse_merge!({ key: args.first.to_s.camelize(:lower).to_sym })
    args << options
    super(*args)
  end
  def self.belongs_to(*args)
    options = args.extract_options!
    options.reverse_merge!({ key: args.first.to_s.camelize(:lower).to_sym })
    args << options
    super(*args)
  end
end

```

Here's the current code for associations in AMS: https://github.com/rails-api/active_model_serializers/blob/aaa60bfdc131cbd96b322037f56bce99a9f36de5/lib/active_model/serializer/associations.rb

sebbean:
hey ya'll

Just wanted to point out: https://github.com/domchristie/humps as a nice way to not have to change the API format (I'm partial to snake_case for API data) to fit the front end library (React in my case).

dkniffin:
[...]

PR #534, opened 2014-03-30
Support lowerCamel key format

kylefritz:
You can specify that serializers use the lower-camel key format at the config, class or instance level.

``` ruby

ActiveModel::Serializer.setup do |config|
  config.key_format = :lower_camel
end

class BlogLowerCamelSerializer < ActiveModel::Serializer
  format_keys :lower_camel
end

BlogSerializer.new(object, key_format: :lower_camel)
```

Relevant to issue #398. Pull request #436 also offered a solution for changing the key format but that approach is now in conflict with head/master.

### DIFF
--- a/lib/active_model/array_serializer.rb
+++ b/lib/active_model/array_serializer.rb
@@ -20,16 +20,13 @@ module ActiveModel
       @meta            = options[@meta_key]
       @each_serializer = options[:each_serializer]
       @resource_name   = options[:resource_name]
-      @key_format      = options[:key_format]
+      @key_format      = options[:key_format] || options[:each_serializer].try(:key_format)
     end
     attr_accessor :object, :scope, :root, :meta_key, :meta, :key_format
 
     def json_key
-      if root.nil?
-        @resource_name
-      else
-        root
-      end
+      key = root.nil? ? @resource_name : root
+      key_format == :lower_camel ? key.camelize(:lower) : key
     end
 
     def serializer_for(item)
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -117,14 +117,16 @@ end
       @except        = Array(options[:except]) if options[:except]
       @key_format    = options[:key_format]
     end
-    attr_accessor :object, :scope, :root, :meta_key, :meta
+    attr_accessor :object, :scope, :root, :meta_key, :meta, :key_format
 
     def json_key
-      if root == true || root.nil?
+      key = if root == true || root.nil?
         self.class.root_name
       else
         root
       end
+
+      key_format == :lower_camel ? key.camelize(:lower) : key
     end
 
     def attributes
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -35,7 +35,7 @@ end
 class Comment < Model
 end
 
-class Blog < Model
+class WebLog < Model
 end
 
 ###
@@ -66,10 +66,10 @@ class CommentSerializer < ActiveModel::Serializer
   attributes :content
 end
 
-class BlogSerializer < ActiveModel::Serializer
+class WebLogSerializer < ActiveModel::Serializer
   attributes :name, :display_name
 end
 
-class BlogLowerCamelSerializer < BlogSerializer
+class WebLogLowerCamelSerializer < WebLogSerializer
   format_keys :lower_camel
 end
--- a/test/integration/action_controller/serialization_test.rb
+++ b/test/integration/action_controller/serialization_test.rb
@@ -194,6 +194,22 @@ module ActionController
       end
     end
 
+    class LowerCamelArraySerializerTest < ActionController::TestCase
+      class WebLogController < ActionController::Base
+        def render_array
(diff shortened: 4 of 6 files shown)


zavan:
Is this removed from v0.10.0.rc1?

joaomdmoura:
Hey @zavan actually it wasn't implemented on 0.10 version.

zavan:
Thanks @joaomdmoura. Will it be implemented?

joaomdmoura:
@zavan we haven't talked about it yet.
When it was implemented it made sense because of ember.js. But, as far as I know, Ember is also updating its format to use JSON API (that 0.10.0 currently supports).
Do you have a particular need related to it?

kylefritz:
I implemented this feature to support javascript frameworks in general (not specifically ember).

If you're working with serialized data from javascript, it makes sense to use the key naming format for that language.

joaomdmoura:
I know @kylefritz :smile: I just was pointing out that back than it make ember integration simpler.

We haven't received a lot of requests related to it and we are deeply focus on make this integration with Rails 5 happen. But if you want, you can open a PR bringing this feature to 0.10 we would definitely check it :smile:

kylefritz:
cool. we're locked on an older AMS version. I'll drop in a new PR soon-ish. good luck with the rails 5 integration.

zavan:
Hi guys, I have a few old projects using this format (some angular, some
without frameworks), because it feels better than working with underscores
in Javascript, I wanted to update their AMS version, that's why I asked. I
think this is an important feature for anyone using Javascript without a
standardized format, but no hurry, everything is fine :)

Em seg, 1 de jun de 2015 12:36, Kyle Fritz <notifications@github.com>
escreveu:

zavan:
Kyle, it would be very useful if the format got inherited by subclasses,
for example, I have a project with 2 APIs, a legacy one with underscores,
and a new one with lowerCamel,  so in one of them, I need to specify the
format in every serializer.
I'll wait for your PR, and if necessary, work on it to provide this.

Em seg, 1 de jun de 2015 13:05, Felipe Zavan <emailbloggers@gmail.com>
escreveu:

grin:
would be great to have this working in 0.10 :+1:

kylefritz:
@zavan i totally agree about the inheritance. i'm not sure why that never worked.

i also had an issue getting the "top-level" key to be formatted correctly. for example

``` ruby
#some controller
render json: important_things, serializer: ImportantThingsSerializer
```

returns

```javascript
{important_things: [{correctlyFormattedKey: true}, {correctlyFormattedKey: true}..]}
```

I'll see if I can re-implement this method against the 0.10 branch. Suggestions welcome.

joaomdmoura:
@kylefritz just a heads up, the PR #958, that will probably be merged soon, will change how the ```root``` is implemented right now, so you should wait for it to be merged to work on top of it in case you really want to move forward with this PR.

kylefritz:
oh great info @joaomdmoura. thanks for the heads up.

hugopeixoto:
I'm using the JSON API adapter with ember-data and I'm having a small compatibility issue.
Attributes generated by AMS are using underscores (as they are left untouched), while ember-data/jsonapi uses the dasherize function for the attribute lookup.

The jsonapi spec does not mention anything about word separators, but the examples do use dashes.

Should this transformation be handled specifically by the jsonapi adapter? Or should AMS provide a general way of applying key transformations?

My current workaround was to define a base serializer:
````ruby
class BaseSerializer < ActiveModel::Serializer
  def attributes *args
    Hash[super.map do |key, value|
      [key.to_s.dasherize.to_sym, value]
    end]
  end
end
````

joaomdmoura:
hey @hugopeixoto,
just checked with json api folks.

> dashes are just a recommendation, not part of the spec
[reference](http://jsonapi.org/recommendations/#naming)

I'll give it some thought. 
Could you please open another issue based on this?

jfelchner:
@joaomdmoura I'm just throwing my 2 cents in here, but IMO the entire point of JSON API is to avoid bike shedding.

If the JSON API [recommends dashes](http://jsonapi.org/recommendations/#naming) and the adapter is a JSON API adapter, then just do dashes.  I may not prefer it, you may not prefer it, "Rails" may not prefer it, but in the grand scheme of things whether we use dashes vs underscores vs camelCase vs whatever the new hotness is, doesn't really matter.

What _does_ matter is that everyone joins together on that consensus so that we can move on to more important things.

joaomdmoura:
@jfelchner totally agree. As I said in my last comment, indeed this is the recommendation. We should follow it. We can keep it up on your PR.

datibbaw:
First of all, thanks for this! :)

I've been looking around to find out what the "default value" for key_format actually is; our serializers are configured globally to use :lower_camel as the key format, but there are some places where we have to deal with backward compatibility and as such we should use the default snake case.

For now I've settled for something recognisable, i.e. `format_keys :underscored`, but I was just curious to hear from a more authoritative source ;-)

om-nishu-trantor:
It is not working for version `0.10.6`

bf4:
@om-nishu-trantor Please open a new issue if you're looking for help or to help others. Commenting on an issue closed 3 years ago in general is considered 'necromancy', and should only be done with care.  In this case, I believe the PR refers to a version which is a long distant cousin of yours.

PR #535, opened 2014-03-31
Serialize all attributes when attributes invoked with no args

mmosche2:
Implements functionality for serializing all attributes when the attributes class method is invoked with no arguments.


Darkside73:
Do not work in 0.10

PR #609, opened 2014-08-26
Feature/support polymorphic associations v2

polmiro:
@steveklabnik finished rebasing onto the 0.9-stable branch.

I had to make some changes to the implementation. The logic for the actual serialization is now in the Serializer / ArraySerializer instead of the Association objects. I think this is also where this logic should actually belong.

https://github.com/rails-api/active_model_serializers/pull/423

### DIFF
--- a/lib/active_model/array_serializer.rb
+++ b/lib/active_model/array_serializer.rb
@@ -15,6 +15,7 @@ module ActiveModel
       @object          = object
       @scope           = options[:scope]
       @root            = options.fetch(:root, self.class._root)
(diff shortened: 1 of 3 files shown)


mankind:
@polmiro will you be adding this polymorphic feature to the 0.8 branch too, since it will become master in the near future.

kurko:
@mankind `0.8` is not the same as `master`. They're similar in interface, but different internally. I'd advise you to add this to the `master` branch, which will become `0.10` and the official release soon.

PR #623, opened 2014-09-01
Allow ArraySerializer to inherit :each_serializer root

gauthier-delacroix:
**This PR replaces #619 as there was too many conflicts after multiple rebases.**

This change make ArraySerializer inherit `each_serializer`'s root when no specific root is set.

I'ts just a way to avoid definig root twice and keep things DRY.

JSON API now requires the same (plural) root for single resource and collections.

### DIFF
--- a/lib/active_model/array_serializer.rb
+++ b/lib/active_model/array_serializer.rb
@@ -14,11 +14,12 @@ module ActiveModel
     def initialize(object, options={})
       @object          = object
       @scope           = options[:scope]
-      @root            = options.fetch(:root, self.class._root)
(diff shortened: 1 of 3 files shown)


bf4:
I think this is an important PR for the 0.10 release... as it is now, the :root option is never passed into the serializer because it's partitioned as an adapter opt

It seems the docs about setting `ActiveModel::Serializer.root = true` and how that's required for the meta key to be added is missing.  Especially since the array serializer calls `serializable_hash` without any options

remear:
What else needs to be done here?

PR #655, opened 2014-09-19
Fix Rails Master

wouterw:
Looks like rails master is now calling `_render_with_renderer_json` instead of `_render_option_json` as of
https://github.com/rails/rails/commit/4dfe140ef3fbed34e6ae8af3b32402630a4690a1


sandstrom:
@wouterw Thanks! :boat:

Issue #657, opened 2014-09-20
undefined method attribute for serializer


kbullaughey:
I have a problem with AMS (0.9.0 stable) overshadowing my attributes. 

I have an attribute called `root` which gets overshadowed by `ActiveModel::Serializer#root`. According to the readme, I thought I could do the following:

```
attribute :original_root, key: :root

def original_root 
  object.root 
end
```

But it seems there is no method `attribute`. And with the plural version of this method, it's not obvious how one would specify a different name for just one attribute. 

Are the docs out of date? Or do I need to wait for 0.10.0 for this to work?

Any suggested workarounds? I'd rather not rename my property in the database. This was working fine before I upgraded from 0.8.0 to 0.9.0 (at least I think that's what broke it).

jeremyhaile:
@joaomdmoura I like the new behavior, but as others have mentioned - the README still references the 'attribute' method.

joshkinabrew:
I have a serializer that is using the old `attribute` method, but can someone let me know what the new syntax would be for this line? I too was bitten by the docs :frowning: 
```ruby
attribute :active_ingredients, key: "ActiveIngredient"
```

partmendoza:
attributes :ActiveIngredient
def ActiveIngredient
  object. active_ingredients
end

jfragoulis:
I cannot find any solid examples on how to replace the `attribute` method, for the life of me. Can anyone provide some or update the `README.md` ?

```ruby
class MySerializer < ActiveModel::Serializer
  # from
  attribute :id_path, key: :path
  
  # to
  def path
    object.id_path
  end
```

Is this the way ?

joaomdmoura:
Hey ppl, sorry for take soo long to come back to this. We've being really busy with the new version 0.10.0. Btw I strongly recommend you to give it a try :smile: 
@jfragoulis yes, you are right.

FYI the `key` feature is back already on 0.10.0

PR #658, opened 2014-09-20
gracefully handling name collisions with ActiveModel::Serializer methods

kbullaughey:
I posed issue #657 about an issue I was having where my model property, `root`, collided with `ActiveModel::Serializer#root`. The problem lay in how Serializer#attributes only defines an accessor method if one doesn't already exist. 

This is tricky because a major feature of AMS is that one can define methods in descendent classes and when registering an attribute with this name, it should call that method. But how does one know which methods are on the serializer as part of the model's serialization design scheme, and which ones are part of ActiveModel::Serializer itself.

In my proposed fix, I register accessor methods prefixed with `_accessor_for_` and then call these when it seems appropriate, namely if the serializer doesn't have a method by the given name or when it does, but that method was defined on ActiveModel::Serializer itself (using `method(name).owner == ActiveModel::Serializer`)

This works, but feels very hackish. It's also fragile. I'm assuming I only care about collisions that are defined on `ActiveModel::Serializer` and not modules included therein or in ancestors, etc. 

I'm sure there's a more robust way to do this. 

In particular, I liked the idea of specifying a name separate from the method name, like I mentioned in #657:

```
attribute :original_root, key: :root
def original_root 
  object.root 
end
```

But I couldn't find the implementation of `attribute` and couldn't see any easy way to implement this. Plus it seemed like conflicts should be handled gracefully from the start. 

I'm sure there is a good, elegant solution out there, I just couldn't think of it  without substantial refactoring (which appears to be going on anyway for 0.10.0).

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -123,7 +123,7 @@ end
 
         attrs.each do |attr|
           define_method registry_name(attr) do
-            object.send attr
+            object && object.send(attr)
           end unless method_defined? registry_name(attr)
 
           @_associations[attr] = klass.new(attr, options)


kbullaughey:
In case anyone is interested in a work-around, I was able to solve my case as follows:

    def attributes
      hash = super
      hash[:root] = object.root
      hash
    end

holyketzer:
@kbullaughey nice workaround, thank you!

I use it for add ```object``` field in JSON, because there is no another way.
```ruby
def attributes
  super.merge(object: 'model_name')
end
```
It is impossible to define `def object` method because it overrides `object` in serialiser internals.

Issue #605, opened 2014-08-25
0.9 no longer automatically removes ? from boolean methods


dangalipo:
So this is more of an issue when using AMS in a class not backed by active model.

In 0.8 you could specify a boolean method and AMS would strip the ? in the JSON output how ever it no longer does this.

Example:

``` ruby
class Foo
  include ActiveModel::Serialization

  def has_a_bar?
    true
  end

  def active_model_serializer
    FooSerializer
  end
end

class FooSerializer < ActiveModel::Serializer
  attributes :has_a_bar?
end

#0.8
FooSerializer.new(Foo.new).to_json # "{\"foo\":{\"has_a_bar\":true}}"
#0.9
FooSerializer.new(Foo.new).to_json #"{\"foo\":{\"has_a_bar?\":true}}"
```

pedromtavares:
Could we get this behavior ported to master please?

NullVoxPopuli:
for a work around for now, you could do:

```ruby
  attributes :has_a_bar
  def has_a_bar
    object.has_a_bar?
  end
```

pedromtavares:
Yeah, but sort of a pain to rewrite 20 serializers with around 10 of these methods each after upgrading.

NullVoxPopuli:
I know it's a pain, but I think including the `?` makes more sense, as far as getting what you specify.

@rails-api/ams ?

Sija:
@NullVoxPopuli: I disagree, no way to use 'em sensibly from js-side of things (`object.is_visible` vs `object['is_visible?']`)

pedromtavares:
Yeah it doesn't make sense since '?' isn't used in Javascript. I'm trying to figure out a monkey patch in master for this (my entire Ember app is broken ATM), first time reading AMS code :)

NullVoxPopuli:
what if someone, for some reason wanted a `?`, by omitting it by default, how would one specify to keep it?

Sija:
`?` is IMO as much part of ruby's syntactic sugar as it's attribute name so it's more about compatibility with every other client-side language than with ruby <-> ruby API

NullVoxPopuli:
@Sija, that's a good reason / argument -- I could get on board with that. 
I'm curious what other @rails-api/ams members think.

bf4:
Just reading in email, but would be best to strip '?' From json key.  If previous behavior, def need to fix.  Failing test?

B mobile phone

NullVoxPopuli:
[...]

PR #662, opened 2014-09-22
Add auto-stripping of question mark for attributes

jastkand:
This is a pull-request for issue #605. There is a little mess in this part

``` ruby
striped_attrs = attrs.map do |attr|
  symbolized = attr.is_a?(Symbol)

  attr = attr.to_s.gsub(/\?\Z/, '')
  attr = attr.to_sym if symbolized
  attr
end
```

I had to add condition of checking attribute type to make tests from this commit https://github.com/rails-api/active_model_serializers/commit/06e4c2c9d601a067124e2653ff66e0c8c5320867 pass.
As we can see if we pass string params to `attributes` method than output hash will be generated with string param as a key.

Example:

``` ruby
  attributes 'name', :gender
end
```

``` ruby
UserSerializer.new(user).as_json # => { 'name' => 'John Doe', :gender => 'male'  }
```

I think that this is a little bit strange. It would be much predictable to generate symbolized hash no matter what kind of param was passed to `attributes` method. Tell me if you agree with this and I'll fix this pull-request.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -79,10 +79,12 @@ end
       end
 
       def attributes(*attrs)
-        @_attributes.concat attrs
-
         attrs.each do |attr|
-          define_method attr do
+          striped_attr = strip_attribute attr
+
+          @_attributes << striped_attr
+
+          define_method striped_attr do
             object.read_attribute_for_serialization attr
           end unless method_defined?(attr)
         end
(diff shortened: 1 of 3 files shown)


hut8:
Any chances of this being ported to 0.10.x?

remear:
@hut8 There was a discussion on it. https://github.com/rails-api/active_model_serializers/pull/1563

Both that PR and this issue are closed, so if there's something you want to discuss further, I'd suggest opening a new issue with your question(s) and use-case(s).

PR #693, opened 2014-10-23
Cache Support at AMS 0.10.0

joaomdmoura:
This is a implementation of cache support to the new version of AMS.

It started originally at the mailing list:
[AMS: Caching](https://groups.google.com/forum/#!topic/rails-api-core/FRy7GMbGEMI) with *Matthijs Langenberg*.
[New Cache Support on AMS](https://groups.google.com/forum/#!topic/rails-api-core/kxP0WxWzFEg). with me

This implementation proposes x different approaches of the old versions of AMS.

- A different syntax. (no longer need the cache_key method)
- An options argument that have the same arguments of ```ActiveSupport::Cache::Store```, plus a ```key``` option that will be the prefix of the object cache on a pattern ```"#{key}-#{object.id}"```.
- It cache the objects individually and not the whole Serializer return, re-using it in different requests (as a show and a index method for example.)

There are many benefits on this approach:

- You can easily know which will be your final cache key, so you can expire it with any custom code.
- You can use all options of ```ActiveSupport::Cache::Store``` (:expires_in for example)
- All objects will be cached individually and re-used on any other return needed
- It keeps you Serializer clean by not depending on old ```cache_key``` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ serializers:
 
 ```ruby
 class PostSerializer < ActiveModel::Serializer
+  cache key: 'posts', expires_in: 3.hours
   attributes :title, :body
 
   has_many :comments
@@ -245,6 +246,37 @@ You may also use the `:serializer` option to specify a custom serializer class, 
 
 The `url` declaration describes which named routes to use while generating URLs
 for your JSON. Not every adapter will require URLs.
+
+## Caching
+
+To cache a serializer, call ```cache``` and pass its options.
+The options are the same options of ```ActiveSupport::Cache::Store```, plus
+a ```key``` option that will be the prefix of the object cache
+on a pattern ```"#{key}/#{object.id}-#{object.updated_at}"```.
+
+**[NOTE] Every object is individually cached.**
+**[NOTE] The cache is automatically expired after update an object but it's not deleted.**
+
+```ruby
+cache(options = nil) # options: ```{key, expires_in, compress, force, race_condition_ttl}```
+```
+
+Take the example bellow:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  cache key: 'post', expires_in: 3.hours
+  attributes :title, :body
+
+  has_many :comments
+
+  url :post
+end
+```
+
+On this example every ```Post``` object will be cached with
+the key ```"post/#{post.id}-#{post.updated_at}"```. You can use this key to expire it as you want,
+but in this case it will be automatically expired after 3 hours.
 
 ## Getting Help
 
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -50,6 +50,20 @@ module ActiveModel
         json[meta_key] = meta if meta && root
         json
       end
+
+      private
+
+      def cached_object
+        klass = serializer.class
+        if klass._cache
+          _cache_key = (klass._cache_key) ? "#{klass._cache_key}/#{serializer.object.id}-#{serializer.object.updated_at}" : serializer.object.cache_key
+          klass._cache.fetch(_cache_key, klass._cache_options) do
+            yield
+          end
+        else
+          yield
+        end
+      end
     end
   end
 end
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@ coverage
 doc/
 lib/bundler/man
 pkg
+Vagrantfile
+.vagrant
 rdoc
 spec/reports
 test/tmp
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -10,6 +10,9 @@ module ActiveModel
       attr_accessor :_attributes
       attr_accessor :_associations
       attr_accessor :_urls
+      attr_accessor :_cache
+      attr_accessor :_cache_key
+      attr_accessor :_cache_options
     end
 
     def self.inherited(base)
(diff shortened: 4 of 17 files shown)


BenMorganIO:
Hey, has this been actively worked on lately? If caching is required for some apps, would 0.8 be a better fit?

README.md
```diff
@@ -150,6 +151,37 @@ get its `Comment`s as well.
...
+
+```ruby
+cache(options = nil) # options: ```{key, expires_in, compress, force, race_condition_ttl}```
+```
+
+Take the example bellow:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  cache key: 'post', expires_in: 3.hours
```
kurko: @joaomdmoura could we make `key` optional? Then we could use `object.cache_key` by default. Makes sense?

kurko:
Could you rebase this? We want to move forward with caching :grin:

README.md
joaomdmoura: I'm not sure if I have already done it :confused: But I'll check it.

joaomdmoura:
haha Actually I already started it 3 times :disappointed: but the tests always end up broken.
I'll try to really do this asap. :smile:

kurko:
I'm so sorry :( Getting this in is my #1 priority now, I promise.

Could you also squash your commits into one with a nice title and description?

joaomdmoura:
@kurko It's done! Rebased, working and squashed.

lib/active_model/serializer/adapter.rb
```diff
@@ -50,6 +50,20 @@ def include_meta(json)
...
         json
       end
+
+      private
+
+      def cached_object
+        klass = serializer.class
+        if klass._cache
+          _cache_key = klass._cache_key || serializer.object.class.name.downcase
+          klass._cache.fetch("#{_cache_key}-#{serializer.object.id}", klass._cache_options) do
```
kurko: I think I wasn't clear in my previous comment (https://github.com/rails-api/active_model_serializers/pull/693/files#diff-04c6e90faac2675aa89e2176d2eec7d8R252). Basically, I believe we should use `model.cache_key` here when key is not used (docs in http://api.rubyonrails.org/classes/ActiveRecord/Integration.html). If you don't specify a key, it'd use `object.cache_key`:

```
Person.find(5).cache_key  # => "people/5-20071224150000"
```

The problem with using only an id here is that if we update the model, it'll reuse an old cache because `update_at` is disregarded. With `cache_key`, it'll regard `update_at`. With a custom `key`, I believe `updated_at` should also be included in this identifier.

What do you think?

BenMorganIO: :+1:

joaomdmoura:
@kurko totally agree. I'll work on this.

joaomdmoura:
@kurko there is a down side on adding the updated_at to the key that I would like to discuss.
One of the reason I didn't  added this before was:

On the later versions of cache on AMS there wasn't a way to expire an specific cache on an abstracted logic, because there was no way to know the cache key.
I added the key option and defined this simple "key-id" pattern in order to make it simple to expire this cache on any part of your application, like a background job for example.

Do you think there is no need to worry about it? Or do you see some workaround?

kurko:
@joaomdmoura 

1. using `updated_at`, changed objects would automatically make caches obsolete. I believe this is what people would expect to be the convention.
2. expiring caches wouldn't be that hard. I'm pretty sure you could find objects with a Glob and wildcard.
3. people can still disregard this and set their own keys. Perhaps another option would be sense if you feel uncomfortable, e.g  `cache key: "post", identifier: :id` vs `cache key: "post", identifier: [:id, :updated_at]`

joaomdmoura:
@kurko yeah, u're probably right.
I may not be taking into account the most expected behaviour.
I'll move forward on this feature so that we can finally merge it.

lib/active_model/serializer/adapter.rb
```diff
@@ -1,3 +1,4 @@
+require 'pry'
```
nwjsmith: A wild `pry` appears

nwjsmith:
Is there a way for this cache-ing scheme to support something like [Fragment/Russian Doll Caching](http://edgeguides.rubyonrails.org/caching_with_rails.html#fragment-caching)? If not, this leaves AMS at a serious disadvantage to JBuilder for higher-performance apps.

One of the things that is nice about the old `def cache_key`-based caching scheme was that the key could be generated dynamically. You can, for example, generate a key based on the cache key for associated objects, or the current Git commit. Will there still be hooks in this new caching scheme to allow for that?

lib/active_model/serializer/adapter.rb
BenMorganIO: Observe, `pry` in its natural habitat.

joaomdmoura:
@nwjsmith thank you for your interest on this.
1. None older version of AMS had this, but I totally agree with you that it would be an awesome feature.
A work around I've being doing on past project is Composite Caching (http://robots.thoughtbot.com/fast-json-apis-in-rails-with-key-based-caches-and).
This PR does not have this feature, but I confess that I'm tempted to add this soon!

2. Unfortunately not. The cache key can be partially defined dynamically, using the ```key``` option.
This may also be an improvement to this cache implementation, there is room for this feature.

lib/active_model/serializer/adapter.rb
joaomdmoura: hahahahahah sorry people, already removed it. :smile:

joaomdmoura:
@kurko just letting you know that it's ready. No pressure hahah :grin:

kurko:
:heart:

sandstrom:
Awesome! Smooth :boat:

BenMorganIO:
YAY

kenips:
:heart: :heart: :heart: 

@kurko can we get a 0.10.0.alpha released soon?

kurko:
@kenips we need to get fragment cache first :) Check #802.

joaomdmoura:
@kenips It's already WIP. I'll try to get it done next week.
The first version is already working but there is some optimisations that I need to work on, and also think how I will deal with relationships.

PR #695, opened 2014-10-24
Small README tweak, add Built in Adapters section

nikz:
Hi All,

Spent a wee while today trying to get the JSON generated work properly for Ember.

Eventually (after trying to set `config.root`, instantiating my own Serializers in my controllers, etc. etc.) I figured out that the intent is to use JsonApiAdapter :smile: 

So I added a small section to the README just to correct that default, and I additionally added a section on the different adapters for the next person.

Hope it's OK - let me know if you want tweaks.

Cheers,

Nik

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ end
 ```
 
 Generally speaking, you as a user of AMS will write (or generate) these
-serializer classes. By default, they will use the JsonApiAdapter, implemented
+serializer classes. By default, they will use the JsonAdapter, implemented
 by AMS. If you want to use a different adapter, such as a HalAdapter, you can
 change this in an initializer:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,6 +85,35 @@ end
...
+
+```ruby
+ActiveModel::Serializer.config.adapter = :json_api
+```
+
+The main difference between the default `:json` and `:json_api` adapters is that 
+`:json_api` includes a "root" element (usually the class name of the serialized 
+object). This is roughly equivalent to Rails' `include_root_in_json` setting and
+useful for compatibility with frameworks such as Ember.
+
```
kurko: I think this paragraph is not needed. A simple link to http://jsonapi.org would avoid confusion because the differences are enormous.

JoshSmith:
How relevant is this PR still today? @kurko?

kurko:
I think it's stale. @nikz ping

Closing this for now. Please reopen if you have some time to change it slightly.

Issue #661, opened 2014-09-21
Should one need to specify both embed: :objects and embed_in_root: false?


kbullaughey:
It seems redundant to need to specify both `embed: :objects` and `embed_in_root: false`. It seems like `embed: :objects` should imply `embed_in_root: false`, no? Would there be any case where one would want both? At least it seems like a better default is to let the former imply the latter.

I'm using 0-9-stable.

PR #698, opened 2014-10-24
Updating read me to inform deprecation of embed method #661

joaomdmoura:
It also replaces the include by embed. It's related to #661


joaomdmoura:
I'm not sure it still valid, so I'm closing it.

Issue #507, opened 2014-02-09
Should enforce snake_case attributes by default


eccegordo:
When I run the following generator:

```
rails g serializer student firstName
```

results in the following output in `app/serializers/student_serializer.rb`

```
class StudentSerializer < ActiveModel::Serializer
  attributes :id, :firstName
end
```

If feasible, this should probably enforce the snake_case convention. 

```
class StudentSerializer < ActiveModel::Serializer
  attributes :id, :first_name
end
```

The camelCased attribute presents subtle issues for users of Ember Data who are expecting snake_cased attributes in the JSON output.

Some related discussion here https://github.com/dockyard/ember-appkit-rails/issues/183

I recognize that there may be larger more subtle issues involved with enforcing the specific snake_case behavior. But if a user types in a camelCased attribute in their generator command they may introduce subtle issues that are not completely obvious.

Wondering if perhaps one solution is to keep the explicitly camelCased attribute (respect user intention in the generator) but inject a transformation of the attribute for the user.

```
class StudentSerializer < ActiveModel::Serializer
  attributes :id, :firstName, :first_name

  def first_name
    object.firstName
  end
end
``` 

This behavior perhaps could be hidden behind a special flag `--mixedcase`

joaomdmoura:
I'm closing this one, mostly because it's out of date and we are already handling it on other PR's a discussions.

PR #699, opened 2014-10-26
Snake case attributes #507

joaomdmoura:
Adding automatically treat to **CamelCased** attributes when using AMS Serializer gerenator.
It replaces the **CamelCased** attribute by a **snake_cased** one and defining a method to access the original attribute inside the serializer. For example:

```ruby
  attributes :full_title, :body

  has_many :comments

  url :post

  def full_title
    object.fullTitle
  end
end
```

As I wrote on README:
> **CamelCased** attributes will be converted to **snake_case** to ensure a AMS
> approach consistence. It will be done be automaticly creating a method in the
> generated serializer.

It is related to feature proposal #507 and other projects as [Ember APIKit Rails](dockyard/ember-appkit-rails#183)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# ActiveModel::Serializers 
- 
+# ActiveModel::Serializers
+
 [![Build Status](https://travis-ci.org/rails-api/active_model_serializers.svg)](https://travis-ci.org/rails-api/active_model_serializers)
 
-ActiveModel::Serializers brings convention over configuration to your JSON generation. 
+ActiveModel::Serializers brings convention over configuration to your JSON generation.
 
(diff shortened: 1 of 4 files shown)


kurko:
This is a clever addition, but I don't think it belongs into AMS. We can't guarantee that everyone will use snakecase. This PR, though, would make AMS not agnostic.

Perhaps we could think about some generator hook, so that you can add the generator to your app and AMS could check if you have a certain directory or something?

JoshSmith:
@joaomdmoura Since we're trying to remain agnostic, what do you think of the response above? Also, projects like `ember-appkit-rails` are now deprecated entirely.

I'd recommend the linter approach mentioned in #507.

kurko:
I'm giving this a close because it's stale for so long.

joaomdmoura:
Oo I didn't even remembered this PR hahahah sorry about that. Make yourself comfortable to close it :stuck_out_tongue_closed_eyes:

PR #725, opened 2014-11-11
Support has_one to be compatible with 0.8.x

ggordon:
The 0.8.x branch used ```has_one``` for associations, it looks like this was changed to ```belongs_to``` in 0.10.x. This patch adds support for ```has_one``` in addition to ```belongs_to``` for backwards compatibility.

### DIFF
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -23,7 +23,6 @@ module ActiveModel
         end
       end
 
-
       def setup
         @author = Author.new(name: 'Steve K.')
         @author.bio = nil
@@ -43,11 +42,11 @@ module ActiveModel
         @comment_serializer = CommentSerializer.new(@comment)
       end
 
-      def test_has_many
+      def test_has_many_and_has_one
         assert_equal(
           { posts: { type: :has_many, association_options: { embed: :ids } },
             roles: { type: :has_many, association_options: { embed: :ids } },
-            bio: { type: :belongs_to, association_options: {} } },
+            bio: { type: :has_one, association_options: {} } },
           @author_serializer.class._associations
         )
         @author_serializer.each_association do |name, serializer, options|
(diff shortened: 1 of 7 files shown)


JoshSmith:
@ggordon would you mind updating the description with your reasoning behind this?

ggordon:
@JoshSmith done.

JoshSmith:
Is the `belongs_to` association no longer tested? It looks like you removed the associations test for that.

ggordon:
I only added tests for has_one, I didn't delete anything.

test/serializers/associations_test.rb
```diff
@@ -44,7 +44,7 @@ def test_has_many
         assert_equal(
           { posts: { type: :has_many, options: { embed: :ids } },
             roles: { type: :has_many, options: { embed: :ids } },
-            bio: { type: :belongs_to, options: {} } },
```
JoshSmith: Am I misunderstanding, or is this not removing the the test for the `belongs_to`?

ggordon:
Sorry it's been a while since I looked at this... That change is in the 'test_has_many' method so it should be irrelevant. The test 'test_belongs_to' is testing the belongs_to association on the comment serializer, which hasn't changed.

JoshSmith:
You don't think this belongs in a `test_has_one`? Just curious, since it feels kind of hidden for someone without all the context of this PR and reasoning behind it.

ggordon:
I could add a test to associations_test.rb, I guess I was only focused on jsonapi so I put it in adapter/json_api/has_one_test.rb

JoshSmith:
Sure, whatever you think is best.

Appreciate the backwards compatibility on this, though! :love_letter:

ggordon:
Didn't see any easy way to separate the has_many and has_one tests, so i renamed it to test_has_many_and_has_one.

JoshSmith:
Awesome, that makes sense to me.

kurko:
@ggordon could you update the README.md and CHANGELOG.md?

ggordon:
README.md and CHANGELOG.md updated. Also I was sometimes getting a failure in unrelated code, I think it was an id collision in the Post model ids and the order the tests were executed. I had to make a workaround to get the tests to consistently pass.

PR #735, opened 2014-11-19
Allows providing a `module_namespace` option

jejacks0n:
This is passed through to the active record object, so that it can prepend the module namespace to the serializer name before constantizing it.

:white_check_mark: Allows easy versioning for serializers!
If you want to version your serializers, your controller can now do so easily.

:white_check_mark: Allows creating a better controller/view pattern!

application_controller.rb

``` ruby
  def default_serializer_options
    {module_namespace: self.class.name.deconstantize}
  end
```

path structure

```
|- controllers
|  |- api
|     |- v1
|         posts_controller.rb
|- serializers
   |- api
      |- v1
          posts_serializer.rb
```


brancusi:
+1

PR #742, opened 2014-11-29
Serialize polymorphic, embed-in-root relationships by class name instead of association name

carsomyr:
Since entities are embedded at the root level, polymorphic associations should use the class name under polymorphism instead of the association name. `has_one` and `has_many` relationships have been updated to reflect this, and `DefaultSerializer` now represents polymorphic types just as `Serializer` would.

This change brings active_model_serializers much closer to supporting Ember Data sideloading with polymorphism and the [`ActiveModelAdapter`](http://emberjs.com/api/data/classes/DS.ActiveModelAdapter.html). On the Rails side, you need an initializer of the form:

    ActiveModel::Serializer.setup do |config|
      config.embed = :ids
      config.embed_in_root = true
    end

... and a `key` option in associations to ensure that `key_name` gets serialized as `key_name` instead of `key_name_id`:

    has_one :content, polymorphic: true, key: "content"

Admittedly, I made this PR to bring an end to monkey patching of our active_model_serializer / Ember Data transport layer, but I do believe that people would want this type of embed-in-root behavior regardless of what consumes the JSON. Also, getting an API server and JavaScript MVC framework to play well together would be a nice showpiece for complex serialization done right, i.e., with polymorphism *and* sideloading.

### DIFF
--- a/lib/active_model/default_serializer.rb
+++ b/lib/active_model/default_serializer.rb
@@ -12,12 +12,17 @@ module ActiveModel
     def initialize(object, options={})
       @object = object
       @wrap_in_array = options[:_wrap_in_array]
+      @polymorphic = options[:polymorphic]
     end
 
     def as_json(options={})
       instrument('!serialize') do
         return [] if @object.nil? && @wrap_in_array
         hash = @object.as_json
+
+        hash = {:type => type_name(@object), type_name(@object) => hash} \
+          if @polymorphic && !@object.nil?
+
         @wrap_in_array ? [hash] : hash
       end
     end
@@ -27,6 +32,10 @@ module ActiveModel
     private
     def instrumentation_keys
       [:object, :wrap_in_array]
+    end
+
+    def type_name(elem)
+      elem.class.to_s.demodulize.underscore.to_sym
     end
   end
 end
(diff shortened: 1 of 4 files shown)


mankind:
@carsomyr  and @kurko, I switched from 0.9  to  0.8 because @steveklabnik  said it will eventually become the official branch, I actually need this feature right now, so rebasing to 0.8 will be great.

@carsomyr thanks for working on it.

kurko:
@carsomyr would it be possible to rebase against `master`? `master` will be `0.10` and the official release, besides being way cleaner. `0.8` and `0.9` will be deprecated.

@mankind could you try `master` (0.10) and see if it works for you? We're eventually going to deprecate all other branches. `0.8` is not going to be the official branch. `master` is pretty stable, lacking cache though (we're working on it).

mankind:
@kurko  thanks for letting me know. I will switch to **0.10**.

carsomyr:
@kurko Ok, I'll try rebasing on `master`.

remear:
This PR was issued toward 0-9-stable and later proposed it be directed toward 0.10. If this changeset is still desired, please open a new PR against master.

PR #744, opened 2014-12-02
Capsulate serializers and enable identifier instead of default id.to_s

TheWudu:
This PR contains:
a) Enable use of identifier to override default id.to_s
b) Enable use of own serializer capsulated in modules
c) Enable use of object_type to override default name (including modules)

According a: We need the possibility to define an own identifier, because the nested models are mainly identified by its type and not it's id. With this change the default behaviour is not changed and it is possible to use another content in the links section. ( e.g. { id: xxx, type: yyy } )

According b): We do not  have a rails/rails-api structure and the Serializers are inside <Application>::Serializer, and the models are inside <Application>::Model. The original code would expect the serializer to be defined in <Application>::Model::<model>Serializer. This would mix serializer and models.

According c) Because of the modules the type_name would look like "<application>/model/<model>" which looks ugly and does represent the internal structure of the classes, not the content. 

We are not really happy with the current solution, because the model has to define the serializer it is used by, but we can't add the "to-use-serializer" to the call, because of the implicit nesting. The object_type method may be moved to the serializer instead of the model.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -142,7 +142,9 @@ module ActiveModel
     def each_association(&block)
       self.class._associations.dup.each do |name, options|
         association = object.send(name)
-        serializer_class = ActiveModel::Serializer.serializer_for(association)
+        serializer_class = options[:options][:serializer] if association
+        serializer_class ||= ActiveModel::Serializer.serializer_for(association)
+
         serializer = serializer_class.new(association) if serializer_class
 
         if block_given?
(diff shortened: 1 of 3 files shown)


JoshSmith:
@TheWudu ping. Are you able to provide tests?

TheWudu:
Hi!

Sorry for my late answer, a million things to do .. We have to adapt to the latest json-api, so I have to check what is currently possible with the serializer, and what needs to be adapted, and for the next PR i will provide tests.

TheWudu:
I close the pull request, we forked the master again and will adapt the now needed changes, i will create multiple smaller pull requests out of it. Including tests.

PR #750, opened 2014-12-07
pass serializable_options to associations

ashanbrown:
@RobinDaugherty had committed some changes included in 0.9.1 to forward options passed to `as_json` to serialize_object, bringing back the behavior from 0.8.  I definitely welcomed these changes, as I had been stuffing options into `scope` to get around this, but these options, however, are still not available to associations.  I believe they should be passed to serialize_object for associations and these changes do just that.

I was afraid of breaking stuff so the code goes out of it's way continue to support serialize_object methods that cannot accept options.

I've included some tests and some helpers as well to help create serializer classes so that I don't need to define all my serializers in the poro file.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -197,6 +197,7 @@ end
     def embedded_in_root_associations
       associations = self.class._associations
       included_associations = filter(associations.keys)
+      options = Array.wrap(serialization_options.presence)
       associations.each_with_object({}) do |(name, association), hash|
         if included_associations.include? name
           association_serializer = build_serializer(association)
@@ -209,7 +210,7 @@ end
               hash = hash[association.embed_in_root_key] ||= {}
             end
 
-            serialized_data = association_serializer.serializable_object
+            serialized_data = association_serializer.serializable_object(*options) # always sending options may break some tests/serializers
             key = association.root_key
             if hash.has_key?(key)
               hash[key].concat(serialized_data).uniq!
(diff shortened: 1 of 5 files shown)


dontfidget:
Just wondering if anyone has any comments on this?  Again, the idea is to automatically pass options through to association serializers.  If we don't think it should be default behavior, maybe there could at least be an option to pass the options on through.  Thanks.

gkop:
Unfortunately this patch doesn't fix my issue with `has_many ..., embed_in_root: true`; the serializable options are not passed through to the the association serializers.

dontfidget:
I thought I included a test for that.  Any idea how it's failing (i.e. how your use case is different from the spec I added)?  Thanks.

gkop:
Would you rebate or merge the 0.9 stable branch and make sure your tests still pass?

dontfidget:
I've rebased my changes on 0-9-stable and the tests do pass locally for me so presumably they'll pass on CI.  I am sort of wondering if there should be an option on associations to forward the options to them, rather than changing existing behavior.  But either way, I definitely would like an easy way to to forward associations to options.

ashanbrown:
Just want to check in on this.  My goal is make the serializable options pass through to an association serializer without having to explicitly instantiate those serializers with the desired optinos (i.e. by just using has_many).  This seemed like a good way to go about it, but I'm open to other options.  Thanks.

macfanatic:
Would love to use this feature, have a few serializers that make use of a `has_many :something, serializer: MySerializer` setup but fails due to the `scope` being nil when called on the collection.

NullVoxPopuli:
is this still applicable?

bf4:
I think this was fixed

PR #759, opened 2014-12-13
Fix method overwriting of `ActionController::TestCase` (It breaks Rails 4.2.0's controller test)

tricknotes:
This PR came from https://github.com/rails-api/active_model_serializers/pull/620 .
This commit fixes the following error on Rails 4.2.0: https://github.com/emberjs/ember-rails/issues/408#issuecomment-66860483

I feel happy if you could bump new version with this commit :pray:


tricknotes:
It seems this commit is landed in [0-9-stable](https://github.com/rails-api/active_model_serializers/tree/0-9-stable) branch.

PR #767, opened 2015-01-05
Adds support for meta attribute

kurko:
Currently, `0.10.0.pre` (master) doesn't support `meta` option in `render`. This
way, there's no way to support features such as pagination. `0.9` had
this feature in place.

This adds support for it, as well as fixes small things in README.md.

This won't support `meta` in array responses because arrays don't have
keys, obviously. Also, the response should have a `root` key, otherwise
no `meta` will be included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,3 @@
+### 0.10.0
+
+  * adds support for `meta` and `meta_key` [@kurko]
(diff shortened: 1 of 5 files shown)


guilleiguaran:
why this was closed? :grin:

kurko:
I'm testing locally and it didn't work. Found out that JSONAPI will use ArraySerializer (which shouldn't have a `root` and so no `meta`), but adds a `root` at the end. So relying on the Serializer's `root` isn't right. Will fix this and reopen.

PR #768, opened 2015-01-05
Adds support for meta attribute

kurko:
Currently, 0.10.0.pre doesn't support `meta` option in `render`. This
way, there's no way to support features such as pagination. `0.9` had
this feature in place.

This adds support for it, as well as fixes small things in README.md.

This won't support `meta` in array responses because arrays don't have
keys, obviously. Also, the response should have a `root` key, otherwise
no `meta` will be included.

In some cases, for example using JsonApi, ArraySerializer will result in
a response with a `root`. In that case, `meta` will be included.

### DIFF
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -18,7 +18,8 @@ module ActiveModel
       end
 
       def as_json(options = {})
-        serializable_hash(options)
+        hash = serializable_hash(options)
+        include_meta(hash)
       end
 
       def self.create(resource, options = {})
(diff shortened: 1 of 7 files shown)


kurko:
@guilleiguaran tested and working :grin:

lib/active_model/serializer/adapter.rb
```diff
@@ -30,6 +31,27 @@ def self.create(resource, options = {})
...
+        serializer.meta_key || "meta"
+      end
+
+      def root
+        serializer.json_key
+      end
+
+      def include_meta(json)
+        puts "meta: #{meta.inspect}" if $debug
+        puts "root: #{root.inspect}" if $debug
```
guilleiguaran: please delete the ``puts`` :)

kurko:
@guilleiguaran :+1:

PR #769, opened 2015-01-07
Regression Fix: Don't use array serializer when rendering a hash

markquezada:
90343ce introduced a bug when rendering basic hashes. Since hashes respond to `each` they were being incorrectly rendered as multi-level arrays after that commit.

This PR simply checks if the resource is not a hash before using the array serializer. I'm not entirely convinced using `each` is the right approach here in the first place, and I dislike checking for a hash explicitly, but this restores the previous functionality without removing the functionality introduced in the above commit.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -87,8 +87,8 @@ module ActionController
       if serializer = options.fetch(:serializer, default_serializer(resource))
         options[:scope] = serialization_scope unless options.has_key?(:scope)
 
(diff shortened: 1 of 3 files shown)


kurko:
Can you rebase, @markquezada. We released `0.9.3` with a series of fixes and I believe a similar patch was introduced, but I'm not sure.

markquezada:
@kurko Looks like the offending commit was reverted in 577db35. I'll close this PR.

PR #771, opened 2015-01-11
Make linked resource type names consistent with root names

sweatypitts:
Using the JSON API adapter, when models are coming from a Rails engine and are namespaced, the namespace was reflected in linked resource types but not in the root type, leading to inconsistent resource names. This PR removes the namespace from linked resources. For example:

Given models like this:

```ruby
module Pizza

  class Dough < ActiveRecord::Base
    belongs_to :pie
  end

  class Pie < ActiveRecord::Base
    has_one :crust, class_name: 'Pizza::Dough'
  end

end
```

Serialization could end up looking like this:

```javascript
{
  linked: {
    'pizza/doughs': [{
      id: 456
    }]
  },
  pies: {
    id: 123,
    links: {
      crust: {
        id: 456,
        type: 'pizza/dough'
      }
    }
  }
}
```

But should look like this:

```javascript
{
  linked: {
    'doughs': [{
      id: 456
    }]
  },
  pies: {
    id: 123,
    links: {
      crust: {
        id: 456,
        type: 'dough'
      }
    }
  }
}
```

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -126,7 +126,7 @@ module ActiveModel
 
         def serialized_object_type(serializer)
           return false unless Array(serializer).first
-          type_name = Array(serializer).first.object.class.to_s.underscore
+          type_name = Array(serializer).first.object.class.to_s.demodulize.underscore
           if serializer.respond_to?(:first)
             type_name.pluralize
           else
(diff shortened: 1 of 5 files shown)


guilleiguaran:
can we have a test for this to avoid future regressions?

sweatypitts:
Indeed, incoming.

sweatypitts:
@guilleiguaran How's this look?

guilleiguaran:
Looks great for me :+1:

PR #773, opened 2015-01-12
Make json api adapter 'include' option accept an array

sweatypitts:
I'm proposing this change because it just makes more sense (to me). I'm not sure why we're accepting a comma-delimited string and then splitting it into an array instead of just accepting an array in the first place. This still allows a comma-delimited string for legacy reasons. I did not include a test for this on purpose but can add it if you'd like.

I have updated the readme to reflect the usage change. It's pretty straight-forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ resources in the `"linked"` member when the resource names are included in the
 `include` option.
 
 ```ruby
+  render @posts, include: ['authors', 'comments']
+  # or
   render @posts, include: 'authors,comments'
 ```
 
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -121,7 +121,9 @@ module ActiveModel
         end
 
         def check_assoc(assoc)
-          @options[:include].split(',').any? do |s|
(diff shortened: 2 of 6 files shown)


kurko:
> I did not include a test for this on purpose but can add it if you'd like.

I see many tests in there though. 

Anyway, could you leave at least one test for `render @posts, include: 'authors,comments'`, without array?

sweatypitts:
Sorry, the way I wrote that was a little confusing. I meant that I did not write tests for the comma-delimited string behavior. But yes, I can add tests for that.

sweatypitts:
@kurko This should be good to go now.

kurko:
> We can’t automatically merge this pull request.

While you rebase this, do you mind adding to the README the stringfied version of `include` too, instead of just the array version? Something like,

> `include: "comments,comments.user"` is also supported.

sweatypitts:
@kurko K, rebased again and added the additional documentation. :smiley:

kurko:
Needs rebase :grin:

sweatypitts:
@kurko Rebased again! Please merge quickly before I need to rebase again ahhhhhhh! :smile:

kurko:
hahaha

PR #774, opened 2015-01-13
Fix nested include attributes

nhocki:
When the requests asked for a nested attribute in the `include` and it's missing or empty, don't break because the type of the object can't be determined.

If the request is for a collection and none of the elements has the attribute, it will not be added to the `linked` key, similar to what happens with simple includes.


kurko:
LGTM.

PR #775, opened 2015-01-13
Revert #665

nateberkopec:
This reverts commit 90343cea4dcdd84abd1a86cf3175e26bd0b7ca81.


kurko:
@nateberkopec thanks. Can you open this against 0-9-1 instead? This branch has tons of commits more, bunch of fixes, but had to be reverted. With 0-9-1 you shouldn't see conflicts (I think).

Issue #766, opened 2014-12-28
Controller tests fail with 0.9.2 and Rails 4.2


brandonhilkert:
I just recently upgraded an app to Rails 4.2. All my controllers tests are failing with:

```ruby
  1) Error:
MarketingControllerTest#test_should_get_landing:
NoMethodError: undefined method `clear' for nil:NilClass
    test/controllers/marketing_controller_test.rb:5:in `block in <class:MarketingControllerTest>'
```

Here are the details:

https://gist.github.com/brandonhilkert/2f39218578b27e264efe

If i remove `active_model_serializers`, the tests pass.

The inclusion of `ActionController::TestCase.send(:include, ::ActionController::SerializationAssertions)` in the gem is causing this. If I comment it out, the tests again pass.

It appears when `super` is called from https://github.com/rails-api/active_model_serializers/blob/0-9-2/lib/action_controller/serialization_test_case.rb#L25, when it goes back to rails (https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/test_case.rb#L64), it fails out with the `clear` calls on those instance variables.

It's still not clear to me why the inclusion of this would causing those vars to be nil, but maybe it's more obvious to you.

The master branch passes, but I'm not sure if I should be relying on this given that it's not released yet.

kapso:
+1 seeing the same, but `0.9.1` has an issue #746 so cant use that

quekshuy:
+1 affects our controller tests as well.

kurko:
Released https://rubygems.org/gems/active_model_serializers/versions/0.9.3. It's basically 0.9.1 (tons of bugfixes) minus the commit that caused tons of regressions. Please test and let us know how it goes :grin:

Issue #641, opened 2014-09-06
Doesn't work on Rails 4.2.0 beta


bigardone:
Hi!
I've been playing around with **Rails 4.2.0beta** and **active_model_serializers**, but it doesn't work. When rendering json in a controller it doesn't use the generated serializer.
This is the serializer:
https://github.com/bigardone/rails_and_react/blob/master/app/serializers/person_serializer.rb
And this the controller:
https://github.com/bigardone/rails_and_react/blob/master/app/controllers/people_controller.rb

Am I doing anything wrong or missing anything?

Thanks in advance!

barberj:
@igagnidz's fix didn't work for me. I'm on rails (4.2.0) non beta. I have ended up just using jbuilder which ships with rails now.

dbackeus:
+1

We are using active_model_serializers with mongoid together with respond_with (now responders gem) and didn't work out of the box after the rails 4.2 upgrade. Explicitly initialising our serializers seems ok as a work around for now.

kapso:
Running into similar issues with Rails 4.2. Is there a release date for `0-9-stable`.

Thanks much!

iRonin:
thx @igagnidz! :+1:

bf4:
Taking @igagnidz comment, applied as follows

```ruby
# Gemfile
gem "active_model_serializers", "~> 0.9.2"

# config/initializers/0_extensions.rb
Dir[Rails.root.join("lib/extensions/*.rb")].each do |file|
  require file
end

# lib/extensions/active_model_serializers.rb
gem "active_model_serializers", "= 0.9.2"
# blow up when this patch it no longer needed
# see https://github.com/rails-api/active_model_serializers/issues/641#issuecomment-68196151
require "active_model_serializers"
module RailsApiSerializationPatch
  def _render_with_renderer_json(json, options)
    serializer = build_json_serializer(json, options)

    if serializer
      super(serializer, options)
    else
      super(json, options)
    end
  end
end
ActiveSupport.on_load(:action_controller) do
  include RailsApiSerializationPatch
end
```

MarshalGeazipp:
thx @igagnidz and @bf4 ! :+1:

dirkkelly:
:+1: @igagnidz and @bf4

ankit8898:
Patch works thanks @igagnidz and @bf4 :+1:

sb8244:
Confirmed patch works for me

yagooar:
[...]

PR #776, opened 2015-01-13
Revert #665

nateberkopec:
This reverts commit 90343cea4dcdd84abd1a86cf3175e26bd0b7ca81.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -87,7 +87,7 @@ module ActionController
       if serializer = options.fetch(:serializer, default_serializer(resource))
         options[:scope] = serialization_scope unless options.has_key?(:scope)
 
-        if resource.respond_to?(:each)
+        if resource.respond_to?(:to_ary)
           options[:resource_name] = controller_name 
           options[:namespace] = namespace_for_serializer if namespace_for_serializer
         end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -56,7 +56,7 @@ end
       attr_reader :key_format
 
       def serializer_for(resource, options = {})
-        if resource.respond_to?(:each)
+        if resource.respond_to?(:to_ary)
           if Object.constants.include?(:ArraySerializer)
             ::ArraySerializer
           else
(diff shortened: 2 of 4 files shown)


kurko:
@nateberkopec does this work on your app?

nateberkopec:
@kurko Our 4.2 app's test suite is all green. https://github.com/rails-api/active_model_serializers/issues/766 is definitely fixed by this.

kurko:
@nateberkopec ok, go with branch in your app for now. Will talk to @steveklabnik to release 0.9.3 after this is merged.

dontfidget:
Thanks for trying to take this approach.  I've been trying to track down one other behavior that I think changed with 0.9.2 but I can't seem to reproduce it in test and I just wanted to make sure no one has seen this.  It had been my belief that you could embed serializer definitions within other serializer definitions and it would favor the embedded one.  For example,

```
class UserInfoSerializer < ActiveModel::Serializer
    attributes :nested

    def nested
      'NESTED!'
    end
  end

  has_one :user
end
```

If you did this, I believe that the nested UserSerializer would supercede a UserSerializer defined outside of the serializer.  However, I can't get exhibit behavior in tests again any version in 0.9.x.

Am I wrong in believing that this ever was the case or could this be related to the changes introduced to support automatic reloading of serializers in development?  Thanks.

kurko:
Not sure, but I think this should be in a separate issue.

dontfidget:
OK, I'll open a separate note.  I just mentioned it here because this pull request seems to be ultimately it getting back to the original behavior, and I thought this might have been a regression.

indirect:
I've been using this in my app today with Rails 4.2; it's fixed our issues and hasn't had any problems across four developers on four machines. :shipit:

recurrence:
:+1: in my environment.  Solves a considerable headache :)

meskyanichi:
I wasn't seeing any errors. However, my issue was #641, where the serializers were silently ignored and ALL attributes on my models were always serialized, instead of only the ones specified in the serializers.

Using this PR I am no longer seeing this issue in Rails 4.2.0. :+1:

indirect:
@kurko think we can get a release? :grinning:

kurko:
Perhaps yes :)

@klabnik ping

kurko:
Er.. @steveklabnik

kurko:
Released https://rubygems.org/gems/active_model_serializers/versions/0.9.3.

PR #780, opened 2015-01-15
add column_names method

BenMorganIO:
If you are using a large Rails application and you can't keep track of the columns names to whitelist, sometimes its easier just to use a method to white list all the column names for you and use a black list instead. This would be true for a Rails engine such as [Spree](https://github.com/spree/spree).

This method currently has no tests. Not sure how testing should be done in this app.

Usage:

```ruby
class ArticleSerializer < ActiveModel::Serializer
  attributes *column_names
end
```

Result is all columns in the `Article` class are now available to be serialized for the `ArticleSerializer`.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -36,6 +36,10 @@ module ActiveModel
       end unless method_defined?(key)
     end
 
+    def self.column_names
+      to_s.sub("Serializer", "").constantize.column_names
+    end
+
     # Defines an association in the object that should be rendered.
     #
     # The serializer object should implement the association name
@@ -167,10 +171,6 @@ module ActiveModel
       serializer = options.fetch(:options, {}).fetch(:serializer, nil)
       opts[:serializer] = serializer if serializer
       opts
-    end
-
-    def column_names
-      to_s.sub("Serializer", "").constantize.column_names
     end
 
     private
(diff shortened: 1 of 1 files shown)


lib/active_model/serializer.rb
```diff
@@ -36,6 +36,10 @@ def self.attribute(attr, options = {})
       end unless method_defined?(key)
     end
 
+    def self.column_names
+      to_s.sub("Serializer", "").constantize.column_names
```
filipebarcos: I don't know if this is a valid requirement or not, but instead of relying on the serializer name and doing string manipulation, you could just use `resource.class`

BenMorganIO: @filipebarcos Thanks for the timbit. I'll try that.

beauby:
I personally feel we should not make that too easy, so as not to encourage beginners to do it.
You could manually do something like 
```ruby
attributes MyModel.attribute_names
```
which IMO has very little overhead.

Closing this for now, feel free to reopen if needed.

BenMorganIO:
@beauby that's actually a great solution!

bf4:
Looks like data would leak with that. Just use `adapter: false` if you want no transformation

beauby:
@bf4 What do you mean "data would leak"? Using `adapter: false` would not give nested serialization...

Issue #772, opened 2015-01-12
Array render not working anymore on 4.2.0


solher:
Hi,

I just noticed that my error messages are not rendered anymore since 4.2.0 (and AMS master, otherwise, serializers are never reached with 0.9).

I've got a :
"NoMethodError (undefined method `new' for nil:NilClass)"

when I try to do :
"render json: [ 'No session found for the current user' ]"

It seems to be looking for a serializer for the object but it finds nothing since an array isn't a "custom" class.
Maybe it is just because what I'm doing is bad practice, I don't know...
Thanks for the help !

kurko:
It shouldn't fail like this. Will take a look later. It's bug and needs to be fixed. If you want to make it work right now, I'd suggest having something more descriptive. Have you checked http://jsonapi.org? Their error format would work fine with AMS.

ryansch:
I just hit my head on this in our default error handling for the api.  The problem here is trying to render out any type that doesn't have a corresponding Serializer inside of an array.

Normal flow:
We'll try `render json: 'a string'`.  AMS fails to find a serializer for the string literal and the action controller extension bails out on https://github.com/rails-api/active_model_serializers/blob/master/lib/action_controller/serialization.rb#L41 and uses rails' built in render.

Error'd flow:
We'll try to `render json: ['a string']`.  This time AMS is able to find an ArraySerializer for the array but then fails to find the serializer class in https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/array_serializer.rb#L13-L17 and raises the error reported in this issue.

AMS should be able to catch this error and fall back to the default render.

ryansch:
To be clear: We usually return something like `{ errors: ['some error'] }`.  I discovered this behavior due to a bug in one of our controllers.

nessur:
This version of AMS lacks any support for a 'default' serializer, which in 0.9.0 allowed any rendering of a ruby class that responds to 'as_json'. So, AMS 0.10.0.rc1 can't serialize: String, Hash, Time, Integer

from Pry, in our rspec test suite:
```
from: /Users/me/.rvm/gems/ruby-2.1.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer/array_serializer.rb @ line 17 ActiveModel::Serializer::ArraySerializer#initialize:

     9: def initialize(objects, options = {})
    10:   options.merge!(root: nil)
    11:
    12:   @objects = objects.map do |object|
[...]

PR #782, opened 2015-01-18
Serailize plain arrays

mitchlloyd:
This PR is in response to https://github.com/rails-api/active_model_serializers/issues/772

I wanted to learn about ActiveModelSerializers so I thought I'd try to tackle a bug. Unfortunately the one I picked lead me to some pretty big design questions. 

Specifically:

Is there a way to avoid invoking all the guts of AMS on a "plain" array?

```ruby
  def show
    # Would be nice to avoid AMS altogether
    render json: ['one', 'two', 'three']
  end
end
```

and what does it even mean to have primitive values in an array with serializable objects?

```ruby
  def show
    post = Post.find(params[:id])
    render json: ['look at this post', post, 3]
  end
end
```

I'm sending this PR because it at least provides a failing test for the issue. If you have some helpful insights or If you think this PR might be usable let me know and I can work on any needed adjustments. Otherwise, feel no remorse about closing :)

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -5,6 +5,7 @@ module ActiveModel
     extend ActiveSupport::Autoload
     autoload :Configuration
     autoload :ArraySerializer
+    autoload :PrimitiveSerializer
     autoload :Adapter
     include Configuration
 
@@ -107,7 +108,7 @@ module ActiveModel
       else
         options
           .fetch(:association_options, {})
-          .fetch(:serializer, get_serializer_for(resource.class))
+          .fetch(:serializer, get_serializer_for(resource))
       end
     end
 
@@ -221,7 +222,11 @@ module ActiveModel
 
     attr_reader :options
 
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(resource)
+      klass_serializer(resource.class) || primitive_serializer(resource)
+    end
+
+    def self.klass_serializer(klass)
       serializers_cache.fetch_or_store(klass) do
         serializer_class_name = "#{klass.name}Serializer"
         serializer_class = serializer_class_name.safe_constantize
@@ -229,10 +234,15 @@ module ActiveModel
         if serializer_class
           serializer_class
         elsif klass.superclass
-          get_serializer_for(klass.superclass)
+          klass_serializer(klass.superclass)
         end
       end
     end
 
+    def self.primitive_serializer(resource)
+      if PrimitiveSerializer.can_serialize?(resource)
+        PrimitiveSerializer
+      end
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


kurko:
Take a look at https://github.com/rails-api/active_model_serializers/blob/master/lib/action_controller/serialization.rb#L22. You can redefine `user_adapter?` in your controller and disable it temporarily.

Solher:
Works great for me ! Thanks a lot. But your fork isn't going to be merged ?

kurko:
@mitchlloyd does my answer help you?

mitchlloyd:
@kurko I think you may have misinterpreted my intent. I was trying to get the ball rolling on fixing an issue that you identified as a bug. I'm not scratching my own itch. 

Are you suggesting that issue #772 could be solved with better documentation?

I was mainly looking for feedback on this PR.

hderms:
I was going to fix the same issue #772 as @mitchlloyd but realized he already had done so. Is there any obstacle to getting this merged in?

kurko:
@mitchlloyd so, if you have `MyUser` model and no `MyUserSerializer`, it will set `DefaultSerializer`? I'm not sure about that. If you're serializing a model that has no serializer, I believe it should fail.

The original bug (#722) was regarding arrays, instead.

We need more opinions on this.

mitchlloyd:
@kurko I can see how throwing an exception when an object doesn't have a serializer could be a better experience than falling back to the default serialization behavior of Rails in some cases. I agree it would be nice to get more opinions on this though. If that's the conclusion then I'd be happy to change the PR to implement that behavior.

However, in this particular case (from #772):

`render json: [ 'No session found for the current user' ]`

Should the user see an error? `Error: no StringSerializer was defined`. Probably not. Here the OP just expected the default JSON serialization behavior from Rails which seems pretty reasonable to me.

Another possibility might be to do a check upfront for primitive values like `String` and `Hash` and not attempt to lookup a serializer for those values. That seems like it could be a little fragile, but let me know what you think.

kurko:
@mitchlloyd yes, I think we should whitelist those primitive types and have a baked-in serializer for them. As for the rest, my opinion is that we should just raise.

mitchlloyd:
This new `PrimitiveSerializer` handles integers, strings, hashes, and symbols. I kept this check after the primary use case of class-name-lookup for objects with user-defined serializers. I didn't want to add anything that would hurt the performance of the primary use case of the gem.

We still have the behavior where users will get an sub optimal error (`NoMethodError (undefined method `new' for nil:NilClass`) if they don't have a serializer defined for a given class. However, that's a different issue that probably warrants a new PR.

bf4:
Propose to close as fixed by #962

PR #783, opened 2015-01-18
Properly formatting passed in serializer namespace

antstorm:
This fix allows to pass `namespace: :api`, with a lowercase `:api`. Old format, of course, is still supported.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -110,7 +110,7 @@ end
 
       def build_serializer_class(resource, options)
         "".tap do |klass_name|
-          klass_name << "#{options[:namespace]}::" if options[:namespace]
+          klass_name << "#{options[:namespace].to_s.classify}::" if options[:namespace]
           klass_name << options[:prefix].to_s.classify if options[:prefix]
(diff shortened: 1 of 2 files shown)


antstorm:
Any chance of getting this in?

saneshark:
@antstorm I'm not sure that this really adds anything useful and it would kind of make things confusing if it were ever introduced into the Readme as such, because you could have more than one module in the namespace and there would be no way to present that if you pass in a symbol, right?

```ruby
namespace: 'Api::V1'
# vs.
namespace: :api ???
```

antstorm:
@saneshark sure, there's no way of specifying two or more nested modules in a symbol. My reasoning behind this change is to keep things consistent with Rails, which has `namespace :api` in the routes, that is then used to prefix your controllers with `Api::`.

It might be hard to compare these two, because router does more than just prefixes controller classes, but still I think that's a valid point.

kurko:
We need more people chiming in here to make a decision.

groyoh:
This issue does not seem to have moved for quite a while. I'm closing it. If somebody like this should be re-opened, feel free to leave a comment.

PR #790, opened 2015-01-27
pass context to ArraySerializer

lanej:


kurko:
Do you mind adding a test, a mention in the documentation and a little bit of context on why you want this merged? Why `context` and not something more general, that other people could use? Perhaps `options` would be a better fit, which you could then use `context` in your app (via hash)?

lanej:
will do

lanej:
@kurko context is available for singular serializers.  My issue was that context is not "distributed" to the singular serializers when going through an array serializer.

In my opinion, this more of a bug fix to make the collection / singular context availability consistent.

kurko:
Thanks for the explanation. I'm not using 0.9 for years now, so I needed some `context` :grin:

lanej:
:rimshot:

PR #791, opened 2015-01-28
Fixes undefined method `pluralize' for false:FalseClass

philsturgeon:
Had some errors where the `serialized_object_type` method was returning `false` and the `add_linked()` method didn't seem to care. This might not be the best solution, but it works for me.


philsturgeon:
Looks like @nhocki has already fixed this in #774, but he did a better job and actually wrote tests too.

PR #793, opened 2015-01-29
not serialize keys with nil value

andreywerewolf:
Most clients IOS / Android crashing if the key value nil.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -142,7 +142,7 @@ module ActiveModel
         end
 
       attributes.each_with_object({}) do |name, hash|
-        hash[name] = send(name)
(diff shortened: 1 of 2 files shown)


kurko:
@andreywerewolf are these in-house iOS/Android apps? I believe it's something you have fix on your clients. There are mobile apps consuming our API with nil values and we never heard complaints.

Removing attributes doesn't seem right to me. `nil` has significance, it means the attribute has no value. Removing it means it doesn't exist.

JoshSmith:
I agree with @kurko on this. I vote for closing this issue as this should be handled in your apps. We have the same experience.

andreywerewolf:
ok)))

PR #794, opened 2015-01-30
add to_param for correct URL generation

carlesjove:
I don't think that this needs tests, but I can write them if necessary. Just make sure that the correct URL is generated when using Rails' `url_helpers` in tests.


bf4:
FWIW, @carlesjove I removed this in #981 because I couldn't tell where it was used :)

carlesjove:
It was used in another PR that I made, which wasn't merged yet. If that PR
is going to be merged I'll see if I can do without it, or start a
conversation on the issue

PR #798, opened 2015-01-31
Fix lost test `test_include_multiple_posts_and_linked`

donbobka:
Just found that test `test_include_multiple_posts_and_linked` was lost due to copy/paste method with same name.


JoshSmith:
:+1:

sweatypitts:
:+1:

PR #799, opened 2015-01-31
Fields option should filter linked associations

donbobka:
This feature implements filtering of associations with option `fields`. 
IMHO, linked associations are like fields and they should be filtered too, when option `fields` used

Unfortunately, I found nothing about it at http://jsonapi.org/format/. What do you think about it?

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -156,8 +156,16 @@ module ActiveModel
       end
     end
 
-    def each_association(&block)
-      self.class._associations.dup.each do |name, options|
+    def associations(options = {})
+      if options[:fields]
+        self.class._associations.slice(*options[:fields])
+      else
+        self.class._associations.dup
+      end
+    end
+
+    def each_association(options = {}, &block)
+      self.associations(options).each do |name, options|
         next unless object
 
         association = object.send(name)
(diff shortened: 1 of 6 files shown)


ggordon:
I think the include parameter http://jsonapi.org/format/#fetching-includes is the mechanism for filtering associations.

donbobka:
Include parameter is about including, not filtering.
Example:
```ruby
render json: topic
```
```json
{
  "topics": {
    "id": 1,
    "name": "Example topic", 
    "links": {
      "posts": [1, 2, 3]
    }
  }
}
```
How I can filter `topics['links']['posts']`?

For these purpose you have a `fields` parameter. It's about filtering. But without this PR that's doesn't work.
Example:
```ruby
render json: topic, fields: [:id, :name]
```
```json
{
  "topics": {
    "id": 1,
    "name": "Example topic", 
  }
}
```

donbobka:
Look at commit 4fa22c004b3a3f21442fe0980eb8643011e74bc9. There you can find some tests.

donbobka:
But if look at problem from another side: may be linked associations should be opt-in?

```ruby
class TopicSerializer < ActiveModel::Serializer
  attribute :name
  belongs_to :user
  has_many :posts
end

render json: topic
```
Result

```json
{
  "topics": {
    "id": 1,
    "name": "Example topic", 
  }
}
```

bf4:
ref https://github.com/rails-api/active_model_serializers/issues/1058

joaomdmoura:
I'm closing his one for now, it's outdated, but we can still use it on #1058 as @bf4 pointed out.

joaomdmoura:
Btw this is also done on https://github.com/rails-api/active_model_serializers/pull/999 that might be closed soon as well

Issue #648, opened 2014-09-11
undefined method `belongs_to`


colby-swandale:
I am using rails 4.1.4 with the latest active_model_serializers from rubygems.

In the documentation there are examples that show the usage of `belongs_to` but when i try it myself rails complains the method is undefined. Am i missing something? Here is Serializer class i am trying this on.

```
class UserSerializer < ActiveModel::Serializer
  attributes :first_name, :last_name, :email, :state, :uuid
  belongs_to :tenant
end
```

mrusa:
I have got the same problem, using Ruby 2.1.2, Rails 4.1.9 and active_model_serializers 0.9.2 on nginx with passenger.

On my local machine everything works fine but on production there is the above-mentioned error. The passenger error page shows im using active_model_serializers 0.9.2, so there's the correct version running on production. Im not sure what im doing wrong?

My serializers look like the following:

```ruby
class InvoiceSerializer < ActiveModel::Serializer
  attributes :id, :invoice_number, :locked, :active

  has_many :invoice_positions
end
```

```ruby
class InvoicePositionSerializer < ActiveModel::Serializer
  attributes :id, :invoice, :quantity

  belongs_to :invoice
end
```

pferdefleisch:
I just double checked and it turns out `belongs_to` doesn't exist.
If you look at the `0.9.2` README [here](https://github.com/rails-api/active_model_serializers/tree/0-9-2)
you will see this

> Serializers are only concerned with multiplicity, and not ownership. belongs_to ActiveRecord associations can be included using has_one in your serializer.

So think about what you are really doing with your `belongs_to`
If you will never be using `InvoicePositionSerializer` by itself then it is unnecessary
If you will, then use a `has_one` like @chabgood mentioned

colby-swandale:
i'm going to go ahead and close this issue.

mrusa:
Ah ok, so i was reading the wrong readme.. I was just irritated that everything worked on my local environment. Thanks for your help

gnapse:
The README on master (the one that is shown when someone visits the repository's home page, still shows `belongs_to` in an example, and no example with `has_one`.

jimboslicethat:
I know this is a closed issue, but has anyone experienced problem with the has_one method behaviour changing in production? I understand that we are supposed to use has_one instead of belongs_to. I'm just wondering if anyone has run into problems with has_one. See my stackoverflow question for examples: 

[...]

PR #800, opened 2015-02-01
Change 'belongs_to' in README for 'has_one' (see #648)

gnapse:


JoshSmith:
:+1:

JoshSmith:
@joaomdmoura looks good to me, but I'd appreciate another :+1: if it is.

joaomdmoura:
It started at #648 
After reading it I don't think it's the right solution.
As was mentioned, at [previous](https://github.com/rails-api/active_model_serializers/tree/0-9-2) versions of AMS belongs_to wasn't supported

> Serializers are only concerned with multiplicity, and not ownership. belongs_to ActiveRecord associations can be included using has_one in your serializer.

The master README represents a reference to 0.10.0 version and it supports ```belongs_to```:

> The generated seralizer will contain basic `attributes` and `has_many`/`has_one`/`belongs_to` declarations, based on the model.

One of the key points on AMS is to follow Rails conventions, so I really appreciate to keep this kind of *syntax sugars*

Some of the lines updated on this PR are an practical example using ```PostSerializer``` and  ```CommentSerializer```

```ruby
  cache key: 'posts', expires_in: 3.hours
  attributes :title, :body

  has_many :comments

  url :post
end
```

```ruby
class CommentSerializer < ActiveModel::Serializer
  attributes :name, :body

  belongs_to :post

  url [:post, :comment]
end
```

They already have the usual ```has_many - belongs_to``` relationship. Use the same syntax to AMS make it a lot easier to understand. Because of that I'm not sure if this PR should be merged. Hope it makes sense.

I think that the problem pointed out at #648 is related to a miss understanding that the master now representes a new implementation. Maybe we could try to point it out at the README that it's for 0.10.0.

JoshSmith:
Ah, I see. Thanks for the clarification @joaomdmoura. I'd misunderstood.

I feel that this should not be merged in that case, as it's not reflective of what's actually happening.

@gnapse appreciate the PR, but does this clarify things?

gnapse:
I only submitted this PR because the library did not support `belongs_to` previously, yet the README file had an example using it. I think at the time I did it, the version supporting it was not yet released as stable, and I was confused to see the example in the README, but it didn't work for me on the latest stable version at the time.

However I agree with @joaomdmoura that it should be supported. It's clearer to mimic the semantics of the underlaying data models, even if this library is not concerned with ownership, but plurality. I'm glad that it was added to more recent versions. In that case it's ok to disregard this PR completely.

JoshSmith:
Thanks @gnapse. I think we'll go ahead and close this then.

PR #807, opened 2015-02-07
Add Overriding attribute methods section to README.

alexstophel:


guilleiguaran:
can we add same for associations?

alexstophel:
It looks like a section for associations already exists.

Issue #802, opened 2015-02-02
Fragment Cache


joaomdmoura:
Considering my PR #693 
Cache Composition (http://robots.thoughtbot.com/fast-json-apis-in-rails-with-key-based-caches-and) is one of the work-arounds for the absence of fragmente cache on AMS.

As we know, a fragment cache implementation would be extremely efficient to improve API's that have some dynamic attributes that may vary on every request.

Assuming that PR #693 will be merged soon, this should be a new feature to the cache implementation, my proposal is to follow Rails conventions, adding two new options to ```cache``` method: ```only``` and ```except```.

Example:
```ruby
class PostSerializer < ActiveModel::Serializer
  cache key: 'post', expires_in: 3.hours, only: [:title]
  attributes :title, :body

  has_many :comments

  url :post
end
```

In the example, only the ```title``` attribute would be cached, everything else would be retrieved from the ```Post``` object. ```Except``` would work exactly the opposite.

Despite of being a simple new feature it isn't quite easy and small. So I'm not moving forward with this on PR #693 

I'll try to work on this asap, but would like to hear some thoughts about it.
cc/ @kurko @steveklabnik @nwjsmith

guilleiguaran:
I just want to say that if this is done active_model_serializers will be shipped by default in new Rails 5.0 apps.

joaomdmoura:
@guilleiguaran haha :+1: it will be done!

kurko:
Go go go, @joaomdmoura :grin:

antonioparisi:
:+1:

joaomdmoura:
I'm not sure if I should also enable users to cache relationships (all their attributes). @kurko any thoughts?

joaomdmoura:
@kurko, never mind haha I ended up doing it.
So now you can also specify ```only``` or ```except``` with associations names, it will fragment cache it.

ex. 
```ruby
LikeSerializer = Class.new(ActiveModel::Serializer) do
  cache only: [:post]
  attributes :id, :time

  belongs_to :post
end
``` 

The ```post``` part of the response will be cached, but it won't cache the ```post``` **object**. So, your ```PostSerializer``` will not use this cache because it may not contain exactly what you want.

Will squash, document it and open the PR probably tomorrow.

antonioparisi:
Excellent work!

On Wednesday, 11 February 2015, João Moura <notifications@github.com> wrote:

[...]

PR #810, opened 2015-02-12
Adding Fragment Cache to AMS

joaomdmoura:
It's an upgrade based on the new Cache implementation #693.
Referenced: #802 

It allows to use the Rails conventions to cache specific attributes or associations, as you will see in the example ahead. 

It's optimized to re-use the cache in ever way as possible, in order to do this, the cached data will shared and resued on different methods, also the Serializers that are cached will be used when referenced as relationship at other serializer.

If there is some API endpoint that shouldn't be fully cached, you can still optimise it, using Fragment Cache on the attributes and relationships that you want to cache.

You can define the attributes by using ```only``` or ```except``` option on cache mehtod.

Example:

```ruby
  cache key: 'post', expires_in: 3.hours, only: [:title]
  attributes :title, :body

  has_many :comments

  url :post
end
```

It also work when defining a key to an attribute, in this case you should use the original name of the attribute (no worries, the key still will work):

```ruby
AlternateBlogSerializer = Class.new(ActiveModel::Serializer) do
  cache only: [:name]
  attribute :id
  attribute :name, key: :title
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,10 @@ The options are the same options of ```ActiveSupport::Cache::Store```, plus
 a ```key``` option that will be the prefix of the object cache
 on a pattern ```"#{key}/#{object.id}-#{object.updated_at}"```.
 
+The cache support is optimized to use the cached object in multiple request. An object cached on an ```show``` request will be reused at the ```index```. If there is a relationship with another cached serializer it will also be created and reused automatically.
+
 **[NOTE] Every object is individually cached.**
+
 **[NOTE] The cache is automatically expired after update an object but it's not deleted.**
 
 ```ruby
@@ -294,6 +297,27 @@ end
 On this example every ```Post``` object will be cached with
 the key ```"post/#{post.id}-#{post.updated_at}"```. You can use this key to expire it as you want,
 but in this case it will be automatically expired after 3 hours.
+
+### Fragmenting Caching
+
+If there is some API endpoint that shouldn't be fully cached, you can still optmise it, using Fragment Cache on the attributes and relationships that you want to cache.
+
+You can define the attribute by using ```only``` or ```except``` option on cache method.
+
+**[NOTE] Cache serializers will be used at their relationships**
+
+Example:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  cache key: 'post', expires_in: 3.hours, only: [:title]
+  attributes :title, :body
+
+  has_many :comments
+
+  url :post
+end
+```
 
 ## Getting Help
 
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -1,3 +1,5 @@
+require 'active_model/serializer/adapter/fragment_cache'
+
 module ActiveModel
   class Serializer
     class Adapter
@@ -32,7 +34,37 @@ module ActiveModel
         "ActiveModel::Serializer::Adapter::#{adapter.to_s.classify}".safe_constantize
       end
 
+      def fragment_cache(*args)
+        raise NotImplementedError, 'This is an abstract method. Should be implemented at the concrete adapter.'
+      end
+
       private
+
+     def cache_check(serializer)
+        @cached_serializer = serializer
+        @klass             = @cached_serializer.class
+        if is_cached?
+          @klass._cache.fetch(cache_key, @klass._cache_options) do
+            yield
+          end
+        elsif is_fragment_cached?
+          FragmentCache.new(self, @cached_serializer, @options, @root).fetch
+        else
+          yield
+        end
+      end
+
+      def is_cached?
+        @klass._cache && !@klass._cache_only && !@klass._cache_except
+      end
+
+      def is_fragment_cached?
+        @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
+      end
+
+      def cache_key
+        (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{@cached_serializer.object.updated_at}" : @cached_serializer.object.cache_key
+      end
 
       def meta
         serializer.meta if serializer.respond_to?(:meta)
@@ -49,20 +81,6 @@ module ActiveModel
       def include_meta(json)
         json[meta_key] = meta if meta && root
         json
-      end
-
-      private
-
-      def cached_object
-        klass = serializer.class
-        if klass._cache
-          _cache_key = (klass._cache_key) ? "#{klass._cache_key}/#{serializer.object.id}-#{serializer.object.updated_at}" : serializer.object.cache_key
-          klass._cache.fetch(_cache_key, klass._cache_options) do
-            yield
-          end
-        else
-          yield
-        end
       end
     end
   end
--- a/lib/active_model/serializer/adapter/fragment_cache.rb
+++ b/lib/active_model/serializer/adapter/fragment_cache.rb
@@ -1,0 +1,78 @@
+module ActiveModel
+  class Serializer
+    class Adapter
+      class FragmentCache
+
+        attr_reader :serializer
+
+        def initialize(adapter, serializer, options, root)
+          @root       = root
+          @options    = options
+          @adapter    = adapter
+          @serializer = serializer
+        end
+
+        def fetch
+          klass = serializer.class
+          # It will split the serializer into two, one that will be cached and other wont
+          serializers = fragment_serializer(serializer.object.class.name, klass)
+
+          # Instanciate both serializers
+          cached_serializer     = serializers[:cached].constantize.new(serializer.object)
+          non_cached_serializer = serializers[:non_cached].constantize.new(serializer.object)
+
+          cached_adapter     = @adapter.class.new(cached_serializer, @options)
+          non_cached_adapter = @adapter.class.new(non_cached_serializer, @options)
+
+          # Get serializable hash from both
+          cached_hash     = cached_adapter.serializable_hash
+          non_cached_hash = non_cached_adapter.serializable_hash
+
+          # Merge both results
+          @adapter.fragment_cache(cached_hash, non_cached_hash)
+        end
+
+        private
+
+        def cached_attributes(klass, serializers)
+          cached_attributes     = (klass._cache_only) ? klass._cache_only : serializer.attributes.keys.delete_if {|attr| klass._cache_except.include?(attr) }
+          non_cached_attributes = serializer.attributes.keys.delete_if {|attr| cached_attributes.include?(attr) }
+
+          cached_attributes.each do |attribute|
+            options = serializer.class._attributes_keys[attribute]
+            options ||= {}
+            # Add cached attributes to cached Serializer
+            serializers[:cached].constantize.attribute(attribute, options)
+          end
+
+          non_cached_attributes.each do |attribute|
+            options = serializer.class._attributes_keys[attribute]
+            options ||= {}
+            # Add non-cached attributes to non-cached Serializer
+            serializers[:non_cached].constantize.attribute(attribute, options)
+          end
+        end
+
+        def fragment_serializer(name, klass)
+          cached     = "#{name.capitalize}CachedSerializer"
+          non_cached = "#{name.capitalize}NonCachedSerializer"
+
+          Object.const_set cached, Class.new(ActiveModel::Serializer) unless Object.const_defined?(cached)
+          Object.const_set non_cached, Class.new(ActiveModel::Serializer) unless Object.const_defined?(non_cached)
+
+          klass._cache_options       ||= {}
+          klass._cache_options[:key] = klass._cache_key if klass._cache_key
+
+          cached.constantize.cache(klass._cache_options)
+
+          cached.constantize.fragmented(serializer)
+          non_cached.constantize.fragmented(serializer)
+
+          serializers = {cached: cached, non_cached: non_cached}
+          cached_attributes(klass, serializers)
+          serializers
+        end
+      end
+    end
+  end
+end
--- a/test/adapter/json_api/linked_test.rb
+++ b/test/adapter/json_api/linked_test.rb
@@ -1,11 +1,11 @@
 require 'test_helper'
-
 module ActiveModel
   class Serializer
     class Adapter
       class JsonApi
         class LinkedTest < Minitest::Test
           def setup
+            ActionController::Base.cache_store.clear
             @author1 = Author.new(id: 1, name: 'Steve K.')
             @author2 = Author.new(id: 2, name: 'Tenderlove')
             @bio1 = Bio.new(id: 1, content: 'AMS Contributor')
@@ -103,8 +103,9 @@ module ActiveModel
                   }
                 }, {
                   id: "1",
-                  content: "AMS Contributor",
+                  rating: nil,
                   type: "bios",
+                  content: "AMS Contributor",
                   links: {
                     author: { linkage: { type: "authors", id: "1" } }
                   }
@@ -119,8 +120,9 @@ module ActiveModel
                   }
                 }, {
                   id: "2",
-                  content: "Rails Contributor",
+                  rating: nil,
                   type: "bios",
+                  content: "Rails Contributor",
                   links: {
                     author: { linkage: { type: "authors", id: "2" } }
                   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,3 +4,5 @@
   * adds method to override association [adcb99e, @kurko]
   * adds `has_one` attribute for backwards compatibility [@ggordon]
   * updates JSON API support to RC3 [@mateomurphy]
+  * adds fragment cache support [@joaomdmoura]
+  * adds cache support to attributes and associations [@joaomdmoura]
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -10,41 +10,54 @@ module ActiveModel
 
     class << self
       attr_accessor :_attributes
+      attr_accessor :_attributes_keys
       attr_accessor :_associations
       attr_accessor :_urls
       attr_accessor :_cache
+      attr_accessor :_fragmented
       attr_accessor :_cache_key
+      attr_accessor :_cache_only
+      attr_accessor :_cache_except
       attr_accessor :_cache_options
     end
 
     def self.inherited(base)
       base._attributes = []
+      base._attributes_keys = {}
       base._associations = {}
       base._urls = []
     end
 
     def self.attributes(*attrs)
+      attrs = attrs.first if attrs.first.class == Array
       @_attributes.concat attrs
 
       attrs.each do |attr|
         define_method attr do
           object && object.read_attribute_for_serialization(attr)
-        end unless method_defined?(attr)
+        end unless method_defined?(attr) || _fragmented.respond_to?(attr)
       end
     end
 
     def self.attribute(attr, options = {})
       key = options.fetch(:key, attr)
+      @_attributes_keys[attr] = {key: key} if key != attr
       @_attributes.concat [key]
       define_method key do
         object.read_attribute_for_serialization(attr)
-      end unless method_defined?(key)
+      end unless method_defined?(key) || _fragmented.respond_to?(attr)
+    end
+
+    def self.fragmented(serializer)
+      @_fragmented = serializer
     end
 
     # Enables a serializer to be automatically cached
     def self.cache(options = {})
-      @_cache = ActionController::Base.cache_store if Rails.configuration.action_controller.perform_caching
-      @_cache_key = options.delete(:key)
+      @_cache         = ActionController::Base.cache_store if Rails.configuration.action_controller.perform_caching
+      @_cache_key     = options.delete(:key)
+      @_cache_only    = options.delete(:only)
+      @_cache_except  = options.delete(:except)
       @_cache_options = (options.empty?) ? nil : options
     end
 
@@ -141,12 +154,12 @@ module ActiveModel
     attr_accessor :object, :root, :meta, :meta_key, :scope
 
     def initialize(object, options = {})
-      @object   = object
-      @options  = options
-      @root     = options[:root] || (self.class._root ? self.class.root_name : false)
-      @meta     = options[:meta]
-      @meta_key = options[:meta_key]
-      @scope    = options[:scope]
+      @object     = object
+      @options    = options
+      @root       = options[:root] || (self.class._root ? self.class.root_name : false)
+      @meta       = options[:meta]
+      @meta_key   = options[:meta_key]
+      @scope      = options[:scope]
 
       scope_name = options[:scope_name]
       if scope_name && !respond_to?(scope_name)
@@ -183,22 +196,29 @@ module ActiveModel
       attributes += options[:required_fields] if options[:required_fields]
 
       attributes.each_with_object({}) do |name, hash|
-        hash[name] = send(name)
+        unless self.class._fragmented
+          hash[name] = send(name)
+        else
+          hash[name] = self.class._fragmented.public_send(name)
+        end
       end
     end
 
     def each_association(&block)
       self.class._associations.dup.each do |name, association_options|
         next unless object
-
         association_value = send(name)
 
         serializer_class = ActiveModel::Serializer.serializer_for(association_value, association_options)
 
-        serializer = serializer_class.new(
-          association_value,
-          options.merge(serializer_from_options(association_options))
-        ) if serializer_class
+        if serializer_class
+          serializer = serializer_class.new(
+            association_value,
+            options.merge(serializer_from_options(association_options))
+          )
+        elsif !association_value.nil? && !association_value.instance_of?(Object)
+          association_options[:association_options][:virtual_value] = association_value
+        end
 
         if block_given?
           block.call(name, serializer, association_options[:association_options])
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -1,3 +1,5 @@
+require 'active_model/serializer/adapter/json/fragment_cache'
+
 module ActiveModel
   class Serializer
     class Adapter
@@ -6,30 +8,44 @@ module ActiveModel
           if serializer.respond_to?(:each)
             @result = serializer.map{|s| self.class.new(s).serializable_hash }
           else
-            @result = cached_object do
-              @hash = serializer.attributes(options)
-              serializer.each_association do |name, association, opts|
-                if association.respond_to?(:each)
-                  array_serializer = association
-                  @hash[name] = array_serializer.map { |item| item.attributes(opts) }
-                else
-                  if association
-                    @hash[name] = association.attributes(options)
-                  else
-                    @hash[name] = nil
+            @hash = {}
+
+            @core = cache_check(serializer) do
+              serializer.attributes(options)
+            end
+
+            serializer.each_association do |name, association, opts|
+              if association.respond_to?(:each)
+                array_serializer = association
+                @hash[name] = array_serializer.map do |item|
+                  cache_check(item) do
+                    item.attributes(opts)
                   end
                 end
+              else
+                if association
+                  @hash[name] = cache_check(association) do
+                    association.attributes(options)
+                  end
+                elsif opts[:virtual_value]
+                  @hash[name] = opts[:virtual_value]
+                else
+                  @hash[name] = nil
+                end
               end
-              @hash
             end
+            @result = @core.merge @hash
           end
 
           if root = options.fetch(:root, serializer.json_key)
             @result = { root => @result }
           end
-
           @result
         end
+      end
+
+      def fragment_cache(cached_hash, non_cached_hash)
+        Json::FragmentCache.new().fragment_cache(cached_hash, non_cached_hash)
       end
     end
   end
--- a/lib/active_model/serializer/adapter/json/fragment_cache.rb
+++ b/lib/active_model/serializer/adapter/json/fragment_cache.rb
@@ -1,0 +1,15 @@
+module ActiveModel
+  class Serializer
+    class Adapter
+      class Json < Adapter
+        class FragmentCache
+
+          def fragment_cache(cached_hash, non_cached_hash)
+            non_cached_hash.merge cached_hash
+          end
+
+        end
+      end
+    end
+  end
+end
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -1,3 +1,5 @@
+require 'active_model/serializer/adapter/json_api/fragment_cache'
+
 module ActiveModel
   class Serializer
     class Adapter
@@ -26,13 +28,15 @@ module ActiveModel
               end
             end
           else
-            @hash = cached_object do
-              @hash[:data] = attributes_for_serializer(serializer, @options)
-              add_resource_links(@hash[:data], serializer)
-              @hash
-            end
+            @hash[:data] = attributes_for_serializer(serializer, @options)
+            add_resource_links(@hash[:data], serializer)
           end
           @hash
+        end
+
+        def fragment_cache(cached_hash, non_cached_hash)
+          root = false if @options.include?(:include)
+          JsonApi::FragmentCache.new().fragment_cache(root, cached_hash, non_cached_hash)
         end
 
         private
@@ -43,7 +47,7 @@ module ActiveModel
           resource[:links][name][:linkage] += serializers.map { |serializer| { type: serializer.type, id: serializer.id.to_s } }
         end
 
-        def add_link(resource, name, serializer)
+        def add_link(resource, name, serializer, val=nil)
           resource[:links] ||= {}
           resource[:links][name] = { linkage: nil }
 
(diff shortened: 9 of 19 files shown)


joaomdmoura:
@kurko and it's alive ;) made with :heart: 
Tests passing, Commits Squashed :+1:

README.md
```diff
@@ -294,6 +295,25 @@ On this example every ```Post``` object will be cached with
 the key ```"post/#{post.id}-#{post.updated_at}"```. You can use this key to expire it as you want,
 but in this case it will be automatically expired after 3 hours.
 
+### Fragmenting Caching
+
+If there is some API endpoint that shouldn't be fully cached, you can still optmize it, using Fragment Cache on the attributes and relationships that you want to cache.
```
kurko: Typo, `optime`.

README.md
```diff
@@ -294,6 +295,25 @@ On this example every ```Post``` object will be cached with
...
+
+You can define the attribute or relationships by using ```only``` or ```except``` option on cache mehtod.
```
kurko: Typo, `mehtod`.

lib/active_model/serializer/adapter.rb
```diff
@@ -55,15 +55,97 @@ def include_meta(json)
...
+          serializers[:cached].constantize.attribute(attribute, options)
+        end
+
+        non_cached_attr.each do |attribute|
+          options = @serializer.class._attributes_keys[attribute]
+          options ||= {}
+          serializers[:non_cached].constantize.attribute(attribute, options)
+        end
+        fragment_associations(serializers, associations)
+      end
```
kurko: Could you extract all of these methods into another class? `Cache` perhaps? e.g

```
def cached_hash_root(cached_hash, non_cached_hash)
  cache.hash_root(cached_hash, non_cached_hash)
end

# ...

private

def cache
  Cache.new(self)
end
```

Otherwise, it's going to be hard to be maintained.

joaomdmoura: Totally agree. I'll extract it to an standalone class and include it in the Adapter. If you have another suggestions just let me know

joaomdmoura: Totally agree. I'll extract it to an standalone class and include it in the Adapter. If you have another suggestions just let me know

joaomdmoura:
@kurko, Updated, thank you for the review! Just let me know if you find anything else.

joaomdmoura:
@kurko, Updated, thank you for the review! Just let me know if you find anything else.

lib/active_model/serializer/adapter.rb
```diff
@@ -1,7 +1,10 @@
+require 'active_model/serializer/cache'
+
 module ActiveModel
   class Serializer
     class Adapter
       extend ActiveSupport::Autoload
+      include ActiveModel::Serializer::Cache
```
kurko: So, my suggestion was to really move the code out of the former class. Instead, you just moved it to a new file, but the code is still inside the Adapter.

In my opinion, the main problem with this approach is that you can't unit test the code you added, but only have integration tests, which aren't really good for refactoring the code later. My suggestion was to use composition, or in other words, call the cache class such as:

```ruby
def cached_object
   CachedResource.new(serializer).load
end
```

With that, you can unit test `CachedResource` separately. Besides that, the adapter has a huge amount of methods doing things that are not the adapter's responsibility.

What do you think? cc @guilleiguaran

joaomdmoura: Got it. Indeed. Serialization_test is getting too big.

> Besides that, the adapter has a huge amount of methods doing things that are not the adapter's responsibility.

Convinced me

JoshSmith:
@joaomdmoura what can I do to help here?

thibaudgg:
@joaomdmoura looks awesome. :heart_eyes: 
One question, will [`fetch_multi`](http://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-fetch_multi) be supported in the future? It would be really useful (and faster!) for fetching collection or association from cache in one read. :dash: 
Thanks!

joaomdmoura:
@JoshSmith tks man, but there are just some details missing, I'm working on it already :smile: but there are some improvements that I would like to work on in the near future that you can focus too, like the @thibaudgg suggestion.

@thibaudgg Tks! Unfortunately the new cache feature does not use ```fetch_multi```. But it's a great suggestion! It's more related with my last PR #693 adding the new cache feature, if you read this PR you will realize that I cache each object separately. Indeed change the way the ```Adapter``` works when retrieving an cached collection to use ```fetch_multi``` could be an improvement, maybe also a great opportunity to start benchmarking AMS features.

thibaudgg:
@joaomdmoura `fetch_multi` support would be a big plus against `jbuilder` and `jbuilder_cache_multi`! Thanks.

lib/active_model/serializer/adapter.rb
joaomdmoura: @kurko, to avoid performance loss I'll just remove the fragment cache logic, that correspond for 90% of all code, agreed?:
```ruby
ef cached_object
         klass = serializer.class
        if klass._cache && !klass._cache_only && !klass._cache_except
          _cache_key = (klass._cache_key) ? "#{klass._cache_key}/#{serializer.object.id}-#{serializer.object.updated_at}" : serializer.object.cache_key
          klass._cache.fetch(_cache_key, klass._cache_options) do
            yield
          end

        elsif klass._cache_only && !klass._cache_except || !klass._cache_only && klass._cache_except
          FragmentCache.new(self, serializer).fetch
        else
          yield
        end
      end
```

kurko: Sounds good, that's a good start. I don't think, though, that another class will result in any significant loss.

***

On another not, I think we need to name things better here; we need to better group the code. For instance, I have no idea from a glance what the following is:

```ruby
klass._cache_only && !klass._cache_except || !klass._cache_only && klass._cache_except
```

I'd rather have


```ruby
def cached_object
  # ...
  elsif fragment_cache?
      FragmentCache.new(self, serializer, @options, @root).fetch
  else
  #...
end

private

def fragment_cache?
  klass._cache_only && !klass._cache_except || !klass._cache_only && klass._cache_except
end
```

joaomdmoura: The loss would be result of passing the block correspondent to the yield to order function. But yeah it's nothing big.

---------------------------------

Agreed. Even ```cached_object`` isn't a good name, I'm working on this too.

joaomdmoura:
@kurko, It's done.
- Extracted the fragment logic to a new class ```FragmentCache```
- Refactored some code from #693 
- Wrote some tests related to ```FragmentCache```

JoshSmith:
:beers:

kurko:
Great work here, @joaomdmoura! Before merging, I want to spend more time with it. I'll deploy it to our staging env and see how it behaves and whether I can catch bugs or not. I'd advise you doing the same over there :D

lib/active_model/serializer/fragment_cache.rb
```diff
@@ -0,0 +1,99 @@
...
+        Object.const_set cached, Class.new(ActiveModel::Serializer) unless Object.const_defined?(cached)
+        Object.const_set non_cached, Class.new(ActiveModel::Serializer) unless Object.const_defined?(non_cached)
+
+        klass._cache_options ||= {}
+        klass._cache_options[:key] = klass._cache_key if klass._cache_key
+        cached.constantize.cache(klass._cache_options)
+
+        serializers = {cached: cached, non_cached: non_cached}
+        cached_attributes_and_association(klass, serializers)
+        return serializers
```
kurko: There's no need for this `return`.

lib/active_model/serializer/fragment_cache.rb
```diff
@@ -0,0 +1,99 @@
...
+        else
+          hash = (core_cached[1]) ? core_cached[1].merge(core_non_cached[1]) : core_non_cached[1]
+        end
+        hash.merge no_root_non_cache.merge no_root_cache
+      end
+
+      def fragment_associations(serializers, associations)
+        associations.each do |association|
+          options = ",#{association[1][:association_options]}" if association[1].include?(:association_options)
+          eval("#{serializers[:non_cached]}.#{association[1][:type].to_s}(:#{association[0]}#{options})")
```
kurko: Could you clarify what's going on here? A comment in this case would be helpful, I think, as I can't really understand what's happening.

This `eval` ringed a lot of alarm sounds in my head. What is it doing? Is it really needed?

lib/active_model/serializer/fragment_cache.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        core_cached = cached_hash.first
+        core_non_cached = non_cached_hash.first
+        no_root_cache = cached_hash.delete_if {|key, value| key == core_cached[0] }
+        no_root_non_cache = non_cached_hash.delete_if {|key, value| key == core_non_cached[0] }
+
+        if @root
+          hash[@root] = (core_cached[1]) ? core_cached[1].merge(core_non_cached[1]) : core_non_cached[1]
+        else
+          hash = (core_cached[1]) ? core_cached[1].merge(core_non_cached[1]) : core_non_cached[1]
```
kurko: This line is exactly the same as the one above. Could if be something like:

```ruby
cached_resource = (core_cached[1]) ? core_cached[1].merge(core_non_cached[1]) : core_non_cached[1]
hash = if @root
         { @root => cached_resource }
       else
         cached_resource
       end
```

Which allows you to remove the line 34.

lib/active_model/serializer/fragment_cache.rb
```diff
@@ -0,0 +1,99 @@
...
+        serializers = fragment_serializer(@serializer.object.class.name, klass)
+
+        cached_serializer = serializers[:cached].constantize.new(@serializer.object)
+        non_cached_serializer = serializers[:non_cached].constantize.new(@serializer.object)
+
+        cached_hash = @adapter.class.new(cached_serializer, @options).serializable_hash
+        non_cached_hash = @adapter.class.new(non_cached_serializer, @options).serializable_hash
+
+        if @serializer.root && @adapter.class == ActiveModel::Serializer::Adapter::JsonApi
+          cached_hash_root(cached_hash, non_cached_hash)
```
kurko: If you're adding code specific for `JsonApi`, I think you should push that into the adapter. You could use polymorphism.

You could create `ActiveModel::Serializer::Adapter::FragmentCache` and `ActiveModel::Serializer::Adapter::JsonApi::FragmentCache`. Then here you could just do:

```ruby
adapter.fragment_cache(cached_hash, non_cached_hash)
```

This would also get rid of `cached_hash_root` from this class, specially given it's `JsonApi` specific.

lib/active_model/serializer/fragment_cache.rb
```diff
@@ -0,0 +1,99 @@
...
+        else
+          non_cached_hash.merge cached_hash
+        end
+      end
+
+      private
+
+      def cached_hash_root(cached_hash, non_cached_hash)
```
kurko: Given this method is JsonApi specific, we'd probably have to add a new method for each adapter. That's yet another argument for the polymorphic approach I mentioned above.

Also, it should be named something like `json_api_cache_root` or something like that.

test/adapter/json_api/linked_test.rb
```diff
@@ -44,8 +45,8 @@ def test_include_multiple_posts_and_linked
             @second_comment.post = @first_post
             @second_comment.author = nil
             assert_equal([
-                           { title: "Hello!!", body: "Hello, world!!", id: "1", links: { comments: ['1', '2'], author: "1" } },
-                           { title: "New Post", body: "Body", id: "2", links: { comments: [], :author => "2" } }
+                           { :id=>"1", :title=>"Hello!!", :body=>"Hello, world!!", :links=>{:comments=>["1", "2"], :blog=>"999", :author=>"1"} },
+                           { :id=>"2", :title=>"New Post", :body=>"Body", :links=>{:comments=>[], :blog=>"999", :author=>"2"} }
```
kurko: Can we get rid of the hash-rocket version just for consistency? We have it differently everywhere. Also, it feels so Ruby 1.9 :grin:

lib/active_model/serializer/fragment_cache.rb
joaomdmoura: Yeah, I know, I just felt that I should make it explicitly.
Never mind I'm removing it. :stuck_out_tongue_closed_eyes:

lib/active_model/serializer/fragment_cache.rb
joaomdmoura: Commenting. No there is no need, I just forgot it there while I was reviewing the code, I should have replaced it with ```.send```

lib/active_model/serializer/fragment_cache.rb
joaomdmoura: Agreed, but I transformed it into a ternary. Fewer lines, still understandable. 
```hash = (@root) ? { @root => cached_resource } : cached_resource```

File: lib/active_model/serializer/fragment_cache.rb (hunk 8 above)
Comment (joaomdmoura) [inline lib/active_model/serializer/fragment_cache.rb]: Yeah, you are definitely right, there is no sense to have a method to a specific adapter using ```FragmentCache``` abstract as it is.
This is the only change I haven't worked yet, I'll give it some thought yet. But polymorphism indeed seems a good way to go.

File: lib/active_model/serializer/fragment_cache.rb (hunk 6 above)
Comment (guilleiguaran) [inline lib/active_model/serializer/fragment_cache.rb]: public_send please :)

File: lib/active_model/serializer/fragment_cache.rb (hunk 8 above)
Comment (kurko) [inline lib/active_model/serializer/fragment_cache.rb]: Ping me if you need help with it.

File: lib/active_model/serializer/adapter.rb [hunk 11]
```diff
@@ -32,8 +35,36 @@ def self.adapter_class(adapter)
         "ActiveModel::Serializer::Adapter::#{adapter.to_s.classify}".safe_constantize
       end
 
+      def fragment_cache(*args)
```
joaomdmoura: @kurko abstracted method. Duck typing strategy.

lib/active_model/serializer/adapter.rb
```diff
@@ -32,8 +35,36 @@ def self.adapter_class(adapter)
...
+        raise NotImplementedError, 'This is an abstract method. Should be implemented at the concrete adapter.'
+      end
+
       private
 
+      def cache_check
```
joaomdmoura: Updated the old ```cache_object``` to ```cache_check```.
Also abstracted some of its behaviours into other methods. (bellow)

lib/active_model/serializer/adapter/fragment_cache.rb
```diff
@@ -0,0 +1,90 @@
+module ActiveModel
```
joaomdmoura: I commented everything that could make it easy to support.

lib/active_model/serializer/adapter/fragment_cache.rb
```diff
@@ -0,0 +1,90 @@
...
+          # Instanciate both serializers
+          cached_serializer     = serializers[:cached].constantize.new(@serializer.object)
+          non_cached_serializer = serializers[:non_cached].constantize.new(@serializer.object)
+
+          # Get serializable hash from both
+          cached_hash     = @adapter.class.new(cached_serializer, @options).serializable_hash
+          non_cached_hash = @adapter.class.new(non_cached_serializer, @options).serializable_hash
+
+          # Merge both results
+          @adapter.fragment_cache(cached_hash, non_cached_hash)
```
joaomdmoura: The duck typing implementation that was defined at ```adapter.rb``` similar to ```serializable_hash```

lib/active_model/serializer/adapter/fragment_cache.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        # Add associations to non-cached Serializer
+        def fragment_associations(serializers, cached_attr, non_cached_attr)
+          associations            = @serializer.each_association
+          cached_associations     = associations.select {|k,v| cached_attr.include?(k)}
+          non_cached_associations = associations.select {|k,v| !cached_attr.include?(k)}
+
+          # Add cached associations to cached Serializer
+          cached_associations.each do |k,v|
+            serializers[:cached].constantize.public_send(v[:type], k, v[:association_options])
```
joaomdmoura: Using public_send @guilleiguaran :smile: thank you for point it out!

joaomdmoura:
Rebased with :heart: PR fully updated.

kurko:
Talking to @joaomdmoura, we decided that it's better if we start by making the cache separate by resource. Right now, it's caching the entire response, including associations.

Say we have `posts` with `comments` and you want to update `comments`. Currently, you'll have to manually invalidate the respective `post` cache. You'll have to write code outside AMS to invalidate that. We want to start simple.

With the new decision, the main resource (`post`) will be cached separately from each `comment`. Then, when a comment is updated, AMS should load the new cache automatically. That will mean we'll have to do some processing to fetch caches, but will be simpler to start with.

Although the current way is faster, it'll be much more complicated for end users. We'll start with the simpler approach, which should be fast enough for most applications, and then move to the integral cache, which will require us to put some effort in better documenting it.

thibaudgg:
@kurko sounds great!

joaomdmoura:
Finally. It's done! @kurko 
Updated the implementation as we have being discussing, added some new tests too.
It's rebased and squashed :smile:

JoshSmith:
@joaomdmoura  :muscle: :heart_eyes:

thibaudgg:
:+1:

brancusi:
Nice!

kurko:
I tried this in an app and couldn't make it work. Basically, when I add `cache only: [:name]` to my serializer, I get this:

```
  1) Error:
UsersTest::GET /users/{id}#test_0001_returns a user by id:
NoMethodError: undefined method `my_virtual_attribute' for #<User:0x0000011126d1c8>
    app/controllers/users_controller.rb:6:in `show'
    test/test_helper.rb:50:in `get'
    my_app_test_path_test.rb:346:in `block (4 levels) in <class:UsersTest>'
```

My serializer is like this:

```
  cache only: [ :name ]

  attributes :id,
             :name,
             :my_virtual_attribute

  def my_virtual_attribute
    :some_custom_value
  end
end
```

If I leave `cache` out, it works fine. I tested against f7fc56f05184. For your reference, the PR that adds support for virtual associations (not attributes) is https://github.com/rails-api/active_model_serializers/pull/792.

When fixing, please make sure that both virtual attributes and associations are working.

Great work, by the way.

kurko:
Needs to be rebased 😬

joaomdmoura:
@kurko fixed, new tests written, rebased and squashed. :smile:

kurko:
FYI, I'm trying to backport this PR into a fork that supports RC1 so I can test this in my app. If it works there, I'll merge it. This is so central to AMS; it has to work perfectly.

lib/active_model/serializer/adapter.rb
```diff
@@ -32,8 +34,38 @@ def self.adapter_class(adapter)
...
       end
 
+      def fragment_cache(*args)
+        raise NotImplementedError, 'This is an abstract method. Should be implemented at the concrete adapter.'
+      end
+
       private
 
+      def cache_check(serializer)
+        @serializer = serializer
```
kurko: This mutation is breaking a test in https://github.com/ride/active_model_serializers/blob/master-rc1-with-fragment-cache/test/action_controller/json_api_linked_test.rb#L146-L150 (it's a branch which I'm backporting this into RC1 because that's what my company is using).

Basically, halfway through the serialization (within a execution of `serializable_hash`), `@serializer`, which was `PostSerializer`, becomes `AuthorSerializer` after this. That same test is passing in this PR, so I'm not really sure what's going on, but regardless of that, mutating a variable that's used to figure out how to serialize stuff is bad and can lead to nasty behaviors in the future. If we're going to mutate, we need to simply instantiate a new object then or make sure the serialization has finished.

@joaomdmoura if you're available, let's discuss this over chat :)

kurko:
@joaomdmoura first of all, great work here! This got better and better every iterations. I tried this in my app and tests are green :grin: 

However, I need to show you something. I'm probably doing something wrong, but I ran some tests and got this:

```
# app/serializers/users.rb
class UserSerializer < ActiveModel::Serializers
  cache only: [
      :name, :email, :primary_role, :secondary_roles, 
      :mobile_number, :date_of_birth
    ]

  # ...
end

# test/integration/users_test.rb
Benchmark.bm do |x|
  x.report do
    1000.times do
      get "/users/#{user.id}", nil, oauth_headers(:route_matching)
    end
  end
end

Results:

               user        system    total     real
no cache       21.550000   1.820000  23.370000 ( 28.894494)
cache          22.270000   1.910000  24.180000 ( 30.448672)
fragment cache 32.320000   2.210000  34.530000 ( 43.051876)

(cache means `only: []` wasn't used in the serializer)
```

Do you have any idea why this is?

joaomdmoura:
Great @kurko 
I'm really glad we pulled it off.
The benchmark issue ( #832 ) now would be awesome, btw I'm already working on it.
The cache obviously improved overall performance. 

I'm sad about the fragment cache result, but as we discussed it would be really useful for attributes that involves some kind of logic, as virtual attributes on the serializer itself. Would be great to benchmark this case too :smile: 

But there is some improvement we can work on to optimize fragment cache speed, once benchmark tests are done will be way easier to keep track of it.

kurko:
Awesome. I'm testing it and if I don't find issues in a day or two, I'll merge this :smile:

Catalectic:
Any update on this? Really looking forward to this :)

kurko:
So, does any one have any problems with this? If not, I'm merging this pretty soon (likely tomorrow).

guilleiguaran:
@kurko can you paste the code you used for benchmarks in a gist?

kurko:
```ruby
# both fragment
cache only: [:field1, field2, :etc]

# and not fragment
cache

# then, in an integration test:
Benchmark.bm do |x|
  x.report do
    1000.times do
      get "/users/#{user.id}", nil, oauth_headers(:route_matching)
    end
  end
end
```

joaomdmoura:
\o/ :laughing:

lib/active_model/serializer/adapter/fragment_cache.rb
```diff
@@ -0,0 +1,78 @@
...
+          non_cached_attributes.each do |attribute|
+            options = serializer.class._attributes_keys[attribute]
+            options ||= {}
+            # Add non-cached attributes to non-cached Serializer
+            serializers[:non_cached].constantize.attribute(attribute, options)
+          end
+        end
+
+        def fragment_serializer(name, klass)
+          cached     = "#{name.capitalize}CachedSerializer"
```
mib32: If different serializers share the same model, but has totally different attributes, would their caches collide, even if the key is different? I believe i see it on my machine. If you believe this could happen, i can present those tests to you, but i need time to set it up.

maybe it should go something like "#{serializer.name}CachedSerializer"?

joaomdmoura: Yeap, I think it could happen. Not the cache_key itself, because it takes the file path into account but the Class name would colide.
I'll check it on the master codebase, create a new issue for that and update it.

mib32: @joaomdmoura thanks, that would be great! 
Here is the stack trace of the error i've got, when the classes collide, in case you would like to look at it. The latter called serializer tries to look for an attribute that was held by the first serializer, but not in the latter. Am i abusing the way serializers should be used, if i'm the first one who's got that error?

```
Failure/Error: get '/markers/local.json?coordinates=55.75,36.00&limit=10'
     NoMethodError:
       undefined method `name' for #<MarkerDistanceSerializer:0x007f80339b6eb8>
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer.rb:155:in `public_send'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer.rb:155:in `block in attributes'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer.rb:151:in `each'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer.rb:151:in `each_with_object'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer.rb:151:in `attributes'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/json.rb:15:in `block in serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter.rb:51:in `block in cache_check'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/cache.rb:299:in `block in fetch'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/cache.rb:585:in `block in save_block_result_to_cache'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/cache.rb:547:in `block in instrument'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/notifications.rb:166:in `instrument'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/cache.rb:547:in `instrument'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/cache.rb:584:in `save_block_result_to_cache'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/cache.rb:299:in `fetch'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter.rb:50:in `cache_check'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/json.rb:14:in `serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/flatten_json.rb:6:in `serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/fragment_cache.rb:27:in `fetch'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter.rb:54:in `cache_check'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/json.rb:14:in `serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/flatten_json.rb:6:in `serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/json.rb:10:in `block in serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/array_serializer.rb:6:in `each'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/array_serializer.rb:6:in `each'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/json.rb:10:in `map'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/json.rb:10:in `serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter/flatten_json.rb:6:in `serializable_hash'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/active_model/serializer/adapter.rb:24:in `as_json'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/json/encoding.rb:35:in `encode'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/json/encoding.rb:22:in `encode'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/core_ext/object/json.rb:37:in `to_json_with_active_support_encoder'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/renderers.rb:116:in `block in <module:Renderers>'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/bundler/gems/active_model_serializers-e7d3323d2352/lib/action_controller/serialization.rb:51:in `block (2 levels) in <module:Serialization>'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/renderers.rb:45:in `block in _render_to_body_with_renderer'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/renderers.rb:41:in `_render_to_body_with_renderer'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/renderers.rb:37:in `render_to_body'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/abstract_controller/rendering.rb:25:in `render'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/rendering.rb:16:in `render'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/instrumentation.rb:44:in `block (2 levels) in render'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/core_ext/benchmark.rb:12:in `block in ms'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/core_ext/benchmark.rb:12:in `ms'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/instrumentation.rb:44:in `block in render'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/instrumentation.rb:87:in `cleanup_view_runtime'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/activerecord-4.2.3/lib/active_record/railties/controller_runtime.rb:25:in `cleanup_view_runtime'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/actionpack-4.2.3/lib/action_controller/metal/instrumentation.rb:43:in `render'
     # /Users/antonmurygin/.rvm/gems/ruby-2.2.2/gems/remotipart-1.2.1/lib/remotipart/render_overrides.rb:14:in `render_with_remotipart'
     # ./app/controllers/markers_controller.rb:20:in `block (2 levels) in local'
```

bf4:
@kurko I'm trying to write a benchmark with positive results for caching but haven't been able to show differences with it on or off.  I'm probably missing something.  I see you had a working benchmark at one point. help?

https://github.com/rails-api/active_model_serializers/pull/810#issuecomment-89870165

```plain
Update: here are the numbers I got:

               user        system    total     real
no cache       21.550000   1.820000  23.370000 ( 28.894494)
cache          16.870000   1.580000  18.450000 ( 21.429540)
fragment cache 22.270000   1.810000  24.080000 ( 28.504920)

(cache means `only: []` wasn't used in the serializer)
```

https://github.com/rails-api/active_model_serializers/pull/810#issuecomment-94940858

```plain
# both fragment
cache only: [:field1, :field2, :etc]

# and not fragment
cache

# then, in an integration test:
Benchmark.bm do |x|
  x.report do
    1000.times do
      get "/users/#{user.id}", nil
    end
  end
end
```

beauby:
@bf4 Any change when changing [this line](https://github.com/rails-api/active_model_serializers/pull/810/files#diff-5624f9cf35be4dc254412d979fd293c4R48) from
```ruby
end unless method_defined?(key) || _fragmented.respond_to?(attr)
```
to
```ruby
end unless method_defined?(key) || _fragmented.respond_to?(key)
```
?

bf4:
@beauby The number in the comment https://github.com/rails-api/active_model_serializers/pull/810#issuecomment-165063161 were a quote of @kurko .  I haven't been able to replicate much.. see #1393

PR #811, opened 2015-02-12
Reimplement serialization scope and scope_name

mateomurphy:
I've more or less directly dropped in the code, and I'm open to discussing alternate approaches; I will provide tests once the direction is settled.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -8,6 +8,16 @@ module ActionController
 
     ADAPTER_OPTION_KEYS = [:include, :fields, :root, :adapter]
 
+    included do
+      class_attribute :_serialization_scope
+      self._serialization_scope = :current_user
+    end
+
+    def serialization_scope
+      send(_serialization_scope) if _serialization_scope &&
+        respond_to?(_serialization_scope, true)
+    end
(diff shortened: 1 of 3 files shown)


JoshSmith:
This looks pretty straightforward to me.

@kurko what do you think?

JoshSmith:
This is probably the only thing keeping me from upgrading to AMS master.

I'd like thoughts from @kurko or @guilleiguaran or @joaomdmoura before just giving this the :+1:

lib/action_controller/serialization.rb
```diff
@@ -8,6 +8,16 @@ module Serialization
 
     ADAPTER_OPTION_KEYS = [:include, :fields, :root, :adapter]
 
+    included do
+      class_attribute :_serialization_scope
+      self._serialization_scope = :current_user
+    end
+
+    def serialization_scope
+      send(_serialization_scope) if _serialization_scope &&
```
joaomdmoura: if possible, use public_send instead :smile:

JoshSmith:
Thanks for help @joaomdmoura :beers:

joaomdmoura:
No worries. You can count on me :beers:

PR #812, opened 2015-02-13
Returning DefaultSerializer when nil is passed to serializer_for

nstruthers:
This avoids a costly `_const_get` search for a non-existent `NilClassSerializer`. We hit a performance issue when serializing thousands of records that had a `has_one` association that could potentially be nil--the `_const_get` search was responsible for about 26 seconds of a 28 second `to_json` call. It can be fixed by adding an empty `NilClassSerializer`, but this seems like the better fix.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -56,7 +56,9 @@ end
       attr_reader :key_format
 
       def serializer_for(resource, options = {})
-        if resource.respond_to?(:to_ary)
+        if resource.nil?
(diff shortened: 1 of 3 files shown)


JoshSmith:
Can you add a test for this?

nstruthers:
No problem, will do.

JoshSmith:
Thanks!

nstruthers:
Ok--I've added the test to ensure that `build_serializer` returns a DefaultSerializer when the association is nil.

PR #813, opened 2015-02-14
Allow to define custom serializer for given class

jtomaszewski:
.. by defining #serializer_class method in serialized object's class. Resolves #515.

This is a really small fix and has a test - so I don't see any reasons why it shouldn't be merged.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -56,7 +56,9 @@ end
       attr_reader :key_format
 
       def serializer_for(resource, options = {})
-        if resource.respond_to?(:to_ary)
+        if resource.respond_to?(:serializer_class)
+          resource.serializer_class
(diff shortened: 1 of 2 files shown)


JoshSmith:
I say :+1:

lib/active_model/serializer.rb
```diff
@@ -56,7 +56,9 @@ def format_keys(format)
       attr_reader :key_format
 
       def serializer_for(resource, options = {})
-        if resource.respond_to?(:to_ary)
+        if resource.respond_to?(:serializer_class)
+            resource.serializer_class
```
kurko: We're using 2 spaces indentation.

jtomaszewski: Fixed (force pushed a reworked commit)

PR #822, opened 2015-02-25
Replace has_one with attribute in template

bf4:
has_one is no longer supported

this way we generate valid templates

### DIFF
--- a/lib/generators/serializer/templates/serializer.rb
+++ b/lib/generators/serializer/templates/serializer.rb
@@ -3,6 +3,6 @@ class <%= class_name %>Serializer < <%= parent_class_name %>
   attributes <%= attributes_names.map(&:inspect).join(", ") %>
 end
 <% association_names.each do |attribute| -%>
-  has_one :<%= attribute %>
+  attribute :<%= attribute %>
 <% end -%>
 <% end -%>
--- a/test/serializers/generators_test.rb
+++ b/test/serializers/generators_test.rb
@@ -46,7 +46,7 @@ class SerializerGeneratorTest < Rails::Generators::TestCase
     run_generator
     assert_file "app/serializers/account_serializer.rb" do |serializer|
       assert_match(/^  attributes :id, :name, :description$/, serializer)
-      assert_match(/^  has_one :business$/, serializer)
+      assert_match(/^  attribute :business$/, serializer)
     end
   end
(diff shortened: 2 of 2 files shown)


lib/generators/serializer/templates/serializer.rb
```diff
@@ -3,6 +3,6 @@ class <%= class_name %>Serializer < <%= parent_class_name %>
   attributes <%= attributes_names.map(&:inspect).join(", ") %>
 end
 <% association_names.each do |attribute| -%>
-  has_one :<%= attribute %>
+  attribute :<%= attribute %>
```
kurko: I believe it should be `belongs_to`, not `attribute`.

bf4: If you say so.  There's basically no usage tests around this code.

The generator defines [association_names](https://github.com/rails-api/active_model_serializers/blob/a8243761fedbff9877a957fe9b5f21e644bb8fee/lib/generators/serializer/serializer_generator.rb#L21) as

```ruby
def association_names
  attributes.select { |attr| attr.reference? }.map { |a| a.name.to_sym }
end
```

[belongs_to](https://github.com/rails-api/active_model_serializers/blob/a8243761fedbff9877a957fe9b5f21e644bb8fee/lib/active_model/serializer.rb#L24)

```ruby
    # Defines an association in the object that should be rendered.
    #
    # The serializer object should implement the association name
    # as a method which should return an object when invoked. If a method
    # with the association name does not exist, the association name is
    # dispatched to the serialized object.
    def self.belongs_to(*attrs)
      associate(:belongs_to, attrs)
    end
```

Whereas attribute

```ruby
    def self.attribute(attr, options = {})
      @_attributes.concat [key]
      define_method key do
      end unless method_defined?(key)
    end
```

I'm not really sure where `references?` comes from. Seems to be a schema definition for a `has_many`, no?

kurko:
Nevermind. I misread this PR before.

bf4:
Thanks!

bf4:
@kurko Any desire to revert this since was re-added in https://github.com/rails-api/active_model_serializers/pull/725 ? I'm happy to submit a PR.

kurko:
Yes, definitely.

bf4:
Done

Issue #832, opened 2015-03-09
AMS Benchmark tests


joaomdmoura:
AMS represents a crucial part at an application and it can impact performance (good or badly).
The goal is to make AMS an tool that not only helps you to bring convention over configuration to your JSON generation, but also give you the tools that might help your apps performance.

In order to do this would be great to have some benchmak test that help us to evaluate how much a feature / bug_fix / refactoring impact overall performance. It would be great to keep track of the difference between versions, and help to convince people to update their AMS.

I'm not so sure how would be the best way to do this, I'm thinking about how would be possible to add it to the workflow with our already existing tests. Before start something would be great to hear some thoughts about the idea, if it's valid or not, and how to implement it.

JoshSmith:
I think it's a great idea, especially after seeing this last week: https://is-ember-fast-yet.firebaseapp.com/

I don't feel one way or another about the implementation of it, but I think in general the concept is a good one. :+1:

joaomdmoura:
Great! @kurko also liked the concept. I'll put it in my queue :smile: 
btw #833 is exactly the kind of thing that would make a benchmark suite amazing

JoshSmith:
That's a great example @joaomdmoura.

What are your thoughts about how to implement this?

joaomdmoura:
Right now I've 2 different implementation on my mind.

1. Use a new file to keep history and ```Benchmark``` module ([link](http://ruby-doc.org/stdlib-2.2.0/libdoc/benchmark/rdoc/Benchmark.html)). This implementation will take into account the code it self. 

2. Use and external service like [Loader.io](http://loader.io/), that is freemium and simulate real requests.

I'm tending to the first option, but either way, I'm thinking about move this to a new OS project. A gem that helps you to built a benchmark test suite. Then add it as dependency to AMS.

What do you think about it @JoshSmith ?

JoshSmith:
I would prefer the first option, I think, (or something like it) since that seems closer to our code and doesn't rely on any external service.

Have you also seen https://github.com/rails/rails-perftest?

joaomdmoura:
Agreed. I didn't knew it, but seem great. I'll check it out to see if we can use it or at least have some ideas.

remear:
The benchmark foundation has been added. Benchmarking work will be ongoing and is currently being tracked over in #1575

PR #833, opened 2015-03-10
Cache serializers for class

lsylvester:
Currently, each time a model is serializer is tries to look up the Serializer class using `safe_constantize`. This can be slow when serializing a large number of models.

This adds a cache to return the serializer class for each class, which reduced the amount of time looking up the serializer classes.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -199,11 +199,11 @@ module ActiveModel
       opts
     end
 
-    private
-
     def self.serializers_cache
-      @serializers_cache ||= Threadsafe::Cache.new
+      @serializers_cache ||= ThreadSafe::Cache.new
     end
+
+    private
 
     def self.get_serializer_for(klass)
       serializers_cache.fetch_or_store(klass) do
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -1,3 +1,5 @@
+require 'thread_safe'
+
 module ActiveModel
   class Serializer
     extend ActiveSupport::Autoload
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -201,14 +201,20 @@ module ActiveModel
 
     private
 
-    def self.get_serializer_for(klass)
-      serializer_class_name = "#{klass.name}Serializer"
-      serializer_class = serializer_class_name.safe_constantize
+    def self.serializers_cache
+      @serializers_cache ||= Threadsafe::Cache.new
+    end
 
-      if serializer_class
-        serializer_class
-      elsif klass.superclass
-        get_serializer_for(klass.superclass)
+    def self.get_serializer_for(klass)
+      serializers_cache.fetch_or_store(klass) do
+        serializer_class_name = "#{klass.name}Serializer"
+        serializer_class = serializer_class_name.safe_constantize
+
+        if serializer_class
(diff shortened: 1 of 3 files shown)


joaomdmoura:
:+1: It seems good to me. @JoshSmith ?

lib/active_model/serializer.rb
```diff
@@ -199,16 +199,22 @@ def serializer_from_options(options)
       opts
     end
 
+    def self.serializers_cache
+      @serializers_cache ||= ThreadSafe::Cache.new
```
JoshSmith: I'm not very familiar with this. Can you explain what this does and whether there are any potential side effects?

joaomdmoura: btw it would generate a new dependency with ```thread_safe``` gem right?
Is there some reason to not use ```ActionController::Base.cache_store```

JoshSmith: I think that's correct, and I see no reference to `ThreadSafe` elsewhere. That concerns me a little.

lsylvester: `Threadsafe` is a dependancy of `ActiveSupport`, so it is already indirectly as dependancy. I can add it is explicitly as a dependancy.

The reason that to use it instead of `ActionController::Base.cache_store` is that is in purely in memory so is faster than going to an external cache like memcached which `ActionController::Base.cache_store` would do.

joaomdmoura: Got it. Well, it seem cool for me as we are talking about small things.
Can it cause some problem to not-quit-amazing-servers with less memory ir is it derisive?

JoshSmith: I share @joaomdmoura's concern about low memory servers (being as I'm on a `1x` Heroku dyno). How much memory are we talking?

Yet another reason why #832 is a great idea.

lsylvester: I don't think that it will cause any memory issues. It is only storing mappings between existing model classes and serializer classes, and keeping the those classes in memory is going to use more memory that the mappings. 

Rails already uses `ThreadSafe::Cache`s to store some similar things like constant lookups and controller names.

kurko:
Summoning @guilleiguaran.

lib/active_model/serializer.rb
joaomdmoura: Indeed ```ActiveSupport::TimeZone.zones_map``` in Rails 4.1 is using ```ThreadSafe::Cache```, as you can see [here]9https://github.com/rails/rails/blob/8764ef95dce3cbd0a58186e3cc1da90cc07862eb/activesupport/lib/active_support/values/time_zone.rb#L189).
Okay, I think it's enough to make me feel comfortable about it. cc/ @kurko @JoshSmith @guilleiguaran

JoshSmith: I agree then, but really appreciate the discussion on it. Thanks for obliging us @lsylvester! :trophy:

kurko: Could you remove the hardcoded dependency? That way, if Rails removes it AMS tests will explode and we'll me notified that we need to do something about it (either leave it or replace it, in case there is a better alternative).

lsylvester: @kurko the dependency has been removed.

joaomdmoura: :+1: great work @lsylvester 
Thank you!

PR #836, opened 2015-03-11
Makes passed in options accessible inside serializers

kurko:
In some cases, we want to pass arguments from the controller and we want
to serializer a resource according to that. This allows serializers to
use the `options` method to retrieve whatever was passed in via
arguments.


nhocki:
:heart_eyes: :+1:

philduffy:
:+1:

joaomdmoura:
It seems cool to me! :+1:

PR #840, opened 2015-03-13
Add url option to assocation for jsonapi

### LINKED ISSUE #834: JSONAPI resource relation urls
Currently, the jsonapi adapter adds resource relationships as link objects with ids and types. The jsonapi spec also allows the the resource relationship to be defined as a URL (see http://jsonapi.org/format/#document-structure-resource-relationships). For example:

```json
{
  "type": "articles",
  "id": "1",
  "links": {
    "comments": "http://example.com/articles/1/comments"
  }
}
```
There is currently no way to do this using ActiveModel::Serializers as far as I can tell. 

I propose that there should be a `:url` option for the assoications methods `has_many` and `belongs_to` that can be used to generate an resource url instead of linked objects, so that the above resource could be generated by the following serializer:

```ruby
class ArticleSerializer
  has_many :comments, url: [:article, :comments]
end
```

I am not sure if this option is useful for the other serializers, so feedback on this would be appreciated.

If you are happy with this suggestion I can work on the implementation for this.

lsylvester:
This adds the url option to associations to tell the jsonapi adapter to use a url instead of a link object. 

The adapter assumes that the correct url can be generated using `named_routes`.

In order to generate the urls, the adapter is given the controller instance to use to generate the urls. Let me know if I should be exposing url generation abilities is some other way.

Fixes #834

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -42,7 +42,7 @@ module ActionController
 
           @_serializer_opts[:scope] ||= serialization_scope
           @_serializer_opts[:scope_name] = _serialization_scope
-
+          @_adapter_opts[:url_helper] = self
           # omg hax
           object = serializer.new(resource, @_serializer_opts)
           adapter = ActiveModel::Serializer::Adapter.create(object, @_adapter_opts)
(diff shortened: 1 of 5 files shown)


lsylvester:
This may need to change depending on feedback for json-api/json-api#453

joaomdmoura:
FYI
The tests are not actually broken.
It's and inconsistence related to to cache feature at Travis. I've discussed it with @kurko and I am already looking into it.

PR #841, opened 2015-03-13
Fix issue with embedding multiple associations under the same root key

antstorm:
When trying to embed multiple associations under the same root key AMS will crash trying to call `has_key?` on `Array`, because current merge conflict resolving strategy will return an `Array` instead of `Hash`.

Unfortunately can not use `deep_merge` here, because of rails 3.x and it's old `deep_merge` implementation (doesn't accept block).

A test reproducing the issue is included.

I'm also successfully using this solution on a project that relies heavily on JSON API and is properly covered with tests.


kurko:
Looks good to me.

PR #849, opened 2015-03-19
Implement a method cache for faster dispatch and allowing super calls

byroot:
The code is heavily borrowed from ActiveRecord.

It should make the attribute methods faster even though I didn't write any benchmark to back that claim.

The real reason I implemented this, is to be able to use `super` in attribute methods:

```ruby
  attributes :name

  def name
    super.titleize
  end
end
```

Regards.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -6,10 +6,11 @@ module ActiveModel
     autoload :Configuration
     autoload :ArraySerializer
     autoload :Adapter
+    autoload :AttributeMethods
     include Configuration
+    include Serializer::AttributeMethods
 
     class << self
-      attr_accessor :_attributes
       attr_accessor :_associations
       attr_accessor :_urls
       attr_accessor :_cache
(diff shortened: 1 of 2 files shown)


kurko:
This is interesting. @guilleiguaran @joaomdmoura can you chime in?

byroot:
Oh I forgot to mention. This technique only works for ruby 2.0+. ActiveRecord up to 4.2 had a fallback. I don't know if it's worth implementing it since 1.9.3 is EOL. But I can do it if it's important.

bf4:
@byroot We've dropping support for Ruby < 2.0 in master, so this is again viable.  Would you be interested in rebasing this off of current master? Let me know if I can help.

byroot:
I'll try to find some time.

bf4:
Awesome!

B mobile phone

byroot:
So I looked into it, and given the way it's implemented now with `AttributeReader` etc, this optimization wouldn't help as much. I actually think it would be worse.

The idea was to generate attributes accessor dynamically via `class_eval`, but to share cache them in a module.

If we were doing that on AttributeReader, it would generate a `singleton_class` for every attributes, which is quite slow IIRC.

bf4:
@byroot Oh, but this PR is great! the attributes internal api is still in flux.  See e.g. https://github.com/rails-api/active_model_serializers/pull/1370  I hope there's still something we get improve (would need a benchmark, of course, likehttps://github.com/rails-api/active_model_serializers/pull/233

PR #850, opened 2015-03-20
Use association value for determining serializer used

mateomurphy:
Ensures overridden association value works when orignal association does not return a result.


JoshSmith:
I just ran into this myself, and I say :+1:

PR #852, opened 2015-03-20
Fix options merge order in `each_association`

mateomurphy:
Custom association serializers were getting clobbered when using an each serializer.


JoshSmith:
:+1:

Issue #829, opened 2015-03-08
JSONAPI: support RC3 features


kurko:
- [ ] new structure with `data`
    - [ ] include `type` in every resource
- [ ] rename `linked` to `included`

joaomdmoura:
:+1: @kurko let me know if you need any help with this :smile:

steveklabnik:
We just labeled an RC3, and are gonna stop changing so that you all can implement stuff :heart: 


https://github.com/json-api/json-api/issues/484

zekefast:
Guys, do you need any help with that? Please, let me know! I want to faciliate development of 0.10.x and json-api standard support.

kurko:
I won't have time to work on this until mid april. @zekefast do you have some time to craft a PR? I have time to review 😄

zekefast:
@kurko, Yes, I do.

So, we need look into the difference of current implementation and RC3 and fill it right?
Do you have any info about currently implemented version of JSON-API standard?

mateomurphy:
I'm currently implementing these changes for my own use, I can also submit a PR when I'm done

kurko:
@zekefast current master is supporting RC1. We have to move into using `data` as root key. @mateomurphy can you coordinate with @zekefast :D

mateomurphy:
@zekefast do you have anything done already, or when were you planning on working on this? I've started already as our app needs to support RC3.

jasonkriss:
I haven't started any work on this yet, but it's something that I'd love to land asap. I'd be happy to help. @mateomurphy is there a way we could see what you've done so far?

mateomurphy:
I can put together a PR tomorrow

jasonkriss:
That'd be great. Thanks @mateomurphy!

zekefast:
@mateomurphy, I haven't done anything yet. I can look into your PR today and we can then coordinate work further. Please, let me know what's lack in your PR and what part of spec is missing in it.

Thanks!

mateomurphy:
Mostly done: https://github.com/rails-api/active_model_serializers/pull/853 the main thing remaining are the URLs in links

kurko:
yay closing

zekefast:
@mateomurphy, @kurko thank you guys!

PR #853, opened 2015-03-20
RC3 Updates for JSON API

mateomurphy:
* Implements latest structure (`data` root, `included`, and linkages)
* id and type are present on all resources, the latter is named consistently

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -164,6 +164,14 @@ module ActiveModel
       end
     end
 
+    def id
+      object.id if object
+    end
+
+    def type
+      object.class.to_s.demodulize.underscore.pluralize
+    end
+
     def attributes(options = {})
       attributes =
         if options[:fields]
@@ -171,6 +179,8 @@ module ActiveModel
         else
           self.class._attributes.dup
         end
+
+      attributes += options[:required_fields] if options[:required_fields]
 
       attributes.each_with_object({}) do |name, hash|
         hash[name] = send(name)
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -16,16 +16,14 @@ module ActiveModel
         end
 
         def serializable_hash(options = {})
-          @root = :data
-
           if serializer.respond_to?(:each)
-            @hash[@root] = serializer.map do |s|
-              self.class.new(s, @options.merge(top: @top, fieldset: @fieldset)).serializable_hash[@root]
+            @hash[:data] = serializer.map do |s|
+              self.class.new(s, @options.merge(top: @top, fieldset: @fieldset)).serializable_hash[:data]
             end
           else
             @hash = cached_object do
-              @hash[@root] = attributes_for_serializer(serializer, @options)
-              add_resource_links(@hash[@root], serializer)
+              @hash[:data] = attributes_for_serializer(serializer, @options)
+              add_resource_links(@hash[:data], serializer)
               @hash
             end
           end
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -35,11 +35,9 @@ module ActiveModel
         private
 
         def add_links(resource, name, serializers)
-          type = serialized_object_type(serializers)
           resource[:links] ||= {}
-
           resource[:links][name] ||= { linkage: [] }
-          resource[:links][name][:linkage] += serializers.map { |serializer| { type: type, id: serializer.id.to_s } }
+          resource[:links][name][:linkage] += serializers.map { |serializer| { type: serializer.type, id: serializer.id.to_s } }
         end
 
         def add_link(resource, name, serializer)
@@ -47,9 +45,7 @@ module ActiveModel
           resource[:links][name] = { linkage: nil }
 
           if serializer && serializer.object
-            type = serialized_object_type(serializer)
-
-            resource[:links][name][:linkage] = { type: type, id: serializer.id.to_s }
+            resource[:links][name][:linkage] = { type: serializer.type, id: serializer.id.to_s }
           end
         end
 
@@ -58,17 +54,15 @@ module ActiveModel
 
           resource_path = [parent, resource_name].compact.join('.')
 
-          if include_assoc?(resource_path) && resource_type = serialized_object_type(serializers)
-            plural_name = resource_type.pluralize.to_sym
-            @top[:linked] ||= {}
-            @top[:linked][plural_name] ||= []
+          if include_assoc?(resource_path)
+            @top[:included] ||= []
 
             serializers.each do |serializer|
               attrs = attributes_for_serializer(serializer, @options)
 
               add_resource_links(attrs, serializer, add_included: false)
 
-              @top[:linked][plural_name].push(attrs) unless @top[:linked][plural_name].include?(attrs)
+              @top[:included].push(attrs) unless @top[:included].include?(attrs)
             end
           end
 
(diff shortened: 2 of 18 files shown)


lib/active_model/serializer.rb
```diff
@@ -164,6 +164,14 @@ def json_key
       end
     end
 
+    def id
+      object.id if object
+    end
+
+    def type
+      object.class.to_s.demodulize.underscore.pluralize
```
kurko: 1. I believe this should go into the adapter because it's not really used by anyone else and it's a JSONAPI need.
2. Having this in the serializer, we're forced to keep it public, which I think doesn't make much sense.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -35,58 +35,40 @@ def serializable_hash(options = {})
...
-            resource[:links][name][:ids] += serializers.map{|serializer| serializer.id.to_s }
-          end
+          resource[:links][name] ||= { linkage: [] }
+          resource[:links][name][:linkage] += serializers.map { |serializer| { type: serializer.type, id: serializer.id.to_s } }
         end
 
         def add_link(resource, name, serializer)
           resource[:links] ||= {}
-          resource[:links][name] = nil
+          resource[:links][name] = { linkage: nil }
```
kurko: What if resource has no links. Will it still have `{ linkage: nil }`? I believe that we should not have links if we have no associations. What do you think?

kurko: nvm

test/action_controller/serialization_scope_name_test.rb
```diff
@@ -29,12 +29,12 @@ def render_new_user
 
   def test_default_scope_name
     get :render_new_user
-    assert_equal '{"users":{"admin?":false}}', @response.body
+    assert_equal '{"data":{"admin?":false,"id":"1","type":"test_users"}}', @response.body
```
kurko: 1. The type here should be `"type": "user"` (or `users`).

2. Given it is type, in singular, I tend to prefer `user`. Just personal experience.

lib/active_model/serializer.rb
kurko: Oh, I see what you're doing. You are:

1. relying on the serialized object's type, not the serializer name.
2. leaving it here so people can override the inferred type.

Is that it? I'm onboard with that.

kurko:
Superb job here. I'm really glad this was done. Could you:

1. edit README to mention we support RC3, and
2. add an entry to the CHANGELOG?

test/action_controller/serialization_scope_name_test.rb
lsylvester: I think that the resource type should stay pluralized as the urls in a rails app are normally pluralized and the jsonapi api spec recommends that the URL for a collection of resources be formed from the resource type.

lib/active_model/serializer.rb
mateomurphy: Yes, easily overriding the type is one of the main reasons I put this here. One can create an `ApplicationSerializer` and use different inflection rules.

I put the id here as well because it made thing more symmetrical, but I can remove it.

test/action_controller/serialization_scope_name_test.rb
mateomurphy: 1. As noted above, it's `test_user` because that's the class of the resource, and I think it makes more sense to use that to determine the type, allowing one make generic serializers that decorate different models. But I am open to discussion on this. In this particular case, I think it would make sense to rename `TestUser` to `User` so that the type string is `users`.

2. I would prefer singular as well, but I decided to follow the examples in the spec which are all plural. But this is also why I wanted to make it very easy to change the inflection rules use.

lib/active_model/serializer.rb
```diff
@@ -172,6 +180,8 @@ def attributes(options = {})
           self.class._attributes.dup
         end
 
+      attributes += options[:required_fields] if options[:required_fields]
```
mateomurphy: Just noticed that this could lead to duplicates, I'll fix it.

JoshSmith:
@kurko ping. Looks like changelog and readme were updated.

lib/active_model/serializer.rb
mateomurphy: nvm it's fine

JoshSmith:
@mateomurphy can you rebase?

mateomurphy:
@JoshSmith Yep, done

lib/active_model/serializer.rb
zekefast: Wouldn't be a good idea to add comments or `GOTCHA`'s to those methods with explanation why the was placed here and what the original intention was behind that decision?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -16,7 +16,7 @@ def initialize(serializer, options = {})
         end
 
         def serializable_hash(options = {})
-          @root = (@options[:root] || serializer.json_key.to_s.pluralize).to_sym
+          @root = :data
```
zekefast: As this fixed by standard whould be better to move that into `initialize` method?

yawboakye: :+1:

mateomurphy: Actually, there's no reason to store the root, as it's not used elsewhere

lib/active_model/serializer.rb
mateomurphy: Good idea, I was already planning on adding it to the documentation.

However, we need to figure out what to do about the `id` here. The main issue is that the jsonapi adapter requires an id attribute, so how do we handle the cases where the user doesn't add the `id` attribute to the configuration (or any other required attribute)? Raise an error, or add the `id` transparently? If the latter, I can implement a generic solution, something like

```
    def attributes(options = {})
      attributes =
        if options[:fields]
          self.class._attributes & options[:fields]
        else
          self.class._attributes.dup
        end

      attributes += options[:required_fields] if options[:required_fields]

        hash[name] = read_attribute(name)
      end
    end

    def read_attribute(attr)
      if respond_to?(attr)
        send(attr)
      else
      end
    end
```

test/action_controller/serialization_scope_name_test.rb
mateomurphy: @kurko  I've refactored the test code a little bit, the output of type is now `"users"` as expected

lib/active_model/serializer.rb
yawboakye: I'd prefer raising an error when the `id` attribute is missing since in some cases the `id` used to build the URL isn't the resource's original `id` but a slug maybe.

lib/active_model/serializer.rb
```diff
@@ -164,6 +164,14 @@ def json_key
       end
     end
 
+    def id
+      object.id if object
```
yawboakye: The JSON API spec requires resource `id` to be a string.

mateomurphy: It's converted to a string in the adapter (https://github.com/rails-api/active_model_serializers/pull/853/files#diff-ae1e38e96a8bf28ae37918a65c448c73R89)

yawboakye: Okay, great :+1:

kurko:
LGTM. HERE WE GO!

kurko:
Reference https://github.com/rails-api/active_model_serializers/issues/829

mateomurphy:
I found some bugs right after you merged this. I fixed them here: https://github.com/rails-api/active_model_serializers/pull/858

PR #863, opened 2015-03-28
Add issue stats to README

JoshSmith:
We should probably make it clearer how active the development is and also how quickly we close issues. Will help motivate us to do better and also clarify to everyone else that we are _really_ active.

screen shot 2015-03-28 at 12 29 25 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ActiveModel::Serializers
 
-[![Build Status](https://travis-ci.org/rails-api/active_model_serializers.svg)](https://travis-ci.org/rails-api/active_model_serializers)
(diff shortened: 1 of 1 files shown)


kurko:
I don't have an opinion about it, except that people can see Github's Graph with our level of activity. We need more people agreeing with it. cc @guilleiguaran

JoshSmith:
@joaomdmoura @guilleiguaran what do you think?

saneshark:
Generally speaking, I'm :+1: but with 3 major versions of this gem, at least 2 of which are in active development and the main one not quite ready for prime-time, I don't know how much of the intended value you want to provide will actually be realized. 

It might be disingenuous to suggest that there are X number of PRs and X number of Issues closed on master, when it is suggested that one use 0.9.x vs 0.10.x (presumably the one getting the most PRs)

joaomdmoura:
I'm not sure if I have an opinion about it too. But I'm okay with that, mostly because idk if the "last commit" date is enough to satisfy those that want to know how actively we are working.

NullVoxPopuli:
I think this would be neat. 

Needs rebase, though.

Issue #825, opened 2015-03-03
Specify association serializer with conditional?


eliduke:
Hello AMSers!

I need some help and I hope I have come to the right place.

I am using AMS 0.8.3 and I am trying to use a conditional in a serializer to specify which serializer to use for a certain association. The standard json end point has a list of users with basic information. Then I want to have an "export" end point where more detailed user information would be output into a CSV or something like that.

Here's what I have so far...

```
class PrizeController < ApplicationController

def entrants
  render json: @prize, export: true, serializer: PrizeEntrantsSerializer
end
```

I'm passing in the `export: true` option thinking that I can use that in the conditional.

```
class PrizeEntrantsSerializer < ActiveModel::Serializer

  attributes :id, :event, :event_prize_type,
  has_many :entrants, serializer: which_serializer

  def which_serializer
    @options[:export] == true ? PrizeEntrantExportSerializer : PrizeEntrantSerializer
  end

end
```

That didn't work. I get `undefined local variable or method 'which_serializer'`, so I tried this:

```
class PrizeEntrantsSerializer < ActiveModel::Serializer

  attributes :id, :event, :event_prize_type

  if @options[:export] == true
    has_many :entrants, serializer: PrizeEntrantExportSerializer
  else
    has_many :entrants, serializer: PrizeEntrantSerializer
  end

end
```

That doesn't work either. I get `undefined method '[]' for Nil Class`. I am able to pass in the `@options` successfully because I tested it out by adding as an attribute and it does return correctly. Is it just not possible to have conditionals in a serializer?

Thanks for your help!

mehlah:
Hello guys,

I created a Rails app that reproduces what's described by @eliduke https://github.com/mehlah/ams-issue-825/

I'm suggesting a fix at #865, to support Symbols for `:serializer` option, that refers to a method to call.
I'm making myself familiar with AMS code and architecture, so I'm not sure this is the right direction. If so, I think we should support passing a Proc too.
@JoshSmith thoughts?
Is this something to port to 0.10.x/master?

kurko:
@eliduke `has_many` is a class method whereas `which_serializer` is an instance method. Try `def self. which_serializer` instead. Should work.

beauby:
@eliduke Did @kurko's suggestion help?

eliduke:
@beauby Thanks for pinging me again on this. In the end I switched things up and found a different solution, but now that this is back in my brain I will take a look and see if @kurko's suggestion works.

[...]

PR #865, opened 2015-03-30
`:serializer` option now can take a symbol as a method to call.

mehlah:
Fix suggestion for #825

This is handy for specifying a custom serializer class based on
conditionals.

### DIFF
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -48,7 +48,13 @@ module ActiveModel
 
         def target_serializer
           serializer = option(:serializer)
-          serializer.is_a?(String) ? serializer.constantize : serializer
+          if serializer.is_a?(String)
+            serializer.constantize
+          elsif serializer.is_a?(Symbol)
+            source_serializer.send(serializer)
+          else
+            serializer
+          end
         end
 
         def source_serializer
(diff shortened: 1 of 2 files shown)


webattitude:
:+1:

lib/active_model/serializer/associations.rb
```diff
@@ -48,7 +48,13 @@ def option(key, default=nil)
 
         def target_serializer
           serializer = option(:serializer)
-          serializer.is_a?(String) ? serializer.constantize : serializer
+          if serializer.is_a?(String)
+            serializer.constantize
+          elsif serializer.is_a?(Symbol)
+            source_serializer.send(serializer)
+          else
+            serializer
```
bf4: Interesting point. I'm not really sure what I think about it. 

This code could be

```ruby
case serializer
when String then serializer.constantize
when Symbol then source_serializer.send(serializer)
else serializer
end
```

NullVoxPopuli: this looks a bit cleaner (the case/when)

remear:
@rails-api/ams Do we want to continue with this?

mehlah:
Pls let me know if you want me to update the code and rebase it /cc @remear

beauby:
I have very little knowledge of 0.8. cc @joaomdmoura ?

PR #866, opened 2015-03-30
Fix the failing `test_render_json_with_callback`

mehlah:
Since Rails 4.0.10, a JS comment is prepended to JSONP callbacks, to address CVE-2014-4671.
Introduced in Rails at https://github.com/rails/rails/pull/16109
The build is still :red_circle: for `0-8-stable` due to other errors (I'm taking a look)


remear:
0-8-stable currently has no errors so it was probably fixed along the way. Closing this.

PR #867, opened 2015-04-02
Add include nil option

TheWudu:
Add option: include_nil, which is false by default. Added tests.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -183,7 +183,7 @@ module ActiveModel
       attributes += options[:required_fields] if options[:required_fields]
 
       attributes.each_with_object({}) do |name, hash|
-        hash[name] = send(name)
+        hash[name] = send(name) if !send(name).nil? || options[:include_nil]
       end
     end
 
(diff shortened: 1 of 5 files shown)


kurko:
Let me understand, you don't want to get nil attributes in your response and you're making it the default for everyone? If that's the case, I'm -1. Also, I believe this will affect the JSON API adapter.

If you make it default to true, then it makes more sense.

TheWudu:
I think it is useless to return nil values, but I already thought that someone would have a problem this way. If the owner or the most people say it would be better to have a "prevent_nil" instead of "include_nil" I would change this.

bf4:
Given that JSON has null, I think it's reasonable to use nil, no?  If you want to remove nil values from the hash, that seems to me to be on the end-users side. e.g. `as_json.reject! {|_,v| v.nil? }`

joaomdmoura:
I agree. Let's go for `prevent_nil` approach. :smile: 
Btw, you might need to rebase it, and taking into account the date of the PR there will be some conflicts, for sure. Sorry about that.

bf4:
@TheWudu I'm not convinced this needs to be a feature of AMS.  Have you tried something like the below with a custom serializer superclass?

```ruby
class TheWuduSerializer < ActiveModel::Serializer
    super.reject! {|_,v| v.nil? }
  end
end

class AuthorSerializer < TheWuduSerializer
end
```

beauby:
Seems like an edge case to me.

bf4:
Closing due to lack of response. Please reopen if you'd like to continue discussion and development.  However, based on my understanding, I'm also not convinced this should be a feature in AMS.

PR #868, opened 2015-04-03
Fixed a bug that appears when a nil association is included

groyoh:
Take a serializer that has an association with a serializer specified for this association, e.g.:

    PostPreviewSerializer = Class.new(ActiveModel::Serializer) do
      attributes :title, :body, :id
      belongs_to :author, serializer: AuthorPreviewSerializer
    end

When serialized with JsonApi adapter, if this association is included and the value associated is `nil` then `included` would look like this: `"included":[{"id":"", type:"nil_classes"}]`.

I changed the behaviour so that it only non nil values are added to `included`.


kurko:
Thanks @groyoh!

PR #872, opened 2015-04-13
AMS Benchmark tests #832

joaomdmoura:
Adding a benchmak test structure to help contributors to keep track
of how their PR will impact overall performance.

It enables developers to create test inside of tests/benchmark.

This implementation adds a rake task: ```rake benchmark``` that checkout
one commit before, run the test of  tests/benchmark, then mover back to
the last commit and run it again. By comparing the benchmark results between
both commits the contributor will notice if and how much  his contribution
will impact overall performance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,15 @@ install:
   - bundle install --retry=3
 
 env:
-  - "RAILS_VERSION=4.0"
-  - "RAILS_VERSION=4.1"
-  - "RAILS_VERSION=4.2"
-  - "RAILS_VERSION=master"
+  global:
+    - JRUBY_OPTS=-Xcext.enabled=true
+  matrix:
+    - "RAILS_VERSION=4.0"
+    - "RAILS_VERSION=4.1"
+    - "RAILS_VERSION=4.2"
+    - "RAILS_VERSION=master"
 
 matrix:
-  include:
-    - rvm: jruby-19mode
-      gemfile: gemfiles/Gemfile.rails-3.2.x
-      env: "JRUBY_OPTS=-Xcext.enabled=true"
   allow_failures:
     - rvm: ruby-head
     - env: "RAILS_VERSION=master"
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,10 @@ env:
   - "RAILS_VERSION=master"
 
 matrix:
+  include:
+    - rvm: jruby-19mode
+      gemfile: gemfiles/Gemfile.rails-3.2.x
+      env: "JRUBY_OPTS=-Xcext.enabled=true"
   allow_failures:
     - rvm: ruby-head
     - env: "RAILS_VERSION=master"
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,8 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in active_model_serializers.gemspec
 gemspec
 
-gem "minitest"
+gem 'minitest'
+gem 'rugged'
 
 version = ENV["RAILS_VERSION"] || "4.2"
 
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'minitest'
-gem 'rugged'
+gem 'git'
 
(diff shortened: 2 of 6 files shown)


kurko:
May I suggest something? Not everyone creates PR's with one commit. Should we perhaps run it against master instead? Or maybe something like `rake benchmark master`? What do you think about it?

joaomdmoura:
Totally agreed @kurko I thought about the issue but didn't came up with anything. Good idea ;)

joaomdmoura:
@kurko and it's done ;)
I think that I also fixed that travis ci thing I told you, the tests are supposed to work with the new dependency, but accordingly to their status page they are having some performance issues today.

kurko:
Tests failing. Restarting.

joaomdmoura:
Finally! @kurko can you review it for me? :stuck_out_tongue_closed_eyes:

.travis.yml
```diff
@@ -15,12 +15,15 @@ install:
   - bundle install --retry=3
 
 env:
-  - "RAILS_VERSION=4.0"
-  - "RAILS_VERSION=4.1"
-  - "RAILS_VERSION=4.2"
-  - "RAILS_VERSION=master"
+  global:
+    - JRUBY_OPTS=-Xcext.enabled=true
```
bf4: housekeeping: I don't think is a good idea to set JRUBY_OPTS this way since [travis uses `JRUBY_OPTS="--server -Dcext.enabled=false -Xcompile.invokedynamic=false"` now](http://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) ([cext.enabled=false is the future](http://docs.travis-ci.com/user/languages/ruby/#JRuby%3A-C-extensions-support-is-disabled)) and ref: https://github.com/rspec/rspec-dev/pull/115

kurko: @joaomdmoura ^

Gemfile
```diff
@@ -3,7 +3,8 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in active_model_serializers.gemspec
 gemspec
 
-gem "minitest"
+gem 'minitest'
+gem 'git'
```
kurko: I believe you should use the gemspec's `development_dependency` method, no?

Rakefile
```diff
@@ -9,4 +12,35 @@ Rake::TestTask.new do |t|
...
+  puts "------------------------------------\n\n"
+
+  puts "------------------------------------~> (Actual Branch) #{ref}"
+  puts actual
+  puts "------------------------------------"
+end
+
+def run_benchmark_spec(ref)
+    @git.checkout(ref)
+    response = Benchmark.realtime { Rake::Task['benchmark_tests'].invoke }
```
kurko: So, when you switch to a branch, do you run the tests only once? Given there's warm up and all that, would it make sense to run it 2 or 3 times each branch in order to get a better average?

.travis.yml
joaomdmoura: :cry:
I'll find a workaround

Gemfile
joaomdmoura: :sweat_smile: yeap

Rakefile
joaomdmoura: I made some tests and there was no huge difference, anyway, I'll give it a try again after made the updates required :smile:

kurko: Cool. Just remember that tests are usually very fast, but when in production and getting hit by thousands of users, the minor difference can make a huge difference.

beauby:
What's the status on this one?

bf4:
@joaomdmoura have you been able to demonstrate difference in caching vs. non caching in a branch and performance of each across two branches?  I've been futzing with this a bit and haven't been able to show a difference.

bf4:
https://github.com/rails-api/active_model_serializers/pull/1393 is based on this, and is on master, should run across commit history without conflicts

kurko:
This is stale. @joaomdmoura any thoughts on this? Any plans to continue?

bf4:
I now have a working implementation at https://github.com/rails-api/active_model_serializers/pull/1393 that I'm pretty happy with, for the benchmarking part, and some maturing revision-running code I'll be pushing up soon.

PR #879, opened 2015-04-23
Added serializer namespacing option

groyoh:
I added an `namespace` option to the serializer associations and array serializer. This option can be used to define in which `module` the correct serializer may be found.  This option is pretty useful when dealing with versionized serializer and polymorphic associations. I added some information on how to use it in the README, e.g. :

> You may use the `:namespace` options to specify the `module` where the serializer(s) of an
association may be found, for example:
```ruby
has_many :comments, namespace: Blogging::Serializers
```
The serializer used then depends on the class of the object to be serialized e.g. if a
comment is an instance of class `AdminComment`, the serializer would use
`Blogging::Serializers::AdminComment` to serializer this comment.

To make it clear, I do not use rails so I don't know if this would fit well with rails. I'm opened to any kind of discussions and changes. And of course, serializers here are used without `Serializer` suffix because that's the way we use it in our team but this behavior may as well be changed (adding a `suffix` option for example).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,3 +6,4 @@
   * updates JSON API support to RC3 [@mateomurphy]
   * adds fragment cache support [@joaomdmoura]
   * adds cache support to attributes and associations [@joaomdmoura]
+  * adds a `:namespace` option to association and array serializer [@groyoh]
--- a/README.md
+++ b/README.md
@@ -249,6 +249,16 @@ The `has_many`, `has_one`, and `belongs_to` declarations describe relationships 
 resources. By default, when you serialize a `Post`, you will get its `Comment`s
 as well.
 
+You may use the `:namespace` options to specify the `module` where the serializer(s) of an
+association may be found, for example:
+
+```ruby
+  has_many :comments, namespace: Blogging::Serializers
+```
+The serializer used then depends on the class of the object to be serialized e.g. if a
+comment is an instance of class `AdminComment`, the serializer would use
+`Blogging::Serializers::AdminComment` to serializer this comment.
+
 You may also use the `:serializer` option to specify a custom serializer class, for example:
 
 ```ruby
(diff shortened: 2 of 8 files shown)


cfeckardt:
This is useful to me!

I want to selectively use different serializers for different endpoints in my app.
One of these associations are polymorphic, and should use different serializers, depending on which endpoint is hit. This PR solves this problem. :+1:

dpmccabe:
This commit fixed my issue from #916. Thanks!

joaomdmoura:
Just letting you guys know that I'm aware of this PR, but before merge it I have to discuss the implementation with other AMS members, it might take some weeks yet, but this is on my radar.
For now you can still workaround it by explicitly specifying the Serializer inside you action.

bryanrite:
This would be very useful for me as well, as different versions of an API use differently namespaced serializers and working with polymorphic associations is very difficult. Thanks!

bf4:
Closing as it appears to have been resolved by https://github.com/rails-api/active_model_serializers/pull/1225  Sorry for the lost work here :(  Please re-open if there's more to discuss

groyoh:
@bf4  I completely forgot about this issue over time. Unfortunately #1225 does not solve our issue. We are mostly using sinatra apps and follow our own structure which does not fit with Rails and AMS convention. Basically, our structure is defined like so (purely fictional example):

```ruby
module Model
  class User; end
  class Computer; end
  class Laptop < Computer; end
  class Server < Computer; end
end

module Serializer
  module V1 # for API version 1
    class User
      has_many :computers # can be Laptop or Server
    end
    class Laptop
    end
    class Server
    end
  end

  module V2 # for API version 2
    class User
      has_many :computers # can be Laptop or Server
    end
    class Laptop
    end
    class Server
    end
  end
end
```

So there are 5 issues here that we have to tackle:
* We want to define many different serializer namespaces (one per API version)
* The serializers and models are not defined in the same namespace
* The serializers do not end with `Serializer` (this is actually a different concern but I still wanted to point it out)
* The `has_many :computers` is a polymorphic association which can have instances of `Laptop` and  `Server`
* The `Laptop` and `Server` serializer do not fit as nested class in the `User` serializer because they can be used outside of the scope of the `User` serializer. So #1225 does not apply too much in this case.

Our solution was to define a `namespace` option which could be used in two different ways:
* `ArraySerializer.new(computers, namespace: Serializer::V2)` which would look up the serializer for the array elements in the `Serializer::V2` module.
* With a relationship:

```ruby
module Serializer
  module V1
    class User
      has_many :computers, namespace: Serializer::V1
    end
  end
end

serializer = V1::User.new(user) # this will lookup the `Laptop` and `Server` serializers within the `Serializer::V1` namespace
```

Do you have any opinion on how to solve these issues with having a `namespace` option or strictly respecting Rails/AMS conventions? Maybe I missed a feature that could help us here. As always, if my example was not clear enough I could give in more details.

bf4:
Maybe we can write failing tests for each scenario. There is also another pr for adding a namespace option to the serialization context like 0.9 had. That should help, I think.  https://github.com/rails-api/active_model_serializers/pull/1436

B mobile phone

PR #888, opened 2015-04-27
Changed duplicated test name in action controller test

groyoh:
I fixed a test in `test/action_controller/explicit_serializer_test.rb` which had the same name of another test declared before. This basically prevented the first test to run.


joaomdmoura:
Tks @groyoh !
I'm merging it! :smile:

PR #890, opened 2015-04-27
Remove unused method `def_serializer`

JustinAiken:
While looking at the tests noticed that this wasn't used anymore.


joaomdmoura:
Tks @JustinAiken 
I'm merging it.

PR #892, opened 2015-04-28
Fixed a bug that appeared when json adapter serialize a nil association

groyoh:
Take the following serializers e.g.:

```ruby
AuthorPreviewSerializer = Class.new(ActiveModel::Serializer) do
  attributes :id, :name
end    
PostPreviewSerializer = Class.new(ActiveModel::Serializer) do
  attribute :id
  # note that we specify a serializer for the association
  belongs_to :author, serializer: AuthorPreviewSerializer
end
```
If a post is serialized with Json adapter and its author is nil, then the post would be rendered as : 
```ruby
{ 
  id: 43, 
  author: { id: nil, name: nil }
}
```
instead of
```ruby
{ 
  id: 43, 
  author: nil
}
```

This happens only when the `serializer` option is set for the association.

### DIFF
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -23,7 +23,7 @@ module ActiveModel
                   end
                 end
               else
-                if association
+                if association && association.object
                   @hash[name] = cache_check(association) do
                     association.attributes(options)
                   end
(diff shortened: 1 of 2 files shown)


philipgiuliani:
Thanks a lot! I just ran into this. Would be cool to get it merged!

joaomdmoura:
Tks @groyoh ! I'm merging it :+1:

PR #896, opened 2015-05-02
JSON Adapter Refactoring

adomokos:
I pulled the logic from the `serializable_hash` into private methods. This way it's a lot easier to understand what the code does. No logic has changed, I verified it with the passing unit tests.

### DIFF
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -15,24 +15,8 @@ module ActiveModel
             end
 
             serializer.each_association do |name, association, opts|
-              if association.respond_to?(:each)
-                array_serializer = association
-                @hash[name] = array_serializer.map do |item|
-                  cache_check(item) do
-                    item.attributes(opts)
-                  end
-                end
-              else
-                if association
-                  @hash[name] = cache_check(association) do
-                    association.attributes(options)
-                  end
-                elsif opts[:virtual_value]
-                  @hash[name] = opts[:virtual_value]
-                else
-                  @hash[name] = nil
-                end
-              end
+              populate_hash_for_array_serializer(name, association, opts)
+              populate_hash(name, association, opts, options)
             end
             @result = @core.merge @hash
           end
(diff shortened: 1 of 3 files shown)


lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +31,32 @@ def serializable_hash(options = {})
       def fragment_cache(cached_hash, non_cached_hash)
         Json::FragmentCache.new().fragment_cache(cached_hash, non_cached_hash)
       end
+
+      private
+
+      def populate_hash_for_array_serializer(name, association, opts)
+        return unless association.respond_to?(:each)
```
joaomdmoura: IMHO I'm not sure about moving the conditionals to inside this private methods.
It just not seems to me that this conditional handling would be their responsibility. Running the same conditional twice is a smell of that.

joaomdmoura:
I reviewed it seems okay for me. Just pointed out a small change, that is to keep the ```association.respond_to?(:each)``` inside ```serializable_hash```
Would be awesome if you could rebase it with master :smile:

lib/active_model/serializer/adapter/json.rb
adomokos: The reason I moved it to a private method was to eliminate the nested conditionals in the iterator. I think the code is easier to read this way, and moving the guard condition back into the iterator defeats the purpose of this PR.

joaomdmoura: The nested conditional itself still exists, and now it will run twice the same conditional to each interaction. 
The duplicated line itself is a bad smell. [link1](http://www.laputan.org/pub/patterns/fowler/smells.pdf) [link2](https://www.cqse.eu/publications/2010-code-similarities-beyond-copy-paste.pdf)

> Such cloned code is considered harmful for
two reasons: (1) multiple, possibly unnecessary, duplicates
of code increase maintenance costs [16, 24] and, (2) inconsistent
changes to cloned code can create faults and, hence,
lead to incorrect program behavior [13].

I totally understand the motivation here :smile: but I disagree that keeping the conditional inside ```serializable_hash``` defeats the purpose of this PR, creating the private functions indeed help to keep it clean and organised.

PR #897, opened 2015-05-03
Allow to define custom serializer for given class

imanel:
Resolves #515.
Adjusted [commit from 0-9-stable](https://github.com/rails-api/active_model_serializers/commit/d8b78a36cfde1bef6dc23c6a64f6b92c4597258d) so it will work with master.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -115,7 +115,9 @@ module ActiveModel
     end
 
     def self.serializer_for(resource, options = {})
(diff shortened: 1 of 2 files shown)


ekampp:
This seems like a good, clean implementation, and would keep serialization logic for the object out of the controller.

joaomdmoura:
A tiny and clean patch, taking into consideration it was already merge into old AMS version, I'm merging it.

PR #906, opened 2015-05-08
Fragment cache fails for models inside a module

balauru:
It will fail when trying to set constant because `Test::NestedRoleCachedSerializer` is not a correct constant name.

### DIFF
--- a/test/adapter/fragment_cache_test.rb
+++ b/test/adapter/fragment_cache_test.rb
@@ -20,6 +20,13 @@ module ActiveModel
           }
           assert_equal(@role_hash.fetch, expected_result)
         end
+
+        def test_this_works
+          role            = Test::NestedRole.new(name: 'Great Author', description:nil)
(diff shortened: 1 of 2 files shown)


balauru:
@joaomdmoura I can propose a solution to this one if you want me to, I've already have a draft on it, my idea is to define the constant on the module (if one exists) instead of defining it directly on Object.

joaomdmoura:
@balauru Tks! Totally :smile:
I'll probably dive into it on the weekend, so yeah you should post your proposal here, if you think you are ready to put together a PR just let me know.

balauru:
I've added a solution that uses unit tests as a starting point, will clean it up, any feedback?

bf4:
@alexandru-calinoiu Looks good. Sorry this has gotten lost in the shuffle. Are you interested in picking it up again?

NullVoxPopuli:
@alexandru-calinoiu can you rebase this? looks good, though!

PR #909, opened 2015-05-12
Defining Json-API Adapter as Default

joaomdmoura:
As we decided, it's time to push JSON-API forward by making it the default adapter into AMS.
We really believe this is one of the greatest features into 0.10.x version, and that could help a lot of different projects from different technologies.
Despite of being a *small* change, we expect to see more people using it's conventions once we merge it.
@steveklabnik @guilleiguaran @kurko

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@ AMS does this through two components: **serializers** and **adapters**.
 Serializers describe _which_ attributes and relationships should be serialized.
 Adapters describe _how_ attributes and relationships should be serialized.
 
+By default AMS will use the JsonApi Adapter that follows RC3 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
+Check how to change the adapter in the sections bellow.
+
 # RELEASE CANDIDATE, PLEASE READ
 
 This is the master branch of AMS. It will become the `0.10.0` release when it's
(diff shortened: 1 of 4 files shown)


joaomdmoura:
@guilleiguaran o/

guilleiguaran:
@joaomdmoura \o

spastorino:
We would need to revert this in order to be able to use AMS by default in Rails. :heart:

steveklabnik:
This has changed?

joaomdmoura:
@steveklabnik we have just talked trough it on IRC.
You can check the conversation there but the idea is to make it easier to integrate with the usual JS frameworks around that are not using json api **yet**. 
Ember is already moving forward with it, but the ideia would be to avoid push it at once but make it really easy to use.

steveklabnik:
:/

steveklabnik:
So, basically, the reason that JSON API should be the default is multifold, but one of the main ones is that convention over configuration is the heart of AMS.

You pick _some_ kind of thing as the default, and so choosing an underspecified, ad-hoc old version doesn't make nearly as much sense to me. Classic AMS output was changed in JSON API because it isn't particularly great, a starting point.

Issue #904, opened 2015-05-07
Fragment cache + array payload compounds non-cached keys


apolzon:
If you define a serializer with a fragment cache, such as

```ruby
class TestSerializer < ActiveModel::Serializer
  cache key: "test", except: [:field1]
  attributes :field1, :field2
end
```

and then render in your controller, an array:
```ruby
render json: resources, each_serializer: TestSerializer
```

the fragment cache's call to attribute (line 52) will continually add field1, such that if resources has 20 objects, by the end, field1 will be called 20 times for the last invocation of the TestSerializer.

I believe ActiveModel::Serializer.attribute should only concat the key onto @_attributes if it is not already present.
i.e., line 45 (in master) should be:
```ruby
      @_attributes.concat [key] unless @_attributes.include? key
```

Also, since the @_attributes variable is at the class level, subsequent requests compound the issue even further.

groyoh:
Nice catch! It was indeed a bug which would also happen simply by calling multiple times `.attributes` or `.attribute` with the same attribute name. I don't know if this has any consequence on the behavior itself, but might slow the overall system since we iterate over the `@_attributes` variable when calling `#attributes`.

I created a PR to fix this (#914).

joaomdmoura:
Tks @apolzon and @groyoh !
Indeed it was a great catch, and an important issue.
But it didn't impacted the performance so much, it remained the same. :smile: Just ran some benchmark test against an older master commit  and also against ```0.10-stable``` 
:heart:

PR #914, opened 2015-05-18
Prevent possible duplicated attributes in serializer

groyoh:
This PR is based on #904 . 

The issue was that calling `ActiveModel::Serializer.attributes` or `ActiveModel::Serializer.attribute` methods multiple times with the same attribute would add this attribute again and again to the `@_attributes` variable. As the `attribute` method is called by the `FragmentCache` multiple times, this might be a important issue.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -31,6 +31,7 @@ module ActiveModel
     def self.attributes(*attrs)
       attrs = attrs.first if attrs.first.class == Array
       @_attributes.concat attrs
+      @_attributes.uniq!
 
       attrs.each do |attr|
         define_method attr do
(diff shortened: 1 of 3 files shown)


joaomdmoura:
Indeed this a important issue! I'm deeply interest on this.
Tests falling, I'm restarting it.

lib/active_model/serializer.rb
```diff
@@ -31,6 +31,7 @@ def self.inherited(base)
     def self.attributes(*attrs)
       attrs = attrs.first if attrs.first.class == Array
       @_attributes.concat attrs
+      @_attributes.uniq!
```
joaomdmoura: I'm not sure with we need to worry about the ```uniq!``` here

groyoh: Well, I can see 3 use cases:
* a user calls `attributes :title, :title`
* a user calls: 
```ruby
attributes :title
attributes :title
```
* a future feature to call `attributes` in a similar way the fragment cache

The first two are not really serious, but the last one might go unseen like the fragment cache.

PR #915, opened 2015-05-19
Fix root/meta for ArraySerializer

ryansch:


ryansch:
I've rebased this on master and squashed in a commit to make both calls to `#assert_equal` use the correct argument order.

lib/active_model/serializer/array_serializer.rb
```diff
@@ -21,11 +22,11 @@ def initialize(objects, options = {})
       end
 
       def json_key
-        @objects.first.json_key if @objects.first
-      end
-
-      def root=(root)
-        @objects.first.root = root if @objects.first
+        if root == true && @objects.first
+          @objects.first.class.root_name
```
joaomdmoura: I'm not sure about using the ```root_name``` here. ArraySerializer is pretty abstract, take a lott at this piece of code for example:

```ruby
@objects = objects.map do |object|
  serializer_class = options.fetch(
    :serializer,
    ActiveModel::Serializer.serializer_for(object)
  )
  serializer_class.new(object, options.except(:serializer))
end
```

By checking this code we can presume that the objects might not be instances from the same class. Wrap them by the class name from its first object doesn't would be right.

ryansch: The problem here is the jsonapi adapter.  It sets `serializer.root = true` in its initializer.  I'm not sure how to deal with heterogenous arrays.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -4,6 +4,11 @@ module ActiveModel
   class Serializer
     class Adapter
       class Json < Adapter
+        def initialize(serializer, options = {})
+          super
+          serializer.root = true if @options[:root]
```
bf4: this logic doesn't make sense to me.  why are we only checking if the value of options[:root] is truthy?

lib/active_model/serializer/array_serializer.rb
```diff
@@ -4,9 +4,10 @@ class ArraySerializer
       include Enumerable
       delegate :each, to: :@objects
 
-      attr_reader :meta, :meta_key
+      attr_accessor :meta, :meta_key, :root
```
bf4: what's the read to change the reader to an accessor?

lib/active_model/serializer/array_serializer.rb
joaomdmoura: We could define a default ```data``` root by default is there isn't one defined. But I'm not convinced yet about the root implementation on ```ArraySerializer``` :stuck_out_tongue_closed_eyes:

ryansch:
I'm going to retire this PR.  We ended up ditching both jsonapi and AMS 0.10.x in favor of 0.8.3 for now.  More of a 'just works' approach.

joaomdmoura:
:cry:

bf4:
@joaomdmoura This is fixed in #936, no?

joaomdmoura:
@bf4 I haven't test, so I'm not sure about it. But supposedly yes.

PR #918, opened 2015-05-20
Add rescue_from handler to clear state

### LINKED ISSUE #917: rescue_from handler fails with AMS
ActionController::Serialization caches the serializer discovered in `#get_serializer` in `@_serializer`.  When the render call raises an error and ends up in a rescue_from handler, the serializer is still set to whatever was discovered before the error was raised.  This usually prevents the correct error response from being generated.

ryansch:
Fixes #917

### DIFF
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -211,3 +211,9 @@ end
 Spam::UnrelatedLinkSerializer = Class.new(ActiveModel::Serializer) do
   attributes :id
 end
+
+RaiseErrorSerializer = Class.new(ActiveModel::Serializer) do
+  def json_key
+    raise StandardError, 'OOPS'
+  end
+end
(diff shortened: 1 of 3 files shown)


test/fixtures/poro.rb
```diff
@@ -211,3 +211,9 @@ def self.root_name
 Spam::UnrelatedLinkSerializer = Class.new(ActiveModel::Serializer) do
   attributes :id
 end
+
+RaiseErrorSerializer = Class.new(ActiveModel::Serializer) do
+  def json_key
+    raise StandardError, 'OOPS'
```
JoshSmith: Hah, can we write a different error message? :laughing:

joaomdmoura:
Tks @ryansch, I'm merging it! :smile:

PR #919, opened 2015-05-20
Adding configuration option to exclude `null` values from output

navinpeiris:
This adds a configuration options that allows the user to specify whether `null` values should appear in the resultant media type, and by default it's set to include `null` values so that existing behaviour isn't changed.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -53,6 +53,14 @@ module ActionController
       end
     end
 
+    def rescue_with_handler(exception)
+      @_serializer = nil
+      @_serializer_opts = nil
+      @_adapter_opts = nil
+
+      super(exception)
+    end
+
     module ClassMethods
       def serialization_scope(scope)
         self._serialization_scope = scope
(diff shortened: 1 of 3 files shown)


joaomdmoura:
@navinpeiris I know there is a lot of different opinions about the ```nulll``` return, a lot of mixed feelings :stuck_out_tongue_closed_eyes: 

We, as a team, are trying to push JSON-API forward, because we believe that it's an amazing convention that can help a lot of different people from different technologies. We [just made](https://github.com/rails-api/active_model_serializers/pull/909) it the default adapter, and the ```null ``` return is one of [its conventions](http://jsonapi.org/format/).

So I'm not in favor of adding thins configuration. Of course we could still add it only to the JSON adapter, but I'm not sure if it worths. 
cc @steveklabnik @guilleiguaran @kurko

navinpeiris:
Cool, thanks for that @joaomdmoura. Yeps, JSON-API does look like the bomb, but unfortunately we've got to stick with our current format for awhile.

Would this be much of an issue though considering it's provided as a configuration option for the user to chose which way they want to go, and by default it'll stay true to the spec as well?

I'm just trialing converting from ROAR to AMS to take advantage of fragmented caching, so this'll help making the responses a 100% compatible with the previous, but saying that, this is more of a nicety than a must, so feel free to close this off if it's not inline with where you peeps want to head

PR #920, opened 2015-05-20
Allow :key to be specified for attributes overridden in the serializer

navinpeiris:
This makes the following example code function as expected:

```ruby
  attribute :num_article, key: :articlesCount

  def num_articles
    object.articles.count
  end
end
```

An alternative to the above is to specify the attribute and corresponding method as `articlesCount`, but this allows us to follow ruby code conventions while allowing us to customize the response as expected by the client.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -45,7 +45,11 @@ module ActiveModel
       @_attributes_keys[attr] = {key: key} if key != attr
       @_attributes << key unless @_attributes.include?(key)
       define_method key do
-        object.read_attribute_for_serialization(attr)
+        if attr != key && self.class.method_defined?(attr)
+          self.send(attr)
+        else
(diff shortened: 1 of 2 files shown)


joaomdmoura:
@navinpeiris I understand the motivation here, but in IMHO it seem a little bit too much for me.
I'm more in favor of stick with the conventions used on older versions, seems easier to understand and enables you to define all attributes in one line, event the virtual ones using the ```attributes``` method.
Let's keep it open for a while so other rails-api members can comment if they want to.

navinpeiris:
Cool, thanks again @joaomdmoura. Again, this is a nicety and doesn't hold us back, so please feel free to close this if you'd rather keep it as is

joaomdmoura:
I'm closing it for now. @kurko make yourself comfortable to comment if you want to.
Tks for the support @navinpeiris  :smile:

PR #924, opened 2015-05-21
Fixing issue where fragment cache calls underlying attribute methods multiple times even when cached

navinpeiris:
Fragment cache was calling `Serializer#attributes` method multiple times when trying to trying to resolve cached/non-cached attribute keys, which in turn was calling those respective methods in the underlying ActiveModel object, which results is more calls to the underlying object than if the object wasn't cached.

This has now been fixed by getting the fragment cache to use the `Serializer::_attributes` variable which can be used to resolve the cached/non-cached attribute keys.

I would have loved to add some tests into this but my Minitest - Fu is non-existent, and I spent way too much time already trying to figure out how to verify the call counts to specific methods, any help on this would be greatly appreciated.

### DIFF
--- a/lib/active_model/serializer/adapter/fragment_cache.rb
+++ b/lib/active_model/serializer/adapter/fragment_cache.rb
@@ -35,8 +35,9 @@ module ActiveModel
         private
 
         def cached_attributes(klass, serializers)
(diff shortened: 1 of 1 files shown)


joaomdmoura:
Great @navinpeiris you're right, I'm there is no need for new tests here, it just changes the actual implementation. :smile: 
Thanks you so much, and sorry for take so long to review it!

PR #930, opened 2015-05-27
Reverting PR #909

joaomdmoura:
Reverting PR #909 but also and adding json api usage advise on readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ AMS does this through two components: **serializers** and **adapters**.
 Serializers describe _which_ attributes and relationships should be serialized.
 Adapters describe _how_ attributes and relationships should be serialized.
 
-By default AMS will use the JsonApi Adapter that follows RC4 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
+By default AMS will use the **Json Adapter**. But we strongly advise you to use JsonApi Adapter that follows RC4 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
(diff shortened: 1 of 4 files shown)


steveklabnik:
:cry: :cry: :cry: :cry:

joaomdmoura:
:sob: sorry @seveklabnik , but I know we will define it as default soon! :disappointed: 
In the mean while we should promote it on some JS frameworks! It will help us into build a a standard and push jsonapi forward.

steveklabnik:
I think this works _against_ that goal :/

spastorino:
As I told you guys, let's not try to accomplish everything in one move.
AMS with JSON API as default won't be accepted in Rails. AMS with json adapter by default and an easy way to switch to JSON API will be.

spastorino:
I'm totally fine with both ways for me it's just a switch I have to change to use what I need.
The thing is right now JSON API is not accepted as default in Rails.
If AMS' default adapter is JSON API I'm perfectly fine but I can't merge Rails API unless I completely remove AMS as a dependency on Rails side and I think that's much worse.

joaomdmoura:
It's indeed worse Steve. This is why I agree with Santiago on this. Being integrated into Rails might even help  us in the long run when promoting json app. 





João M. D. Moura

joaomdmoura:
As I discussed on IRC with @spastorino, it seem that you @steveklabnik, also agreed on this for now, so I'm merging it.

PR #939, opened 2015-06-04
Use a more precise generated cache key

aaronlerch:
This change updates the default generated cache key (when the `key: 'foo'` option is specified and the object's `#cache_key` implementation isn't used) to use a more precise version of `updated_at`. By default, it was invoking `updated_at.to_s` which ends up with a cache key value like `cache-key-prefix/807130-2015-05-27 18:11:05 UTC`

This effectively means, if someone specifies a cache key in the serializer (which is really just a cache key prefix value -- I'll save that discussion for another PR perhaps ;) ) it will only be cached with a precision to the minute. So if I have a model and it gets cached, and I update it within the same minute, the cache will not be considered invalid and my model changes will not be reflected.

The ActiveRecord implementation of the default cache key, as you can see here https://github.com/rails/rails/blob/master/activerecord/lib/active_record/integration.rb#L55 , allows some customization but defaults to `updated_at.utc.to_s(:nsec)` which gives a much greater precision. In the example I gave above, it would be `cache-key-prefix/807130-20150527181105000000000`.
Since this gem doesn't have the dependencies to leverage the `:nsec` format helper, I leveraged the same format string to use here. (Currently using a magic string, since the only duplicate copies are in tests.)

I specifically did *not* update the cache_key to use `#utc`, but if that's desired I can update it to use that as well.

I also fixed up some of the tests. The test harnesses were written specifically using `updated_at.to_i` which allowed some implicit assumptions about cache invalidation to be encoded into the tests, which failed (rightly so) when I made the change to use a higher precision value.

:sparkles: :clock3: :sparkles: 

Thanks!
Aaron

### DIFF
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -72,7 +72,9 @@ module ActiveModel
       end
 
       def object_cache_key
-        (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{@cached_serializer.object.updated_at}" : @cached_serializer.object.cache_key
+        object_time_safe = @cached_serializer.object.updated_at
+        object_time_safe = object_time_safe.strftime("%Y%m%d%H%M%S%9N") if object_time_safe.respond_to?(:strftime)
+        (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{object_time_safe}" : @cached_serializer.object.cache_key
       end
 
       def meta
--- a/test/action_controller/serialization_test.rb
+++ b/test/action_controller/serialization_test.rb
@@ -98,13 +98,12 @@ module ActionController
 
         def render_fragment_changed_object_with_relationship
           comment = Comment.new({ id: 1, body: 'ZOMG A COMMENT' })
+          comment2 = Comment.new({ id: 1, body: 'ZOMG AN UPDATED-BUT-NOT-CACHE-EXPIRED COMMENT' })
           author = Author.new(id: 1, name: 'Joao Moura.')
-          post = Post.new({ id: 1, title: 'New Post', body: 'Body', comments: [comment], author: author })
-          post2 = Post.new({ id: 1, title: 'New Post2', body: 'Body2', comments: [comment], author: author })
-          like = Like.new({ id: 1, post: post, time: 3.days.ago })
+          like = Like.new({ id: 1, likeable: comment, time: 3.days.ago })
 
           generate_cached_serializer(like)
-          like.post = post2
+          like.likable = comment2
           like.time = DateTime.now.to_s
 
           render json: like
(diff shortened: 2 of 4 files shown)


lib/active_model/serializer/adapter.rb
```diff
@@ -63,7 +63,7 @@ def is_fragment_cached?
       end
 
       def cache_key
-        (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{@cached_serializer.object.updated_at}" : @cached_serializer.object.cache_key
+        (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{@cached_serializer.object.updated_at.strftime("%Y%m%d%H%M%S%9N")}" : @cached_serializer.object.cache_key
```
aaronlerch: Unfortunately this means if `updated_at` is `nil`, you're gonna have a Bad Time. If that's something we should consider, I can update this method to be a bit more resilient.

test/action_controller/serialization_test.rb
```diff
@@ -265,7 +264,7 @@ def test_render_with_cache_enable
         assert_equal expected.to_json, @response.body
 
         get :render_changed_object_with_cache_enabled
-        assert_equal expected.to_json, @response.body
+        assert_not_equal expected.to_json, @response.body
```
aaronlerch: This used to be able to be `assert_equal` because the cache key wasn't a high enough resolution to detect the change. This was actually incorrect (and/or bad caching), so at a minimum I changed the test to assert on the correct result, which it now receives.

test/fixtures/poro.rb
```diff
@@ -4,11 +4,11 @@ def initialize(hash={})
   end
 
   def cache_key
-    "#{self.class.name.downcase}/#{self.id}-#{self.updated_at}"
+    "#{self.class.name.downcase}/#{self.id}-#{self.updated_at.strftime("%Y%m%d%H%M%S%9N")}"
   end
 
   def updated_at
-    @attributes[:updated_at] ||= DateTime.now.to_time.to_i
```
aaronlerch: This usage of `#to_i` was one of the root causes behind the tests passing when they shouldn't have, and (incidentally) the tests not testing for real-world scenarios where `updated_at` isn't an integer, but is an instance of `Time` (or more likely an `ActiveSupport::TimeWithZone` instance.)

lib/active_model/serializer/adapter.rb
joaomdmoura: Yeah, that would be awesome if we could make it more resilient :smile:

joaomdmoura:
@aaronlerch sorry for taking so long to review it, I was rushing with some new feature myself, but I'm really in favor of this PR.
Could you just try to avoid threat the updated_at as nil thing and rebase it with master? :smile:

aaronlerch:
No problem -- yeah, I'll make the change and update the pull request. I'm travelling for a few days so I'll get to it as soon as I'm able. Thanks!

joaomdmoura:
Great! Thank you! :heart:

aaronlerch:
Hold on, I think I messed up my attempt to rebase off master.

aaronlerch:
Okay, assuming Travis likes my pull request, it should be ready now @joaomdmoura -- including the change to make the use of `#updated_at` more resilient. Thanks for your patience!

joaomdmoura:
Amazing @aaronlerch! I'm merging it! :heart:

PR #948, opened 2015-06-10
AMS supports JSONAPI 1.0 instead of RC4

SeyZ:


joaomdmoura:
Hm... @SeyZ I'll just double check if it's being fully supported before merge this. :smile:

SeyZ:
of course @joaomdmoura thanks!

joaomdmoura:
Indeed, as far as I checked we are compliant to 1.0
But as I check I realised that there are some new partes we could start implementing :smile:
I'm merging it!

PR #949, opened 2015-06-10
Fix: Don't pass serializer option to associated serializers

### LINKED ISSUE #870: Problem with each_serializer: foo and has_many associations
So it seems like if you do something like
``` ruby
render json: foo, each_serializer: FooSerializer

#with this serializer
class FooSerializer
   attributes :something
   has_many :bars
end
class BarSerializer
end
```
you will get an error like `undefined method 'something' for #<Bar:0x007f8c08d5d1f8>`

The regression appears to have been introduced between the following commits:

https://github.com/rails-api/active_model_serializers/compare/a8243761fedbf...48ed7

I believe the issue is in lib/active_model/serializer.rb in `def each_assocation` where we do `options.merge(serializer_from_options(association_options))` which causes the association to be serialized with the option `each_serializer: FooSerializer`

edwardloveall:
Fixes #870

Commit af81a40 introduced passing a serializer's 'options'
along to its associated model serializers.

Thus, an explicit 'each_serializer' passed to render for a
singular resource would be passed on as the implicit 'serializer'
for its associations.

With @bf4

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -217,7 +217,7 @@ module ActiveModel
         if serializer_class
           serializer = serializer_class.new(
             association_value,
-            options.merge(serializer_from_options(association_options))
+            options.except(:serializer).merge(serializer_from_options(association_options))
           )
         elsif !association_value.nil? && !association_value.instance_of?(Object)
           association_options[:association_options][:virtual_value] = association_value
--- a/test/action_controller/adapter_selector_test.rb
+++ b/test/action_controller/adapter_selector_test.rb
@@ -3,7 +3,7 @@ require 'test_helper'
 module ActionController
   module Serialization
     class AdapterSelectorTest < ActionController::TestCase
-      class MyController < ActionController::Base
+      class AdapterSelectorTestController < ActionController::Base
         def render_using_default_adapter
           @profile = Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1' })
           render json: @profile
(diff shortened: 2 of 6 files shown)


bf4:
[Test failures](https://travis-ci.org/rails-api/active_model_serializers/builds/66297930) are a race condition in the specs (what, ivars everywhere? yup) to be address in another issue/PR.

joaomdmoura:
Would be awesome if we could try to make this one green before merge it :smile:

bf4:
@joaomdmoura Like I wrote above, the specs actually pass, but intermittently fail due to global state.  I'll write a PR against master, right now, if that helps.

bf4:
Created PR into this branch https://github.com/edwardloveall/active_model_serializers/pull/1

bf4:
@joaomdmoura wee, all green! (not sure if https://github.com/rails-api/active_model_serializers/commit/14439aada49f9444098521a0163ea24cc6e5ba7b and https://github.com/rails-api/active_model_serializers/issues/951 require more attention. I'm inclined to ignore until it's hard to work around.)

lib/active_model/serializer.rb
```diff
@@ -217,7 +217,7 @@ def each_association(&block)
         if serializer_class
           serializer = serializer_class.new(
             association_value,
-            options.merge(serializer_from_options(association_options))
+            options.except(:serializer).merge(serializer_from_options(association_options))
```
bf4: This is the simplest thing that worked without doing a bunch of refactoring. (We considered adding a `::new_association_serializer` that wraps `::new` but it wasn't a quick win.)

joaomdmoura:
Great! Really awesome paired work here!
Indeed a quick win :smile:, simple and solves the problem.
Merging it.

PR #950, opened 2015-06-11
WIP - Deserializer implementation

joaomdmoura:
We decided that it's time to work with deserialization. :tada::tada: :tada:  

AMS aim to bring convention over configuration to JSON generation, but as it evolves into a more complete solution to build great APIs it makes sense to also handle deserialization.

The initial proposal of Deserialization is to convert your json input into something the ```ActiveRecord```  understands.

It will play really nice with json-api, enabling us to follow it conventions when creating and updating objects using AMS.

How it works:
- [x] Automatically handle strong parameters, permitting the parameters that should be deserialized
- [x] It identify the attributes that should be deserialized by your ```attributes``` method inside an existing serializer
- [x] Allows you to override the parameters that should be serialized (ex. ```deserialize :title, :body```)
- [ ] Work with the existing adapters conventions

- JSON
 - [ ] Working CRUD
 - [ ] Parsing attributes
 - [ ] Parsing associations automatically

- JSON API
 - [x] Working CRUD
 - [x] Parsing attributes
 - [x] Parsing associations automatically

Usage:

**post_serialization.rb**
```ruby
class PostSerialization < ActiveModel::Serializer
  attributes :id, :title, :body
  belongs_to :user, root: :rows
  params :title, :body
end
```

**post_controller.rb**
```ruby

  def create
    @post = Post.new(post_params)

    if @post.save
      render json: @post, status: :created, location: @post
    else
      render json: @post.errors, status: :unprocessable_entity
    end
  end

  private

  def post_params
    PostSerialization.deserialize(params)
  end
end
```

*This is a Working In Progress yet* :stuck_out_tongue_closed_eyes:


kurko:
Would it make sense to include `ActiveModel::Deserializer` on load, instead of manually doing it in every controller?

saneshark:
:+1: 

Some thoughts I have:
1) introducing some sort of type validation around strong parameters 
2) some mechanism to sanitize or make uniform, during serialization for things such as dates. Ideally they should always be passed in as iso8601 (I assume this is what deserialize could be used for)
3) unique exceptions that can be rescued appropriately when deserialization fails, params are invalid

How will this work with PORO objects, presenter objects that accept parameters for multiple underlying objects as part of a renormalization of how the underlying data is stored vs presented?

bf4:
I guess if we're going to add another sort of transformation, I'd really like to be able to generate the serializer/deserializer from a json_schema such what a hypermedia client library like [heroics](https://github.com/interagent/heroics) or translator such as [swagger](https://github.com/swagger-rb/swagger-rb) do.

And/or maybe use transformation 'service' object such as  https://github.com/solnic/transproc

joaomdmoura:
Hey ppl, just letting you know I've being working on this. I've just updated the initial PR.

You might notice that the implementation changed a little bit. We've being improving it with Rails Team. :smile: 

@saneshark thanks for share you thoughts, indeed we will be able to help user into sanitizing their data too with this. Maybe in next steps. **For now**, deserialization would still work with PORO project because it just return and instance of the object.

@bf4 Indeed, as I said, deserialization will enable us to do a lot more, new features that will make it even easier to write APIs, I don't think the PR will handle those features right now, but some of those might follow this one, once finished and merged.

btw **this is still WIP** :smile: cc @spastorino

Yoshyn:
Hi.

It's a great idea to implement deserialization ! :)

I've got one question on it : why did you had a params key in the serializer? Don't you think that it's up to the controller to manage the authorized params?

I my mind, the _params method, it's like the attr_accessible (not doing the same thing but i the idea : it give to other object that the controlleur the definition of what is accessible or not).

joaomdmoura:
Hey @Yoshyn, tks for your comments :smile:

```params``` key is a way of override the attributes you want to deserialize. If the attributes will be the same already defined on ```attributes``` you don't need to use it. As I'm evolving it's becoming pretty clear how it will need to work close with ```StrongParameters``` and this is why I started the ```params``` approach.

I haven't updated the PR yet, but the next update will introduce the first JSONAPI crud-working implementation. I'll indeed create a deserializer module that will be included into ```Serializer``` but the already existing JSON API Adapter will handle the JSON API deserialization logic.

Here is an updated examples of the controller:

```ruby

  def create
    @post = PostSerialization.deserialize(post_params)

    if @post.save
      render json: @post, status: :created
    else
      render json: @post.errors, status: :unprocessable_entity
    end
  end

  private

  def post_params
    PostSerialization.permit(params)
  end
end
```

fsmanuel:
@joaomdmoura i took your permit approach and added a `normalize` method to transform the json into something rails can handle:

```ruby
      def permit(attributes, associations)
        relationships = associations.reduce({}) do |memo, assoc|
          memo[assoc] = { data: [:type, :id] }
          memo
        end

        permitted = params.require(:data).permit(
          attributes: attributes,
          relationships: relationships
        )

        normalize(permitted)
      end

      def normalize(params)
        relationships = params['relationships'].reduce({}) do |memo, assoc|
          if assoc.last['data'].is_a? Array
            memo["#{assoc.first}_ids"] = assoc.last['data'].map do |record|
              record['id']
            end
          elsif assoc.last['data'].is_a? Hash
            memo["#{assoc.first}_id"] = assoc.last['data']['id']
          end

          memo
        end

        params['attributes'].merge(relationships)
      end
```

README.md
```diff
@@ -4,9 +4,9 @@
 
 ActiveModel::Serializer brings convention over configuration to your JSON generation.
 
-AMS does this through two components: **serializers** and **adapters**.
-Serializers describe _which_ attributes and relationships should be serialized.
-Adapters describe _how_ attributes and relationships should be serialized.
+AMS does this through tree components: **serializers** and **adapters**.
```
fsmanuel: **deserializers**

joaomdmoura: Actually it was a TYPO :smile:
It will still be only two components (probably), but I might need to change serializer's name

joaomdmoura:
Tks for sharing @fsmanuel! The implementation I'm working on will be a little bit different because I want to keep part of the logic on the adapter but the result will be similar :smile:

bf4:
One thing that might be worth taking advantage of is that active record objects that mix in [ActiveModel::Serialization](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb) also include [ActiveModel::Serializers::JSON](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/json.rb) which not only adds an `as_json` method, from `from_json`, as well.  Might we make use of the `from_json` here?

philipp-spiess:
:+1: nice work!

I would favor a strong params approach per default. Imagine you have a `UserSerializer` with a working implementation. No you decide to add the `admin` flag to the serializer. If the params are automatically permitted based on the serializer, the application is now vulnerable although you might not even think it does (you wouldn't think so in a vanilla rails >4.0 setup).

joaomdmoura:
#### Update
*Still a WIP*

:tada:**This updates marks the first working version of create / update using JSON API conventions**:tada:
Thank you all!

--------------

I'm updating this implementation, still not sure if the deserialize should have the ability to build a model, mostly because it wouldn't need with this new implementation around `strong_parameters` convention.

As you will check on this update the `deserializer` method is called inside the `post_params`:

```ruby
def post_params
  PostSerialization.deserialize(params)
end
```

Your object creation would stay the same:

```ruby
Post.new(post_params)
# or
Post.create(post_params)
```

Because everything, including `strong_parameters` usual implementation would be handled inside `post_params`.

as @philipp-spiess  pointed out, the permitter parameter can be defined inside the `Serialization` (old `serializer`) file:

```ruby
class PostSerialization < ActiveModel::Serializer
  attributes :id, :title, :body
  belongs_to :user, root: :rows

  # permitted_params
  params :title
end
```

In this case as you can see, only `title` param is permitted.

@bf4 thank you so much for pointing it out! I'll definitely check it, I think we can use it! 

It's still a WIP, and I'm still not sure about the ability to build a model that I was working with before but ended up removed.

For those that want to check it and give it a try I'm using a [dummy app](https://github.com/joaomdmoura/AMS-Deserializer-DSL).

philipp-spiess:
I'm thinking of the following example: I'm having a user management application which allows admins to create and update users. When creating a user, it's possible to choose an email address and a password - but when updating it, it should only be possible to change the email address.

I don't think that it's easy to maintain if I need two Serialization classes now, an `UserCreateSerialization` and a `UserUpdateSerialization`. This is the kind of use case where I always liked `strong_params` - you can whitelist parameters for specific requests.

What's the return type of `Serialization#deserialize`? If this is a plain `ActionController::Parameters` we could still chain another `permit` I guess which results in the following:

```ruby
def user_upate_params
  UserSerialization.deserialize(params).permit(:email)
end
````

which would be a very good compromise for me.

**Comment by jfelchner** [comment]
Personally IMO I think that `ActionController::Parameters` has an abysmal DSL and should not be used if it can be helped.

File: test/fixtures/poro.rb [hunk 2]
```diff
@@ -1,6 +1,10 @@
 class Model
   FILE_DIGEST = Digest::MD5.hexdigest(File.open(__FILE__).read)
 
+  def self.create(params)
+    @attributes = params
+  end
```
bf4: elsewhere in this PR we have

```ruby
 @post = Post.create(post_params)
```

which leads me to believe `create` should be `new(params)`, no?

lib/active_model/serializer.rb
```diff
@@ -102,7 +111,16 @@ def self.adapter
...
+      name.demodulize.underscore.sub(/_serialization$/, '') if name
+    end
+
+    def self.deserialize(params)
+      association_keys = _reflections.map(&:name)
+      permitted_params = adapter.params(@_params, association_keys) || @_params
+      permitted_key    = adapter.root || root_name
+
+      sanitazed_params = params.require(permitted_key.to_sym).permit(permitted_params)
+      adapter.parse(sanitazed_params)
```
bf4: Now that #954 is merged, I imagine this method should be moved into `ActiveModel::SerializableResource.deserialize(resource, options)` to parallel https://github.com/bf4/active_model_serializers/blob/df140293d3010ecbbac50061c22e03a154d1a761/lib/active_model/serializable_resource.rb#L15-L27

lib/active_model/serializer.rb
```diff
@@ -102,7 +111,16 @@ def self.adapter
...
-      name.demodulize.underscore.sub(/_serializer$/, '') if name
+      name.demodulize.underscore.sub(/_serialization$/, '') if name
+    end
+
+    def self.deserialize(params)
+      association_keys = _reflections.map(&:name)
+      permitted_params = adapter.params(@_params, association_keys) || @_params
+      permitted_key    = adapter.root || root_name
+
+      sanitazed_params = params.require(permitted_key.to_sym).permit(permitted_params)
```
bf4: Is this an implicit dependency on `strong_params`? 

> params.require(permitted_key.to_sym).permit(permitted_params)

(also, typo 'sanitazed_params')

lib/active_model/serializer.rb
```diff
@@ -162,7 +180,7 @@ def self.serializers_cache
 
     def self.get_serializer_for(klass)
       serializers_cache.fetch_or_store(klass) do
-        serializer_class_name = "#{klass.name}Serializer"
+        serializer_class_name = "#{klass.name}Serialization"
         serializer_class = serializer_class_name.safe_constantize
```
bf4: @joaomdmoura how about we move the renaming commit into its own PR for clarity? :)

```diff
-        serializer_class_name = "#{klass.name}Serializer"
+        serializer_class_name = "#{klass.name}Serialization"
```

test/fixtures/poro.rb
joaomdmoura: It's because I started to write some [tests for deserialization](https://github.com/rails-api/active_model_serializers/pull/950/files#diff-81a0ce879bbbddb9593f3e0faa8c16bbR9)

lib/active_model/serializer.rb
joaomdmoura: Yeap, I totally agree! :smile:

lib/active_model/serializer.rb
joaomdmoura: It's! I want to remove the need for the user to declare it (him/her)self.
btw how do you feel about it?
Good catch with the TYPO!

lib/active_model/serializer.rb

test/fixtures/poro.rb
bf4: Right, but don't you want `Model.create(params)` to return an instance of that model with those attributes?

lib/active_model/serializer.rb
bf4: so, a couple things, how I feel about using strong_params

1. I personally never really got comfortable with strong params
1. AMS will start raising unpermitted parameter exceptions and such, which might be unexpected? The serializer, IIRC, ignores extra info, which I think is being 'liberal in what you accept'.
1. deserialize will no longer take in a hash, but an instance of [`ActionController:: Parameters`](https://github.com/rails/strong_parameters#use-outside-of-controllers) (which I suppose we could do that transformation here...)
1. it would [add a dependency on action_controller/action_pack which AMS currently does not have](https://github.com/rails/strong_parameters/blob/42397cc023b78a671580e35ba6013b6f4a8fe54a/strong_parameters.gemspec)
1. You could just `params.fetch(root) { fail KeyError, "Missing root in #{params.inspect}" }.slice(permitted_params)` (after requiring [Hash#slice](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/core_ext/hash/slice.rb))
```

**Comment by GCorbel** [comment]
I have a problem with your pull request.

I send this data `{"data":{"relationships":{"users":{"data":[{"type":"users","id":"2"}]}},"type":"trails"}}` to my application and I deserialize with `TrailSerialization.deserialize(params)`. It works until [this line](https://github.com/joaomdmoura/active_model_serializers/blob/deserializer-implementation/lib/active_model/serializer/adapter/json_api.rb#L23).

In my case, `a` is `["users", {"data"=>[{"type"=>"users", "id"=>"2"}]}]`. So `a.first` is `user` and `a.first['data'] is nil`. Even if I search the last value, I think your code will not work with an array.

Let me know if I can help.

Thanks for your help.

**Comment by joaomdmoura** [comment]
Good catch, awesome help!! I'll update this PR tonight adn will try to include a fix for it! :smile:

**Comment by beauby** [comment]
Cool work, I'll start testing it right now. Anything I can help with?

**Comment by saneshark** [comment]
Can this be adapted to work with CanCanCan? load_and_authorize_resource

**Comment by joaomdmoura** [comment]
@beauby tests are the most helpful thing right, because there are a lot of edge cases involved and I can't cover then all. :sweat_smile:  
I'm using a [dumb application](https://github.com/joaomdmoura/AMS-Deserializer-DSL) in case you want to follow up.

@saneshark I'll need to check the CanCan, I think it might work together, but I'm not sure, I'll try to build a sample scenario on the dumb app covering it.

**Comment by joaomdmoura** [comment]
@GCorbel just tested the example you mentioned [here](https://github.com/rails-api/active_model_serializers/pull/950#issuecomment-132391163) it seems to be working, actually if you check that line I do `a.shift` on the begging of the line, that should remove the user from the array.
Could you share the exception it same and some code with me through a gist?

File: lib/active_model/serializer.rb (hunk 5 above)
Comment (bf4) [inline lib/active_model/serializer.rb]: Bump

File: lib/active_model/serializer/adapter/json_api.rb [hunk 6]
```diff
@@ -4,9 +4,31 @@ module ActiveModel
...
+             relationships[assoc][@root] = DEFAULT_ATTRIBUTES
+           end
+
+          return :type, {attributes: permitted}, {relationships: relationships}
+        end
+
+        def self.parse(params)
+          attrs, assoc = {}, {}
+          attrs = params['attributes'] if params['attributes']
+          assoc = params['relationships'].map {|a| {a.shift => a.first['data']['type'].camelize.safe_constantize.find(a.shift['data']['id'])}} if params['relationships']
```
bf4: Is this a hard dep on ActiveRecord?

GCorbel:
@joaomdmoura, you're right. I checked the code and decomposed it to find the bug and I introduced a new one.

This is what I do : 

    curl 'http://localhost:3000/api/v1/trails' -H 'Origin: http://localhost:4200' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: fr,en-US;q=0.8,en;q=0.6' -H 'CSP: active' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/43.0.2357.130 Chrome/43.0.2357.130 Safari/537.36' -H 'Content-Type: application/vnd.api+json' -H 'Accept: application/vnd.api+json' -H 'Referer: http://localhost:4200/users' -H 'Connection: keep-alive' --data-binary '{"data":{"relationships":{"users":{"data":[{"type":"users","id":"2"}]}},"type":"trails"}}' --compressed

This is a copy/paste from Chrome.

This is my code :  

    module Api
      module V1
        class TrailsController < ApplicationController
          def show
            render json: Trail.find(params[:id])
          end

          def create
            render json: { trail: Trail.create(trail_params) }
          end

          def update
            trail = Trail.find(params[:id])
            render json: trail.update_attributes(trail_params)
          end

          private

          def trail_params
            TrailSerialization.deserialize(params)
          end
        end
      end
    end

And this is my error : 

    TypeError (no implicit conversion of String into Integer):
      app/controllers/api/v1/trails_controller.rb:21:in `trail_params'
      app/controllers/api/v1/trails_controller.rb:9:in `create'

Not very useful so I decomposed your code to this : 

        def self.parse(params)
          attrs, assoc = {}, {}
          require 'pry'; binding.pry
          attrs = params['attributes'] if params['attributes']
          assoc = params['relationships'].map do |a|
            key = a.shift
            value = a.first['data']['type'].camelize.safe_constantize.find(a.shift['data']['id'])
            { key => value }
          end if params['relationships']
          assoc.reduce attrs, :merge
        end

I fails at this line `value = a.first['data']['type'].camelize.safe_constantize.find(a.shift['data']['id'])`.

`a.first['data']` gives `[{"type"=>"users", "id"=>"2"}]` so with `a.first['data']['type']` I have `TypeError: no implicit conversion of String into Integer`.

I think it's a more useful description.

GCorbel:
def self.parse(params)
          attrs, assoc = {}, {}
          attrs = params['attributes'] if params['attributes']
          assoc = params['relationships'].map do |a|
            key = a.shift
            data = a.first['data']

            value = if data.kind_of? Array
              data.map do |row|
                row['type'].camelize.singularize.safe_constantize.find(row['id'])
              end
            else
              data['type'].camelize.safe_constantize.find(row['id'])
            end
            { key => value }
          end if params['relationships']
          assoc.reduce attrs, :merge
        end

Seems to work. It's uggly but it works. What you think ?

beauby:
@GCorbel Why the call to `.singularize`? The JSONAPI spec says that the type can be either singular or plural, but should always remain the same.
Also, doing something along the lines of
```ruby
data = [data] unless data.kind_of? Array
value = data.map do |row| row['type'].camelize.safe_constantize.find(row['id'])
```
instead of
```ruby
value = if data.kind_of? Array
  data.map do |row|
    row['type'].camelize.singularize.safe_constantize.find(row['id'])
  end
else
  data['type'].camelize.safe_constantize.find(row['id'])
end
```
might be clearer.

GCorbel:
Because if the type is "Users", constantize will search for the model Users and fail because of the plural name. By convention, the model name is singular.

            data = [data] unless data.kind_of? Array
            value = data.map { |row| row['type'].camelize.singularize.safe_constantize.find(row['id']) }

Works too.

I think relation can also have attributes. With this code, it will be replaced by the record in the database. I think `{"data":{"relationships":{"users":{"data":[{"type":"users","id":"2", "attributes": {"nickname": "GCorbel"}}]}},"type":"trails"}}` will not work. I think we should build a new params and reparse it with a recursive function. I will do some tests.

GCorbel:
value = data.map { |row| row['attributes'].merge(id: row['id']) }

Works too and it save and SQL request. I used `permit!` to accept all attributes. It may be a security issue. Can relationships be nested? I suppose. I think the recursive function will be useful in this case.

GCorbel:
def self.parse(params)
          attrs, assoc = {}, {}
          attrs = params['attributes'] if params['attributes']
          attrs[:id] = params['id'] if params['id'].present?
          assoc = params['relationships'].map do |a|
            key = a.shift
            data = a.first['data']

            data = [data] unless data.kind_of? Array
            value = data.map {|row| parse(row)}

            { key => value }
          end if params['relationships']
          assoc.reduce attrs, :merge
        end

Works and is recursive so relationships can have relationships. This line `attrs[:id] = params['id'] if params['id'].present?` allow to add the id to the result because it was missing.

I will stop to spam all of you for today... Sorry for the inconvenience.

saneshark:
it would be worth checking out how CanCan, linked above, builds associations. Also it should really only build the association if accepts_nested_attributes_for is specified on the model, right?

https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/controller_resource.rb#L29-L37

beauby:
@GCorbel Oh, I understand for the `.singularize` I was just puzzled why you would call it only when `data` is an array.

However, according to JSONAPI spec:
> Resource linkage MUST be represented as one of the following:
> - null for empty to-one relationships.
> - an empty array ([]) for empty to-many relationships.
> - a single resource identifier object for non-empty to-one relationships.
> - an array of resource identifier objects for non-empty to-many relationships.

and 

> Resource Identifier Objects
> A "resource identifier object" is an object that identifies an individual resource.
> A "resource identifier object" MUST contain type and id members.
> A "resource identifier object" MAY also include a meta member, whose value is a meta object that contains non-standard meta-information.

So no attributes inside a resource identifier object.

lib/active_model/serializer.rb
beauby: +1, moving the renaming commit into its own PR would make this PR much clearer.

GCorbel:
@beauby, I tink your're right. Maybe we can check with  @wycats for confirmation.

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: Yeap, deserialization itself will be deeply dependent of AR.
The idea is to deserialize to something AR understands.
One of the features we are aiming to have on Rails 5

beauby: As @GCorbel noticed,
```ruby
assoc = params['relationships'].map {|a| {a.shift => a.first['data']['type'].camelize.safe_constantize.find(a.shift['data']['id'])}} if params['relationships']
```
should be
```ruby
assoc = params['relationships'].map {|a| {a.shift => a.first['data']['type'].camelize.singularize.safe_constantize.find(a.shift['data']['id'])}} if params['relationships']
```
since the JSONAPI `type` attribute can be plural (and it it serialized in a plural form by active_model::serializer), but the corresponding `ActiveModel` class name is singular.

GCorbel:
What can we do with recursive nested attributes. Do we just ignore them?

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: Awesome, I'm updating it.

beauby:
There's nothing to be done as it is not part of the spec.

Le vendredi 21 août 2015, Guirec Corbel <notifications@github.com> a écrit :

> What can we do with recursive nested attributes. Do we just ignore them?
>
> —
> <https://github.com/rails-api/active_model_serializers/pull/950#issuecomment-133233690>
> .
>


-- 
Lucas Hosseini
lucas.hosseini@gmail.com

joaomdmoura:
Hey everyone, thanks for all the help, I reading through everything u comment, and updating the implementation.
The `singularize` thing was a good catch!

@saneshark I also just installed CanCan on my dumb project to check it out how to make both work together.

I'll might split the renaming into a new PR, indeed it might make it easier to review.

lib/active_model/serializer/adapter/json_api.rb
beauby: Also, as per the specification, the `data` attribute of a relationship object can be an array of resource identifiers. Hence, that part should actually be something along the lines of:
```ruby                                                                                                                                                                                                       
        def self.parse(params)
          attrs, assoc = {}, {}
          attrs = params['attributes'] if params['attributes']
          assoc = params['relationships'].map do |rel|                                                                                                                                                      
            key, data = rel.shift, rel.first['data']                                                                                                                                                        
            value = if data.kind_of? Array                                                                                                                                                                  
                      data.map { |ri| self.parse_resource_identifier ri }                                                                                                                                   
                    elsif data                                                                                                                                                                              
                      self.parse_resource_identifier data                                                                                                                                                   
                    else                                                                                                                                                                                    
                      nil                                                                                                                                                                                   
                    end                                                                                                                                                                                     
            {key => value}                                                                                                                                                                                  
          end if params['relationships']                                                                                                                                                                    
          assoc.reduce attrs, :merge
        end

      private:
        def self.parse_resource_identifier(ri)                                                                                                                                                              
          ri['type'].camelize.singularize.safe_constantize.find(ri['id'])                                                                                                                                   
        end         
```

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -4,9 +4,31 @@ module ActiveModel
...
     class Adapter
       class JsonApi < Adapter
+        @root = :data
+        DEFAULT_ATTRIBUTES = [:type, :id]
+
+        def self.params(permitted, associations)
+          relationships = {}
+          associations.each do |assoc|
+             relationships[assoc] = {}
+             relationships[assoc][@root] = DEFAULT_ATTRIBUTES
```
beauby: This does only allow a single resource identifier object. However, the spec allows for *either* a single resource identifier object, *or* an array of resource identifier objects. Not sure how to handle the situation with strong parameters though.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -4,9 +4,31 @@ module ActiveModel
...
+           end
+
+          return :type, {attributes: permitted}, {relationships: relationships}
```
beauby: `:id` should be allowed here, in order to handle client-generated IDs (http://jsonapi.org/format/#crud-creating).

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -4,9 +4,31 @@ module ActiveModel
...
+        end
+
+        def self.parse(params)
+          attrs, assoc = {}, {}
+          attrs = params['attributes'] if params['attributes']
+          assoc = params['relationships'].map {|a| {a.shift => a.first['data']['type'].camelize.singularize.safe_constantize.find(a.shift['data']['id'])}} if params['relationships']
```
beauby: The `.find()` call here is actually not needed, as we can get away with just the IDs. I think something along the lines of: https://gist.github.com/beauby/6e6fa23d09aa0745ad58 would give an ActiveRecord-friendly payload, while not relying explicitly on it.

GCorbel:
@joaomdmoura , I sill have the same problem with arrays. The id is missing also.

@beauby , in this line : `key, data = rel.shift.camelize, rel.first['data']` you don't need to camelize because you don't need to constantize. `{"users": ..}` is transformed it `{"Users_ids": ...}` which is wrong.  I this it have to be `user_ids`.

beauby:
@GCorbel Spot on.

I made a separate PR (#1070) incorporating various elements discussed here, and adding a few tests, in case somebody is willing to add more tests.

GCorbel:
@beauby Rock on :+1:  !!

joaomdmoura:
@beauby as I mentioned in the other PR, I can't merge that one by itself, the best thing to do is to open a PR to my branch on my fork, then we can keep everything organized here and merge it all in once.

beauby:
@joaomdmoura There should not be any need to merge with your PR as I based my branch on master and cherry-picked all the relevant parts of your PR (leaving out the renaming Serializer -> Serialization). 
If you need to have it in a separate branch you might as well just clone it in a new branch, which would avoid the merging headaches.

beauby:
@joaomdmoura Let me know, if you really prefer the merge option, I'll do it later today.

joaomdmoura:
@beauby the PR is about deserialization as well, there is no need to that piece of code to exist without this code as well.
Yesterday I also removed the naming change on this PR. :smile: 
Would also be nice to be able to concentrate the effort and discussion on a single place.
Btw I didn't reviewed it yet, but tks for the effort! I'm sure it is a good work!

beauby:
Done :+1:

test/deserializers/json_api_integration_test.rb
```diff
@@ -0,0 +1,69 @@
+require 'test_helper'
+
+module ActionController
+  module Serializer
+    class JsonApiIntegrationTest < ActionController::TestCase
+      class JsonApiIntegrationTestController < ActionController::Base
+        def create_resource
+          with_adapter ActiveModel::Serializer::Adapter::JsonApi do
+            @post = Post.create(post_params)
+            render json: @post, status: :created
```
beauby: How come the response here is not serialized in JsonApi format?

beauby:
@bf4 Any thought on how to encapsulate deserialization in `ActiveModel::SerializableResource`? Currently it does not seem obvious to me, as it seems to me it is the responsibility of the serializer.

joaomdmoura:
`ActiveModel::SerializableResource` becomes a middleware that handles the integration between `Adapters` and `Serializers`, IMO it could be responsible for deserialization as well, taking into consideration the logic is splitted between `Adapters` and `Serializers`. I'll probably move the `deserialization method to it.

beauby:
@joaomdmoura I get that, and I'd personally appreciate the symmetry of having both `serialize` and `deserialize` methods at the same place. However, we'd still need to pass the serializer to the method, something like `ActiveModel::SerializableResource.deserialize(PostSerializer, params)`, right? Or am I missing something?

bf4:
@beauby The interface would be `ActiveModel::SerializableResource.deserialize(json_document, options)` and it would need to infer the serializer for each root, not unlike how it infers the serializer from the resource.  (maybe also see https://github.com/rails-api/active_model_serializers/pull/950#discussion_r36134190 )

beauby:
@bf4 Oh, right. Great then!

Gemfile
```diff
@@ -3,7 +3,11 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in active_model_serializers.gemspec
 gemspec
 
-gem "minitest"
+group :development do
+  gem "minitest"
+  gem "sqlite3", platform: :ruby
+  gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
+end
```
bf4: These being test dependencies should be in the gemspec, no?

beauby: I would agree, but since "minitest" was there already, I did as the Romans do.

lib/active_model/serializer.rb
andrewhavens: I would agree that the deserializer should not be responsible for determining which params are required/permitted. That is the developers responsibility. However, it would be nice to adhere to the same API that Rails provides (so in the case of a JSON API request body, parse the body and add the data to the action controller params hash).

joaomdmoura: @andrewhavens you still will be abel to costumize it but through serializers, how does it feels to you?

Gemfile
joaomdmoura: > I did as the Romans do.

:joy: 
Yup.

joaomdmoura:
It took longer than I expected but this is rebased, finally.

lib/active_model/serializer.rb
beauby: @joaomdmoura As mentioned [there](https://github.com/joaomdmoura/active_model_serializers/pull/6) I forgot to remove this method while merging our forks (you can see that currently the method `deserialize` is defined twice).

test/action_controller/explicit_serializer_test.rb
```diff
@@ -1,14 +1,14 @@
...
-  module Serialization
+  module Serializer
     class ExplicitSerializerTest < ActionController::TestCase
       class ExplicitSerializerTestController < ActionController::Base
         def render_using_explicit_serializer
           @profile = Profile.new(name: 'Name 1',
                                  description: 'Description 1',
                                  comments: 'Comments 1')
-          render json: @profile, serializer: ProfilePreviewSerializer
+          render json: @profile, serializer: ProfilePreviewSerializer
```
beauby: Any reason for this extra space?

GCorbel:
It seems to work for me. Thanks for your great job!

beauby:
So, what's the status on this one?

Dschee:
Yeah, same question here, when will the deserializer implementation for the JSON API adapter be merged into the master? Maybe it should be merged before implementation for the JSON adapter ist ready – otherwise we'll need to wait even longer until we can use this on the official gem.

Also I suggest to include some usage to the README.

Dschee:
I have also just read a lot more into this and tried to get deserialization to work on my real-world app that I'm developing right now – but there's **at least two things that don't work** yet:

1. The **`deserialize` method** doesn't work properly (both with and without whitelisted `password`)
2. A related object **doesn't get build alongside my object** (if no id is specified it should be build)

Although the last point isn't necessarily a part of the JSON API specification (as far as I could see) there still needs to be a solution on how to deal with **bidirectional presence validations** like this: A `User` who needs to have at least one `spoken_languages` set and a `SpokenLanguage` which belongs to a `user` and therefore needs him set, too. I could, of course, remove validations to get it working, but that's not the way to go, is it?

You can check out the entire project for a **working example** [here](https://github.com/Flinesoft/MovieAssistor/tree/feature/json-api), but I'm quoting the parts in question below, too, too keep things clean. When trying to fix this it's probably easiest though to just checkout the projects `json-api` branch and run the `rspec` command.

---

Here's my Serializers:
``` ruby
##
# Describes serialization of a User to a JSON format and vice versa according to the JSON API specification.
  attributes :id, :first_name, :last_name, :nickname, :email, :gender, :birthday, :country_code, :image

  has_many :spoken_languages
end

##
# Describes serialization of a SpokenLanguage to a JSON format and vice versa according to the JSON API specification.
class SpokenLanguageSerializer < ActiveModel::Serializer
  attributes :id, :language_code

  belongs_to :user
end
```

Here's my Controller:
``` ruby
##
# The API V1 controller managing users, logins/logouts, registrations, password resets and the like.
class Api::V1::UsersController < Api::V1Controller

  # Creates a new user with given parameters and renders a JSON response.
  #
  # @return [void]
  def create
    @user = User.new(user_params)
    if @user.save
      render json: object, status: :created, location: object
    else
      render json: { errors: @user.errors }, status: :unprocessable_entity
    end
  end

private

  # Deserializes the users parameters into ActiveRecord accessable format and filters unpermitted values.
  #
  # @return [Hash] permitted attributes for a user
  def user_params
    UserSerializer.deserialize(params, [:password])
  end

end

```

And here's my spec:

``` ruby
it 'responds with created including data for valid data' do
  post :create, {
    data: {
      type: 'users',
      attributes: attributes_for(:user), # FactoryGirl factory is working (linter-approved)
      relationships: {
        spoken_languages: { data: [{ type: 'spoken_languages', attributes: attributes_for(:spoken_language) }] }
      }
    }
  }
  expect(response).to have_http_status(:created)
  # more expectations
end
```

beauby:
@Dschee Thanks for the thoughts you put into this one.
Concerning your first point, the feature is clearly not ready for production yet. You could help by opening an issue stating how and when the deserialization fails.
Concerning your second point, the JSONAPI spec explicitly states:
> If a relationship is provided in the relationships member of the resource object, its value MUST be a relationship object with a data member. The value of this key represents the linkage the new resource is to have. (http://jsonapi.org/format/#crud-creating)

and

> Resource linkage MUST be represented as one of the following:
> + null for empty to-one relationships.
> + an empty array ([]) for empty to-many relationships.
> + a single resource identifier object for non-empty to-one relationships.
> + an array of resource identifier objects for non-empty to-many relationships.
>
> (http://jsonapi.org/format/#document-resource-object-linkage)

Therefore, a related object should not be built. Only existing objects should be supplied. In your case, you could either change your API to comply to the spec, or build a custom adapter based on the JSONAPI one. 
However, it would probably make sense with the JSON adapter.

sahal2080:
Not sure what's the status here as of now, so I just wanted to let those like me who were looking for AMS-y deserializing about an alternative [Deserializer](https://github.com/gaorlov/deserializer/tree/has_many) gem. It pretty much follows the API conventions of AMS. Hope it helps others and it might even benefit the core devs in picking/not picking an approach, API, etc.

beauby:
Thanks for the heads up @sahal2080!

beauby:
Should we close this now that #1248 has been merged?

kurko:
@joaomdmoura?

joaomdmoura:
@kurko we implemented a new one, that was merged already and will be on RC4 soon :smile:

benlieb:
> Not sure what's the status here as of now, so I just wanted to let those like me who were looking for AMS-y deserializing about an alternative [Deserializer](https://github.com/gaorlov/deserializer/tree/has_many) gem. It pretty much follows the API conventions of AMS. Hope it helps others and it might even benefit the core devs in picking/not picking an approach, API, etc.

This link is now here [Deserializer](https://github.com/gaorlov/deserializer/)

PR #952, opened 2015-06-12
added option to include nested associations

iMacTia:
Devs can now use a new parameter in Serializer classes when defining associations (has_many, has_one, ...): `include_nested_assosiations`.

If this option is true, serialization will "follow" the association including not only its attributes, but also its associations.

### DIFF
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -19,13 +19,17 @@ module ActiveModel
                 array_serializer = association
                 @hash[name] = array_serializer.map do |item|
                   cache_check(item) do
-                    item.attributes(opts)
+                    opts[:include_nested_associations] ?
+                      self.class.new(item).serializable_hash :
+                      item.attributes(opts)
                   end
                 end
               else
                 if association && association.object
                   @hash[name] = cache_check(association) do
-                    association.attributes(options)
+                    opts[:include_nested_associations] ?
+                      self.class.new(association).serializable_hash :
+                      association.attributes(options)
                   end
                 elsif opts[:virtual_value]
                   @hash[name] = opts[:virtual_value]
--- a/test/serializers/include_nested_associations_test.rb
+++ b/test/serializers/include_nested_associations_test.rb
@@ -5,7 +5,7 @@ module ActiveModel
     class IncludeNestedAssociationsTest < Minitest::Test
       def setup
         @author = Author.new(name: 'Steve K.')
-        @author.bio = Bio.new(content: 'Hello!', rating: 5)
+        @author.bio = Bio.new(id: 1, content: 'Hello!', rating: 5)
 
         @post = Post.new({ title: 'New Post', body: 'Body' })
 
@@ -24,7 +24,12 @@ module ActiveModel
       def test_no_option_is_passed_in
         @post_serializer = PostSerializerWithNestedAssociations.new(@post)
         json = ActiveModel::Serializer::Adapter::Json.new(@post_serializer).as_json
-        assert json[:author][:bio] != nil
+        expected_return = {
+          id: 1,
+          content: 'Hello!',
+          rating: 5
+        }
+        assert_equal json[:author][:bio] , expected_return
       end
     end
   end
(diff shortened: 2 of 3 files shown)


joaomdmoura:
Great @iMacTia! I was really looking forward to implement it!
Could you write so tests to cover the usual and some corner cases? :smile:

iMacTia:
Added some basic tests, unfortunately was my first time with Minitest, I found myself a bit disoriented @_@

test/serializers/include_nested_associations_test.rb
```diff
@@ -0,0 +1,31 @@
...
+      def test_no_nested_association_are_included_by_default
+        @post_serializer = PostSerializer.new(@post)
+        json = ActiveModel::Serializer::Adapter::Json.new(@post_serializer).as_json
+        assert_nil json[:author][:bio]
+      end
+
+      def test_no_option_is_passed_in
+        @post_serializer = PostSerializerWithNestedAssociations.new(@post)
+        json = ActiveModel::Serializer::Adapter::Json.new(@post_serializer).as_json
+        assert json[:author][:bio] != nil
```
joaomdmoura: @iMacTia could you just assert the content itself? So we also validate that is matches the expected one :smile: 
```ruby
expected_return = {
  ...
}
assert_equal json[:author][:bio] , expected_return
```

lib/active_model/serializer/adapter/json.rb
```diff
@@ -19,13 +19,17 @@ def serializable_hash(options = {})
                 array_serializer = association
                 @hash[name] = array_serializer.map do |item|
                   cache_check(item) do
-                    item.attributes(opts)
+                    opts[:include_nested_associations] ?
+                      self.class.new(item).serializable_hash :
+                      item.attributes(opts)
```
bf4: This might be a good time to extract `add_association` to its own method, like the json_api adapter has

joaomdmoura: :+1:

bf4:
@iMacTia need any help bringing this to completion? I'm happy to pair

iMacTia:
I thought you guys were working on it (extracting `add_association` or whatever), I'm currently using my branch in my project and it works like a charm. CI is not passing because of tests, but I'm not very good with the tools you've chosen...

joaomdmoura:
Yeap @iMacTia, indeed I'm inclined to merge the [Associations Refactoring](https://github.com/rails-api/active_model_serializers/pull/985), but I still think that nested associations are a need, we would still be willing to merge it. Maybe on top or the associations refactoring after I merge it.

iMacTia:
Hi @joaomdmoura, just wanted to know what's the plan for this feature, are you still waiting for something from me? We've been using my branch in production for some months now, seems working with no problems. I see a similar problem has been addressed for side-loading, but looks to me like a different feature after all

casiodk:
Hi what is the status here. I would also like to see the possibilty of deep nesting associations, as I´m converting from jbuilder to AMS right now...

iMacTia:
Hi @casiodk, Just asked the same yesterday :)
Hopefully we'll get back from them shortly.
In the meantime, trust me if I tell you that you made the best choice in switching to AMS.

casiodk:
Yes it seems like the best choice, but not as flexible

joaomdmoura:
Hey @iMacTia and @casiodk, first of all thank you for the support and interest on this. :smile: 

Some ppl have being addressing this with different approaches, I think this PR is pretty outdated right now, and I liked the discussion on #1113, and reviewed #1114. 
So I'm closing this one in favor of #1114, It solves the same problem, but is up-to-date and also do some nice refactor on the adapter.

I totally agree with your comment on #1113 

> Recursive loop problem should be fixed as well, the problem is that if you want a standard behaviour (like you always want the comment author when you get a comment), you have to copy-and-paste the same "include" option in all serialisers, where with the include_nested_associations you delegate this to the serializer itself and you just need to manage special cases.

This might be a real need for some ppl, @NullVoxPopuli even proposed a solution for that [here](https://github.com/rails-api/active_model_serializers/issues/1113#issuecomment-137649920). Let's go for nested association and see if more ppl back up the `nest everything` approach, if this get backed then we can work on it, if you really wanted it, you could work on a PR on top of #1114 once it gets merged, it won't be a priority but would check it out :smile:

PR #953, opened 2015-06-12
use model name to determine the type

lsylvester:
I propose changing the `Serializer#type` from being based on the classes model_name instead of being based on the class itself.

This provides two advantages:

* It allows the type to be customized at the model layer (currently can only be done at the serializer layer), and the type will be consistent with the other places that rails uses names (like routes).
* The `model_name` memoized so the underscoring and pluralization are only done once, so it should be faster.

It changes the behaviour of namespaced models (was previously demodulized), but the introduction of the demodulization was to change it from `spam/unrelated_links`. I don't know if there is any reason that `unrelated_links` would be preferable to `spam_unrelated_links`

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -188,7 +188,7 @@ module ActiveModel
     end
 
     def type
-      object.class.to_s.demodulize.underscore.pluralize
+      object.class.model_name.plural
     end
 
     def attributes(options = {})
--- a/test/adapter/json_api/linked_test.rb
+++ b/test/adapter/json_api/linked_test.rb
@@ -203,7 +203,7 @@ module ActiveModel
             assert_equal expected, alt_adapter.serializable_hash[:included]
           end
 
-          def test_ignore_model_namespace_for_linked_resource_type
+          def test_underscore_model_namespace_for_linked_resource_type
             spammy_post = Post.new(id: 123)
             spammy_post.related = [Spam::UnrelatedLink.new(id: 456)]
             serializer = SpammyPostSerializer.new(spammy_post)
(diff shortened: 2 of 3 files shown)


joaomdmoura:
@lsylvester Thank you for this PR.
I'm okay with the proposal, it makes sense to me. @kurko any opinions about it? I checked [json api docs](http://jsonapi.org/format/#document-resource-object-identification) and it still is consistent to it.
Could you please rebase it with master to fix the conflicts?

lsylvester:
@joaomdmoura I have rebased this against master

kurko:
:+1: I like it.

fbernier:
I like the idea but see a problem with this implementation. Currently, it's quite easy to use AMS outside of active_model by simply including ActiveModel::Serialization into an object, for example: Hashie::Mashie wrapping an Elasticsearch result.

I haven't tested if this commit breaks anything but model_name is active_model specific and might make it more complicated to use outside of active_model.

Is this really what you guys want?

joaomdmoura:
@fbernier indeed, this is a good point. 

As AMS is becoming part of Rails itself it's hard to deny that it's deeply associated with ActiveModel, even our tests are all focused into models, but in the other hand I'm really happy that people are able to use AMS in unusual ways :smile: 

It seems that this PR itself isn't breaking the usage of AMS as long as you are serializing an ActiveRecord Object, as we are calling ```model_name ``` into the ```object``` instance that is being serialized, this isn't the only dependency of AR, a lot of other methods into AMS expect the object being serialized to be an AR object, so I think that you will probably be good to go. :smile:

kurko:
@fbernier if this breaks in any way, just create a method called `model_name`. Perhaps the name `model_name` could be broader, such as `entity_name` so that it doesn't feel like you're dealing with Rails on your ES results.

You can also override the `serializer#type`.

lsylvester:
@fbernier You can add the `model_name` method to non AR classes by adding `extend ActiveModel::Naming`

Issue #945, opened 2015-06-09
Serialize Object/Array when no AMS Serializer is defined


kayakyakr:
I'm receiving an issue, in latest master, where I receive ```NoMethodError (undefined method `new' on nil:NilClass.)``` when I attempt to serialize something like, ['Result is invalid'].

From everything I can tell, it's rolling through the get_serializer_for method, failing to find a serializer for my object, but trying to create an instance of that serializer anyway. It should fall back to a default serializer.

Sorry if this has been addressed, I checked through the issues but didn't see anything that stood out to me. As a workaround, I'm having to specify "render json: obj, adapter: false" when I want to render a json-ready hash rather than a rails model.

bf4:
Needs more info.  What is obj? Which ref are you on? (paste output of `bundle show active_model_serializers`) Which Ruby?  Which Rails?  What does your serializer look like?  Could you provide some more of the stack trace?

kayakyakr:
active_model_serializers-7b0a85fdda85
rbx 2.2.10
rails 4.1

code is called as:

```Ruby
render json: {error: 'Result is Invalid'}
# throws exception because serializer is nil
```

Stack trace is less than interesting and I don't currently have an example set up to cause the error, but digging through the code:

```Ruby
#action_controller/serialization.rb:23
@_serializer ||= ActiveModel::Serializer.serializer_for(resource) # @_serializer is nil when resource is not explicitly specified.

#action_controller/serialization.rb:47
object = serializer.new(resource, @_serializer_opts) # calls new on serializer when it may be nil
```

Is ```render json: {}``` supposed to fall back to a more appropriate renderer?

bf4:
perfect. So, the bug is there's no serializer for Hash. solutions: call 
`to_json` on your hash, and the resulting string will be your response 
body :)

kayakyakr:
That's a pretty ugly solution. I like my adapter: false workaround better.

bf4:
I suppose we could argue this is a bug in AMS in that if it can't find a serializer for non-string object, it should either just pass it through untouched or call to_json on it depending on the context.

kayakyakr:
That would be the best, I think.

I also noticed that if you pass it a model that you haven't built a serializer for (yet or at all), it fails in the same way. Previous versions of AMS either passed those objects on to a lower renderer or just called to_json on them. Also chokes in the same way with an association without a serializer.

bf4:
[...]

PR #954, opened 2015-06-12
Encapsulate serialization in ActiveModel::SerializableResource

bf4:
Usage: `ActiveModel::SerializableResource.serialize(resource, options)`

tl;dr as seen in [test/serializable_resource_test.rb](https://github.com/rails-api/active_model_serializers/blob/df140293d3010ecbbac50061c22e03a154d1a761/test/serializable_resource_test.rb) `ActiveModel::SerializableResource.new(@resource)` can now be tested outside of a controller instead of the (more or less) equivalent code `ActiveModel::Serializer::Adapter.create(ProfileSerializer.new(@resource))`

ref: discussion in https://github.com/rails-api/active_model_serializers/pull/936#discussion-diff-31732511

This PR includes

1. a simple interface for building the adapter
2. a simpler interface for serializing a resource outside of a controller for e.g. testing, background jobs, etc.  
3. the SerializableResource has all the expected methods:  as_json/to_json/serializable_hash
4. Uses the `get_serializer` controller method which may possibly better extensibility, and possibly better compatibility with Rails serializer proposal in https://groups.google.com/d/topic/rubyonrails-core/K8t4-DZ_DkQ/discussion
5. Deprecates the controller method `use_adapter?` and displays a warning when falsy.
6. Deprecates calling ActionController::Serialization::ADAPTER_OPTION_KEYS in favor of ActiveModel::SerializableResource::ADAPTER_OPTION_KEYS
7. Provides a `ActiveModel::SerializableResource.serialize` method which implies #950 may add the interface `ActiveModel::SerializableResource.deserialize`


lib/action_controller/serialization.rb
```diff
@@ -18,34 +19,25 @@ def serialization_scope
...
-          @_serializer_opts[:scope_name] = _serialization_scope
-
-          # omg hax
-          object = serializer.new(resource, @_serializer_opts)
-          adapter = ActiveModel::Serializer::Adapter.create(object, @_adapter_opts)
+        builder = ActiveModel::Serializer::Builder.new(resource, options)
+        if builder.serializer?
+          builder.serialization_scope ||= serialization_scope
+          builder.serialization_scope_name = _serialization_scope
+          adapter = builder.adapter
```
bf4: :100: right?

lib/action_controller/serialization.rb
```diff
@@ -53,11 +45,22 @@ def use_adapter?
...
+    #     exception_model = ActiveModel::Serializer.build_exception_model({ errors: ['Internal Server Error'] })
+    #     render json: exception_model, status: :internal_server_error
+    #   end
+    # OR remove method as it doesn't do anything right now.
     def rescue_with_handler(exception)
-      @_serializer = nil
-      @_serializer_opts = nil
-      @_adapter_opts = nil
-
       super(exception)
```
bf4: Method wasn't being used, so wrote a comment to that effect, as it's outside the scope of this PR.  (There's a test in `test/action_controller/rescue_from_test.rb` but it doesn't depend on any code in this method and passes when the method is removed.

lib/action_controller/serialization.rb
```diff
@@ -18,34 +19,25 @@ def serialization_scope
...
-      @_serializer ||= ActiveModel::Serializer.serializer_for(resource)
-
-      if @_serializer_opts.key?(:each_serializer)
-        @_serializer_opts[:serializer] = @_serializer_opts.delete(:each_serializer)
-      end
-
-      @_serializer
+      fail ActiveModel::Serializer::Error, "ActionController::Serialization##{__method__} has been removed. "\
+        "Please use ActiveSupport::Serializer#build instead"
     end
```
bf4: Thought this was better than a straight-up deprecation just in case anyone is using these methods. But this won't handle people that override the methods and expect them to have some effect.  I'm not sure the best way to handle that.

lib/action_controller/serialization.rb
```diff
@@ -18,46 +19,49 @@ def serialization_scope
...
+          if builder.serializer?
+            builder.serialization_scope ||= serialization_scope
+            builder.serialization_scope_name = _serialization_scope
+            builder.adapter
+          else
+            resource
+          end
         end
+        super(serializable_resource, options)
       end
```
bf4: the Controller using the builder block form to conditionally return the adapter or the resource

bf4:
If this looks good, I'll update the README as well.

There are other places the builder can (should?) be applied in the tests, but nothing that I thought was necessary for this PR.

I didn't add explicit tests for the builder since it's essentially an extracted method object and is covered by existing tests.  Eventually, some tests should be extracted/added to it.

This PR moves us closer to 1) having a simple interface for building the adapter 2) using the simpler interface for testing serializers in our apps outside of the controller (e.g. [some hacks I've written](https://github.com/rails-api/active_model_serializers/issues/878#issuecomment-96073030) )

lib/active_model/serializer.rb
```diff
@@ -2,10 +2,12 @@
 
 module ActiveModel
   class Serializer
+    Error = Class.new(StandardError)
```
joaomdmoura: @bf4 any particular reason to implement this ```Error``` class across the PR?

lib/action_controller/serialization.rb
```diff
@@ -18,46 +19,49 @@ def serialization_scope
...
-          super(adapter, options)
-        else
-          super(resource, options)
+        serializable_resource = ActiveModel::Serializer::build(resource, options) do |builder|
+          if builder.serializer?
+            builder.serialization_scope ||= serialization_scope
+            builder.serialization_scope_name = _serialization_scope
+            builder.adapter
+          else
+            resource
```
joaomdmoura: Does it affects #945 or #955 somehow?

lib/action_controller/serialization.rb
joaomdmoura: I'm not sure about this method either, I'm in favor of removing it. 
@kurko do u know if is there some corner-tricky-case here?

lib/active_model/serializer.rb
```diff
@@ -28,6 +30,18 @@ def self.inherited(base)
...
+    # Otherwise, yields the builder and returns the contents of the block
+    def self.build(resource, options = {})
+      builder = Builder.new(resource, options)
+      if block_given?
+        yield builder
+      else
+        builder.adapter
+      end
+    end
+
```

joaomdmoura:
@bf4 Thank you for your work on this.
I liked the implementation itself, but so many deprecations :cry:
As we are on RC1 I think we could merge it, but first I would like to hear thoughts of other AMS members.

Summoning @kurko @guilleiguaran :

ppl just to clarify. The idea here is to make it easier to use AMS outside controllers. Test scenarios or background jobs for example.

Right now it's a kind messy have to instantiate the ```serializer```, the ```adapter``` and then call ```.as_json``` into it.

I made some pontual comments into the PR itself, but overall I liked the implementation, just worries about the deprecations and how it can impact other developers. Wanted to hear your thoughts about it :smile:

lib/action_controller/serialization.rb
kurko: This is good. We're currently in an RC, so people will have time to see these deprecations.

lib/action_controller/serialization.rb
kurko: I've never crossed this method myself, so I lack context. Are you absolutely sure it's not used anywhere? If it's not, I'd advise removing it and then adding a solution in another PR.

lib/action_controller/serialization.rb
bf4: There's no behavior change in this method, just using the extracted code, so no.

bf4: However, I think that we should try to use the builder interface as much as possible in specs from now on, passing in :adapter, :serializer, :each_serializer, etc., and only instantiating adapters and serializers directly in builder tests (which don't yet exist of course), and possibly in serializer tests.  I think it would be a big win if we could narrow the necessary/recommended/whatever public interface to the builder and maybe something else, but I think passing `adapter: false` would still be better in those cases.

lib/action_controller/serialization.rb
bf4: @kurko did you look at my comments in the code above the method? I explain what I know about it. Pasting below for convenience.

> Method wasn't being used, so wrote a comment to that effect, as it's outside the scope of this PR. 
> (There's a test in test/action_controller/rescue_from_test.rb but it doesn't depend on any code in this method and passes when the method is removed.)

> ```diff
+    # Tries to rescue the exception by looking up and calling a registered handler.
+    #
+    # Possibly Deprecated
+    # TODO: Either Decorate 'exception' and define #handle_error where it is serialized
+    # For example:
+    #   class ExceptionModel
+    #     include ActiveModel::Serialization
+    #     def initialize(exception)
+    #     # etc
+    #   end
+    #   def handle_error(exception)
+    #     exception_model = ActiveModel::Serializer.build_exception_model({ errors: ['Internal Server Error'] })
+    #     render json: exception_model, status: :internal_server_error
+    #   end
+    # OR remove method as it doesn't do anything right now.
     def rescue_with_handler(exception)
-      @_serializer = nil
-      @_serializer_opts = nil
-      @_adapter_opts = nil
```

File: lib/action_controller/serialization.rb (hunk 3 above)
Comment (bf4) [inline lib/action_controller/serialization.rb]: Glad you support that. I'm really unhappy with straight-up removing them, but didn't come up with any code solutions that I thought were an improvement over just failing.

Although, I think my implementation is wrong, in that if someone overrides these methods, they won't get a failure.  So, that leaves me with doing a method_defined? check at boot/inheritance and/or taking a look at method missing.. ugh.. I'm going to take another look to see if I can keep the code alive a little longer.

File: lib/active_model/serializer.rb (hunk 5 above)
Comment (bf4) [inline lib/active_model/serializer.rb]: I added it so that I could raise an AMS-specific failure in the 'deprecated' controller methods, and there wasn't any Error class already defined.  I wanted to make sure when testing, for example, that I'm asserting that I'm raising a Runtime error and that it actually is the one from the method, rather than an accident elsewhere in the code... I could check the exception message, I think.

**Comment by bf4** [comment]
For easy of review, I'm going to push changes as new commits, and rebase  when done.

File: lib/active_model/serializer.rb [hunk 8]
```diff
@@ -2,10 +2,12 @@
 
 module ActiveModel
   class Serializer
+    Error = Class.new(StandardError)
     extend ActiveSupport::Autoload
     autoload :Configuration
     autoload :ArraySerializer
     autoload :Adapter
+    autoload :Builder
```
kurko: Now that I look at it, shouldn't we be moving away from `autoload` to a thread safe world?

bf4: autoload is threadsafe for some time now. See https://github.com/headius/thread_safe/issues/15#issuecomment-28936120 and even if it weren't `extend ActiveSupport::Autoload` would make it so.

lib/active_model/serializer.rb
```diff
@@ -159,7 +173,10 @@ def initialize(object, options = {})
       @meta_key   = options[:meta_key]
       @scope      = options[:scope]
 
+      # SCOPE
       scope_name = options[:scope_name]
+      # Adds a class method with ':scope_name' that calls ':scope()'
+      # But isn't called anywhere
```
bf4: I don't intend to include this comment in the final PR, but wanted to bring attention to scope/scope_name now being dead code, in case you'd like me to remove calling it from the controller and the builder.

lib/active_model/serializer/adapter.rb
```diff
@@ -16,16 +16,17 @@ def initialize(serializer, options = {})
         @options = options
       end
 
-      def serializable_hash(options = {})
+      def serializable_hash(options = nil)
         raise NotImplementedError, 'This is an abstract method. Should be implemented at the concrete adapter.'
```
bf4: See https://github.com/bf4/active_model_serializers/commit/6ed30e3493a97f2232b9d2cda11abeae84a1e81e for why I changed the options to nil here and elsewhere

lib/active_model/serializer/builder.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter])
+
+      def initialize(resource, options = {})
+        @resource = resource
+        @adapter_opts, @serializer_opts =
+          options.partition { |k, _| ADAPTER_OPTION_KEYS.include? k }.map { |h| Hash[h] }
+      end
+
+      delegate :serializable_hash, :as_json, :to_json, to: :adapter
```
bf4: convenience methods. Calling ActiveModel::Serializer.build without a block returns the builder, but you can still call to_json, as_json, serializable_hash on it.

test/action_controller/serialization_test.rb
```diff
@@ -378,6 +375,45 @@ def test_cache_expiration_on_update
...
+      def test_fail_calling_use_adapter_on_controller_that_overrides_use_adapter
+        controller = Class.new(ImplicitSerializationTestController) {
+          def use_adapter?
+            false
+          end
+        }.new
+        assert_raises(ActiveModel::Serializer::Error) {
+          controller.use_adapter?
+        }
+      end
```
bf4: The two failing controller tests are because the 'deprecated' controller methods that raise the error are being overridden, and hence not called.  I'm not sure how to handle that, so adding this failing test to call attention to it.

lib/action_controller/serialization.rb
joaomdmoura: #917 and #918

bf4: So, remove it, fantastic

bf4:
Moved functionationality from `render_with` to `get_serializer` both for less deprecations, possibly better extensibility, and possibly better compatibility with Rails serializer proposal in     https://groups.google.com/d/topic/rubyonrails-core/K8t4-DZ_DkQ/discussion

I think the current state is better than before.

bf4:
While looking at this, I found some of the 0-8-stable stuff interesting

- benchmarking serialization https://github.com/rails-api/active_model_serializers/pull/233 (I think this has basically been moved into Rails)
- https://github.com/rails-api/active_model_serializers/blob/0-8-stable/DESIGN.textile
- Usage stuff from the README https://github.com/rails-api/active_model_serializers/blob/0-8-stable/README.md

Anyhow, I plan on rebasing this and removing a bunch of stuff.  Let me know what you think.

joaomdmoura:
@bf4 just letting u know.
I'm aware the effort you are putting into this. It's getting bigger haha so I'm going to try to review it this week in order to move forward.

btw do you have another name suggestion different than ```builder```? :grin:

About the topics u brought related to 0.8.x
- Benchmark is important to me, I made a PR some while ago [here](https://github.com/rails-api/active_model_serializers/pull/872). I still need to pick this up.
- The stuff from README and Design were great, I spotted that you added some similarities on README, besides of being really useful, I'm not sure if we should move it to another file to keep README as cleaner as possible

bf4:
I'm open to suggestions re: naming

Would you like me to rebase it to less conmits now? Whatever is easier to review for you.  I'll clean up the conmits either way when we're happy with it.

Right now, my thinking is mostly around how the words are a bit confusing how we build the serializer from the resource, then build the adapter from that serializer instance, so that the 'serializer' for a resource is really the adapter instance.

That's why I like adding Serializer#build, because then if we want to change Serializer to ResourceSerializer and Adapter to SchemaSerializer we can keep the interface stable.

The changes to the readme are just a spike for discussion and to remind me to finish that, since it needs cleanup.

bf4:
In an attempt to speed this along, I extracted parts of it to other PRs, and rebased to simplify it.

bf4:
Failure seems transient, related to global state issues present in master https://travis-ci.org/rails-api/active_model_serializers/jobs/68195121 Possibly related to #961

lib/active_model/serializer/builder.rb
```diff
@@ -0,0 +1,71 @@
+require "set"
+module ActiveModel
+  class Serializer
+    class Builder
```
kurko: So, I think this PR is a great step in organizing things. The current serializer is a huge mess.

With that said, I just want to leave registered in the interwebs a comment here about nomenclatures that we think we need to improve, if not in this PR, in the future. I think it's a bad idea to objects/classes with anything ending in `*er` because you'll end up with things like `Builder#build`. `SerializedHash#build` expresses much better its intent. I've written about it and more in http://alexsquest.com/texts/complexity-in-software-4-the-art-of-naming. There's also http://www.carlopescio.com/2011/04/your-coding-conventions-are-hurting-you.html.

Not a blocker, just something to improve, perhaps in another PR.

bf4: Do you have a suggestion for a better name?  I believe AMS used to have a build_as_json method, or something like that.  I actually don't like builder all that much either, but honestly, I find the adapter being what we call serializable_hash on, rather than the serializer, more confusing, which is part of why I'd like to encapsulate their combined behavior.

(I'm actually thinking of collecting common class names from Ruby projects and seeing what works and what doesn't. This is of interest to me as well).

bf4: In fact, I would name this class `Serializer` if there weren't already one... `ResponseSerializer`? `FinalSerializer`? `SerializableResource`?

bf4: Renamed this to SerializableResource... could use some :eyes:

lib/action_controller/serialization.rb
```diff
@@ -18,48 +19,37 @@ def serialization_scope
...
-          @_serializer_opts[:scope_name] = _serialization_scope
-
+      ActiveModel::SerializableResource.serialize(resource, options) do |serializable_resource|
+        if serializable_resource.serializer?
+          serializable_resource.serialization_scope ||= serialization_scope
+          serializable_resource.serialization_scope_name = _serialization_scope
           begin
-            serialized = serializer.new(resource, @_serializer_opts)
+            serializable_resource.adapter
           rescue ActiveModel::Serializer::ArraySerializer::NoSerializerError
```
bf4: This doesn't seem to be the right place or way to catch the failure when instantiating the serializer

bf4:
@kurko @joaomdmoura How can I push this forward?

lib/action_controller/serialization.rb
```diff
@@ -18,48 +19,37 @@ def serialization_scope
...
     end
 
-    def rescue_with_handler(exception)
-      @_serializer = nil
-      @_serializer_opts = nil
-      @_adapter_opts = nil
+    # Deprecated
+    def use_adapter?
+      true
+    end
```
joaomdmoura: I think we could remove the method and the warning. Any specific reasons to keep it @bf4 ?

joaomdmoura:
LGTM @kurko ?

lib/action_controller/serialization.rb
bf4: Only because someone might have used it to turn off the adapter and hence skip the serializer

test/action_controller/serialization_test.rb
```diff
@@ -356,6 +354,28 @@ def test_cache_expiration_on_update
...
+
+      def test_warn_overridding_use_adapter_as_falsy_on_controller_instance
+        controller = Class.new(ImplicitSerializationTestController) {
+          def use_adapter?
+            false
+          end
+        }.new
+        assert_match /adapter: false/, (capture(:stderr) {
+          controller.get_serializer(@profile)
+        })
```
bf4: here's the test for when someone has defined `use_adapter?` as `false`

lib/action_controller/serialization.rb
joaomdmoura: hmmm, okay, lets just not forget of removing it after a while, maybe on official 0.10.x release, I'm not sure. tks :smile:

bf4: I'm happy to add a self-destructing 'raise if AMS.version > 0.10.0' type of thing. That sorta what you're thinking about?

bf4: If this is a blocker for you, I'd rather remove it, merge,  and discuss in subsequent issues or PRs

joaomdmoura: No, not a blocker! :smile: Let's keep it for now. Just wanted to hear @kurko opinion before I merge it because he was following it too

bf4: In the meantime, would you like me to rebase the renamed builder -> serializable_resource?

lib/active_model/serializable_resource.rb
```diff
@@ -0,0 +1,80 @@
...
+    def initialize(resource, options = {})
+      @resource = resource
+      @adapter_opts, @serializer_opts =
+        options.partition { |k, _| ADAPTER_OPTION_KEYS.include? k }.map { |h| Hash[h] }
+    end
+
+    delegate :serializable_hash, :as_json, :to_json, to: :adapter
+
+    # Primary interface to building a serializer (with adapter)
+    # If no block is given, returns the builder, ready for #as_json/#to_json/#serializable_hash
```
bf4: oops, `builder`

lib/action_controller/serialization.rb
kurko: Baby steps :grin:

lib/action_controller/serialization.rb
kurko: Baby steps. No need to cause so many changes in one PR. If you're refactoring, try to avoid change of behaviors in the same go.

kurko:
I think it cleans a lot of things and it's good. I think there's more work to be done, specially because it was really hard to understand all the changes without a lot of mental effort. This is a good step.

bf4:
> 
So, what would you like to see? Just copy the controller logic into a method on SerializableResource?  Tell me what you like this to look like and I'll make it so (and then some followup prs)
>

bf4:
@kurko @joaomdmoura Any blockers on this? Any more discussion needed? Any changes to make? I've updated the PR description a bit (could probably be better written).

joaomdmoura:
@bf4 I've read through the thread again and checked the code.
It seems we don't have anything blocking this, and last inputs were positive. It will probably implies in some refactoring for me at #950, but I'm looking forward to it.
After reading it all seems that we have nothing to keep us form merge it, so I'm doing it.

bf4:
Yay!
>

Issue #877, opened 2015-04-22
[0.10.0rc1] `has_many` doesn't work without serializer defined on association


JustinAiken:
I have this serializer:

```ruby
class BrandSerializer < ActiveModel::Serializer
  attributes ...

  ...
  has_many :style_configurations
...
end
```

`style_configurations` doesn't have an AMS defined, it's just using the standard rails json response.  Under 0.9x this works fine, but on today's new RC, I get this:

```ruby
     NoMethodError:
       undefined method `new' for nil:NilClass
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer/array_serializer.rb:17:in `block in initialize'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/activerecord-4.2.1/lib/active_record/relation/delegation.rb:46:in `map'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/activerecord-4.2.1/lib/active_record/relation/delegation.rb:46:in `map'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer/array_serializer.rb:12:in `initialize'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer.rb:215:in `new'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer.rb:215:in `block in each_association'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer.rb:208:in `each'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer.rb:208:in `each_association'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer/adapter/json.rb:17:in `serializable_hash'
     # /Users/jaiken/.rvm/gems/ruby-2.2.1/gems/active_model_serializers-0.10.0.rc1/lib/active_model/serializer/adapter.rb:23:in `as_json'
[...]

PR #955, opened 2015-06-12
Add some failing tests around has_many assocs ..where no serializer is defined for the thing that is has_many'd

JustinAiken:
As requested in #877

### DIFF
--- a/test/adapter/json/has_many_test.rb
+++ b/test/adapter/json/has_many_test.rb
@@ -17,6 +17,8 @@ module ActiveModel
             @second_comment.post = @post
             @blog = Blog.new(id: 1, name: "My Blog!!")
             @post.blog = @blog
+            @tag = Tag.new(id: 1, name: "#hash_tag")
+            @post.tags = [@tag]
 
             @serializer = PostSerializer.new(@post)
             @adapter = ActiveModel::Serializer::Adapter::Json.new(@serializer)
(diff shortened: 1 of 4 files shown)


JustinAiken:
The three tests are probably redundant, but I can remove which ever ones aren't helpful.

joaomdmoura:
Great @JustinAiken, now we can keep working into this branch using the tests as reference :smile:

bf4:
In looking at this, is the problem that we have a record with a serializer, and one of its associations has a an enumerable object such as an array of hash whose elements don't implement ActiveModel::Serialization? So is the problem in not having a pass-through-serializer that arrayserializer can call or in calling the array serializer in the first place?  I'm been playing with this a bit, and think it might just be better to not use AMS to serialize objects that it doesn't know how to serialize... that's the renderer's job, no?

bf4:
This works inasmuch as it doesn't crash, but I'm not sure if the end behavior is what we want https://github.com/rails-api/active_model_serializers/commit/56a7bd1b6dd9f083b0f3a9d6d24e9dc85efbffac fwiw

bf4:
The commit in this PR is now included in #962. Propose closing (and thanks!) :smile:

joaomdmoura:
Just waiting the merge before closing all issues related to it @bf4

PR #957, opened 2015-06-15
Refactoring - decouple building included and relationships in JSON api adapter.

lsylvester:
This separates the building of relationships from the building of the included resources.

This will allow future work of moving the building of relationships and of the included resources into separate classes, which may help reduce the complexity of the JSON API adapter.


beauby:
Hey @lsylvester, thanks for the work. I believe we ended up doing this in an other PR. Could you confirm this, or rebase this PR if it is not the case?

beauby:
Closing this for now as it has been taken care of in #1103. Feel free to reopen if I missed something.

PR #958, opened 2015-06-15
Splitting json adapter into two 

joaomdmoura:
We discussed about the ```root``` option and how it shouldn't be a config but a different adapter itself.

This PR create a ```FlattenJson``` adapter that doesn't support ```root```, the ```Json``` adapter in the other hand will now have ```root``` by default and you will still be able to manually set it with the ```root``` option.

```FlattenJson``` is now the default adapter, the result still will be the same we had before, when the default was ```Json``` adapter without root.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -6,7 +6,7 @@ module ActionController
 
     include ActionController::Renderers
 
-    ADAPTER_OPTION_KEYS = [:include, :fields, :root, :adapter]
+    ADAPTER_OPTION_KEYS = [:include, :fields, :adapter]
 
     included do
       class_attribute :_serialization_scope
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -168,11 +168,7 @@ module ActiveModel
     end
 
     def json_key
-      if root.nil?
-        self.class.root_name
-      else
-        root
-      end
+      self.class.root_name
     end
 
     def id
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -145,14 +145,6 @@ module ActiveModel
       adapter_class
     end
 
-    def self._root
-      @@root ||= false
-    end
-
-    def self._root=(root)
-      @@root = root
-    end
-
     def self.root_name
       name.demodulize.underscore.sub(/_serializer$/, '') if name
     end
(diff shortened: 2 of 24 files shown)

### CODE REVIEW

**Comment by spastorino** [comment]
Last tiny bit ...

```
curl http://localhost:3000/todos
{"todo":[{"id":1,"title":"Todo 1","completed":false,"order":1},{"id":3,"title":"Todo 8","completed":true,"order":2}]}
```

We would need that one the be `todos` as it was in old AMS, that's the way it plays nice with Ember's Rest adapter.

**Comment by joaomdmoura** [comment]
@spastorino done.
- Removed root option
- Pluralizing root when using ```ArraySerializer```

btw, as I mentioned there is no need to store the result for now.

**Comment by spastorino** [comment]
```
curl http://localhost:3000/todos
{"todo":[{"id":1,"title":"Todo 1","isCompleted":null}]}%
```

**Comment by joaomdmoura** [comment]
@spastorino :cry: tested over here:
(you can check the code [here](https://github.com/joaomdmoura/AMS-Deserializer-DSL))

Can you share ur controller code?
Did you defined ```ActiveModel::Serializer.config.adapter = :json```?
Is it pointed to the last commit?

```
curl http://localhost:3000

{ 
  "posts": [
    {"id":1,"title":"Awesome Post","body":"Indeed a great one!","user":{"id":2,"name":"Mattew Crazy","age":33,"email":"mattew@crazy.com"}},
    {"id":2,"title":"Incredible Post","body":"Indeed another great one!","user":{"id":2,"name":"Mattew Crazy","age":33,"email":"mattew@crazy.com"}}
  ]
}
```

File: lib/action_controller/serialization.rb [hunk 1]
```diff
@@ -6,7 +6,7 @@ module Serialization
 
     include ActionController::Renderers
 
-    ADAPTER_OPTION_KEYS = [:include, :fields, :root, :adapter]
+    ADAPTER_OPTION_KEYS = [:include, :fields, :adapter]
```
bf4: This is a pretty big change.  Prior to this, the root option *always* went to the adapter. Now it will *always* go to the serializer.  Which code path should be dead? root to serializer or root to adapter?

joaomdmoura: Yeah @bf4 this is indeed a big change.
The idea is that ```root``` shouldn't be an option, but another adapter by itself. 
This make it even easier to integrate with some js frameworks.
We also removed the option to set a specific ```root``` key, so the root isn't "going" to the serializer, but is the serializer the one that knows what it should be and the adapter the one the uses it or not.

bf4: I'll have to read through the code some more to make sure I understand the changes as they affect tests that were deleted or whose expectations were changed.  It didn't break https://github.com/rails-api/active_model_serializers/pull/954 (once I rebased it), but I think this is the kind of thing we should include in the README between 'how it works' and 'intended usage'...

joaomdmoura: Yeah, we realized we forgot about the README, but I'm already updating it.
I'm glad it didn't broke #954 :smile:

PR #959, opened 2015-06-17
Update README.md

kangkyu:
I wonder how `ActiveRecord::Serializer` and `ActiveRecord::Serializers` are different? However I found they have different name between the title and code.


joaomdmoura:
Tks! I'm merging it :smile:

Issue #961, opened 2015-06-21
Race condition in ActionController::Serialization::ImplicitSerializerTest#test_render_fragment_changed_object_with_relationship 


bf4:
Should be an easy fix if anyone wants to pick it up.

ref: https://travis-ci.org/rails-api/active_model_serializers/jobs/67701995

```ruby
rails-api/active_model_serializers/test/action_controller/serialization_test.rb:300]:
--- expected
+++ actual
@@ -1 +1 @@
-{"id"=>1, "time"=>"2015-06-21T04:45:29+00:00", "post"=>{"id"=>1, "title"=>"New Post", "body"=>"Body"}}
+{"id"=>1, "time"=>"2015-06-21T04:45:28+00:00", "post"=>{"id"=>1, "title"=>"New Post", "body"=>"Body"}}
```

kurko:
Yeah. We can use Timecop.

PR #970, opened 2015-06-25
Fix transient tests failures

Rodrigora:
These changes fix the issue #961.

I added the timecop gem as suggested by @kurko. 

Freezing the time in the two tests should fix the transient failures.

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -22,5 +22,6 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "rails", ">= 4.0"
   spec.add_development_dependency "bundler", "~> 1.6"
+  spec.add_development_dependency "timecop", ">= 0.7"
   spec.add_development_dependency "rake"
 end
--- a/test/action_controller/serialization_test.rb
+++ b/test/action_controller/serialization_test.rb
@@ -1,3 +1,4 @@
+
 require 'test_helper'
 
 module ActionController
@@ -56,7 +57,7 @@ module ActionController
         end
 
         def update_and_render_object_with_cache_enabled
-          @post.updated_at = DateTime.now
+          @post.updated_at = Time.now
 
           generate_cached_serializer(@post)
           render json: @post
@@ -112,7 +113,7 @@ module ActionController
 
           generate_cached_serializer(like)
           like.likable = comment2
-          like.time = DateTime.now.to_s
+          like.time = Time.now.to_s
 
           render json: like
         end
@@ -224,9 +225,6 @@ module ActionController
       end
 
       def test_render_with_cache_enable
-        ActionController::Base.cache_store.clear
-        get :render_object_with_cache_enabled
-
         expected = {
           id: 1,
           title: 'New Post',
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -15,3 +15,7 @@ if version == "master"
 else
   gem "rails", "~> #{version}.0"
 end
+
+group :test do
+  gem "timecop"
+end
```
bf4: IMHO, any gem required for test to run should be in the gemspec as a development_dependency. your thoughts?

test/action_controller/serialization_test.rb
```diff
@@ -225,11 +222,16 @@ def test_render_with_cache_enable
           }
         }
 
-        assert_equal 'application/json', @response.content_type
-        assert_equal expected.to_json, @response.body
+        ActionController::Base.cache_store.clear
+        Timecop.freeze(DateTime.now) do
```
bf4: You probably mean Time here. I don't think we're running any tests on the pre-Gregorian calendar :) see https://github.com/ruby/ruby/commit/45458752d3f54ac81661d93aa55b8b5805f78138

bf4:
Thanks!

Gemfile
bf4: It too bad we can't just use https://github.com/rails/rails/blob/4-1-stable/activesupport/lib/active_support/testing/time_helpers.rb  since they aren't in Rails 4.0

Rodrigora: @bf4 Ok. You're right. I'll move to gemspec file. I've found this article after your comment: http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

Another options is remove `timecop` dependency and use fixed datetimes like: `Date.new(2015,6,6)`. What do you think?

bf4: I don't think that would work because it wouldn't affect the `updated_at` timestamps that Rails adds on save.  You'd have to call `model.update_column(:updated_at, Time.new(2015,6,6))` after the object was created ( see https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/persistence.rb#L271-L291 ) or whatever the id is in this case

Rodrigora: Ok, but in these tests we don't use ActiveRecord inherited classes in tests.
The updated_at method is created "by hand" : https://github.com/rails-api/active_model_serializers/blob/master/test%2Ffixtures%2Fporo.rb#L21

And, the problem here is that it assumes that all code is executed in the same second. That's why sometimes it fails.

bf4: Good call

active_model_serializers.gemspec
```diff
@@ -22,5 +22,6 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "rails", ">= 4.0"
   spec.add_development_dependency "bundler", "~> 1.6"
+  spec.add_development_dependency "timecop", ">= 0.7"
```
bf4: @kurko Any thoughts on this PR?

joaomdmoura:
LGTM, @Rodrigora could you please squash the commits and rebase it? :grin:

bf4:
@joaomdmoura Looks like @Rodrigora squashed 'em

Rodrigora:
@joaomdmoura, done. :+1:

joaomdmoura:
Great team work here ppl! Congratz! I'm merging it.

active_model_serializers.gemspec
kurko: yup, I had suggested it to avoid time related issues.

kurko: Do you have concerns about timecop?

PR #971, opened 2015-06-25
Restore has_one to generator

bf4:
per #822 since it was readded in #725


joaomdmoura:
tks @bf4 :smile: I'm merging it.

PR #977, opened 2015-06-30
Pagination Serializer

shicholas:
This PR adds the params for pagination as specified in one alternative
of the JSON API spec http://jsonapi.org/format/#fetching-pagination.

The Pagination Serializer, built on top of the ArraySerializer, allows a
user to pass in two params, page_size and page_number to create a links
attribute in the hash. Note that "number and size" is just one
alternative listed in the JSON API spec.

### DIFF
--- a/lib/active_model/serializer/pagination_serializer.rb
+++ b/lib/active_model/serializer/pagination_serializer.rb
@@ -1,0 +1,16 @@
+module ActiveModel
+  class Serializer
+    class PaginationSerializer < ArraySerializer
+      attr_reader :page_size, :page_number
+
+      def initialize(objects, options = {})
+        @page_size   = options.delete(:page_size)
+        @page_number = options.delete(:page_number)
+
+        super
+
+        @objects = @objects.first.instance_variable_get('@objects')
+      end
+    end
+  end
+end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -5,6 +5,7 @@ module ActiveModel
     extend ActiveSupport::Autoload
     autoload :Configuration
     autoload :ArraySerializer
+    autoload :PaginationSerializer
     autoload :Adapter
     include Configuration
 
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -18,6 +18,8 @@ module ActiveModel
         def serializable_hash(options = nil)
           options = {}
           if serializer.respond_to?(:each)
+            add_pagination if serializer.class == PaginationSerializer
+
             serializer.each do |s|
               result = self.class.new(s, @options.merge(fieldset: @fieldset)).serializable_hash(options)
               @hash[:data] << result[:data]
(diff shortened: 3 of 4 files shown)


shicholas:
This is not complete b/c it does not have the URL for the first part of the links. Please help me find where I can add that to this.

shicholas:
hmm I'm not sure how my code is related to the Travis failure.

lib/active_model/serializer/pagination_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+    class PaginationSerializer < ArraySerializer
+      attr_reader :page_size, :page_number
+
+      def initialize(objects, options = {})
+        @page_size   = options.delete(:page_size)
+        @page_number = options.delete(:page_number)
+
+        super
+
+        @objects = @objects.first.instance_variable_get('@objects')
```
shicholas: originally ``@objects`` referred to the ``ArraySerializer`` any thoughts on how to do this better?

joaomdmoura:
Would be cool to follow JSON API paginations specs, but I'm not sure about creating a new serializer for it. Add it as an option to ```ArraySerializer``` seems to make more sense. 

Also not sure about the ```page[size]``` and ```[number]``` implementation, would be nice to check the actual most used pagination implementations and follow the conventions just to make sure it will be easily integrated with gems like [kaminari](https://github.com/amatsuda/kaminari).

Would be great to hear some other thoughts too @kurko @steveklabnik

shicholas:
Sounds good I'm more than happy to code whatever the consensus is.

joaomdmoura:
Awesome @shicholas ! Let's see what they have to say :smile: btw tks for the PR

shicholas:
fwiw, the more I think about this, the more I think it should be the app's responsibility to pass any pagination hash into the ``meta`` key in the ``ArraySerializer`` and the less I like this PR. Since AMS already has support for ``meta``, I think that no extra code is needed here. (but I can add a wiki page instead if you agree).

shicholas:
sorry for the spam, it looks like ``meta`` isn't supported in JSON API, but maybe it should be? (I can make that PR too if you agree)

joaomdmoura:
Hey @shicholas, indeed using ```meta``` seems a good way to go. 
JSON API supports ```meta``` too :smile: just tested it :smile:

shicholas:
oh cool, I misread the readme. thanks!

shicholas:
@joaomdmoura thanks again for all your help! Here is my write-up if anyone else comes across this (albeit this is def not a perfect solution). And please close this PR :) I think the ``meta`` key is much better.

---

### Enabling JSON API Pagination with Rails and AMS

Active Model Serializers has support for a ``meta`` key in the JSON API adapter. This wiki page shows how one can implement pagination that is [compliant with the JSON API spec for pagination](http://jsonapi.org/format/#fetching-pagination).

```ruby
module Pagination
  extend ActiveSupport::Concern

  private

  def pagination
    raise_error_if_invalid_page_params

    {
      self:  "#{base_url}?page[size]=#{params[:page_size]}&"\
      "page[number]=#{params[:page_number]}",
      first: "#{base_url}?page[size]=#{params[:page_size]}&page[number]=1",
      last:  "#{base_url}?page[size]=#{params[:page_size]}&"\
      "page[number]=#{total_pages}",
      prev:  prev,
      next:  last
    }
  end

  def object_count
    fail 'Implement this method in your controller'
  end

  def base_url
    fail 'Implement this method in your controller'
  end

  def total_pages
    @_total_pages ||= (object_count / params[:page_size].to_i).ceil
  end

  def prev
    return nil if params[:page_number].to_i == 1

    "#{base_url}?page[size]=#{params[:page_size]}&"\
      "page[number]=#{params[:page_number].to_i - 1}"
  end

  def last
    return nil if params[:page_number].to_i == total_pages

    "#{base_url}?page[size]=#{params[:page_size]}&"\
      "page[number]=#{params[:page_number].to_i + 1}"
  end

  def raise_error_if_invalid_page_params
    fail if params[:page_size].to_i < 1 || params[:page_number].to_i < 1
  end
end
```

Then in your controller, you can mix this concern, define ``base_url``, and ``object_count`` and use the ``pagination`` method for your meta key.

```ruby
class ProductController < ApplicationController
  include Pagination

  def index
    render json: product_with_pagination, meta: pagination
  end

  private

  def base_url
    products_index_path
  end
  
  def object_count
    Product.count
  end

  def product_with_pagination
    Product.
      limit(params[:page_size].to_i).
      offset((params[:page_number].to_i - 1) * params[:page_size].to_i)
  end
end
```

Now if you invoke the index method with ``:page_number`` and ``:page_size`` params, you should get the meta information contains pagination.

joaomdmoura:
Thank you @shicholas for the support :smile:

PR #978, opened 2015-06-30
fix generators template bug

regonn:
My environment is

```
rails 4.2.3
active_model_serializers 0.10.0.rc2
```

When I executed this code

`$ bin/rails g resource item name:string price:integer user:references`

and generate this code

``` app/serializers/item_serializer.rb
class ItemSerializer < ActiveModel::Serializer
  attributes :id, :name, :price
end
  attribute :user 
```

I think generator template has a bug, and I fixed it.

### DIFF
--- a/lib/generators/serializer/templates/serializer.rb
+++ b/lib/generators/serializer/templates/serializer.rb
@@ -1,8 +1,8 @@
 <% module_namespacing do -%>
 class <%= class_name %>Serializer < <%= parent_class_name %>
   attributes <%= attributes_names.map(&:inspect).join(", ") %>
-end
 <% association_names.each do |attribute| -%>
   has_one :<%= attribute %>
 <% end -%>
+end
 <% end -%>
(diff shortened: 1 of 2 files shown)


joaomdmoura:
@regonn hmm, this is strange. Indeed it seems that the test we have right now don't test the whole file, but only use ```assert_match```.
Could you write a test to this? :smile:

bf4:
@joaomdmoura Partly my bad for not adding one in my previous two revisions of the generator :(

regonn:
@joaomdmoura I added assert_match test.
I couldn't find an example of file tail test at other gem repository. Do you have another idea?

joaomdmoura:
no worries @bf4 :smile: you all have being doing an awesome work here! 
@regonn I think this is the most straightforward way to test it too :smile: 
I'm cool with it, just ran the test against master it was indeed breaking. 
I'm merging it.

Issue #986, opened 2015-07-06
root attribute ignored


mareczek:
```ruby
render json: @current_user, root: 'resource', adapter: :json
```
gives output 
```json
{
  "user": {
    "id": 1,
    "name": "Rambo"
  }
}
```
when expected 
```json
{
  "resource": {
    "id": 1,
    "name": "Rambo"
  }
}
```

joaomdmoura:
Hey @mareczek, :smile: 
Yes, it was meant to be, as you can see on [README](https://github.com/rails-api/active_model_serializers#built-in-adapters):

> FlattenJSON

> It's the default adapter, it generates a json response without a root key. Doesn't follow any specifc convention.

> JSON

> It also generates a json response but always with a root key. The root key can't be overridden, and will be automatically defined accordingly with the objects being serialized. Doesn't follow any specifc convention.

I know it seems hard, but it seemed to us that it would help ppl into being consistent with their objects names and overall APIs structure this way.

I'm not saying that I'll not accept your PR (actually I think it makes sense to bring root option back),  but wanted to understand what is your motivation first :smile:

mareczek:
I have a case where i wan't to mask the root name (it has to do with using different serializers depending on the permissions of the current user).

Thanks for the explanation, overall i think it will not be a bad thing to allow developers to change the root key :+1:

meltingice:
This is currently a roadblock for me. Because there is no longer a way to filter attributes, I have two different serializers for the same object that return different attributes based on the API endpoint that is called. Because of this, the root key for the JSON is wrong when using one of the serializers and there is no way to change it.

I say, either bring back the ability to filter attributes or allow the root key to be changed. Personally I think the latter is a better approach.

bf4:
I think in the short run you could 

```ruby
class BrandBaseSerializer < ActiveModel::Serializer
  def json_key
    "brand-base"
  end
end
```

or make your own adapter

```ruby
require "active_model/serializer/adapter/json"
class MyAdapter < ActiveModel::Serializer::Adapter::Json
  def root
     serializer.root || super
  end
end
ActiveModel::Serializer.config.adapter = MyAdapter
```

meltingice:
Oh cool, both look like good solutions.
Thanks!

--
Ryan LeFevre (@meltingice)
HODINKEE, Inc.
Sr. Software Engineer




[...]

PR #987, opened 2015-07-06
root option is working (fixed #986)

mareczek:
Will be very grateful for a quick resolution.

PR #991, opened 2015-07-08
Add info to README about changing root key for the JSON adapter.

meltingice:
It turns out you actually can change the root key when using the JSON adapter. If you override the `json_key` method, it lets you define a root key that doesn't have to be based on the class name.

Workaround for #986

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,17 @@ Doesn't follow any specifc convention.
 
 #### JSON
 
-It also generates a json response but always with a root key. The root key **can't be overridden**, and will be automatically defined accordingly with the objects being serialized.
-Doesn't follow any specifc convention.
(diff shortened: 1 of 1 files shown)


joaomdmoura:
Yeap, indeed you could do this, but don't want to encourage ppl to do this.

Just saw #986 and understand the need here, but I'm more inclined to merge #987 and bring the functionality back.

meltingice:
Fair enough, that’s definitely a better solution in the end.



--
Ryan LeFevre (@meltingice)
Hodinkee, Inc.
Sr. Software Engineer

On July 8, 2015 at 1:23:44 PM, João Moura (notifications@github.com) wrote:

Yeap, indeed you could do this, but don't want to encourage ppl to do this.

Just saw #986 and understand the need here, but I'm more inclined to merge #987 and bring the functionality back.

Issue #993, opened 2015-07-09
[ams 0.10.0.rc2] ArgumentError - Unknown adapter: json_api


piotrpalek:
Hey, I am adding `ActiveModel::Serializer.config.adapter = :json_api` in `app/config/initializers/active_model_serializers.rb` and when I try to fetch some data I get this error: `ArgumentError - Unknown adapter: json_api. Valid adapters are: [":fragmentcache", ":json", ":flattenjson", ":null", ":jsonapi"]`

wonderbread:
Same issue.

Work around: 
```ruby
# config/initializers/active_model_serializers.rb
ActiveModel::Serializer.config.adapter = ActiveModel::Serializer::Adapter::JsonApi
```

My reason (likely yours too?):
```ruby
# config/initializers/inflections.rb
ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.acronym 'API'
end
```
`.classify` in [this method](rails-api/active_model_serializers/blob/90fb1cf809dfa606181a1e047f30a05efe71912a/lib/active_model/serializer/adapter.rb#L35-L37) is turning it into `jsonAPI` instead of `jsonApi`

piotrpalek:
Good catch, yeah the workaround you posted works, forgot to update the issue with it. Nice catch with the inflector :+1:

bf4:
Do you think you can write a failing test and/or a PR to fix this?

joaomdmoura:
tks you all! great work! Just flaged the PR as good for new contributors.

bf4:
Fixed by #997  and https://github.com/rails-api/active_model_serializers/pull/1006

PR #997, opened 2015-07-10
Added a (failing) test for when inflecting API

wonderbread:
Failing test as per #993.

Not overly sure how to fix this though. Could potentially fail on other inflections too like json -> JSON.


joaomdmoura:
Awesome. Would be cool if someone pick this up and fix it on the same PR so I'm flaging it  as `Good for new contributors` :smile:

bf4:
@joaomdmoura I could only think of two solutions to this: 1) temporarily turn off the inflections 2) change how we look up adapters 3) don't use the inflector

So, I took a stab at making adapters registrable (since I don't really like the hard-coding adapter constant names) https://github.com/wonderbread/active_model_serializers/pull/1

PR #994, opened 2015-07-09
Starting Docs structure

joaomdmoura:
We have being discussing about wiki and docs for a while.
It's time for us to help new and old AMS users to have a proper location to find general documentation and even 'how to' pages.

We decided to move forward with a ```docs/*```. This PR is the initial implementation of it.

PR #998, opened 2015-07-10
Changing root to model class name

joaomdmoura:
As I mentioned on #986 

> Another change I might do is to define the root key based on the model name and not on the serializer one.

This basically changes the default ```root``` used on Json Adapter to be the resource class name instead of the serializer's name.

This is also a prerequisite to other PRs.

example:

This serializer:
```ruby
class FooBarCommentSerializer < ActiveModel::Serializer
  attributes :content
end
```

resulted on this response (using json aadapter):

```json
{
  foo_bar_comment: {
    content: 'ZOMG a comment'
  }
}
```

now it results: 

```json
{
  comment: {
    content: 'ZOMG a comment'
  }
}
```

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -169,7 +169,7 @@ module ActiveModel
     end
 
     def json_key
-      @root || self.class.root_name
+      @root || object.class.model_name.to_s.downcase
     end
 
     def id
--- a/test/adapter/json/belongs_to_test.rb
+++ b/test/adapter/json/belongs_to_test.rb
@@ -39,7 +39,7 @@ module ActiveModel
             serializer = PostPreviewSerializer.new(@anonymous_post)
             adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
 
-            assert_equal({posts: {title: "Hello!!", body: "Hello, world!!", id: 43, comments: [], author: nil}}, adapter.serializable_hash)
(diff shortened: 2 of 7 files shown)


bf4:
I'm somewhat concerned that this appears to be a breaking change in that many tests needed to be changed.  However, I've lost track of if this is a 0.10 issue, so maybe it's ok in the rc?

joaomdmoura:
@bf4 yeap, this is might impact some ppl updating to 0.10.x, but we already know that upgrading from older version won't be easy, there a lot of other breaking changes, and use the resource class name seems the right thing, it helps user to keep a pattern. Do you have any thoughts about it? :smile:

bf4:
I guess I'd like to see some kind of changelog or doc for how the same poro with be rendered given different ams versions (and adapters), but not sure how practical that is.

bf4:
That said, I suppose, that in all the changed tests, the new expectation appears more correct and less surprising from a hypermedia point of view.  e.g. PostWithComments should still have a root key of post(s), not 'post(s)_with_comments'

joaomdmoura:
totally agree @bf4, I think the new doc structure I'm about to merge (the one we have being working on at #994) will be the right place to do it, after we merge it.

> The new expectation appears more correct and less surprising from a hypermedia point of view

Awesome! :smile:

joaomdmoura:
Taking into account that:

- `root` option is back
- this will be needed into `deserialization`
- #999 is waiting waiting for it and might help ppl to filter attributes

I and also after talk to @spastorino and @bf4, that fell comfortable about this, I'm merging it.

bf4:
:+1: :fireworks:

PR #999, opened 2015-07-10
Enabling fields option to json adapter

### LINKED ISSUE #992: How to filter attributes?
Hi, I want to be able to filter over attributes like this :
 
```ruby
   render json: @model, fields: [:fields1, :fields2]
```
The fields options can be found here : https://github.com/rails-api/active_model_serializers/blob/master/lib/action_controller/serialization.rb
```ruby
   ADAPTER_OPTION_KEYS = [:include, :fields, :adapter]
``` 

But this does not work. I try to understand a few the code and write something like this and binding inside the attributes method :
```ruby
ActiveModel::Serializer::Adapter::JsonApi.new(MyModelSerializer.new(Model.find(X), fields: [:id])).serializable_hash(fields: [:fields1])
```
Let's see the attributes method inside serializer: 
```ruby
    def attributes(options = {})
      attributes =
        if options[:fields] #options is nil, @options is {:fields=>[:fields1]}
          self.class._attributes & options[:fields]
        else
          self.class._attributes.dup
        end
    ....
    end
```

I don't understand where i have to pass the fields options in order to the attributes method get it in these params method.
Why there's an options parameter on the attributes method. Why not use the @options var?

joaomdmoura:
It fix #992

```fields``` option was used on Json API adapter but never on Json adapter. This implements the ```fields``` option on it.

It enables you to filter the fields you want on `render` method call:
```ruby
class ProfileSerialization < ActiveModel::Serializer

  has_many :friends
end
```

```ruby
# ProfileController

# Define the resource fields you want in the response
def show
  @profile = Profile.find(params[:id)
   render json: profile, fields: [:name]
end

# You can also define the resource relationships fields you want in the response
def show_with_some_relationship_data
  @profile = Profile.find(params[:id)
   render json: profile, fields: {profile: [:name], friends: [:id]}
end
```

### DIFF
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -14,6 +14,13 @@ module ActiveModel
       def initialize(serializer, options = {})
         @serializer = serializer
         @options = options
+
+        if fields = @options.delete(:fields)
+          @fieldset = ActiveModel::Serializer::Fieldset.new(fields, serializer.json_key)
+          @options[:fieldset] = @fieldset
+        else
+          @fieldset = options[:fieldset]
+        end
       end
 
       def serializable_hash(options = nil)
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -5,13 +5,13 @@ module ActiveModel
     class Adapter
       class Json < Adapter
         def serializable_hash(options = nil)
-          options ||= {}
           if serializer.respond_to?(:each)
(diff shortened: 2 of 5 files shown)


joaomdmoura:
Rebased with master, now with `root` ar class name, tests are supposed to be okay! :smile:

bf4:
ref: https://github.com/rails-api/active_model_serializers/issues/1058

joaomdmoura:
It does not seems to receive a lot of back up so I'm closing it as well in favor of #1058

vasilakisfil:
I can't believe it that this was never merged :) The API for fields was excellent.

PR #1000, opened 2015-07-11
Add ability to override the default adapter on a per-serializer basis

timrogers:
At the moment, the adapter can only be set as app-wide configuration. This doesn't deal well with the case where multiple adapters might be relevant in different cases - for instance, in my context, most API endpoints where we use serializers are designed to match the JSON API spec, but we also have OAuth endpoints that respond with plain old JSON.

This makes it possible to specify an adapter on a per-serializer class basis, like this:

```ruby
class OverriddenAdapterSerializer < ActiveModel::Serializer
  use_adapter :json
end
```

Interested to hear thoughts. This seems to add significant value with little cost in additional code or complexity.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -9,6 +9,7 @@ module ActiveModel
     include Configuration
 
     class << self
+      attr_accessor :_adapter
       attr_accessor :_attributes
       attr_accessor :_attributes_keys
       attr_accessor :_associations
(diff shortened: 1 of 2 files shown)


timrogers:
@joaomdmoura Let me know if I'm missing anything - not contributed to this (awesome!) gem before.

joaomdmoura:
Hey @timrogers, I'm glad you are using AMS and was able to contribute to it :smile: 
You can change that adapter on the controller if you want to, there is an `adapter` option that you can use on rendering as you can check [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/action_controller/serialization.rb#L9).
But make yourself comfortable to contribute in other issues and let me know in case you need any help. :smile:

Issue #995, opened 2015-07-09
Transient test failures


bf4:
- https://travis-ci.org/rails-api/active_model_serializers/builds/70221324
- https://travis-ci.org/rails-api/active_model_serializers/jobs/70249853

```plain
  1) Failure:
ActiveModel::Serializer::CacheTest#test_associations_cache_when_updated [/home/travis/build/rails-api/active_model_serializers/test/serializers/cache_test.rb:102]:
--- expected
+++ actual
@@ -1 +1 @@
-{:id=>12215340, :title=>"New Post", :body=>"Body"}
+nil
```

PR #1003, opened 2015-07-14
Fix transient test failures

Rodrigora:
Hello guys!

This PR fixes the transient test failures reported in #995 . It freezes the time in the two tests which the elapsed time to run can cause the failures.

I used the same solution used in #970.


bf4:
@joaomdmoura Looks fine to me

joaomdmoura:
LGTM too! I'm merging it.

PR #1006, opened 2015-07-14
Fix adapter inflection bug for api -> API

bf4:
## Summary

1. Added a (failing) test for when inflecting API (Rob McFadzean in #997)
2. sub s/API/Api

### Some work done in here was moved in #1017

### DIFF
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -33,7 +33,8 @@ module ActiveModel
       end
 
       def self.adapter_class(adapter)
-        "ActiveModel::Serializer::Adapter::#{adapter.to_s.classify}".safe_constantize
+        adapter_name = adapter.to_s.classify.sub("API", "Api")
+        "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize
       end
 
       def fragment_cache(*args)
@@ -42,7 +43,7 @@ module ActiveModel
 
       private
 
-     def cache_check(serializer)
+      def cache_check(serializer)
         @cached_serializer = serializer
         @klass             = @cached_serializer.class
         if is_cached?
--- a/test/adapter_test.rb
+++ b/test/adapter_test.rb
@@ -38,6 +38,15 @@ module ActiveModel
         adapter = ActiveModel::Serializer::Adapter.create(@serializer, { adapter: :json_api})
         assert_equal ActiveModel::Serializer::Adapter::JsonApi, adapter.class
       end
+
+      def test_inflected_adapter_class_for_known_adapter
+        ActiveSupport::Inflector.inflections(:en){|inflect| inflect.acronym 'API' }
+        klass = ActiveModel::Serializer::Adapter.adapter_class(:json_api)
+
+        ActiveSupport::Inflector.inflections.acronyms.clear
+
+        assert_equal ActiveModel::Serializer::Adapter::JsonApi, klass
+      end
     end
   end
 end


bf4:
cc @kurko @joaomdmoura your thoughts?

lib/active_model/serializer/adapter.rb
```diff
@@ -11,6 +12,47 @@ class Adapter
...
+      class << ActiveModel::Serializer::Adapter
+        def adapter_map
+          ADAPTER_MAP
+        end
+
+        def register(name, klass)
+          adapter_map.update(name.to_s.underscore => klass)
+        end
+
+        def get(adapter)
```
joaomdmoura: @bf4 this seems almost the same logic of the `adapter` [method](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L134)
We should probably remove it from there in order to use this new `get` method, right?

bf4: > 
You mean on the serializer?  I can dedupe that in a new commit or pr
>

joaomdmoura: Yeap, it's exactly the same logic. IMHO it's related to this PR, check this out:

We can replace the actual code:
```ruby
def self.adapter
  adapter_class = case config.adapter
  when Symbol
    ActiveModel::Serializer::Adapter.adapter_class(config.adapter)
  when Class
    config.adapter
  end
  unless adapter_class
    valid_adapters = Adapter.constants.map { |klass| ":#{klass.to_s.downcase}" }
    raise ArgumentError, "Unknown adapter: #{config.adapter}. Valid adapters are: #{valid_adapters}"
  end

  adapter_class
end
```

By something like:
```ruby
def self.adapter
  ActiveModel::Serializer::Adapter.adapter_class(config.adapter)
end
```

bf4: > 
I'll take a look when I'm next at a computer.  Do you like the new interface and behavior?  Or have an alternative in mind?
>

joaomdmoura: cool! Yup, really liked it! If we can make it the only place to deal with this logic then I'm okay with it :smile:

bf4: > 
Cool, I was a little uncertain about the naming

So, the two methods havr slightly different behavior

The one in the adapter would return nil if none found, but the one in the serializer raises an argument error

So, i was thinking of having get and get! on the adapter

Needs some thought where nil is desired


>

bf4: Added another commit to consider.  I've been playing with the interface for success/failurre handling of finding the adapter by name and whether to find it by constants, or safe constantize, etc.  So, this is for consideration as a way to do.

joaomdmoura: @bf4 it's seems to me `nil` is never desired. What are your thoughts about keep the argument error approach and remote `get!`?

bf4: > 
Sounds good.  I'd still like to allow for user defined failure handling
>

bf4: You'll notice that `get` can take a block to handle adapter not found, and is used by `get!`

```ruby
        def get!(adapter)
          get(adapter) do |failure_message| fail ArgumentError, failure_message end
        end
```

bf4:
It was easier to handle the simple case of api being inflected to API rather than Api via sub (not sure that's the most performant way. I didn't investigate), rather than have a whole bunch of changes around that either mutating ` ActiveSupport::Inflector.inflections(:en).acronyms` or using `I18n.with_locale(:ams) { }`, or changing how we look up constants, etc.

bf4:
The stuff with the registered adapter I'll do in another PR, because it also brings up issues with autoloading

joaomdmoura:
Will #1017 eliminate this? Because I'm not sure about it.

bf4:
No, this PR is still required.  I just think [the complexity I was introducing wasn't required for the range of inflection failures](https://github.com/rails-api/active_model_serializers/pull/1006#issuecomment-122378750)

joaomdmoura:
Okay @bf4. Good, it seems the most straightforward solution for me as well. I'm merging it.

PR #1007, opened 2015-07-15
Bug fix for ArraySerializer json_key

jiajiawang:
When the resource is a zero result query,
i.e. post_comments = PostComment.where("1=0")
the json_key will become 'postcomments' rather than 'post_comments'.
Using 'underscore' instead of 'downcase' fixes the error.

### DIFF
--- a/test/array_serializer_test.rb
+++ b/test/array_serializer_test.rb
@@ -38,6 +38,17 @@ module ActiveModel
         assert_equal @serializer.meta, "the meta"
         assert_equal @serializer.meta_key, "the meta key"
       end
+
+      def test_json_key_when_resource_is_empty
+        Array.class_eval do
+          def name
+            'PostComment'
+          end
+        end
+        @post_comments = []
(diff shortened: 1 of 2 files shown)


joaomdmoura:
Awesome @jiajiawang!
Could you add a test to cover it too? So we can make sure it won't easily broke again :smile:

jiajiawang:
Sure

joaomdmoura:
Great! LGTM, I'm merging it.

test/array_serializer_test.rb
```diff
@@ -38,6 +38,17 @@ def test_meta_and_meta_key_attr_readers
         assert_equal @serializer.meta, "the meta"
         assert_equal @serializer.meta_key, "the meta key"
       end
+
+      def test_json_key_when_resource_is_empty
+        Array.class_eval do
```
bf4: :question:

NullVoxPopuli: looks like a fake AR Relation

PR #1010, opened 2015-07-15
add support for string value serializer

vyrak:
I was having issues specifying a string name of a serializer when upgrading to from v0.9.3 to v0.10.0rc2. For example:

``` ruby
  belongs_to :blog, serializer: "CustomBlogSerializer"
end
```

This PR should allow for that again.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -120,7 +120,7 @@ module ActiveModel
     end
 
     def self.serializer_for(resource, options = {})
-      if resource.respond_to?(:serializer_class)
+      serializer = if resource.respond_to?(:serializer_class)
         resource.serializer_class
       elsif resource.respond_to?(:to_ary)
         config.array_serializer
@@ -129,6 +129,8 @@ module ActiveModel
           .fetch(:association_options, {})
           .fetch(:serializer, get_serializer_for(resource.class))
       end
+
+      serializer.respond_to?(:constantize) ? serializer.constantize : serializer
     end
 
     def self.adapter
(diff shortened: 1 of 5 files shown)


vyrak:
hold on, need to account for array serialization, too.

joaomdmoura:
Hey @vyrak thank you. Why exactly you ended up needing to specify the serializer using a `string`?

vyrak:
The code base I was upgrading was setting the serializer option for associations in string form, but as since been updated to not do that, so it doesn't really apply to my use case anymore.

I thought this might be useful to others upgrading from older versions. Maybe it should include a deprecation warning, so we can phase out this old behavior.

jasonayre:
@joaomdmoura To not rely on autoloading to resolve the class upfront. I cant think of a particular instance I can reproduce where it was necessary, but I've been specifying class names as a string in both serializers and models for quite some time as I've had issues when not doing so. (when the class isn't immediately picked up by autoload).

bf4:
> I thought this might be useful to others upgrading from older versions

So, this is breaking change from `v0.9.3 to v0.10.0rc2`.  We could add that to the docs folder? @joaomdmoura

joaomdmoura:
@jasonayre I get it, but it's seems kind of an edge-case.
@bf4 not we should remember to add this to the upgrading tutorial, of course.
@vyrak I'd be glad to merge it if you add the deprecation warning. So we might remove it before release the final 0.10.0 (that might become 1.0). :smile:

beauby:
Could you rebase this one?

NullVoxPopuli:
is this still desired? just for cleanup, I'm going to close this, but feel free to open it again if this is something you still want merged (after rebasing, of course)

PR #1013, opened 2015-07-16
Root option with empty array support

vyrak:
based off #987 
support for `:root` option along with empty array fix

### DIFF
--- a/lib/active_model/serializer/array_serializer.rb
+++ b/lib/active_model/serializer/array_serializer.rb
@@ -5,9 +5,10 @@ module ActiveModel
       include Enumerable
       delegate :each, to: :@objects
 
-      attr_reader :meta, :meta_key
+      attr_reader :root, :meta, :meta_key
 
       def initialize(objects, options = {})
+        @root = options[:root]
         @resource = objects
         @objects  = objects.map do |object|
           serializer_class = options.fetch(
@@ -26,15 +27,8 @@ module ActiveModel
       end
 
       def json_key
-        if @objects.first
-          @objects.first.json_key.pluralize
-        else
-          @resource.name.underscore.pluralize if @resource.try(:name)
-        end
-      end
-
-      def root=(root)
-        @objects.first.root = root if @objects.first
+        key = root || @objects.first.try(:json_key) || @resource.try(:name).try(:underscore)
+        key.try(:pluralize)
       end
     end
   end
--- a/test/array_serializer_test.rb
+++ b/test/array_serializer_test.rb
@@ -6,7 +6,13 @@ module ActiveModel
       def setup
         @comment = Comment.new
         @post = Post.new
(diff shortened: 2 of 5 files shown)


test/array_serializer_test.rb
```diff
@@ -39,15 +45,52 @@ def test_meta_and_meta_key_attr_readers
         assert_equal @serializer.meta_key, "the meta key"
       end
 
-      def test_json_key_when_resource_is_empty
```
joaomdmoura: @vyrak I would like to keep this test, it's testing a totaly different thing.

lib/active_model/serializer/array_serializer.rb
```diff
@@ -26,15 +27,8 @@ def initialize(objects, options = {})
...
-        if @objects.first
-          @objects.first.json_key.pluralize
-        else
-          @resource.name.underscore.pluralize if @resource.try(:name)
-        end
-      end
-
-      def root=(root)
-        @objects.first.root = root if @objects.first
+        key = root || @objects.first.try(:json_key) || @resource.try(:name).try(:underscore)
```
joaomdmoura: :+1:

joaomdmoura:
@vyrak great work! I was about to pick that up, but I'm glad you made it!
I've left a comment on a specific test for you to check it :smile:

test/array_serializer_test.rb
vyrak: Isn't that a duplicated by https://github.com/rails-api/active_model_serializers/pull/1013/files#diff-4511bb323b8555a8103c7964418e0e40R69? I favored that version because it doesn't modify the Array class itself.

joaomdmoura: Nops, it was introduced on #1007 

> The json_key will become 'postcomments' rather than 'post_comments'.
Using 'underscore' instead of 'downcase' fixes the error.

vyrak: Maybe I'm being slow and missing your point, in the `build_named_collection` helper method, it's returning an instance with a name of `"MeResource"` which gets underscored & pluralized to `"me_resources"`. 

If I understand #1007 correctly, it was a problem with an empty collection object with a `name` not getting transformed properly. I believe that `test_json_key_with_resource_with_name_and_no_serializers` is effectively testing that case, right?

joaomdmoura: Oh you are right, I didn't realized that, sorry :grin:

joaomdmoura:
LGTM, I'm merging it and `root` is finally back.

test/array_serializer_test.rb
vyrak: Ah, all good. The test name might've been what was confusing. Maybe it makes more sense to say `test_json_key_with_empty_resource_with_name` instead of `test_json_key_with_resource_with_name_and_no_serializers`. Though it looks like you already merged.

PR #1015, opened 2015-07-17
Prioritize object `type` over serializer `type`

vyrak:
I had some records being serialized that have `type` attributes. With the 0.10.x update, `ActiveModel::Serializer#type` was introduced which ended up ignoring the serialized object's type. This change makes it so if an object responds to `type`, then it will use the object's value, otherwise, it will use the serializer's.

Do note that anyone using the `:json_api` adapter may have to update their serializers, if they have an object with a `type` attribute, but still prefer to use the serializer's fallback type over the object's value.

It may be worth exploring a more robust way of handling these types of scenarios, where the serializer class defines methods that conflict with the underlying object (i.e. `type`, `id`, `json_key`, and possible future ones). Maybe they need to be handled in the adapters that rely on them. Thoughts?

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -177,7 +177,7 @@ module ActiveModel
     end
 
     def type
-      object.class.model_name.plural
+      object.respond_to?(:type) ? object.type : object.class.model_name.plural
     end
 
     def attributes(options = {})
--- a/test/serializers/attribute_test.rb
+++ b/test/serializers/attribute_test.rb
@@ -33,6 +33,29 @@ module ActiveModel
 
         assert_equal([:title], serializer_class._attributes)
       end
+
+      def test_type_attribute
+        serializer_class = Class.new(ActiveModel::Serializer){ attribute :type }
+
+        adapter = ActiveModel::Serializer::Adapter::FlattenJson.new(serializer_class.new(@blog))
+        assert_equal({type: 'blogs'}, adapter.serializable_hash)
+      end
+
+      def test_type_attribute_on_object
+        @blog.define_singleton_method(:type){ 'type' }
+        serializer_class = Class.new(ActiveModel::Serializer){ attribute :type }
+
+        adapter = ActiveModel::Serializer::Adapter::FlattenJson.new(serializer_class.new(@blog))
+        assert_equal({type: @blog.type}, adapter.serializable_hash)
+      end
+
+      def test_type_attribute_on_object_with_nil_value
+        @blog.define_singleton_method(:type){ nil }
+        serializer_class = Class.new(ActiveModel::Serializer){ attribute :type }
+
+        adapter = ActiveModel::Serializer::Adapter::FlattenJson.new(serializer_class.new(@blog))
+        assert_equal({type: @blog.type}, adapter.serializable_hash)
+      end
     end
   end
 end


bf4:
So, I'm trying to think through if the Serializer 'type' and a serialized attribute 'type' have the same semantic meaning.  My gut is they don't, but all the tests fail.  I wonder if any would fail if we added a type attribute to one of the existing serializers in poro.rb...

joaomdmoura:
hey @vyrak I can see the problem here, but I'm not sure this the right solution.
We really want to make json-api really easy to use by automating its conventions.

As you pointed out, this implementation would override the actual `type`, and we would need to warn developers(that use json-api) to update their serializers.

Another solution could be a middle ground, that enables ppl that are **not** using json_api to have their `type` attribute.

Another option would be formally to declare `type` as a protect word (with the others that we have).

How you feel about it?

vyrak:
A middle ground might be the best option. I also agree that `Serializer#type` and the `type` on a serializable object have the different semantic meaning; I believe `Serializer#type` is unnecessary.

I think the logic for convention-based attributes like `type` and `id` should probably be derived in the adapters. In this case, the `JsonApi` adapter. 

In the event that those attributes are explicitly specified for serialization, then the convention-based attributes are still used, unless a mechanism to override it is provided (the mechanism most likely being an adapter specific option); and the serializable object's values are always prioritized, but under the attributes key. For example:

```ruby
class CustomModel < Model
  def id
    "custom id #{super}"
  end

  def type
    "custom type"
  end
end

class CustomModelSerializer < ActiveModel::Serializer
  attributes :id, :type
end
```
The output, when using the `JsonApi` adapter, of an instance of `CustomModel` should be:
```ruby
{
  "type": "CustomModels",
  "id": "1",
  "attributes": {
    "type": "custom type",
    "id": "custom id 1"
  }
}
```
I would also argue that the `json_key` & root name logic in the serializer needs be moved to the adapters as well to keep the `Serializer` footprint as small as possible and avoid future naming conflicts, but maybe I'm getting ahead of myself.

joaomdmoura:
@vyrak great points.

> I think the logic for convention-based attributes like type and id should probably be derived in the adapters. In this case, the JsonApi adapter.

:+1: 

> I would also argue that the json_key & root name logic in the serializer needs be moved to the adapters as well to keep the Serializer footprint as small as possible and avoid future naming conflicts, but maybe I'm getting ahead of myself.

Not sure about this one, but indeed, let's not get ahead of ourselves :smile:

About the example itself, I'm not sure, I understand the logic and the reasons you have, but at the same time, this json response can be confusing, despite of not violate Json API conventions.

vyrak:
A better example would have been having an ActiveRecord that mapped to a table with a type column; that type column can mean anything for that given domain (i.e. type of permission, charge, etc...) and is not necessarily a representative of the table/record/model name.

As for the `id` bit, it is more of an edge case, but it's still good to talk about. There was a mistake in the `id` logic. The override should have been defined in the serializer class; otherwise, the ids in the responses would always match up. So:

```ruby
class CustomModel < Model # really an ActiveRecord with a type column
  def type
    "custom type"
  end
end

class CustomModelSerializer < ActiveModel::Serializer
  attributes :id, :type

  def id
    "custom id #{super}"
  end
end
```

Yes, the response may be confusing, but, like you said, it does not violate the spec. If it was spec'd out, I would be singing a different tune. I believe that restriction should not be introduced by this gem. As it stands now, it should be up to implementers of their API to decide how to structure their responses and document parts where it may be unclear.

joaomdmoura:
All right, despite of any edge case it seems we have middles ground.
What I'm doing is, I'm closing the PR in favor of a new one involving what we have being discussing:

- Let's keep `id` out of it for now.
- Users that **don't use** json-api should be able to have `type` for their own.
- Enable json-api users to have their `type` attributes inside the `attributes` node

After this we can go back to check the next steps, if necessary. :smile: @bf4 @vyrak

PR #1016, opened 2015-07-17
require rails/railtie before subclassing Rails::Railtie

bf4:
Otherwise I cannot bundle console


joaomdmoura:
I've being using `pry` and requiring it on the file I need when working. Any particular reason here?

bf4:
Not sure I follow you. It's just an explicit require in the appropriate place that I found missing when the file was loaded in console before rails/railtie was implicitly required.  I thought it was a no-brainer, honestly.

joaomdmoura:
My bad, u're right. I'm merging it.

PR #1017, opened 2015-07-17
Make Adapters registerable so they are not namespace-constrained

bf4:
Changes:

- Introduce Adapter::lookup for use by Serializer.adapter
- Move Adapter-finding logic from Adapter::adapter_class into Adapter::lookup
- Fix circular load warning as described in https://github.com/rails-api/active_model_serializers/commit/cf6a074a1c62e3f5a30f90a0987b8cf049272953
- Allowed for gem devs to use a [`Gemfile.local`](https://github.com/rails-api/active_model_serializers/pull/1017#discussion_r37602184) for gems not in the gemspec or Gemfile e.g. `pry-byebug` (which was my usecase).
 
Background Ref:
- https://github.com/rails-api/active_model_serializers/pull/1050#issuecomment-133295789
- https://github.com/rails-api/active_model_serializers/pull/1067#discussion_r37710003. 
- The code in this PR was my first attempt to deal with https://github.com/rails-api/active_model_serializers/issues/1006 till I realized a simpler solution.. 

Introduced interfaces:

- non-inherited methods
```ruby
ActiveModel::Serializer::Adapter.adapter_map  # a Hash<adapter_name, adapter_class>
ActiveModel::Serializer::Adapter.adapters     # an Array<adapter_name>
ActiveModel::Serializer::Adapter.register(name, klass) # adds an adapter to the adapter_map
ActiveModel::Serializer::Adapter.lookup(name_or_klass)    # raises UnknownAdapterError error when adapter not found
```

- Automatically register adapters when subclassing

```ruby
     def self.inherited(subclass)
       ActiveModel::Serializer::Adapter.register(subclass.to_s.demodulize,
subclass)
     end
```

- Preserves subclass method `::adapter_class(adapter)`

```ruby
     def self.adapter_class(adapter)
       ActiveModel::Serializer::Adapter.lookup(adapter)
     end
```

- Serializer.adapter now uses `Adapter.lookup(config.adapter)` rather than have
duplicate logic

TODO

Test
- [x] ActiveModel::Serializer::Adapter.adapter_map
- [x] ActiveModel::Serializer::Adapter.adapters
- [x] ActiveModel::Serializer::Adapter.register(name, klass)
- [x] ActiveModel::Serializer::Adapter.get(name_or_klass)
- [x] ActiveModel::Serializer::Adapter::inherited
- [x] ActiveModel::Serializer.adapter

TO CONSIDER

- [x] wrap 'finding adapter by name' logic outside of 'get' method
- [x] Change exception from ArgumentError to UnknownAdapter

TODO in future PR:
- [ ] Change `Adapter` to module (backwards compatible public methods) and make `Adapter::Base` or `Adapter::AbstractClass` 

Full commit message from https://github.com/rails-api/active_model_serializers/commit/cf6a074a1c62e3f5a30f90a0987b8cf049272953

<pre>
Ensure inheritance hooks run
I was seeing transient failures where adapters may not be registered.

e.g. https://travis-ci.org/rails-api/active_model_serializers/builds/77735382

Since we're using the Adapter, JsonApi, and Json classes
as namespaces, some of the conventions we use for modules don't apply.
Basically, we don't want to define the class anywhere besides itself.
Otherwise, the inherited hooks may not run, and some adapters may not
be registered.

For example:

If we have a class Api `class Api; end`
And Api is also used as a namespace for `Api::Product`
And the classes are defined in different files.

In one file:

```ruby
class Api
  autoload :Product
  def self.inherited(subclass)
    puts
    p [:inherited, subclass.name]
    puts
  end
end
```

And in another:

```ruby
class Api
  class Product < Api
    def sell_sell_sell!
      # TODO: sell
    end
  end
end
```

If we load the Api class file first, the inherited hook will be defined on the class
so that when we load the Api::Product class, we'll see the output:

```plain
[ :inherited, Api::Product]
```

However, if we load the Api::Product class first, since it defines the `Api` class
and then inherited from it, the Api file was never loaded, the hook never defined,
and thus never run.

By defining the class as `class Api::Product < Api` We ensure the the Api class
MUST be defined, and thus, the hook will be defined and run and so sunshine and unicorns.

Appendix:

The below would work, but triggers a circular reference warning.
It's also not recommended to mix require with autoload.

```ruby
require 'api'
class Api
  class Product < Api
    def sell_sell_sell!
      # TODO: sell
    end
  end
end
```

This failure scenario was introduced by removing the circular reference warnings in
#1067
</pre>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,8 @@
 source 'https://rubygems.org'
+#
+# Add a Gemfile.local to locally bundle gems outside of version control
+local_gemfile = File.join(File.expand_path('..', __FILE__), 'Gemfile.local')
+eval_gemfile local_gemfile if File.readable?(local_gemfile)
 
 # Specify your gem's dependencies in active_model_serializers.gemspec
 gemspec
--- a/docs/general/adapters.md
+++ b/docs/general/adapters.md
@@ -32,7 +32,7 @@ resources in the `"included"` member when the resource names are included in the
 
 ## Choosing an adapter
 
-If you want to use a different adapter, such as JsonApi, you can change this in an initializer:
+If you want to use a specify a default adapter, such as JsonApi, you can change this in an initializer:
 
 ```ruby
 ActiveModel::Serializer.config.adapter = ActiveModel::Serializer::Adapter::JsonApi
@@ -44,8 +44,59 @@ or
 ActiveModel::Serializer.config.adapter = :json_api
 ```
 
-If you want to have a root key in your responses you should use the Json adapter, instead of the default FlattenJson:
+If you want to have a root key for each resource in your responses, you should use the Json or
+JsonApi adapters instead of the default FlattenJson:
 
 ```ruby
 ActiveModel::Serializer.config.adapter = :json
 ```
+
+## Advanced adapter configuration
+
+### Registering an adapter
+
+The default adapter can be configured, as above, to use any class given to it.
+
+An adapter may also be specified, e.g. when rendering, as a class or as a symbol.
+If a symbol, then the adapter must be, e.g. `:great_example`,
+`ActiveModel::Serializer::Adapter::GreatExample`, or registered.
+
+There are two ways to register an adapter:
+
+1) The simplest, is to subclass `ActiveModel::Serializer::Adapter`, e.g. the below will
+register the `Example::UsefulAdapter` as `:useful_adapter`.
+
+```ruby
+module Example
+  class UsefulAdapter < ActiveModel::Serializer::Adapter
+  end
+end
+```
+
+You'll notice that the name it registers is the class name underscored, not the full namespace.
+
+Under the covers, when the `ActiveModel::Serializer::Adapter` is subclassed, it registers
+the subclass as `register(:useful_adapter, Example::UsefulAdapter)`
+
+2) Any class can be registered as an adapter by calling `register` directly on the
+`ActiveModel::Serializer::Adapter` class. e.g., the below registers `MyAdapter` as
+`:special_adapter`.
+
+```ruby
+class MyAdapter; end
+ActiveModel::Serializer::Adapter.register(:special_adapter, MyAdapter)
+```
+
+### Looking up an adapter
+
+| `ActiveModel::Serializer::Adapter.adapter_map` | A Hash of all known adapters { adapter_name => adapter_class } |
+| `ActiveModel::Serializer::Adapter.adapters`    | A (sorted) Array of all known adapter_names |
+| `ActiveModel::Serializer::Adapter.get(name_or_klass)` |  The adapter_class, else raises an `ActiveModel::Serializer::Adapter::UnknownAdapter` error |
+| `ActiveModel::Serializer::Adapter.adapter_class(adapter)` | delegates to `ActiveModel::Serializer::Adapter.get(adapter)` |
+| `ActiveModel::Serializer.adapter` | a convenience method for `ActiveModel::Serializer::Adapter.get(config.adapter)` |
+
+The registered adapter name is always a String, but may be looked up as a Symbol or String.
+Helpfully, the Symbol or String is underscored, so that `get(:my_adapter)` and `get("MyAdapter")`
+may both be used.
+
+For more information, see [the Adapter class on GitHub](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter.rb)
--- a/docs/general/adapters.md
+++ b/docs/general/adapters.md
@@ -91,9 +91,9 @@ ActiveModel::Serializer::Adapter.register(:special_adapter, MyAdapter)
 
 | `ActiveModel::Serializer::Adapter.adapter_map` | A Hash of all known adapters { adapter_name => adapter_class } |
 | `ActiveModel::Serializer::Adapter.adapters`    | A (sorted) Array of all known adapter_names |
-| `ActiveModel::Serializer::Adapter.get(name_or_klass)` |  The adapter_class, else raises an `ActiveModel::Serializer::Adapter::UnknownAdapter` error |
-| `ActiveModel::Serializer::Adapter.adapter_class(adapter)` | delegates to `ActiveModel::Serializer::Adapter.get(adapter)` |
-| `ActiveModel::Serializer.adapter` | a convenience method for `ActiveModel::Serializer::Adapter.get(config.adapter)` |
+| `ActiveModel::Serializer::Adapter.lookup(name_or_klass)` |  The adapter_class, else raises an `ActiveModel::Serializer::Adapter::UnknownAdapter` error |
+| `ActiveModel::Serializer::Adapter.adapter_class(adapter)` | delegates to `ActiveModel::Serializer::Adapter.lookup(adapter)` |
+| `ActiveModel::Serializer.adapter` | a convenience method for `ActiveModel::Serializer::Adapter.lookup(config.adapter)` |
 
 The registered adapter name is always a String, but may be looked up as a Symbol or String.
 Helpfully, the Symbol or String is underscored, so that `get(:my_adapter)` and `get("MyAdapter")`
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -5,11 +5,11 @@ module ActiveModel
       ADAPTER_MAP = {}
       private_constant :ADAPTER_MAP if defined?(private_constant)
       extend ActiveSupport::Autoload
-      require 'active_model/serializer/adapter/json'
-      require 'active_model/serializer/adapter/json_api'
-      autoload :FlattenJson
-      autoload :Null
       autoload :FragmentCache
+      autoload :Json
+      autoload :JsonApi
+      autoload :Null
+      autoload :FlattenJson
 
       def self.create(resource, options = {})
         override = options.delete(:adapter)
@@ -60,14 +60,14 @@ module ActiveModel
             register(adapter_name, adapter_class)
             adapter_class
           }
-        rescue ArgumentError
+        rescue ArgumentError => e
           failure_message =
             "Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
-          raise UnknownAdapterError, failure_message, $!.backtrace
-        rescue NameError
+          raise UnknownAdapterError, failure_message, e.backtrace
+        rescue NameError => e
           failure_message =
-            "NameError: #{$!.message}. Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
-          raise UnknownAdapterError, failure_message, $!.backtrace
+            "NameError: #{e.message}. Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
+          raise UnknownAdapterError, failure_message, e.backtrace
         end
 
         # @api private
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -1,6 +1,9 @@
 module ActiveModel
   class Serializer
     class Adapter
+      UnknownAdapterError = Class.new(ArgumentError)
+      ADAPTER_MAP = {}
+      private_constant :ADAPTER_MAP if defined?(private_constant)
       extend ActiveSupport::Autoload
       require 'active_model/serializer/adapter/json'
       require 'active_model/serializer/adapter/json_api'
@@ -14,9 +17,71 @@ module ActiveModel
         klass.new(resource, options)
       end
 
+      # @see ActiveModel::Serializer::Adapter.get
       def self.adapter_class(adapter)
-        adapter_name = adapter.to_s.classify.sub('API', 'Api')
-        "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize
+        ActiveModel::Serializer::Adapter.get(adapter)
+      end
+
+      # Only the Adapter class has these methods.
+      # None of the sublasses have them.
+      class << ActiveModel::Serializer::Adapter
+        # @return Hash<adapter_name, adapter_class>
+        def adapter_map
+          ADAPTER_MAP
+        end
+
+        # @return [Array<Symbol>] list of adapter names
+        def adapters
+          adapter_map.keys.sort
+        end
+
+        # Adds an adapter 'klass' with 'name' to the 'adapter_map'
+        # Names are stringified and underscored
+        # @param [Symbol, String] name of the registered adapter
+        # @param [Class] klass - adapter class itself
+        # @example
+        #     AMS::Adapter.register(:my_adapter, MyAdapter)
+        def register(name, klass)
+          adapter_map.update(name.to_s.underscore => klass)
+          self
+        end
+
+        # @param  adapter [String, Symbol, Class] name to fetch adapter by
+        # @return [ActiveModel::Serializer::Adapter] subclass of Adapter
+        # @raise  [UnknownAdapterError]
+        def get(adapter)
+          # 1. return if is a class
+          return adapter if adapter.is_a?(Class)
+          adapter_name = adapter.to_s.underscore
+          # 2. return if registered
+          adapter_map.fetch(adapter_name) {
+            # 3. try to find adapter class from environment
+            adapter_class = find_by_name(adapter_name)
+            register(adapter_name, adapter_class)
+            adapter_class
+          }
+        rescue ArgumentError
+          failure_message =
+            "Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
+          raise UnknownAdapterError, failure_message, $!.backtrace
+        rescue NameError
+          failure_message =
+            "NameError: #{$!.message}. Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
+          raise UnknownAdapterError, failure_message, $!.backtrace
+        end
+
+        # @api private
+        def find_by_name(adapter_name)
+          adapter_name = adapter_name.to_s.classify.tr('API', 'Api')
+          "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize or # rubocop:disable Style/AndOr
+            fail UnknownAdapterError
+        end
+        private :find_by_name
+      end
+
+      # Automatically register adapters when subclassing
+      def self.inherited(subclass)
+        ActiveModel::Serializer::Adapter.register(subclass.to_s.demodulize, subclass)
       end
 
       attr_reader :serializer
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -60,11 +60,7 @@ module ActiveModel
             register(adapter_name, adapter_class)
             adapter_class
           }
-        rescue ArgumentError => e
-          failure_message =
-            "Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
-          raise UnknownAdapterError, failure_message, e.backtrace
-        rescue NameError => e
+        rescue NameError, ArgumentError => e
           failure_message =
             "NameError: #{e.message}. Unknown adapter: #{adapter.inspect}. Valid adapters are: #{adapters}"
           raise UnknownAdapterError, failure_message, e.backtrace
@@ -73,7 +69,7 @@ module ActiveModel
         # @api private
         def find_by_name(adapter_name)
           adapter_name = adapter_name.to_s.classify.tr('API', 'Api')
-          "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize or # rubocop:disable Style/AndOr
+          ActiveModel::Serializer::Adapter.const_get(adapter_name.to_sym) or # rubocop:disable Style/AndOr
             fail UnknownAdapterError
         end
         private :find_by_name
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -17,9 +17,9 @@ module ActiveModel
         klass.new(resource, options)
       end
 
-      # @see ActiveModel::Serializer::Adapter.get
+      # @see ActiveModel::Serializer::Adapter.lookup
       def self.adapter_class(adapter)
-        ActiveModel::Serializer::Adapter.get(adapter)
+        ActiveModel::Serializer::Adapter.lookup(adapter)
       end
 
       # Only the Adapter class has these methods.
@@ -49,7 +49,7 @@ module ActiveModel
         # @param  adapter [String, Symbol, Class] name to fetch adapter by
         # @return [ActiveModel::Serializer::Adapter] subclass of Adapter
         # @raise  [UnknownAdapterError]
-        def get(adapter)
+        def lookup(adapter)
           # 1. return if is a class
           return adapter if adapter.is_a?(Class)
           adapter_name = adapter.to_s.underscore
--- a/lib/active_model/serializer/adapter/flatten_json.rb
+++ b/lib/active_model/serializer/adapter/flatten_json.rb
@@ -1,7 +1,4 @@
-module ActiveModel
-  class Serializer
-    class Adapter
-      class FlattenJson < Json
+class ActiveModel::Serializer::Adapter::FlattenJson < ActiveModel::Serializer::Adapter::Json
         def serializable_hash(options = {})
           super
           @result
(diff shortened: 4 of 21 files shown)


bf4:
@joaomdmoura Ready for review!

test/adapter_test.rb
```diff
@@ -19,16 +19,6 @@ def test_serializer
...
-      def test_adapter_class_for_known_adapter
-        klass = ActiveModel::Serializer::Adapter.adapter_class(:json_api)
-        assert_equal ActiveModel::Serializer::Adapter::JsonApi, klass
-      end
-
-      def test_adapter_class_for_unknown_adapter
-        klass = ActiveModel::Serializer::Adapter.adapter_class(:json_simple)
-        assert_nil klass
-      end
-
```
bf4: moved to adapter_for file

Gemfile
```diff
@@ -1,4 +1,10 @@
 source 'https://rubygems.org'
+#
+# Add a Gemfile.local to locally bundle gems outside of version control
+local_gemfile = File.join(File.expand_path("..", __FILE__), "Gemfile.local")
+if File.readable?(local_gemfile)
+  eval_gemfile local_gemfile
+end
```
bf4: I like using pry-byebug. This let's me do it without interfering with the project

lib/active_model/serializer/adapter.rb
```diff
@@ -11,6 +13,64 @@ class Adapter
 
       attr_reader :serializer
 
+      # Only the Adapter class has these methods.
+      # None of the sublasses have them.
```
joaomdmoura: :+1:

lib/active_model/serializer/adapter.rb
```diff
@@ -11,6 +13,64 @@ class Adapter
...
+          raise UnknownAdapterError, failure_message, $!.backtrace
+        end
+
+        # @api private
+        def find_by_name(adapter_name)
+          adapter_name = adapter_name.to_s.classify.tr("API", "Api")
+          "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize or
+            fail UnknownAdapterError
+        end
+        private :find_by_name
```
joaomdmoura: :+1:

lib/active_model/serializer/adapter.rb
```diff
@@ -11,6 +13,64 @@ class Adapter
...
+      end
+
+      # Automatically register adapters when subclassing
+      def self.inherited(subclass)
+        ActiveModel::Serializer::Adapter.register(subclass.to_s.demodulize, subclass)
```
joaomdmoura: :+1:

lib/active_model/serializer/adapter.rb
```diff
@@ -32,9 +92,9 @@ def self.create(resource, options = {})
         klass.new(resource, options)
       end
 
+      # @see ActiveModel::Serializer::Adapter.get
       def self.adapter_class(adapter)
-        adapter_name = adapter.to_s.classify.sub("API", "Api")
-        "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize
+        ActiveModel::Serializer::Adapter.get(adapter)
```
joaomdmoura: :+1:

test/serializers/adapter_for_test.rb
```diff
@@ -41,10 +43,106 @@ def test_raises_exception_if_invalid_symbol_given
...
+          "json".freeze              => ActiveModel::Serializer::Adapter::Json,
+          "json_api".freeze          => ActiveModel::Serializer::Adapter::JsonApi,
+          "flatten_json".freeze      => ActiveModel::Serializer::Adapter::FlattenJson,
+          "null".freeze              => ActiveModel::Serializer::Adapter::Null
+        }
+      end
+
+      def test_adapters
+        assert_equal ActiveModel::Serializer::Adapter.adapters.sort, [
+          "flatten_json".freeze,
```
joaomdmoura: :+1:

test/serializers/adapter_for_test.rb
```diff
@@ -20,7 +22,7 @@ def test_overwrite_adapter_with_symbol
         adapter = ActiveModel::Serializer.adapter
         assert_equal ActiveModel::Serializer::Adapter::Null, adapter
       ensure
-
+        ActiveModel::Serializer.config.adapter = @previous_adapter
```
joaomdmoura: Why `ensure` just here? I think you added in order to be able to test [this](https://github.com/rails-api/active_model_serializers/pull/1017/files#diff-a217107028e8b59eb12f0e5077b76509R110)
But if that is the case, we probably should also need to ensure the value of it on other tests there are also changing `ActiveModel::Serializer.config.adapter`.
Maybe wrap this kind of tests into some method defined here.

bf4: You'll notice it was an empty ensure before I got to it....  It could go in an after each teardown... This was just a minimal fix

test/serializers/adapter_for_test.rb
bf4: In a future pr I'll go through the app and freeze more strings

bf4:
rebased off of master...

bf4:
You'll likely want to take a look at the commit message in https://github.com/bf4/active_model_serializers/commit/cf6a074a1c62e3f5a30f90a0987b8cf049272953

lib/active_model/serializer/adapter/flatten_json.rb
```diff
@@ -1,19 +1,13 @@
-module ActiveModel
-  class Serializer
-    class Adapter
-      class FlattenJson < Json
-        def serializable_hash(options = {})
-          super
-          @result
-        end
+class ActiveModel::Serializer::Adapter::FlattenJson < ActiveModel::Serializer::Adapter::Json
```
sandstrom: I find the previous style more readable. That's also the syntax (I think) Rails uses internally, perhaps it would be a good idea to align with Rails code style.

Gemfile
sandstrom: I don't think this should be committed.

lib/active_model/serializer/adapter.rb
```diff
@@ -1,12 +1,14 @@
 module ActiveModel
   class Serializer
     class Adapter
+      UnknownAdapterError = Class.new(ArgumentError)
```
sandstrom: I think `class UnknownAdapterError < ArgumentError; end` is a more common syntax.

lib/active_model/serializer/adapter/json/fragment_cache.rb
```diff
@@ -1,16 +1,6 @@
-require 'active_model/serializer/adapter/fragment_cache'
-module ActiveModel
-  class Serializer
-    class Adapter
-      class Json < Adapter
-        class FragmentCache
+class ActiveModel::Serializer::Adapter::Json::FragmentCache
```
sandstrom: Similarly, I find the previous class syntax more readable, and it aligns with Rails coding style.

bf4: > 
Thanks for your comments!  I think aligning with Rails style is a good idea.  Re: this change plz see the commit for details why
>

Gemfile
bf4: Any particular reason?

lib/active_model/serializer/adapter.rb
bf4: > 
They're both pretty common for exceptions, and is pretty much required for structs.  That the poro fixtures do it, though, I think is problematic 
>

lib/active_model/serializer/adapter/json/fragment_cache.rb
sandstrom: *(for reference, the commit message is here https://github.com/bf4/active_model_serializers/commit/cf6a074a1c62e3f5a30f90a0987b8cf049272953)*

To avoid the collision issue I think it's good to only use `module` for namespacing. I think that should avoid the issue?

I also think that automatic registration when subclassing is very magical. My guess is the common 'Rails-way' is to do on-the-fly lookups when an adapter is needed (constantize a string and look in some pre-defined directory, e.g. `app/serializers/adapters`) or to register them explicitly in an initializer.

But I'm no Rails-export, I may be wrong on the conventions. Perhaps @spastorino can weight in?

bf4: Adapter is a class.  For now, adapter can't be a module.  Anyway, the module convention exists to make things work.  This change forces all non-loaded superclasses to be autoloaded when referenced. That's the only way to do it now.

If you think using an inherited hook is magical... Well, it's a pretty fundamental part of Ruby meta programming.  Rails uses these hooks all over the place. If you forget the hook exists, do you prefer the way this pr behaves?  I.e.  You can name your adapter whatever you want.

P.s.  I do intend to fix the name spacing issue after this is merged

sandstrom: I understand that's the only way if the `inherited` hook is used.

I'm thinking that it would be better to register adapters manually, without using the `inherited` hook, i.e. have people call `ActiveModel::Serializer::Adapter.register` directly.

Perhaps some inspiration can be taken from how `ActiveJob` adapters are wired[1].

[1] https://github.com/rails/rails/blob/master/activejob/lib/active_job/queue_adapter.rb

bf4: > I hope I don't come across as negative,

Me as well.  Thanks for bringing it up.  I think I was being a little snippy.

In terms of the feature, I'm not married to the inner workings, but I do like the `Adapter.get` and `Adapter.register` as the source of truth for getting and setting adapters. And that it removes the requirement that a custom adapter be in the ActiveModel::Serializer::Adapter namespace

In terms of borrowing the interface from ActiveJob, I think that's a really great idea!  cc @seuros

bf4: Re: the namespace issue, I'm going to update the PR description to reference it.  I also discussed it in https://github.com/rails-api/active_model_serializers/pull/1050#issuecomment-133295789 and https://github.com/rails-api/active_model_serializers/pull/1067#discussion_r37710003. The code in this PR was my first attempt to deal with https://github.com/rails-api/active_model_serializers/issues/1006 till I realized a simpler solution..

lib/active_model/serializer/adapter/flatten_json.rb
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1017#discussion_r38270180

bf4:
@bolshakov does this meet your needs for using your own adapter?

lib/active_model/serializer/adapter.rb
bf4: There's a possibility here that two adapters will have the same namespace...

lib/active_model/serializer/adapter.rb
joaomdmoura: Yup, both are common, I'm up for both, but let's keep if in order to keep the way we are doing this across the code

lib/active_model/serializer/adapter/json/fragment_cache.rb
joaomdmoura: Great discussion, I also agree the readability is something we need to crave, and indeed we can remove `inherited` hook, and use `register`.
This is a huge refactor / adding some new interfaces already, let's keep it for now, merge it, and work this out on the namespace issue, how do you feel about it?

lib/active_model/serializer/adapter.rb
```diff
@@ -14,9 +16,67 @@ def self.create(resource, options = {})
...
+        def adapters
+          adapter_map.keys.sort
+        end
+
+        # Adds an adapter 'klass' with 'name' to the 'adapter_map'
+        # Names are stringified and underscored
+        def register(name, klass)
+          adapter_map.update(name.to_s.underscore => klass)
+          self
+        end
```
bf4: @joaomdmoura Do we want a `remove` method?  In the tests I faked it.. edge case..

lib/active_model/serializer/adapter.rb
bf4: I split this off just to make it a little easier to read..

lib/active_model/serializer/adapter.rb
joaomdmoura: No, edge case, if someone bring this up then we can check

lib/active_model/serializer/adapter/json/fragment_cache.rb
joaomdmoura: Great discussion, indeed I agree that the previous class syntax was more readable, but have the hook might be a good trade off, this is a big refactor and also add some new behaviors, let's keep it the way it is now, and try to address a way of doing this and gaining back readability on other PR.

docs/general/adapters.md
```diff
@@ -49,3 +49,7 @@ If you want to have a root key in your responses you should use the Json adapter
 ```ruby
 ActiveModel::Serializer.config.adapter = :json
 ```
+
+## Registering an adapter
+
+ActiveModel::Serializer::Adapter.register(:my_adapter, MyAdapter)
```
bolshakov: :+1:

lib/active_model/serializer/adapter/flatten_json.rb
bolshakov: Ruby Style Guide encourage the previous style

bolshakov:
I don't see the reason why I should register adapter before using it.

lib/active_model/serializer/adapter/flatten_json.rb
bf4: even for classes as namespaces? link plz?  (besides that it doesn't work in this case, I think it's wrong)

lib/active_model/serializer/adapter.rb
bf4: I only use the `Class.new` and `Struct.new` forms for dynamic class construction.

- `Foo = Struct.new` because otherwise you get superclass mismatch exceptions
- `Class.new` for one-liner Exception or sentinal objects, e.g. `BadObject = Class.new; if obj == BadOjbect....`.  (or when super serious meta-programming requires it).

bf4:
> I don't see the reason why I should register adapter before using it.

@bolshakov, Thanks for asking. It's because, as the PR subject states: 

> Make Adapters registerable so they are not namespace-constrained

Right now, you can **only** use an adapter **if** it is a constant under `ActiveModel::Serializer::Adapter`.  There's lots of people, myself included, who don't want that restriction.  I've linked to these issues in the PR description, including issue #1006.

bolshakov:
Another option is just pass a class instead of symbol. So if you pass symbol, treat it as an adapter name inside AMS namespace. If you pass class, use it as adapter

bf4:
@bolshakov so how do you imagine the interface.  Right now you can pass a serializer as a class, but an adapter only as a symbol/string of Adapter.constants that is discoverable by safe_constantize.  Do you want to stop supporting adapter names and support only classes?  If you support adapter names, then why not map them once, rather than evaluate them every time you want to use it.  I'd rather pass in `:my_json` than `ActiveModel::Serializer::Adapter::MyJson` and I'd rather pass in `MyJson`, and hey, I could even pass in `my_banana` for `MyJson` because the mapping would no longer depend on the class name... it's unclear to me if you are asking 'what if' and 'why' or 'i wouldn't use this' or 'i wouldn't want to maintain this' or 'I don't think this is an improvement'.

joaomdmoura:
The need to register and adapter might sounds like a pain but it's something that could easily be on the AMS initializer you will need to configs, it also concentrate the adapter handling into a single place.
But I know it might might be tricky for some ppl that are trying to create new adapters, so would be nice to have a `How to` article on the new docs on new adapters creation and usage, can you pull this off @bf4 :grin: 
But this looks solid and good to me.

bf4:
Rebased and made diff easier to read by not changing indentation.

Will add the How To later today.

joaomdmoura:
Nice, I'll wait for it then @bf4. Meanwhile I merged some other PRs so you might need to rebase it again, sorry :smile:

bf4:
@bolshakov So, are we talking about differences in internal implementation or usage?

Here are the changes in usage which I should promote to the description (specific implementation is paraphrased for simplicity):

| | before | after |
|----|----|----|
| `AM::Serializer.adapter` | config.adapter is `:json_api` => `adapter_class = AM::S::Adapter.adapter_class(config.adapter)` <br><br> config.adapter is `Banana` => `adapter_class = Banana` <br><br> `adapter_class && adapter_class or raise ArgumentError` | `AM::S::Adapter.get(config.adapter)` |
| `AM::S::Adapter.adapter_class(adapter)` | `"ActiveModel::Serializer::Adapter::#{adapter.to_s.classify.sub('API', 'Api')}".safe_constantize` | `AM::S::Adapter.get(adapter)` |
| `AM::S::Adapter.register(name, klass)` | ----- | `adapter_map.update(name.to_s.underscore => klass)` |
| `AM::S::Adapter.get(adapter)` | ---- |  adapter is `Banana` => `adapter` <br><br> adapter is `:json_api` => `adapter_map.fetch(adapter.to_s.underscore)` <br><br> &nbsp;&nbsp; if `:json_api` isn't registered, then `"ActiveModel::Serializer::Adapter::#{adapter.to_s.underscore.classify.tr('API', 'Api')}".safe_constantize` `or raise UnknownAdapterError` |

So, the basic behavior is the same.  

1. If `config.adapter` is a class, use it.  
2. else given `config.adapter` or the adapter_options `adapter: adapter`
  1. **new**: if an adapter is registered for that name, use it
  1. else Try to get an adapter by that name in the `AM::S::Adapter` namespace
  1. else raise `s/ArgumentError/UnknownAdapterError`

So, by keeping track of known adapters we 

1. don't need to `safe_constantize` more than one per adapter
2. we can register an adapter with any name we choose, and thus aren't limited to using adapters in the `AM::S::Adapter` namespace


You wrote above:

> So if you pass symbol, treat it as an adapter name inside AMS namespace. If you pass class, use it as adapter

Right now, that's only true for `config.adapter`. Now it will be true for using adapter_options as well, so that's more consistent :)

Plus, you can still pass a symbol in the controller (etc.) to get any adapter.  Without 'registering', we'd have to either only allow symbols/strings for `AM::S::Adapter` namespaced adapters or be willing to constantize arbitrary symbols/strings, which is a security concern, imho.

Plus, it's just nice not to rely on evaluating code at all. :)

lib/active_model/serializer/adapter.rb
```diff
@@ -14,9 +16,67 @@ def self.create(resource, options = {})
...
+
+      # Only the Adapter class has these methods.
+      # None of the sublasses have them.
+      class << ActiveModel::Serializer::Adapter
+        # @return Hash<adapter_name, adapter_class>
+        def adapter_map
+          ADAPTER_MAP
+        end
+
+        # @return Array<adapter_name>
```
bolshakov: Better:

    @return Array<Symbol> list of adapter names

See http://yardoc.org/types.html

lib/active_model/serializer/adapter.rb
```diff
@@ -14,9 +16,67 @@ def self.create(resource, options = {})
...
+        def adapters
+          adapter_map.keys.sort
+        end
+
+        # Adds an adapter 'klass' with 'name' to the 'adapter_map'
+        # Names are stringified and underscored
+        def register(name, klass)
```
bolshakov: @param [Symbol, String] name of the registered adapter
    @param [Class] klass - adapter class itself
    @example 
        AMS::Adapter.register(:my_adapter, MyAdapter)

lib/active_model/serializer/adapter/json/fragment_cache.rb
bf4: I've tried to remove the inherited hook, and replace it with a method in the Adapter class itself, but even then, I'd need to prefix **every** file that may define `class Adapter` rather than reopen it, with something like `adapter = adapter = ActiveModel::Serializer::Adapter # Need to load the Adapter before opening it below`

Not even `Adapter.eager_load!` is sufficient.

lib/active_model/serializer/adapter.rb
bf4: yeah, just me being a little lazy because we're not actually using yardoc (yet?)

lib/active_model/serializer/adapter.rb
```diff
@@ -14,9 +16,67 @@ def self.create(resource, options = {})
...
+          adapter_map.update(name.to_s.underscore => klass)
+          self
+        end
+
+        # @param  adapter [String, Symbol, Class] name to fetch adapter by
+        # @return [ActiveModel::Serializer::Adapter] subclass of Adapter
+        # @raise  [UnknownAdapterError]
+        def get(adapter)
+          # 1. return if is a class
```
bf4: @sandstrom re: ActiveJob [QueueAdapter](https://github.com/rails/rails/blob/a11571cec3213753d63ac3e6b4bb3b97fe2594a6/activejob/lib/active_job/queue_adapter.rb#L27-L60) and [QueueAdapters](https://github.com/rails/rails/blob/a11571cec3213753d63ac3e6b4bb3b97fe2594a6/activejob/lib/active_job/queue_adapters.rb#L123-L133) I think I like `lookup` more than `get`, so thanks for pointing me that way ;)  You think I should change it?

However, I think there are some limitations in that interface in that 

1. activejob queue adapters are only used when assigned via `queue_adapter =`, but there's no way to know if and when an AMS adapter will be used
2. I like `const_get` instead of `safe_constantize`, but we'd still need to rescue a NameError and use the AM::S::Adapter namespace, at least for dynamic lookup
3. I like that you can pass [`interpret_adapter` a class and it can check it for behavior](https://github.com/rails/rails/blob/a11571cec3213753d63ac3e6b4bb3b97fe2594a6/activejob/lib/active_job/queue_adapter.rb#L52-L56), but I think that's outside the scope of this PR
4. I'm glad to get the reminder about private_constant

bf4:
@joaomdmoura @bolshakov @sandstrom All cleaned up!

sandstrom:
@bf4 Awesome work on this!

While we disagreed on some of the details I'm overall very positive to making adapters more flexible. Great! :boat:

bolshakov:
:+1:

bf4:
Failures on appveyor

- https://ci.appveyor.com/project/joaomdmoura/active-model-serializers/build/52/job/0b8lm5r3dr38p05g
- https://ci.appveyor.com/project/joaomdmoura/active-model-serializers/build/52/job/a2t820ffb557ytyc
- https://ci.appveyor.com/project/joaomdmoura/active-model-serializers/build/52/job/ac270hiwxljkrhwd

```plain
 1) Failure:
ActiveModel::Serializer::AdapterForTest#test_adapters [C:/projects/active-model-serializers/test/serializers/adapter_for_test.rb:73]:
--- expected
+++ actual
@@ -1 +1 @@
-["flatten_json", "json", "json_api"]
+["flatten_json", "json", "json_api", "null"]
 ```

- https://ci.appveyor.com/project/joaomdmoura/active-model-serializers/build/52/job/t3erlut52i0t22j3
- https://ci.appveyor.com/project/joaomdmoura/active-model-serializers/build/52/job/fyady8yp0kn2lair

```plain
 1) Failure:
ActionController::Serialization::ImplicitSerializerTest#test_cache_expiration_on_update [C:/projects/active-model-serializers/test/action_controller/serialization_test.rb:394]:
--- expected
+++ actual
@@ -1 +1 @@
-"{\"id\":1,\"title\":\"ZOMG a New Post\",\"body\":\"Body\",\"comments\":[{\"id\":1,\"body\":\"ZOMG A COMMENT\"}],\"blog\":{\"id\":999,\"name\":\"Custom blog\"},\"author\":{\"id\":1,\"name\":\"Joao Moura.\"}}"
+"{\"id\":1,\"title\":\"New Post\",\"body\":\"Body\",\"comments\":[{\"id\":1,\"body\":\"ZOMG A COMMENT\"}],\"blog\":{\"id\":999,\"name\":\"Custom blog\"},\"author\":{\"id\":1,\"name\":\"Joao Moura.\"}}"
 ```

**Comment by bf4** [comment]
Fixed the build

**Comment by joaomdmoura** [comment]
Nice work! Really thank you everyone for helping out with this one! 
It's a big one and will definitely help us into building new functionalities.
Merging.

File: lib/active_model/serializer/adapter/json_api.rb [hunk 17]
```diff
@@ -62,6 +60,12 @@ def resource_identifier_type_for(serializer)
           end
         end
 
+        def add_relationships(resource, name, serializers)
```
beauby: Why, why, why? :D

bf4: > 
Rebase foul
>

PR #1021, opened 2015-07-28
Context aware serializers

### LINKED ISSUE #801: Context-aware predicates in serializers
Re: PR #498, I'd like to open up a discussion on a new potential DSL for serializers.

I've gotten a lot of milage out of this library in the past back by hacking in context-awareness into serializer instances. I'd love it if they were a core feature.

At any moment in the serialization process, a serializer ought to be able to know where it's at in relation to the grand scheme of things. This opens up a few cool options: condensing multiple serializers into one, easily avoiding expensive sub-serializations, etc.

As an example, the DSL I've made for myself behaves as follows:

```ruby
class ContextAwareSerializer < ActiveModel::Serializer

  def initialize(object, options={})
    @associated   = options[:associated]
    @listed       = options[:listed]
    @show_details = options[:details]
    super
  end

  def associated?; @associated; end
  def listed?;     @listed; end

  def nested? listed? or associated?; end
  def show_details?
    if @show_details or @show_details.nil? # If not ~explicitly~ set to false,
      @show_details or not nested?         # only show details if asked for, or at top level
    end
  end

  def filter(keys)
    if show_details?
      super
    else
      super - detail_keys
    end
  end

end
  
class UserSerializer < ContextAwareSerializer

  attributes :id, :uri, :name
  detailed attributes :first_name, :last_name, :bio

  has_one :account
  has_many :posts
  detailed has_many :comments

end
```

I've made it such that serializers instantiated by `ActiveModel::ArraySerializer` receive `options[:listed]`,`ActiveModel::Serializer::Association` passes in `options[:associated]`, and the standard custom options from the controller provide the `options[:details]`.

I'd like to propose incorporating the `nested` DSL: adding the injection of and predicates for `listed?`, `associated?`, and `nested?`.

I think the `detailed` DSL requires more consideration from other people's perspectives and usecases. Certainly, it can be far more easily hacked in than the nested DSL, since you don't have to monkey patch any `active_model_serializer` classes to achieve it. Perhaps it merits a place in the library as a subclass, or disabled-by-default feature. YMMV.

christhekeele:
Closes #801.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,28 @@ class PostSerializer < ActiveModel::Serializer
 end
 ```
 
+### Context-aware predicate methods
+
+If you want to serialize attributes or associations differently depending on the context they're serialized, you can use:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  attributes :id, :body
+
+  has_many :comments
+
+  def body
+    if listed? or associated? # same as `if nested?`
+      object.body.lines.first + '...'
+    else
+      object.body
+    end
+  end
+end
+```
+
+If object is serialized as part of a nested list of objects, `listed?` will be true. If object is serialized as an association of another object, `associated?` will be true. If either are true, `nested?` will be true.
+
 ### Built in Adapters
 
 #### FlattenJSON
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -145,6 +145,15 @@ module ActiveModel
         assert expected_association_keys.include? :writer
         assert expected_association_keys.include? :site
       end
+
+      def test_associated_and_nested_predicates
+        @post_serializer.each_association do |key, association|
+          if key == :blog
+            assert association.associated?
+            assert association.nested?
+          end
+        end
+      end
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,3 +10,4 @@
   * uses model name to determine the type [@lsylvester]
   * remove root key option and split JSON adapter [@joaomdmoura]
   * adds FlattenJSON as default adapter [@joaomdmoura]
+  * adds support for `listed?`, `associated?`, and `nested?` predicates [@christhekeele]
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -241,6 +241,18 @@ module ActiveModel
       @serializers_cache ||= ThreadSafe::Cache.new
     end
 
+    def listed?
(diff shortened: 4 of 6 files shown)


test/serializers/associations_test.rb
```diff
@@ -145,6 +145,15 @@ def test_associations_custom_keys
...
       end
+
+      def test_associated_and_nested_predicates
+        @post_serializer.each_association do |key, association|
+          if key == :blog
+            assert association.associated?
+            assert association.nested?
+          end
+        end
+      end
```
bf4: would be good to have a sanity check assertion that the assertion were called. e.g. before the block set `called=false` and after the assertions set `called=true` then after the block `assert called`

bf4:
So, basically, 

- a resource's associated serializers now know they are `associated?` and 
- the serializers in a collection serializer know they are `listed?`
- and we refer to both associated resources and collected resources as `nested?`

README.md
```diff
@@ -170,6 +170,28 @@ class PostSerializer < ActiveModel::Serializer
...
+
+  def body
+    if listed? or associated? # same as `if nested?`
+      object.body.lines.first + '...'
+    else
+      object.body
+    end
+  end
+end
+```
```
bf4: So, in the example here, when you're serializing a post, you want to use the body attribute, but when serializing a blog, you want the associated posts to use only the first line of the post body?

To me, that sounds like a scenario where I'd do something like

```ruby
 attributes :body
end
  has_many :posts, serializer: BlogPostSerializer
end
class BlogPostSerializer < ActiveModel::Serializer
  def body
    super.lines.first
  end
end
```

Was that an approach you considered?

christhekeele: Originally, yes. Most of it ended up being lots of boilerplate, creating many classes and scattering them throughout my code to achieve a fairly common behaviour: essentially, unless you're serializing a single object for display, only serialize enough information to get to that canonical full-details view, with a few extra display parameters to generate links.

On the other hand, as I've updated to `v0.10.0`, adapters have really helped handle that for me. They provide a better abstraction to handle these concerns; and I no longer think this patch is useful. 

Updating my patch helped keep my tests green, but in the last few days I've discarded it after years of it serving me well. All of the motivations for it, #801, and #498 are covered with adapters, and I, for one, welcome our new `v0.10.0` overlords. 

:keycap_ten:
:raised_hands:
:smile:

joaomdmoura:
@christhekeele have just checked it, I'm glad that you are happy with the 0.10.x version! Just let us know if you need something.

PR #1026, opened 2015-07-30
Bump Version Number to 0.10.0.rc2

jfelchner:
Due to the fact that users need to switch from the released version to `master` occasionally to pull in upstream bugfixes, it's important that this version number stays in sync with the released version.


joaomdmoura:
Agree, @jfelchner good catch, merging.

Issue #974, opened 2015-06-28
JSON API adapter word separator issue


hugopeixoto:
I'm using the JSON API adapter with ember-data and I'm having a small compatibility issue.

The new JSON API adapter currently does not transform the attribute keys in any way. I have an attribute with an underscore in my model, it will be left untouched. ember-data/jsonapi looks up attribute names with dash-separated words. This means that these two adapters cannot be used together without changing one of them.

The jsonapi spec does not specify that words must be dash-separated, but it does recommend it (https://github.com/rails-api/active_model_serializers/pull/534#issuecomment-116296477, [reference](http://jsonapi.org/recommendations/#naming)). ember-data follows this recommendation (https://github.com/emberjs/data/issues/3455).

It would be nice to have a way to use the adapter in the recommended way.
Should this transformation be handled specifically by the jsonapi adapter?
Or should AMS provide a general key transformation option (this feature dropped on 0.10.x: https://github.com/rails-api/active_model_serializers/pull/534)?

EDIT:

My current workaround was to define a base serializer, from which the actual serializers inherit:

```ruby
class BaseSerializer < ActiveModel::Serializer
  def attributes *args
    Hash[super.map do |key, value|
      [key.to_s.dasherize.to_sym, value]
    end]
  end
end
```

hhff:
:+1: I'm having this issue also.  @hugopeixoto - did you find a way around this for the time being?

hhff:
IMO - Ember Data should handle this.

That said - AMS should provide a dasherize option also - as the spec recommends it.  Rails is all about underscores, so we'd have to do something like this:

```
attributes :'full-name'

def full-name
  object.full_name
end

end
```

Not sure if this would even parse...

hugopeixoto:
@hhff: There's a `:key` option in the `#attribute` (singular) method which you could use:

```ruby
attribute :full_name, key: "full-name"
```

I noticed that I forgot to add my current fix to the issue. I have edited the original post with a potential workaround, if anyone runs into this problem.

hhff:
thanks @hugopeixoto !  seems like this should also be a global config, however...

bf4:
@hhff ED does, if you're using the AMS adapter...

hhff:
Thanks @bf4, thinking more about the new JSONAPI adapter, as AMS adapter is being removed from the official ember data package in favour of it.  The suggestions in the above ember data thread are fine tho.

I do think ams should handle this transform for us tho.

[...]

PR #1029, opened 2015-07-31
Allow Key Format to Be Overridden and Have Sensible Defaults

jfelchner:
### Overriding the Key Format

You can specify that serializers use a different style for the format of the
keys that are sent over. This can be done at the config, class or instance
level.

```ruby
# Global
ActiveModel::Serializer.config.key_format = :lower_camel

# Per Serializer
class BlogLowerCamelSerializer < ActiveModel::Serializer
  key_format :lower_camel
end

# Per Instance
BlogSerializer.new(object, key_format: :lower_camel)
```

#### Currently Supported Key Formats ####

* `:unaltered`
* `:lower_camel`
* `:underscore`
* `:dasherize`

The default key format will be set depending on the adapter that you specify.
For example, the JSON adapter defaults to `lower_camel` whereas the JSON API
adapter defaults to `dasherize`.  Any other adapter will default to `unaltered`.

References: #534, https://github.com/emberjs/data/issues/3455
Closes: #974 

Merry Christmas in July everyone. :grinning: :christmas_tree: :santa:


jfelchner:
I looked at the build and everything passed.  No idea why it marked it as a failure.

README.md
```diff
@@ -170,6 +170,35 @@ class PostSerializer < ActiveModel::Serializer
 end
 ```
 
+### Overriding the Key Format
```
joaomdmoura: :+1: It would be awesome if along with this PR you could add some Docs to our [new documentation](https://github.com/rails-api/active_model_serializers/tree/master/docs) we have being working on before release.
Maybe event add some article about integrating it with ember.js using our `How to` section.

lib/active_model/serializer.rb
```diff
@@ -9,8 +9,10 @@ class Serializer
     autoload :Adapter
     autoload :Lint
     autoload :Associations
+    autoload :KeyFormat
```
joaomdmoura: Here you are including the file when it's evaluated, there is no advantage on the autoload. It's not a blocking issue, but in this case you could use the usual `require` or `require_relative` without the autoload. 
We are already updating it to do the same thing with the other included files.

lib/active_model/serializer/array_serializer.rb
```diff
@@ -2,12 +2,22 @@ module ActiveModel
...
+
+      extend ActiveSupport::Autoload
+
+      autoload :Configuration
+      autoload :KeyFormat
+
       include Enumerable
+      include Serializer::Configuration
+      include Serializer::KeyFormat
+
```
joaomdmoura: Same thing related to `autoload` here

lib/active_model/serializer/configuration.rb
```diff
@@ -7,6 +7,7 @@ module Configuration
       included do |base|
         base.config.array_serializer = ActiveModel::Serializer::ArraySerializer
         base.config.adapter = :flatten_json
+        base.config.key_format = nil
```
joaomdmoura: Any particular reason to declare it as `nil`?

joaomdmoura:
:+1: to the PR itself, I made some comments after reviewing it and we also need to check the tests.
Would be great if we could also squash some commits to keep the ones that makes more sense.
Awesome work @jfelchner, get it up! looking forward to have it on 0.10

jfelchner:
@joaomdmoura I'll get on it and comment back here on Monday or Tuesday.  Thanks!

mcmire:
Just want to come by and :+1: this, this looks like a nice addition.

lib/active_model/serializer.rb
jfelchner: @joaomdmoura I never use `autoload`, I was just following convention. :+1: :grinning: 

Done!

lib/active_model/serializer/array_serializer.rb
jfelchner: Done!

lib/active_model/serializer/configuration.rb
jfelchner: @joaomdmoura I just wanted to be explicit so that anyone spelunking in the code can easily look at this file and see what all the global configuration options are.  I'm not tied to it if ya want me to remove it.

jfelchner:
@joaomdmoura ok, I hit a bit of a snag.  I found a bug. :-/

The problem is that we don't want to convert keys for hashes that are inside the values for attributes.  Currently it just goes recursively as deep as it needs to to make sure that everything is converted.

For example if we have this:

```ruby
            @second_post = Post.new(
                            id:                   2,
                            title:                'New Post',
                            body:                 'Body',
                            multi_word_attribute: {
                              multiple_words: 'here are some words'
                            })
```

we don't want `multiple_words` to become `multiple-words`.  The implementation I have _feels_ really nice, but it obviously won't work.

Maybe an abstraction is missing or a refactoring is required to make this the "easy" change?  I don't have enough familiarity with the codebase so I thought you or someone else might be able to provide some insight.

joaomdmoura:
Get it @jfelchner, indeed the behavior you described would be most accurate.
We could accomplish it by calling `format_key` on multiple places: `root`, `attributes` (method), `associations` (method).
I'm not sure how it would like but it seems okay for me in order to keep it simple. 
If you want to, we can try to pair on it.
Let me know your thoughts about it :smile:

lib/active_model/serializer/array_serializer.rb
```diff
@@ -1,13 +1,21 @@
...
       NoSerializerError = Class.new(StandardError)
+
       include Enumerable
+      include Serializer::Configuration
+      include Serializer::KeyFormat
+
       delegate :each, to: :@objects
 
-      attr_reader :root, :meta, :meta_key
+      attr_reader :root, :meta, :meta_key, :options
```
bf4: It doesn't appear the you're using the options reader in any public contexts.  Perhaps we can make it private?  (It's just so the KeyFormat mixin can do `options[:key_format]`, right?)

test/serializers/key_format_test.rb
```diff
@@ -0,0 +1,48 @@
...
+        @first_post.author = @author
+        @second_post.author = @author
+
+        @first_post.comments = []
+        @second_post.comments = []
+      end
+
+      def test_overrides_default_adapter_key_format
+        @serializer = PostWithKeyFormatSerializer.new(@first_post)
+        @adapter = ActiveModel::Serializer::Adapter::JsonApi.new(@serializer)
```
bf4: For future-proofing and consistency, we may want to instead do it the way the controller does (IIRC)

```ruby
serializable_resource = ActiveModel::SerializableResource.serialize(@first_post, adapter: :json_api, serializer: PostWithKeyFormatSerializer)
serializable_resource.as_json
```

e.g. see https://github.com/rails-api/active_model_serializers/blob/b78126565c7c8d29102cc8a8adf66ddf228ff032/test/serializers/meta_test.rb#L115-L117

```ruby
      def load_adapter(options)
        options = options.merge(adapter: :json-api, serializer: PostWithKeyFormatSerializer)
        ActiveModel::SerializableResource.new(@first_post, options)
      end
```

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -39,6 +40,10 @@ def fragment_cache(cached_hash, non_cached_hash)
           JsonApi::FragmentCache.new().fragment_cache(root, cached_hash, non_cached_hash)
         end
 
+        def default_key_format
+          :dasherize
```
bf4: Awesome! Have you benchmarked the performance impact of mutating the hash by default?

README.md
```diff
@@ -170,6 +170,35 @@ class PostSerializer < ActiveModel::Serializer
...
+#### Currently Supported Key Formats ####
+
+* `:unaltered`
+* `:lower_camel`
+* `:underscore`
+* `:dasherize`
+
+The default key format will be set depending on the adapter that you specify.
+For example, the JSON adapter defaults to `lower_camel` whereas the JSON API
+adapter defaults to `dasherize`. Any other adapter will default to `unaltered`.
```
bf4: :100:

lib/active_model/serializer/adapter.rb
```diff
@@ -94,6 +98,38 @@ def include_meta(json)
         json[meta_key] = meta if meta
         json
       end
+
+      def key_format
+        serializer.key_format || default_key_format
```
bf4: So, this essentially allows a serializer to 'override' the default_key_format for the current adapter?

lib/active_model/serializer/adapter.rb
```diff
@@ -94,6 +98,38 @@ def include_meta(json)
...
+      end
+
+      def format_keys(formattable)
+        return formattable if !formattable.respond_to? :each_with_object ||
+                              key_format == :unaltered
+
+        if formattable.respond_to? :keys
+          formattable.each_with_object({}) do |(key, value), formatted_hash|
+            formatted_hash[format_key(key)] = format_keys(value)
```
bf4: idle question sparked by recursion: might we want to cache key conversions (like tail call optimization)? I imagine many of them repeat.

Also, the JSON library has a `recurse_proc` method, which might be useful for doing the key formatting in`def to_json` rather than `def serializable_hash`

joaomdmoura:
Hey @jfelchner, would be awesome if you could read through #1052.
It's deeply related to what you are working on.

I merged it for now but I would like to handle it on this PR following the same approach as well.

Let me know if there is anything I can do in order to make it easier for you :smile:

jfelchner:
@joaomdmoura perfect! I'll take a look. Just been busy the last few days. 😊

jfelchner:
BTW @joaomdmoura I'm absolutely open to pairing on this. 👍🏻

lib/active_model/serializer.rb
```diff
@@ -11,6 +12,7 @@ class Serializer
     autoload :Associations
     include Configuration
     include Associations
+    include KeyFormat
```
joaomdmoura: I like the organization that use external classes bring to the code, but taking into account the complexity of `KeyFormat` Class how do you feel about bring this logic to the `Serializer` itself in other to keep it concise? 
I'd like to heard your thoughts about it :smile:

lib/active_model/serializer/adapter.rb
joaomdmoura: > cache key conversions

:+1: 

> Also, the JSON library has a recurse_proc method, which might be useful for doing the key formatting indef to_json rather than def serializable_hash

`as_json` and yup, it could be nice if possible to centralize this on a single method on the `Adapter`, but idk how it would look like after we discussed [here](https://github.com/rails-api/active_model_serializers/pull/1029#issuecomment-130194632) and on the following comments.

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: yep that would be cool :smile: not required for the PR itself, but I'm curious about it. That remembers me the Benchmark PR I had, I might need to pick that up again.

test/serializers/key_format_test.rb
joaomdmoura: :+1:

joaomdmoura:
Awesome I did just a few comments, because I know the implementation might change after some changes you will do.

> I'm absolutely open to pairing on this.

Cool! Find on Twitter as @joaomdmoura, I will send you my email and we can try to schedule something using screenhero or whatever. But don't wait for me :grin:

lib/active_model/serializer/adapter.rb
```diff
@@ -94,6 +98,38 @@ def include_meta(json)
...
+          end
+        elsif formattable.respond_to? :<<
+          formattable.each_with_object(formattable.class.new) do |item, shovelable|
+            shovelable << format_keys(item)
+          end
+        end
+      end
+
+      def format_key(formattable_key)
+        case key_format
```
timurvafin: Assume we have namespaced object `Admin::User` then key should be `user` instead of `admin/user`. What do you think about adding something like `formattable_key.to_s.demodulize`?

jfelchner:
@timurvafin underscore does that by default.

jfelchner:
@timurvafin oh I see what you're saying, you want the opposite. No I don't think that would be a good idea. They are two different classes and therefore two different types. If someone wants to do that they can override manually by overriding `json_type` on the serializer. Also, this PR doesn't affect values, just the keys.

timurvafin:
@jfelchner probably AMS by default should generate something more json-like for `Admin::Post` then `admin/post`. Maybe something like `admin-post` or `admin_post`?

jfelchner:
@timurvafin we're using AMS with an app with multiple Rails engines and are using `/` between our modules and `-` between words. Unfortunately JSON API isn't clear on what to use and IMO there definitely needs to be a way to differentiate `Admin::User` from `AdminUser` so `/` seemed as good a choice as any.

lib/active_model/serializer/adapter.rb
bf4: @timurvafin Do you have a ref for `Admin::User` -> `user`?  I'm not sure what I would do in such a case.

bf4:
@jfelchner @timurvafin 

> Unfortunately JSON API isn't clear on what to use and IMO there definitely needs to be a way to differentiate Admin::User from AdminUser so / seemed as good a choice as any.

might be worth creating an issue in https://github.com/json-api/json-api/ and linking here

jfelchner:
@bf4 :+1:

bf4:
@jfelchner You want to do it?  An alternative might be that in the e.g. Admin::UserSerializer we require specifying the mapping as however we want it.

jfelchner:
@bf4 yeah I'll do it. And I think that there both needs to be a standard as well as a way to override in the serializer.

jfelchner:
@bf4 also sorry I haven't responded to your comments yet. I haven't had a chance to circle back to this. They were welcomed! 😄

bf4:
@jfelchner no problem.  There's always so much to do!

lib/active_model/serializer/adapter.rb
bf4: ref: https://github.com/json-api/json-api/issues/850

bf4:
@jfelchner I just made a simple issue re: dasherizing Admin::User https://github.com/json-api/json-api/issues/850

joaomdmoura:
I still supper interest on this :smile: are you going to keep working on this @jfelchner ?

beauby:
I think this is a must. As suggested in #1122 (which is JSONAPI specific, but could easily be made to play well with this PR), it might be interesting to offer the possibility to set a chain of formatters, so that people can, for instance, remove some namespacing with a regexp, then singularize the resource name, and then camelcase it.

NullVoxPopuli:
needs rebase

jfelchner:
You all will be happy to know that after weeks of "I'm getting to this", I'm actually working on this _right now_ :)

bf4:
Noting that JSON API recommends `dasherized` keys in http://jsonapi.org/recommendations/#naming

> Member names SHOULD contain only the characters "a-z", "0-9", and the hyphen minus  ("-") as separator between multiple words.

But the spec is looser on what it requires http://jsonapi.org/format/#document-member-names, basically:

> A-Z, a-z, 0-9, -, _,  SPACE (not recommended), any UNICODE character except U+0000 to U+007F (not recommended)
> With some disallowed characters http://jsonapi.org/format/#document-member-names-reserved-characters

Which includes `SOLIDUS, "/"` So I'm thinking maybe it has to be `_`?

jfelchner:
@bf4 I'm sorry I'm not following.  Maybe what has to be a `_`?

bf4:
Sorry, @jfelchner was e.g. `Admin::User` as discussed in https://github.com/rails-api/active_model_serializers/pull/1029#discussion_r37211368

README.md
```diff
@@ -187,6 +187,35 @@ class PostSerializer < ActiveModel::Serializer
...
+
+BlogSerializer.new(object, key_format: :lower_camel)
+```
+
+#### Currently Supported Key Formats ####
+
+* `:unaltered`
+* `:lower_camel`
+* `:underscore`
+* `:dasherize`
```
bf4: might be nice to alias `dasherize` to `jsonapi`

bf4: Also, probably to give an example of each..

lib/active_model/serializer/adapter.rb
```diff
@@ -101,6 +101,10 @@ def fragment_cache(*args)
         raise NotImplementedError, 'This is an abstract method. Should be implemented at the concrete adapter.'
       end
 
+      def default_key_format
+        :unaltered
+      end
```
bf4: I'd stick this in the config, maybe something like https://github.com/rails-api/active_model_serializers/pull/1050/files#diff-ae1e38e96a8bf28ae37918a65c448c73R7

bf4: but generally I think this key formatting should go in e.g. Utils

README.md
```diff
@@ -187,6 +187,38 @@ class PostSerializer < ActiveModel::Serializer
...
+You can specify that serializers use a different style for the format of the
+keys that are sent over. This can be done globally or at the controller action
+level.
+
+Each adapter has a default format, so typically you're either not going to have
+to change this, or you'll edit the global setting.
+
+For example, the JSON adapter defaults to `lower_camel` whereas the JSON API
+adapter defaults to `dasherize`. Any other adapter will default to `unaltered`.
+
```
beauby: I think the JSON API adapter should default to `lower_camel` as well to avoid huge confusion on upgrade.

jfelchner: @beauby I am very much in the other camp on that.  The JSON API standard recommends dasherized keys and the couple JSON API JS libraries I've seen default to expecting _incoming_ requests to be formatted with dasherized keys.

In the vein of making this Just Work with JSON API, I think it should absolutely be dasherized.  Otherwise almost everyone who uses it will need to add the global config option to their projects.

I'll bow to the majority opinion, but I just wanted to voice my concern.

beauby: You're right, if JSON API recommends dashes, we should do dashes by default.

bf4: Yes, see https://github.com/rails-api/active_model_serializers/pull/1029#issuecomment-140972381 and https://github.com/json-api/json-api/issues/850

lib/active_model/serializer/adapter.rb
```diff
@@ -11,6 +13,7 @@ class Adapter
       autoload :Null
       autoload :FlattenJson
       autoload :CachedSerializer
+      include Configuration
```
bf4: So, in https://github.com/rails-api/active_model_serializers/pull/1050/files#diff-ae1e38e96a8bf28ae37918a65c448c73R11 I just use e.g. `ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}`  

the require could also be a problem since it's incompatible with rails autoload

bf4:
Related: https://github.com/rails-api/active_model_serializers/pull/1213

krzkrzkrz:
Any updates on this?

tylerwillingham:
I'm also interested in seeing this make it through. Is there anything I can do to contribute?

andrewhavens:
I came across this PR while looking for a way to specify `model_name` in my serializer. I would like to be able to pass a hash instead of a full model instance to a serializer (to get better performance). Is there already a way to do this in master? Does this PR address this issue?

bf4:
No, this would be a feature request to pass a naked attributes hash directly into the adapter, i think

B mobile phone

clshack:
Any updates on this? 
It's very useful

remear:
Resolved by #1574

ijunaidfarooq:
What is the final results now? This thing is being added to gem?

NullVoxPopuli:
@ijunaidfarooq see: https://github.com/rails-api/active_model_serializers/blob/master/docs/general/rendering.md#key_transform

PR #1030, opened 2015-07-31
allow subclasses of ActiveModel::Serializer to override associations

markus:
This makes it possible for subclasses of ActiveModel::Serializer to override associations based on @options.

### DIFF
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -97,10 +97,14 @@ module ActiveModel
         return unless object
 
         Enumerator.new do |y|
-          self.class._reflections.each do |reflection|
+          reflections.each do |reflection|
             y.yield reflection.build_association(self, options)
           end
         end
+      end
+
+      def reflections
+        self.class._reflections
       end
     end
   end
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -102,6 +102,13 @@ PostSerializer = Class.new(ActiveModel::Serializer) do
   end
 end
 
+CustomizablePostSerializer = Class.new(PostSerializer) do
+  def reflections
+    excludes = @options[:excludes] || []
+    super.select{ |r| excludes.exclude?(r.name) }
+  end
+end
+
 SpammyPostSerializer = Class.new(ActiveModel::Serializer) do
   attributes :id
   has_many :related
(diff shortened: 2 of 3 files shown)


joaomdmoura:
@markus I'm sorry about that but we had made a refactoring into our associations implementation this morning, could you rebase it please?

markus:
@joaomdmoura rebased.

lanej:
+1

joaomdmoura:
Just checked it again. Sorry for taking so long, there was a lot to catch up :grin:

So I can definitely see the need here, but I have some thoughts about the implementation.

Right now we have a `include` options that is used on json-api implementation and it's comparable to this `exclude`, so it might end up confusing someone, but it's just a matter o naming. (it can be hard)

A moth ago I opened a [PR](https://github.com/rails-api/active_model_serializers/pull/999) to make the `fields` option we already had available across all adapters and not only json-api. It basically enables you to filter on the `render` method which attributes and relationship attributes you want to return. I'm not sure if it enables you to not return an entire association as you can do with this `exclude` implementation. 

Despite of being able to do the same thing with both implementations I would be willing to have something the enables you to remove attributes and relationships (or just some of it's attributes), so, exactly the opposite of what `fields` do.

If you decide to go for it, I would also ask you to write some [DOCS](https://github.com/rails-api/active_model_serializers/tree/master/docs) about it as well :smile: We are trying to create this good documentation place for ppl using AMS.

Sorry for the longe reply :smile:

markus:
My intention was just to expose `reflections` as a method that subclasses can override. In my own application I have defined `can_include` as an alternative to `has_many` but my implementation of `can_include` is not ready for a pull request because it does not support other types of associations. The `:excludes` option in the test is just a simple example that overrides `reflections`.

markus:
Here is another example of what you can do with `reflections`:

    class SerializerBase < ActiveModel::Serializer

      class << self
        attr_accessor :_can_include
      end

      def self.can_include(*attrs)
        options = attrs.extract_options!
        self._can_include ||= []
        self._can_include += attrs
        association_type = options.delete(:type) || :has_many
        attrs.push(options)
        send(association_type, *attrs)
      end

      def reflections
        excludes = (self.class._can_include || []) - (@options[:includes] || [])
        self.class._reflections.select{ |r| excludes.exclude?(r.name) }
      end

    end

joaomdmoura:
Hey @markus, meanwhile we have decided to bring the older `filter` feature to the 0.10.x as you can check [here](https://github.com/rails-api/active_model_serializers/issues/1058). 
This will fix this issue, and kill my `fields` PR. 
Can you check if this will work out for you so we can close this and focus on that one? btw it's open to being picked by anyone so far.

markus:
#1058 looks good. I'll close this one.

PR #1031, opened 2015-07-31
Disallow to define multiple associations at once

bolshakov:
I remove this feature since defining multiple associations at once may lead to mess than defined with options:

```ruby
   has_many :comments, :posts, serializer: CoolSerializer
```


joaomdmoura:
Tks @bolshakov, totally agree, I'm merging it.

PR #1033, opened 2015-08-03
[WIP] Attributes refactoring

bolshakov:
* Move attributes related code to separate module 
* Refactor attributes definition

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -3,13 +3,13 @@ require 'thread_safe'
 module ActiveModel
   class Serializer
     extend ActiveSupport::Autoload
+    require_relative 'serializer/associations'
+    require_relative 'serializer/attributes'
 
     autoload :Configuration
     autoload :ArraySerializer
     autoload :Adapter
     autoload :Lint
-    autoload :Associations
-    autoload :Attributes
     include Configuration
     include Associations
     include Attributes
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -9,12 +9,12 @@ module ActiveModel
     autoload :Adapter
     autoload :Lint
     autoload :Associations
+    autoload :Attributes
     include Configuration
     include Associations
+    include Attributes
 
     class << self
-      attr_accessor :_attributes
-      attr_accessor :_attributes_keys
       attr_accessor :_urls
       attr_accessor :_cache
       attr_accessor :_fragmented
@@ -26,33 +26,10 @@ module ActiveModel
     end
 
     def self.inherited(base)
-      base._attributes = self._attributes.try(:dup) || []
-      base._attributes_keys = self._attributes_keys.try(:dup) || {}
       base._urls = []
       serializer_file = File.open(caller.first[/^[^:]+/])
       base._cache_digest = Digest::MD5.hexdigest(serializer_file.read)
       super
-    end
-
-    def self.attributes(*attrs)
-      attrs = attrs.first if attrs.first.class == Array
-      @_attributes.concat attrs
-      @_attributes.uniq!
-
-      attrs.each do |attr|
-        define_method attr do
-          object && object.read_attribute_for_serialization(attr)
-        end unless method_defined?(attr) || _fragmented.respond_to?(attr)
-      end
-    end
-
-    def self.attribute(attr, options = {})
-      key = options.fetch(:key, attr)
-      @_attributes_keys[attr] = { key: key } if key != attr
-      @_attributes << key unless @_attributes.include?(key)
-      define_method key do
-        object.read_attribute_for_serialization(attr)
-      end unless method_defined?(key) || _fragmented.respond_to?(attr)
     end
 
     def self.fragmented(serializer)
@@ -133,25 +110,6 @@ module ActiveModel
 
     def type
       object.class.model_name.plural
-    end
-
-    def attributes(options = {})
-      attributes =
-        if options[:fields]
-          self.class._attributes & options[:fields]
-        else
-          self.class._attributes.dup
-        end
-
-      attributes += options[:required_fields] if options[:required_fields]
-
-      attributes.each_with_object({}) do |name, hash|
-        unless self.class._fragmented
-          hash[name] = send(name)
-        else
-          hash[name] = self.class._fragmented.public_send(name)
-        end
-      end
     end
 
     def self.serializers_cache
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -1,0 +1,103 @@
+require 'active_support/core_ext/class/attribute'
+require 'set'
+
+module ActiveModel
+  class Serializer
+    # Defines an attribute on the object should be rendered
+    #
+    # The serializer object should implement the attribute name
+    # as a method which should return a value when invoked. If a method
+    # with the attribute name does not exist, the attribute name is
+    # dispatched to the serialized object.
+    #
+    module Attributes
+      extend ActiveSupport::Concern
+
+      included do |base|
+        class_attribute :_attributes, instance_reader: false, instance_writer: false
+        class_attribute :_attributes_keys, instance_reader: false, instance_writer: false
+        base._attributes = Set.new
+        base._attributes_keys = {}
+      end
+
+      # @param [Hash] options
+      # @option options [Array<Symbol>] :fields
+      # @option options [Array<Symbol>] :required_fields
+      # @return [Hash<Symbol => any>] attribute name to value mapping
+      #
+      def attributes(options = {})
+        attributes =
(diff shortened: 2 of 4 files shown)


lib/active_model/serializer/attributes.rb
```diff
@@ -0,0 +1,102 @@
+require 'set'
```
joaomdmoura: :+1:

joaomdmoura:
Cool @bolshakov it seems you are on fire to refactor the `Serializer` class.
@bf4 would you mind to help me out reviewing this too :smile: it's always better to have more ppl validating PRs :smile: I'll try to review it later tonight or tomorrow.

bf4:
@bolshakov look good! Thanks! I have two big-picture questions

1. What are your thoughts about the advantages of moving this code into a module, since the interface hasn't changed?  Would you / did you consider using an Attributes object instead?
2. The documentation looks good, but appear to be YARDdoc style to me.  Is that compatible with however we're generating docs now? I know YARD includes RDoc, but I don't know the state of the current tooling or how we're generating the docs.

lib/active_model/serializer.rb
```diff
@@ -9,12 +9,12 @@ class Serializer
     autoload :Adapter
     autoload :Lint
     autoload :Associations
+    autoload :Attributes
     include Configuration
     include Associations
+    include Attributes
```
bf4: If we're going to include it when the file is evaluated, there's no advantage to the autoload, right? Just curious what your thinking here is.

lib/active_model/serializer.rb
```diff
@@ -26,35 +26,12 @@ class << self
     end
 
     def self.inherited(base)
-      base._attributes = self._attributes.try(:dup) || []
-      base._attributes_keys = self._attributes_keys.try(:dup) || {}
```
bf4: I know it's outside the scope of this PR, but when I'm looking at this code, I'm thinking _attributes and _attributes_keys should never be nil, so why not initialize them when defined?  Just curious if that makes any sense.

e.g.

```
class << self
      attr_accessor :_attributes
      attr_accessor :_attributes_keys
end
self._attributes = []
self._attributes_keys = {}
```

Or since we have active support, we might as well use the [class_attribute](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/core_ext/class/attribute.rb#L6-L75) macro and just

```ruby
require 'active_support/core_ext/class/attribute'
class_attribute :_attributes
Serializer._attributes ||= []
class_attribute :_attributes_keys
Serializer._attributes_keys = {}
```

or maybe just [`cattr_accessor`](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/core_ext/class/attribute_accessors.rb)

lib/active_model/serializer.rb
bolshakov: You are right, `include` after `autoload` is completely pointless

bolshakov:
Hi, @bf4 
1. I think the advantage of moving code into a module is a physically separating and grouping related code. It makes code base more clear, and localize changes.  Yes, I tried to use `Attribute` object instead of `_attributes` and `_attributes_keys`, but I faced with pressure which I can't resolve simply. So I decided to leave it as is for now.

2. Yes this is a yard doc style docs. [Yard](https://github.com/lsegal/yard) is rather popular (982 stars) compared to [Rdoc](https://github.com/rdoc/rdoc) (275 stars). 

You can generate docs locally simply by running `yardoc` command (You need to install `yard` gem first)
or view docs on rubydocs.info  http://www.rubydoc.info/github/rails-api/active_model_serializers/master

lib/active_model/serializer.rb
bolshakov: I don't see any case when it makes sense. I treat `ActiveModel::Serializer` class as an abstract class.

bolshakov:
I'm personally not a heavy user of generated YARD docs. But I've found it's very convenient to see class/method docs right in the code then jumping to source.

lib/active_model/serializer.rb
bf4: I'm addressing

```ruby
self.inherited(base)
  base._attributes = self._attributes.try(:dup) || []
  base._attributes_keys = self._attributes_keys.try(:dup) || {}
end
```

Which reads to me says

- When a Serializer is subclassed it inherits a copy of `_attributes` and `_attributes_keys` from its superclass
  - It may mutate its `_attributes` and `_attributes_keys` without the changes affecting any superclasses
  - If the superclass has not set `_attributes` and `_attributes_keys`, assign them defaults of `[]` and `{}` respectively.

Since the only scenario where a Serializer is being subclassed and the `_attributes` and `_attributes_keys` not set is when `ActiveModel::Serializer` itself is being subclassed, rather than assigning defaults in the inherited hook, why not just set the defaults in the `ActiveModel::Serializer` itself? 

The other stuff I wrote about `class_attribute` and `cattr_accessor` are just about simplifying the code around that.

Does that make sense?

bf4:
@bolshakov Thanks.  I have a bunch of technical points to make and since we haven't met, I'm just prefixing this by saying thanks again and :+1:  and :smile: so hopefully it comes across ok.

re: yard, I wasn't addressing popularity, just whether AMS is parsed by a yard-compliant doc parser right now.  I like the brevity of it myself.

re: the module

>  advantage of moving code into a module is a physically separating and grouping related code. It makes code base more clear, and localize changes. 

So, I feel like using modules this way as the end-point is an anti-pattern, since it increases complexity. Here's how I see it

Pros:

- Like you wrote, attribute related code is now separated out and contained

Cons:

- Methods that were on Serializer are essentially inherited (by including the mixin), so the inheritance chain has been extended and there's another object in the system for every serializer
- The Attributes mixin isn't intended to be reused, so it can't be understood on its own. It will always be tightly coupled to the needs of the Serializer.
- Rather than moving the attribute-related code into a module, we could just comment a part of the serializer as dealing with attributes

Comments:

- I know Rails uses this pattern a ton, of 'composing' classes of modules, but I think that's mostly a result of the way it's evolved.  If Rails were rewritten, I think many of those modules would become collaborator classes.  i.e. I prefer composition over inheritance, and modules are inheritance.

lib/active_model/serializer.rb
bolshakov: Good point, I think it's better not to use `try` if we can make code safe without it.

bolshakov:
In ruby modules not only for inheritance, they are also namespaces and monkey patching mechanism. In this particular case I just want to split long file into smaller files since long class is kinda anti-pattern too.

As I mentioned I tried to extract `Attribute` but without success. 

So what do you think, whats better: Move refactored code back to `serializer.rb` or leave it at `Attributes` module?

lib/active_model/serializer.rb
bolshakov: @bf4 I updated PR and with class_attribute it looks much nicer

joaomdmoura: Indeed @bf4, it's a good point! :+1:

joaomdmoura:
Hey everyone, sorry for taken some days to get here, there are a lot of issues and PRs to catch up.
First of all, awesome discussion, great points.

##### Comments, Docs

Our comments right now don't follow any conventions specifically, so I don't think we would be parsed by a yard-compliant doc parser, but despite of the convention it self, I liked the comments it makes it more clear and easy to jump in into code. (this is important for new contributors)

##### The module

There is no right of wrong here, both options are cool, and the pros x cons that @bf4 pointed out makes sense. I think that the have it organized and contained is a great win, but indeed inheritance is an OOP principle that is really particular, and if you stop to think about there are some specific way to apply it ([inherits nothing](http://blog.codinghorror.com/inherits-nothing/) <- Great post about it).

So it's a tradeoff, keep it outside and organized as a module, make it a class itself `Attribute` (that might prove be really hard), or just let it inside `serializer.rb` itself.

We already made the same decision with `Associations`, that was extracted as a `Module`, but I'm not sure we should start to extract other things, for now I would keep the code refactor but inside `serializer.rb` at least until extract it to an `Attribute` class.

How do you feel about it? :smile: 
Sorry for the long reply.

bolshakov:
I understand the motivation not to use inheritance for such kind of things. So let me play a bit more with `Attribute` object. I think it should be useful abstraction.

beauby:
Heads up on this one?

bolshakov:
@beauby  hey, currently I have no enough time for this refactoring, but I'm going to make it later

bf4:
@bolshakov Seems this PR got lost in the shuffle, but still has good stuff in it.  Are you interested in rebuilding it off of master?

bolshakov:
@bf4 I could not manage to finish this refactoring in next two months. So I think it's better to close this PR now and reopen it later.

bf4:
@bolshakov Hope you come back soon!

PR #1034, opened 2015-08-04
allow id attribute to be overriden

lanej:
sometimes serialized id and the model id are not the same.


joaomdmoura:
Hey @lanej, I'm finally reviewing it :smile: 
I also see that you did the same to `type`.
This is something that I was looking forward to have! Would be nice to hear more about what was the scenario that made you do it :smile: 
Great work, I'm merging this and https://github.com/rails-api/active_model_serializers/pull/1037

lanej:
Basic scenario is that I have an old model that has an external id (api presented) and an internal id (database index).  I'm currently using an older version of ams to rewrite the "id" portion of the presentation to expose the external id only.  Using `respond_to?` vs `method_missing` is really the only significant change here.

I have a similar story for #1037

Issue #1040, opened 2015-08-05
Pagination Links on AMS


bacarini:
Hey guys,

I am working on an API and I need to return pagination links on my responses. I am wordering the best way to to that. [JSON-API](http://jsonapi.org/format/#fetching-pagination) specification does not let it clear on how it should display the links top-level.

what is your thoughts about the following example?

```
{
    "data": [..],
    "links": {
        "pages": {
            "first": "?page=1&per_page=1",
            "prev": "?page=2&per_page=1",
            "next": "?page=4&per_page=1",
            "last": "?page=10&per_page=1"
        }
    }
}
```

Anyway, I've opened a [PR](https://github.com/rails-api/active_model_serializers/pull/1041) and I would like to discuss it with you all.

bacarini:
@joaomdmoura, 

I think we can close this issue, since #1041 was merged, no?

joaomdmoura:
Indeed, closed! :smile:

PR #1018, opened 2015-07-25
Adds support for top-level links objects

leandrocp:
It's a generic object which can be used to declare a top-level links object as specified on jsonapi: http://jsonapi.org/format/#document-top-level

```ruby
render json: @post, links: { self: "/posts/1" }
```

PR #1004, opened 2015-07-14
[RFC] JSON API Errors (Initial implementation and roadmap for full feature-set)

bf4:
## Usage

Requires explicit passing of Error serializer

```ruby
# POST /api/{plural_resource_name}
def create  
  set_resource(resource_class.new(resource_params))

  if get_resource.save
    render :show, status: :created, adapter: :json_api
  else
    render json: get_resource.errors, status: :unprocessable_entity, adapter: :json_api, serializer: ActiveModel::Serializer::ErrorSerializer
  end
end
```

PR TODOS:

- DOCS
  - [x] Usage and requirements `jsonapi/errors.md`
 - [x] data and errors cannot co-exist. http://jsonapi.org/format/#document-top-level  [ref](https://github.com/bf4/active_model_serializers/commit/79cd4e2bde0ce8e346fc8b54a07c146d29961143#commitcomment-16526769)
   - [x] requests are transactional, so any part that succeeds needs to be rolled back if any part fails. http://jsonapi.org/format/#crud [ref](https://github.com/bf4/active_model_serializers/commit/79cd4e2bde0ce8e346fc8b54a07c146d29961143#commitcomment-16526769)
  - [x] examples http://jsonapi.org/examples/#error-objects [ref](https://github.com/bf4/active_model_serializers/commit/79cd4e2bde0ce8e346fc8b54a07c146d29961143#commitcomment-16526777)
- [ ] changelog

## Error object:

Serializing errors is different from resources.

- Errors may be an instance of `ActiveModel::Error`,  which we'd want to call `model.errors.messages` on.
  -  Resolution: *true*
- Errors may be given as a hash.
  -  Resolution: *undecided*
- Errors may be given by the end-user as a special 'error' object.
  -  Resolution *undecided*
- `ActiveModel::Error` is necessarily a collection of error objects, each with it's own `'source'` and `'detail'`
  - Resolution: that's how we'll use it.
- [ ] Errors may be automatically generated from the controller.
  - Resolution: *deferred*
  - [ ] [422 for errors](https://github.com/rails-api/active_model_serializers/pull/1004#issuecomment-181708926).. 
  - [ ] but see https://github.com/rails-api/active_model_serializers/pull/1004#discussion_r46471845 for concerns and alternative implementations

## [Full Error API Implementation](http://jsonapi.org/format/#errors):

- [x] error root key to collection of error objects
  - [x] The members data and errors MUST NOT coexist in the same document.
- [x] error objects from ActiveModel::Error
[...]

PR #965, opened 2015-06-24
serializable_hash and as_json should take options = nil

bf4:
per ActiveModel::Serialization#serializable_hash 
https://github.com/rails/rails/blob/96bb004fc6e67cdf1b873f11ad5f8efd06949797/activemodel/lib/active_model/serialization.rb
```ruby
   def serializable_hash(options = nil)
       options ||= {}
```

Otherwise, passing in nil to `as_json` or `serializable_hash` makes things
blow up when passing nil into attributes

Extracted from https://github.com/rails-api/active_model_serializers/pull/954

PR #1041, opened 2015-08-05
Pagination Links

bacarini:
I would like to discuss on how we should display pagination links on AMS.

I think you can relie on `kaminari`and `will_paginate`gems to do so. Having said that, I created a class to build links.

[related issue](https://github.com/rails-api/active_model_serializers/issues/1040)


README.md
```diff
@@ -273,6 +273,18 @@ And you can change the JSON key that the serializer should use for a particular
 The `url` declaration describes which named routes to use while generating URLs
 for your JSON. Not every adapter will require URLs.
 
+## Pagination
+
+If you want pagination links in your response, specify it in the `render`
+
+```ruby
+  render json: @posts, pagination: true
```
bf4: Thanks for this.  Did you try subclassing the `ArraySerializer`? e.g. https://github.com/rails-api/active_model_serializers#2-for-an-array-resource and just setting the meta?  Have you seen pagination gems such as https://github.com/davidcelis/api-pagination ?

bacarini:
Hi @bf4, 

Actually, I got same ideas from `api-pagination` gem. I used to use it, but It does not apply to JSON-API specification. That gem build links on headers, like below.

screen shot 2015-08-06 at 11 15 42 am

I want to insert pagination into the response body. At least, I understood that from [JSON-API](http://jsonapi.org/format/#fetching-pagination). Is it right?

bf4:
Hmm.. well, it should subclass the ArrayAdapter for sure, but besides testing that we can set meta data in a collection, I'm not sure it's the job of this library to expose anything more than the meta option for any pagination libraries to use.

bf4:
My bad, it should be in a links object, not meta http://jsonapi.org/format/#fetching-pagination

> Pagination

> A server MAY choose to limit the number of resources returned in a response to a subset ("page") of the whole set available.

> A server MAY provide links to traverse a paginated data set ("pagination links").

> Pagination links MUST appear in the links object that corresponds to a collection. To paginate the primary data, supply pagination links in the top-level links object. To paginate an included collection returned in a compound document, supply pagination links in the corresponding links object.

> The following keys MUST be used for pagination links:

> - first: the first page of data
> - last: the last page of data
> - prev: the previous page of data
> - next: the next page of data

> The `page` **query parameter** is reserved for pagination. Servers and clients SHOULD use this key for pagination operations.
>  For example, a page-based strategy might use query parameters such as page[number] and page[size], an offset-based strategy might use page[offset] and page[limit], while a cursor-based strategy might use page[cursor].

```js
"links": {
  "self": "http://example.com/posts",
  "first": "http://example.com/posts?page[number]=1",
}
```

bf4:
I'm happy to pair with you to work on the links object. (We'll need tests, too, and might want to start with a PR to the json-api/example page to make sure we're getting it right)

bacarini:
great... So, I will work on it. Should we edit this page example or create another one?

bf4:
add more examples to it.. see e.g. https://github.com/json-api/json-api/pull/828 (error objects) which is a result of me working on error objects here in AMS #1004

lib/active_model/serializer/adapter.rb
```diff
@@ -21,9 +21,12 @@ def serializable_hash(options = nil)
...
       def as_json(options = nil)
-        hash = serializable_hash(options)
-        include_meta(hash) unless self.class == FlattenJson
-        hash
+        serializable_hash(options).tap do |hash|
+          return hash if self.class == FlattenJson
+
+          include_meta(hash)
+          include_pagination_links(hash) if options && options[:pagination]
+        end
```
bf4: I think these changes should be in the json_api adapter, no/

lib/active_model/serializer/array_serializer.rb
```diff
@@ -24,6 +25,7 @@ def initialize(objects, options = {})
         end
         @meta     = options[:meta]
         @meta_key = options[:meta_key]
+        @pagination = ActiveModel::Serializer::Pagination.new(objects) if options[:pagination]
```
bf4: nitpick: `PaginationLinks` is probably a better name. 

Also, I think this code should also be in the [json_api adapter](https://github.com/rails-api/active_model_serializers/blob/03ddeeeec1e4eedafeb68eeccd59e8c3826b859d/lib/active_model/serializer/adapter/json_api.rb#L20-L30)

bf4: (in a subfolder, like the fragment cache)

lib/active_model/serializer/pagination.rb
```diff
@@ -0,0 +1,62 @@
...
+        return :will_paginate if defined?(WillPaginate::CollectionMethods)
+        raise "AMS relies on either Kaminari or WillPaginate." +
+          "Please install either dependency by adding one of those to your Gemfile"
+      end
+
+      def setup_will_paginate
+        WillPaginate::CollectionMethods.module_eval do
+          def first_page?() !previous_page end
+          def last_page?() !next_page end
+        end
```
bf4: monkey patching an external library seems like asking for trouble here.  I might do it in my own app, but I'm uneasy doing it in a public gem.  Is there a way around doing this?

bf4: Maybe wrap the collection object polymorphically? e.g. for will_paginate, the collection wrapper would add those methods?

lib/active_model_serializers.rb
```diff
@@ -2,6 +2,7 @@
 require 'active_model/serializer/version'
 require 'active_model/serializer'
 require 'active_model/serializer/fieldset'
+require 'active_model/serializer/pagination'
 require 'active_model/serializable_resource'
```
bf4: assuming we move the code into the json api adapter, this require should also move

bf4:
Don't forget to add tests! They'll be required before this gets merged, so better now than later, right?

bacarini:
Hey @bf4, I've changed a few things according to your comments. I will add more test later on.

Meanwhile, it would be great some reviews on the new code.

tks

README.md
joaomdmoura: Great! @bf4 about the meta thing, it would work, but it would not follow [json-api conventions](http://jsonapi.org/format/#fetching-pagination).
I think the README might not be the best place for describing this feature, mostly because it's to a particular adapter (json-api) and also because would be awesome to have some examples. 
We are trying to consolidate our [new docs](https://github.com/rails-api/active_model_serializers/tree/master/docs), would be great if instead of adding this to the README you could open a new article there with this PR describing the feature, some examples and some of the gems that ppl can use in order to make it work.

lib/active_model/serializer/adapter.rb
```diff
@@ -21,9 +21,11 @@ def serializable_hash(options = nil)
       end
 
       def as_json(options = nil)
-        hash = serializable_hash(options)
-        include_meta(hash) unless self.class == FlattenJson
-        hash
+        serializable_hash(options).tap do |hash|
```
joaomdmoura: :+1:

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,4 +1,5 @@
 require 'active_model/serializer/adapter/json_api/fragment_cache'
+require 'active_model/serializer/adapter/json_api/pagination_links'
```

lib/active_model/serializer/adapter/json_api/pagination_links.rb
```diff
@@ -0,0 +1,52 @@
...
+
+              unless collection.current_page == collection.total_pages
+                pages[:next] = collection.current_page + FIRST_PAGE
+                pages[:last] = collection.total_pages
+              end
+            end
+          end
+
+          def raise_unless_any_gem_installed
+            return if defined?(WillPaginate) || defined?(Kaminari)
```
joaomdmoura: I'm in favor of using gems that the community is already used to,but would be nice to hear more thoughts about it. @bf4 @spastorino

joaomdmoura:
Great work @bacarini. Btw awesome PR on json-api as well @bf4 ! 

 > Don't forget to add tests! They'll be required before this gets merged, so better now than later, right?  

:+1:  Indeed I see you added a test, would be great to have one covering the `action_controller` flow. You can check how we have being doing it [here](https://github.com/rails-api/active_model_serializers/tree/master/test/action_controller)

bacarini:
@joaomdmoura @bf4, I added test to pagination feature and also included it to `/doc`.

I've exchanged `test/action_controller/json_api_linked_test.rb` to a new different directory, `test/action_controller/json_api/linked_test.rb`. In my opinion, it is clearer and more organized. What do you think guys?

Enriikke:
This is awesome! Thanks for the work on this @bacarini :boom: :+1: :smile:

I would like to make a suggestion. Sometimes pagination is tied to a search or filtered query. Looking at the docs, the returned pagination links look like `?page=1&per_page=1` so any other parameter that was sent with the request is not returned. 

This might be up to individual interpretation but the way the pagination links are returned will require the client to take additional steps before it can use them (i.e. prefixing the api endpoint, adding filters or search query). 

I'm suggesting that pagination links should be returned in a way that no additional processing is needed by the client. I'm working on a project using JSON API that needed pagination links and this is similar to how I approached it:

```ruby
url = request.original_url.sub(/\?.*$/, "")
params = request.query_parameters.merge(page: { number: page, size: size }).to_param

"#{url}?#{params}"
```

In my case, I'm using a `page` hash as suggested in this note http://jsonapi.org/format/#fetching-pagination:

> Note: JSON API is agnostic about the pagination strategy used by a server. Effective pagination
> strategies include (but are not limited to): page-based, offset-based, and cursor-based. The page query
> parameter can be used as a basis for any of these strategies. For example, a page-based strategy
> might use query parameters such as page[number] and page[size], an offset-based strategy might use
> page[offset] and page[limit], while a cursor-based strategy might use page[cursor].

In any case, this is great. I like that this gives the ability to add a root `links` object without overriding `meta`. I will be happy to assist in any way I can :hammer:

bacarini:
According to [a comment on json-api PR](https://github.com/json-api/json-api/pull/834#discussion_r36585777), I added complete URIs, not just query strings. 

It will accept `link/self` like [this PR](https://github.com/rails-api/active_model_serializers/pull/1018) or teh request original url.

@Enriikke, I can see your point, but what do you thing to discuss it on [json-api PR](https://github.com/json-api/json-api/pull/834)?

bacarini:
@Enriikke may bad, I misunderstood your previous comment.

I added new code to merge with urls parameters that already exists.

Enriikke:
Awesome @bacarini I'm looking forward to this PR being merged :+1: :smiley:

lib/action_controller/serialization.rb
```diff
@@ -57,5 +61,13 @@ def serialization_scope(scope)
         self._serialization_scope = scope
       end
     end
+
+    def original_url
+      request.original_url.sub(/\?.*$/, "")
```
bf4: So, this just strips the query params from `request.original_url`?

lib/active_model/serializer/array_serializer.rb
```diff
@@ -5,7 +5,7 @@ class ArraySerializer
       include Enumerable
       delegate :each, to: :@objects
 
-      attr_reader :root, :meta, :meta_key
+      attr_reader :root, :meta, :meta_key, :options, :resource
```
bf4: I see you're now passing options into the array serializer, but also the resource? Is that so that the adapter can get the resource from it? I think we could get around this if we had pagination as one of the adapter options that gets passed through the serializable resource to the adapter, no?  That way the array serializer doesn't need to 'know' that the json adapter wants to add pagination links.

bf4:
@bacarini I'm adding some comments, and want to reiterate that this is awesome work you're doing and a great contribution!

lib/active_model/serializer/array_serializer.rb
bacarini: I just added options and resource to attr_reader, it was already in the `array_serializer`. I needed to do that in order to get info from paigination methos, such as `total_pages` and `current_page`. 

I am not sure, but I think the adapter does not need to know that because of always its reference to one resource and not a collection. Is that makes sense?

but, @bf4 could you give me an example?

lib/action_controller/serialization.rb
bacarini: I am getting only base uri to use it on `pagination link`class.

bacarini:
Thanks @bf4, @Enriikke and @joaomdmoura. hope we make it through.

lib/active_model/serializer/array_serializer.rb
bf4: re: adding `options` and `resource` as public methods on the ArraySerializer so that the json-api adapter may use them

So, in theory, the serializer and array serializer just represent transforming the resource into a serializable hash (of attributes), and the adapter presents that hash according to the type of adapter it is, e.g. json-api adapter.  That's how I'm thinking of it, at least.  If so, then per https://github.com/rails-api/active_model_serializers/blob/03ddeeeec1e4eedafeb68eeccd59e8c3826b859d/lib/active_model/serializable_resource.rb#L5 any options passed to the `SerializableResource` that pertain to the presentation of the resource should go to the adapter, which I think is a good match for `pagination`. That said, `meta` goes to the serializer now, though it is used by the adapter... so, I guess we'll see what makes more sense.  cc @joaomdmoura @kurko

bacarini: I totally agree when we are talking about `pagination` key, but there are also `original_url`and `query_parameters`. Do you think it should go in adapters hash as well?(i.e https://github.com/rails-api/active_model_serializers/blob/03ddeeeec1e4eedafeb68eeccd59e8c3826b859d/lib/active_model/serializable_resource.rb#L5), 

Thus there is no way to `JsonApi` adapter  know about the resources (collection). We can use method `each` which means exactly the same due to this delegate https://github.com/rails-api/active_model_serializers/blob/03ddeeeec1e4eedafeb68eeccd59e8c3826b859d/lib/active_model/serializer/array_serializer.rb#L6. However, I think it quite weird something like it 
```ruby
JsonApi::PaginationLinks.new(serializer.each, options)
```

bf4: Or pseudo-code


```ruby
class JsonApi < Adapter
      add_links(serializer)
  def add_links(serializer)
    links = options.fetch(:links) { {} }
    resource = serializer.instance_variable_get(:@resource)
    add_pagination_links(links, resource) if options[:pagination]

  def add_pagination_links(links, resource)
    links.update(PaginationLinks.new(resource).serializable_hash)
```

bacarini:
@bf4 I removed `resource` and `options` from array_serializer as you said. I do think it is better.
tks.

@joaomdmoura @bf4, any concern about this PR? :smiley:

bacarini:
@bf4 @joaomdmoura [json-api PR](https://github.com/json-api/json-api/pull/834) was merged.

lib/active_model/serializer/adapter.rb
bf4: @joaomdmoura  OT: Might it make more sense for the FlattenJson adapter to implement a no-op `include_meta` so the superclass doesn't need to know about the subclass?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -16,10 +17,9 @@ def initialize(serializer, options = {})
         end
 
         def serializable_hash(options = nil)
-          options = {}
```
bf4: Am curious if you moved the definition of options here to the two places below for style reasons or if something was broken.

lib/active_model/serializer/adapter/json_api/pagination_links.rb
```diff
@@ -0,0 +1,56 @@
...
+            raise "AMS relies on either Kaminari or WillPaginate." +
+              "Please install either dependency by adding one of those to your Gemfile"
```
bf4: nitpick: AMS relies on ... **for pagination**

lib/active_model/serializer/adapter/json_api.rb
bacarini: I needed to use `options` on https://github.com/bacarini/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json_api.rb#L31

lib/active_model/serializer/adapter.rb
bacarini: @bf4 I couldn't agree more with you about this approach. I will be happy to work on it, but in a different PR.

lib/active_model/serializer/adapter/json_api.rb
bf4: The implementation there was a bug, should have been `options ||= {}`. Doesn't make much sense otherwise. That's on me https://github.com/rails-api/active_model_serializers/pull/965/files#diff-ae1e38e96a8bf28ae37918a65c448c73R18

bf4: Since it's critical to you, I'd think @joaomdmoura would support including that bugfix in this feature PR

bacarini: I am quite confused now...hahaha
is the first not the same than the second example below?
```ruby
   options ||= {}
```
```ruby
```

which bug was that?

lib/active_model/serializer/adapter.rb
bf4: Did it in #1048 because

lib/active_model/serializer/adapter/json_api.rb
bf4: Fixed in https://github.com/rails-api/active_model_serializers/pull/1049

bacarini: :+1:

lib/action_controller/serialization.rb
bf4: Or `request.original_url[/\A[^?]+/]` which has the advantage of using an anchor :)

bacarini: @bf4 I don't what I am doing wrong but it seems the exactly the same.

```
(byebug) "http://localhost:3000/author?teste=teste#L21"[/\A[^?]+/]
"http://localhost:3000/author"
(byebug) "http://localhost:3000/author?teste=teste#L21".sub(/\?.*$/, "")
"http://localhost:3000/author"
```

Was it your idea?

bf4: One sec while I whip up a benchmark...

bf4:
```plain
ruby regex.rb
      unanchored sub    59.229k i/100ms
      anchored regex    62.136k i/100ms
-------------------------------------------------
      unanchored sub      1.006M (± 4.4%) i/s -      5.034M
      anchored regex      1.139M (± 5.0%) i/s -      5.717M

Comparison:
      anchored regex:  1138652.9 i/s
      unanchored sub:  1006027.0 i/s - 1.13x slower
```

from

```ruby
      unanchored sub    59.229k i/100ms
      anchored regex    62.136k i/100ms
-------------------------------------------------
      unanchored sub      1.006M (± 4.4%) i/s -      5.034M
      anchored regex      1.139M (± 5.0%) i/s -      5.717M

Comparison:
      anchored regex:  1138652.9 i/s
      unanchored sub:  1006027.0 i/s - 1.13x slower

ruby-2.2.2 ~/personal/benchmarks
📝 $ cat regex.rb
gem "benchmark-ips"
require "benchmark/ips"

URL = "http://localhost:3000/author?teste=teste#L21"
Benchmark.ips do |x|

  x.report("unanchored sub") do |times|
    i = 0
    while i < times
      URL.sub(/\?.*$/, "")
      i += 1
    end
  end

  x.report("anchored regex") do |times|
    i = 0
    while i < times
      URL[/\A[^?]+/]
      i += 1
    end
  end

  x.compare!
end
```

bacarini: great!! :+1: 

changed!

bacarini:
@joaomdmoura I did a rebase. All good to go? :smile:

lib/action_controller/serialization.rb
```diff
@@ -25,6 +25,10 @@ def get_serializer(resource, options = {})
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource#serialize"
         options[:adapter] = false
       end
+      if options[:pagination]
+        options[:original_url] = original_url
+        options[:query_parameters] = query_parameters
+      end
```
bf4: Just thinking about this part of the pagination interface and how someone might want to generate this outside of a controller

 might we want to instead set these as:

```ruby
options[:pagination][:original_url] ||= original_url
options[:pagination][:query_parameters] ||= query_parameters
```

that way options[:pagination] is still truthy and its options are scoped within it...  and is also passed to the adapter as such

On the other hand, since original url is really a fallback for `options[:links][:self]`, maybe this should be
`options[:links][:self] ||= original_url` (which, now that I think about it, we probably only want to strip the params off the original url for building the non-self links, right?)

Another alternative is to pass `render json: @object, links: { pagination: true }` and we just check if `options[:links].delete(:pagination)` is true. That way, there's one less option being passed around, and pagination piggy backs on links, which it has to do anyways.

I rearranged your code as below while thinking this through.  Hope it's helpful.

```ruby
FIRST_PAGE = 1

def options
  @options ||=
    begin
      {
        pagination: {
          original_url: request.original_url,
          query_parameters: request.query_parameters
        }
      }
    end
end

def links
  @links ||= options[:links] || {}
end

def build_pagination_links(collection)
  pagination_links = {}
  return pagination_links if collection.total_pages == FIRST_PAGE


  original_url = options[:pagination][:original_url]
  url_base = original_url[/\A[^?]+/]
  query_parameters = options[:query_parameters] || original_url[/\?.*/] || {}

  current_links = links
  self_link = current_links.fetch(:self) { original_url }

  collection_size            = collection.size
  collection_current_page    = collection.current_page
  collection_total_pages     = collection.total_pages

  link_template = "#{url_base}?%s"
  if collection_current_page != FIRST_PAGE
    pagination_links[:first] = link_template % query_parameters.merge(size: collection_size, number: FIRST_PAGE).to_query
    pagination_links[:prev]  = link_template % query_parameters.merge(size: collection_size, number: (collection_current_page - FIRST_PAGE)).to_query
  end

  if collection_current_page != collection_total_pages
    pagination_links[:next] = link_template % query_parameters.merge(size: collection_size, number: (collection_current_page + FIRST_PAGE)).to_query
    pagination_links[:last] = link_template % query_parameters.merge(size: collection_size, number: collection_total_pages).to_query
  end
end

links.update(build_pagination_links(paginated_resource))
```

bf4:
@kurko @joaomdmoura With this PR, we should change the custom collection serializer (PaginatedSerializer) example in the README, no?  

https://github.com/rails-api/active_model_serializers/blob/8568ed5913edf2c641e33b841dfa28ebbc97b3a4/README.md#2-for-an-array-resource

#### 2. For an array resource:

```ruby
# Or, you can explicitly provide the collection serializer as well
render json: @posts, serializer: PaginatedSerializer, each_serializer: PostPreviewSerializer
```

CHANGELOG.md
```diff
@@ -9,4 +9,5 @@
   * adds cache support to attributes and associations [@joaomdmoura]
   * uses model name to determine the type [@lsylvester]
   * remove root key option and split JSON adapter [@joaomdmoura]
-  * adds FlattenJSON as default adapter [@joaomdmoura]
\ No newline at end of file
+  * adds FlattenJSON as default adapter [@joaomdmoura]
+  * adds support for `pagination links` at top level of JsonApi adapter [@bacarini]
```
joaomdmoura: :+1:

docs/README.md
```diff
@@ -12,6 +12,7 @@ This is the documentation of AMS, it's focused on the **0.10.x version.**
 ## How to
 
 - [How to add root key](howto/add_root_key.md)
+- [How to add pagination links](howto/add_pagination_links.md) (```JSON-API``` only)
```
joaomdmoura: :+1:  Awesome! thank you so much for that!

lib/action_controller/serialization.rb
```diff
@@ -57,5 +61,13 @@ def serialization_scope(scope)
...
       end
     end
+
+    def original_url
+      request.original_url[/\A[^?]+/]
+    end
+
+    def query_parameters
+      request.query_parameters
+    end
```
joaomdmoura: :+1:

lib/action_controller/serialization.rb
joaomdmoura: > ```ruby
options[:pagination][:original_url] ||= original_url
options[:pagination][:query_parameters] ||= query_parameters
```

:+1: 

I'm also in favor or building the self link based on this.
It seems to me that some of the changes proposed here are implemented already, right @bacarini?

File: lib/active_model/serializer/adapter.rb (hunk 6 above)
Comment (joaomdmoura) [inline lib/active_model/serializer/adapter.rb]: #1048 Merged already.

**Comment by joaomdmoura** [comment]
#### Next

All right everyone! Great work here @bacarini!
I'm looking forward to merge it! Just made a comment on top of the @bf4 one [here](https://github.com/rails-api/active_model_serializers/pull/1041#discussion_r36998222)

#### Conflitct

Btw, I'm checking #1018, and this [thread](https://github.com/rails-api/active_model_serializers/pull/1018#discussion_r36938579). I'm still trying to make my mind about it, but in my opinion this PR will kill it, at least the way it is right now.

I agree with @bf4 when we says:

> Well, links is a top-level member of the json-api response object. Pagination is a usage of the links object...

[reference 1](http://jsonapi.org/format/#document-links)
[reference 2](http://jsonapi.org/format/#fetching-pagination)

#### Solution

IMO you are adding the links support and the feature to use it for pagination. @leandrocp might need to rebase #1018 once this one gets merged and add a feature that enable a user to append/override item to the links object.

How do you all feel about it?

#### Details

> With this PR, we should change the custom collection serializer (PaginatedSerializer) example in the > README, no?

@bf4, why exactly? I think I'm not seeing something :grin: despite the fact that I would like to mention the pagination feature on the README as well (@bacarini, haha sorry about that but could you also add a *small* section on README mentioning it? Maybe even link it to the complete doc itself)

**Comment by bacarini** [comment]
First of all, thank you all.

#### Conflict 
Well, @joaomdmoura I don't think this PR will kill #1018 at all. As you can see [here](https://github.com/rails-api/active_model_serializers/pull/1041/files#diff-9eea35f755e2e04ea579c3aebdec641aR54), it gets first the value from `self links`, if it already exists, or the `original_url` in case of not. So, I believe @leandrocp's PR is important to support custom links. I have myself an example to that.
Ex.
Here we have couple of micros-apis but our customer access they through only one endpoint, so this feature would be crucial to return an unique link. Does it make sense?

#### README
I agree to @bf4 on changing the custom collection serializer (`PaginatedSerializer`) example in the README. What do think changing to `CollectionSerializer`?
Ex.
```ruby
# Or, you can explicitly provide the collection serializer as well
render json: @posts, serializer: CollectionSerializer, each_serializer: PostPreviewSerializer
```
Besides that I will add `PaginationLinks` to README no problem, but I want to confirm something before.

@bf4 suggested to use like that `links: { pagination: true }` instead of just `pagination: true`. Which one iit will be? I am still thinking that just `pagination: true` is clearer and better, but it's up to you guys.

docs/howto/add_pagination_links.md
```diff
@@ -0,0 +1,40 @@
+# How to add pagination links
+
+If you want pagination links in your response, specify it in the `render`
+
+```ruby
+  render json: @posts, pagination: true
+```
+
+AMS relies on either `Kaminari` or `WillPaginate`. Please install either dependency by adding one of those to your Gemfile.
```
joaomdmoura: Would be nice to have the links to the projects :smile:

bacarini: :+1:

joaomdmoura:
Cool! Well, for now let's focus on get it merged, then @leandrocp can rebase it and move from there :smile: 

Now I get it, yup, let's change the README example.

hm... Actually, I'd like to propose a change :grin: , instead of using `pagination`, as long as we are depending on other gems architecture, I would go for automatically presume it.

So probably create a method `is_paginated?` and uses it to presume if it should or not have the links and automatically add it follow the conventions. Once the developers already specified that he is paginating the resource it seems to me there is no need to add an option on the `render` and we keep the *Rails magic feelings* :tm:, I know it's different from the actual approach, so I want to hear your thoughts about it.

bacarini:
Although json-api says that pagination links is optional, It is a good ideia.
I will change it. ok?

joaomdmoura:
Yup, it is optional because you might not paginated the resource, but if you do I think the better way is to use the recommended pagination standard :grin: Cool!

bacarini:
@joaomdmoura, @bf4 I've changed and now pagination links will be included in responses automatically as long as the resource is paginated using `Kaminari` or `WillPaginate`. Could you please check it out?

docs/howto/add_pagination_links.md
```diff
@@ -0,0 +1,46 @@
...
+
+```ruby
+#kaminari example
+@posts = Kaminari.paginate_array(Post.all).page(3).per(1)
+render json: @posts
+
+#will_paginate example
+@posts = Post.all.paginate(page: 3, per_page: 1)
+render json: @posts
+```
```
joaomdmoura: :+1:

README.md
```diff
@@ -272,6 +272,11 @@ And you can change the JSON key that the serializer should use for a particular
 
 The `url` declaration describes which named routes to use while generating URLs
 for your JSON. Not every adapter will require URLs.
+## Pagination
+
+Pagination links will be included in your response automatically as long as the resource is paginated using [Kaminari](https://github.com/amatsuda/kaminari) or [WillPaginate](https://github.com/mislav/will_paginate) and if you are using a ```JSON-API``` adapter. The others adapters does not have this feature.
+
+For more information about it, please see in our docs [How to add pagination links](https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/add_pagination_links.md)
```
joaomdmoura: :+1:

test/action_controller/json_api/pagination_test.rb
```diff
@@ -0,0 +1,107 @@
...
+        def test_render_only_prev_and_first_pagination_links
+          expected_links = { "self"=>"#{KAMINARI_URI}?page%5Bnumber%5D=3&page%5Bsize%5D=1",
+            "first"=>"#{KAMINARI_URI}?page%5Bnumber%5D=1&page%5Bsize%5D=1",
+            "prev"=>"#{KAMINARI_URI}?page%5Bnumber%5D=2&page%5Bsize%5D=1"}
+          get :render_pagination_using_kaminari, page: { number: 3, size: 1 }
+          response = JSON.parse(@response.body)
+          assert_equal expected_links, response['links']
+        end
+
+        def test_render_only_last_and_next_pagination_links_with_additional_params
```
joaomdmoura: Do we need to test the `prev` and `first`?

joaomdmoura:
LGTM and ready :+1: just made a one last comment.

test/action_controller/json_api/pagination_test.rb
bacarini: We have different scenarios and I am testing all.

```json
#### Page: 3
"links": {
  "self": "http://example.com/articles?page[number]=3&page[size]=1",
  "first": "http://example.com/articles?page[number]=1&page[size]=1",
  "prev": "http://example.com/articles?page[number]=2&page[size]=1",
  "next": "http://example.com/articles?page[number]=4&page[size]=1",
  "last": "http://example.com/articles?page[number]=13&page[size]=1"
}

#### Page: last
"links": {
  "self": "http://example.com/articles?page[number]=1&page[size]=1",
  "next": "http://example.com/articles?page[number]=4&page[size]=1",
  "last": "http://example.com/articles?page[number]=13&page[size]=1"
}

#### Page: first
"links": {
  "self": "http://example.com/articles?page[number]=1&page[size]=1",
  "first": "http://example.com/articles?page[number]=1&page[size]=1",
  "prev": "http://example.com/articles?page[number]=2&page[size]=1"
}
```

joaomdmoura: `with_additional_params` :grin:

bacarini: :+1:

docs/howto/add_pagination_links.md
```diff
@@ -0,0 +1,46 @@
+# How to add pagination links
+
+Pagination links will be included in your response automatically as long as the resource is paginated and if you are using a ```JSON-API``` adapter. The others adapters does not have this feature.
+
```
bf4: But you could use the PaginationLinks builder with the JSON adapter right? like if I wanted to put links in the 'meta' attribute using a subclass of the json adapter?

bacarini: No, PaginationLinks only will be included to `JSON-API` adapter, but you can still use a custom seriallizer to do so.

Ex.
```ruby
render json: @posts, serializer: PaginatedSerializer, each_serializer: PostPreviewSerializer
```

```ruby
class PaginatedSerializer < ActiveModel::Serializer::ArraySerializer
  def initialize(object, options={})
    meta_key = options[:meta_key] || :meta
    options[meta_key] ||= {}
    options[meta_key] = {
      current_page: object.current_page,
      next_page: object.next_page,
      prev_page: object.prev_page,
      total_pages: object.total_pages,
      total_count: object.total_count
    }
    super(object, options)
  end
end
```

lib/action_controller/serialization.rb
```diff
@@ -25,6 +25,11 @@ def get_serializer(resource, options = {})
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource#serialize"
         options[:adapter] = false
       end
+      if resource.respond_to?(:current_page) && resource.respond_to?(:total_pages)
+        options[:pagination] = {}
+        options[:pagination][:original_url] = original_url
+        options[:pagination][:query_parameters] = query_parameters
+      end
```
bf4: could we maybe instead pass in a `context` so that serialization isn't as tightly coupled to the controller?

e.g. in the controller, just

```ruby
options.fetch(:context) { options[:context] = request }
```
Then in the ArraySerializer

```ruby
class ArraySerializer
   def object # Same interface as Serializer
     @resource
   end
end
```

Then in the JsonApi adapter

```ruby
          options ||= {}
            serializer.each do |s|
               # code
            end
            add_links(options)
          end

        private

        def add_links(options)
          links = @hash[:links] ||= {} # no reason for them to be nil or false
          @hash[:links].update(JsonApi::PaginationLinks.new(object, links).serializable_hash(options))
        end
end
```

Then in the [PaginationLinks builder](https://github.com/rails-api/active_model_serializers/pull/1041#discussion_r36998222)

```ruby
def initialize(collection, links)
  @collection = collection
end
def serializable_hash(options)
  pagination_links = {}
  return pagination_links if !collection.respond_to?(:current_page) && !collection.respond_to?(:total_pages)
  return pagination_links if collection.total_pages == FIRST_PAGE

  context = options.fetch(:context)
  original_url = context.original_url
  url_base = original_url[/\A[^?]+/]
  query_parameters = context.query_parameters.dup || original_url[/\?.*/] || {}

  current_links = links
  self_link = current_links.fetch(:self) { original_url }

  collection_size            = collection.size
  collection_current_page    = collection.current_page
  collection_total_pages     = collection.total_pages


  params = query_parameters.merge!(page: {})
  link_template = "#{url_base}?%s"
  # only add first/previous page links when not on the first page (i.e. we can go back)
  if collection_current_page != FIRST_PAGE
    pagination_links[:first] = link_template % params[:page].merge(size: collection_size, number: FIRST_PAGE).to_query
    pagination_links[:prev]  = link_template % params[:page].merge(size: collection_size, number: (collection_current_page - FIRST_PAGE)).to_query
  end

  # only add next/last page links when not on the last page (i.e. we can go forward)
  if collection_current_page != collection_total_pages
    pagination_links[:next] = link_template % params[:page].merge(size: collection_size, number: (collection_current_page + FIRST_PAGE)).to_query
    pagination_links[:last] = link_template % params[:page].merge(size: collection_size, number: collection_total_pages).to_query
  end
  pagination_links
end
```

bacarini: Hey @bf4 , Is the idea to send the whole request through objects?
We are using just `original_url` in case of it is not already defined in `options[:links][:self]` and `query_parameters `, I really don't think it is good idea, but may have an issue that I am seeing that you are. 

It might be an option if [this feature](https://github.com/rails-api/active_model_serializers/issues/1062) will be approved.

lib/active_model/serializer/adapter/json_api/pagination_links.rb
bf4: See my (too long) comment over in https://github.com/rails-api/active_model_serializers/pull/1041/files#r37346434 but if we have

```ruby
      def initialize(collection)
            raise_unless_any_gem_installed
            @collection = collection
          end
```

rather than checking for WillPaginate or Kaminari, couldn't we check `collection.respond_to?(:total_pages) && collection.respond_to?(:current_page)` since that appears to be the extent of the paginating libs interface we need? (well, we can assume  `collection.size` works correctly if the other two methods are available.)

We'd still have an error message, but it would be that the resource needs these two methods, and we test against both these libraries which support it.

test/action_controller/json_api/pagination_test.rb
```diff
@@ -0,0 +1,116 @@
+require 'test_helper'
+require 'will_paginate/array'
```
bf4: Might we want to use the WillPaginate interface that doesn't monkey-patch Array in our tests?  I imagine this could cause bugs.

test/adapter/json_api/pagination_links_test.rb
```diff
@@ -0,0 +1,104 @@
...
+            new_links = links[:links].each_with_object({}) {|(key, value), hash| hash[key] = "#{value}&teste=teste" }
+            {}.tap do |hash|
+              hash[:data] = [data.values.flatten.second]
+              hash.merge! links: new_links
+            end
+          end
+
+          def test_pagination_links_using_kaminari
+            serializer = ArraySerializer.new(using_kaminari)
+            adapter = ActiveModel::Serializer::Adapter::JsonApi.new(serializer)
```
bf4: See https://github.com/rails-api/active_model_serializers/pull/1018#discussion_r36938727 re: using the SerializableResource interface, tl;dr 

```ruby
 def load_adapter(paginated_collection, options)
        options = options.merge(adapter: :json_api)
        ActiveModel::SerializableResource.new(paginated_collection, options)
      end
```

bf4:
@bacarini Thanks for dealing with my extensive commenting.  I owe you one if you're ever in Chicago.

lib/active_model/serializer/adapter/json_api/pagination_links.rb
bacarini: @bf4 would be [something like it](https://github.com/rails-api/active_model_serializers/pull/1041/files#diff-07e5a5aaa325f8dc2bcf944d12f0db83R28). 

I agree with you. It allows others pagination libs to be used with this feature.

test/action_controller/json_api/pagination_test.rb
bacarini: How we are going to work without `will_paginate/array` ? Because we will need `ActiveRecord` in order to do that.
Could you give me any example, please?

lib/action_controller/serialization.rb
bf4: @bacarini Yeah, the idea is that since we only need to get the current url etc from the context when paginating, instead of manipulating the request object in the controller, and pass in options, we can pass in the `context` (that is, the request object) and only use it when paginating.  The details of it being an `option` param I don't totally like, but it is a pattern used in Rails to have an evaluation context, which I do like a bit better for letting moving the responsibility of understanding the context from the controller to the Link Paginator.  Also, it adds less methods to the controller.

That make sense?

bf4: There's a bunch wrong in the specifics of the code above, I totally admit. I was more interested in the flow.. but e.g.

```ruby
def serializable_hash(options)
  pagination_links = {}
```

is wrong.  those options passed in to serializable_hash should be the pagination_links options.  The thing that has the context should be passed into the initializer, I think.

docs/howto/add_pagination_links.md
bf4: Winning!  That in the docs? (I'm in a car now)

bacarini:
@bf4 no worries! :smiley: 

I`ve changed a few things, could you review? 
tks a lot :+1:

docs/howto/add_pagination_links.md
bacarini: That is not! 
I may do it, but it is another issue, no?

bf4: Up to you, but I think it could be linked from the CollectionSerializer above https://github.com/rails-api/active_model_serializers/pull/1041/files#diff-04c6e90faac2675aa89e2176d2eec7d8R121

test/action_controller/json_api/pagination_test.rb
bf4: Just looking at [will_paginate/array, could you](https://github.com/mislav/will_paginate/blob/master/lib/will_paginate/array.rb)

```ruby
@array = []
options = {}
require 'will_paginate/collection'
page     = options[:page] || 1
per_page = options[:per_page] || WillPaginate.per_page
total    = options[:total_entries] || @array.length

WillPaginate::Collection.create(page, per_page, total) do |pager|
  pager.replace @array[pager.offset, pager.per_page].to_a
end
```

docs/howto/add_pagination_links.md
joaomdmoura: :+1: yup, there will good to have it, maybe a new article on the docs :smile:

test/adapter/json_api/pagination_links_test.rb
joaomdmoura: :+1:

lib/active_model/serializer/adapter.rb
```diff
@@ -21,9 +21,9 @@ def serializable_hash(options = nil)
       end
 
       def as_json(options = nil)
-        hash = serializable_hash(options)
-        include_meta(hash)
-        hash
+        serializable_hash(options).tap do |hash|
+          include_meta(hash)
+        end
```
bf4: This should go away when you rebase, no?

lib/active_model/serializer/adapter/json_api/pagination_links.rb
```diff
@@ -0,0 +1,59 @@
...
+              unless collection.current_page == collection.total_pages
+                pages[:next] = collection.current_page + FIRST_PAGE
+                pages[:last] = collection.total_pages
+              end
+            end
+          end
+
+          def url(options)
+            self_link = options.fetch(:links) {{}}
+            self_link.fetch(:self) {} ? options[:links][:self] : original_url
```
bf4: Looking good.  I'm a little uncertain about what this method is doing.  It seems to be returning `options[:links][:self] || original_url`, but it's being called on each pagination link, and after `pages[:self]` is set, so am I misunderstanding what url this is returning?  It's just the scheme + host + port of the current page, right?

```ruby
page_numbers = {
  self: collection.current_page,
  first: 1,
  prev: collection.current_page - 1,
  next: collection.current_page + 1,
  last: collection.total_pages,
}
# # missing method calls?
# current_page_params = query_parameters.delete(:page)
# options[:links] ||= {}
# url = options[:links][:self] ||= original_url
page_numbers.each_with_object({}) {|(page_name, page_number), pagination_links|
  page_link = {  page: { number: value, size: collection.size } }
  page_params = query_parameters.merge(page_link) # fresh for each link
  query_params = page_params.to_query
  url = options.fetch(:links, {}).fetch(:self, nil) ? options[:links][:self] : original_url
  page_links[page_name] = "#{url}?#{query_params}"
  # require 'uri/http'
  # URI::HTTP.build(scheme: 'http', host: "www.example.com", path:, "/path",
  "query": "query").to_s
}
```

bacarini: :+1:

lib/active_model/serializer/adapter.rb
bacarini: gone!

bacarini:
@joaomdmoura and @bf4 updated docs as suggested.

test/action_controller/json_api/pagination_test.rb
```diff
@@ -0,0 +1,116 @@
...
+        def test_render_only_last_and_next_pagination_links_with_additional_params
+          expected_links = { "self"=>"#{WILL_PAGINATE_URI}?page%5Bnumber%5D=1&page%5Bsize%5D=2&teste=additional",
+            "next"=>"#{WILL_PAGINATE_URI}?page%5Bnumber%5D=2&page%5Bsize%5D=2&teste=additional",
+            "last"=>"#{WILL_PAGINATE_URI}?page%5Bnumber%5D=2&page%5Bsize%5D=2&teste=additional"}
+          get :render_pagination_using_will_paginate, page: { number: 1, size: 2 }, teste: "additional"
+          response = JSON.parse(@response.body)
+          assert_equal expected_links, response['links']
+        end
+
+        def test_render_only_prev_and_first_pagination_links_with_additional_params
```
joaomdmoura: :+1:

joaomdmoura:
LGTM :+1:  only [this](https://github.com/rails-api/active_model_serializers/pull/1041/files#r37348553) still open.
@bf4 I know you are deeply involved in the review of this one so I'll also wit a LGTM from you before merge.

docs/howto/add_pagination_links.md
bf4: @bacarini would you mind updating the will_paginate example to use the ['newer, recommended' syntax that doesn't require the deprecated finders](https://github.com/mislav/will_paginate/issues/322#issuecomment-67785654)?

`@posts = Post.page(3).per_page(1)`

Or even

`@array = Post.all; @posts = WillPaginate::Collection.create(page = 1, WillPaginate.per_page, @array.size) do |pager|  pager.replace @array[pager.offset, pager.per_page].to_a end`

docs/howto/add_pagination_links.md
```diff
@@ -0,0 +1,98 @@
...
+    "self": "http://example.com/articles?page[number]=3&page[size]=1",
+    "first": "http://example.com/articles?page[number]=1&page[size]=1",
+    "prev": "http://example.com/articles?page[number]=2&page[size]=1",
+    "next": "http://example.com/articles?page[number]=4&page[size]=1",
+    "last": "http://example.com/articles?page[number]=13&page[size]=1"
+  }
+}
+```
+
+AMS relies on either [Kaminari](https://github.com/amatsuda/kaminari) or [WillPaginate](https://github.com/mislav/will_paginate). Please install either dependency by adding one of those to your Gemfile.
```
bf4: nitpick: AMS relies on either Kaminari or WillPaginate **for pagination**.

alternative to consider: AMS pagination relies on a paginated collection with the methods `current_page`, `total_pages`, and `size`, such as are supported by both Kaminari or WillPaginate. 

(links left out for ease of me typing :)

test/adapter/json_api/pagination_links_test.rb
```diff
@@ -0,0 +1,114 @@
...
+
+          def expected_response_without_pagination_links
+            data
+          end
+
+          def expected_response_with_pagination_links
+            {}.tap do |hash|
+              hash[:data] = [data.values.flatten.second]
+              hash.merge! links
+            end
```
bf4: overall, I'm happy with this PR and I think it's ok to merge.  I'd like to change this to something like

```ruby
expected_response_without_pagination_links.merge(links)
```

bf4: which I would submit in a follow-up pr to keep this moving

bf4:
@joaomdmoura looks good to me.  The interface is solid and the tests cover the usage well. I'd like to iterate on it a bit, but we can do that in another PR. @bacarini Has put in a lot of work and it will be the basis for more like it, I am sure! :+1: all around!

bf4:
I don't think it necessarily needs to be rebased since I think the commits tell a good story.

bf4:
After this is merged, I think #1018 will need to be repurposed to handle the non-pagination link scenarios and how they may interact with pagination links.

docs/howto/add_pagination_links.md
bacarini: Hey, @bf4 tks for the review.

I believe that, using `WillPaginate::Collection` we are overcomplicated it, since `will_paginate` already do it for us.
http://www.rubydoc.info/github/mislav/will_paginate/WillPaginate/Collection
>```
Array.class_eval do
  def paginate(page = 1, per_page = 15)
    WillPaginate::Collection.create(page, per_page, size) do |pager|
      pager.replace self[pager.offset, pager.per_page].to_a
    end
  end
end
```

What do you think doing this way?

###### Kaminari examples
```ruby
#array
@posts = Kaminari.paginate_array([1, 2, 3]).page(3).per(1)
render json: @posts

#active_record
@posts = Post.page(3).per(1)
render json: @posts
```

###### WillPaginate examples

```ruby
#array
@posts = [1,2,3].paginate(page: 3, per_page: 1)
render json: @posts

#active_record
@posts = Post.page(3).per_page(1)
render json: @posts
```

docs/howto/add_pagination_links.md
bacarini: :+1:

docs/howto/add_pagination_links.md
bf4: Good point

test/adapter/json_api/pagination_links_test.rb
bacarini: done! :+1: :smile:

kalys:
@bf4 Hi. When are you going to merge this PR?

joaomdmoura:
Awesome work @bacarini . I'me merging this.

bf4:
@kalys I don't have commit :) Thanks @joaomdmoura for stewarding this and @bacarini for all the work!

bacarini:
That is great!!!
Tks @bf4 and @joaomdmoura. :+1: 

I am glad we could make it through. :smiley:

leandrocp:
Great work @bacarini :beers:

bacarini:
cheers mate :beers: :smile:

Issue #1046, opened 2015-08-11
root option in render json is not working


PratheepV:
Hi,

I am using the rails-api and ams. But root option in render json is not working. Its always returning the class name only.

```ruby
#gem
gem "active_model_serializers", "~> 0.10.0.rc2"

#config
ActiveModel::Serializer.config.adapter = :json

#controller render
render json: current_user, root: :session

```
At https://github.com/rails-api/active_model_serializers/blob/v0.10.0.rc2/lib/active_model/serializer/adapter/json.rb#L40, are forcing the root as serializer json_key. 

It should respect my root option while rendering the json right.

If I change the code like below its working fine.
```ruby
{ (options[:root] || root) => @result }
```

--Pratheepv

joaomdmoura:
@PratheepV  Would you mind to test it against master? I think it was already fixed.
Thank You. :smile:

PratheepV:
@joaomdmoura , root options is not working for array serializers. Meaning if we have the 

```ruby
render status: 200,
        json: items, root: :apples
```
Getting the response as 

```javscript
  {"": [ {id: 1 , color: 'red'}]}
```

I tried using json_key in serializer also. 

I am using 0.10.0.rc3

PR #1047, opened 2015-08-11
root option in render json is not working

PratheepV:
Bug #1046 .

When I try to override the root option while rendering its not working.

```ruby
render json: current_user, root: :session
```
Getting the response as 

```ruby
{ user: { id:1, name:"Albert", email: {albert@abc.xyz} } }
```

### DIFF
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -37,7 +37,7 @@ module ActiveModel
             @result = @core.merge @hash
           end
 
-          { (options[:root] || root) => @result }
+          { options.fetch(:root) { root } => @result }
         end
 
         def fragment_cache(cached_hash, non_cached_hash)
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -37,7 +37,7 @@ module ActiveModel
             @result = @core.merge @hash
           end
 
-          { root => @result }
+          { (options[:root] || root) => @result }
         end
 
         def fragment_cache(cached_hash, non_cached_hash)


lib/active_model/serializer/adapter/json.rb
```diff
@@ -37,7 +37,7 @@ def serializable_hash(options = {})
             @result = @core.merge @hash
           end
 
-          { root => @result }
+          { (options[:root] || root) => @result }
```
bf4: Thanks for this.

Would you mind including some more info about the bug you're fixing and adding a test?

Did you consider using `options.fetch(:root) { root }` or `options[:root] ||= root` or other ways of checking for a root option? (Curious what considerations led you to this approach)

PratheepV: @bf4 I updated the PR.  Please have a look it once.

joaomdmoura:
@PratheepV  Would you mind to test it against master? I think it was already fixed.
Thank You. :smile:

PratheepV:
@joaomdmoura  Its working fine in master. Will it be available on next release? 

As of now I wrote custom Adapter like this 

```ruby
class TempAdapter < ActiveModel::Serializer::Adapter::Json
    super
    { options.fetch(:root) { root } => @result }
  end
end

ActiveModel::Serializer.config.adapter = TempAdapter
```

Thanks,
Pratheepv

joaomdmoura:
Yup @PratheepV, it will be on the next RC and on the 0.10.x.
I'm closing this one and the associated issue. :smile:

PratheepV:
@joaomdmoura , root options is not working for array serializers. Meaning if we have the 

```ruby
render status: 200,
        json: items, root: :apples
```
Getting the response as 

```javscript
  {"": [ {id: 1 , color: 'red'}]}
```

I tried using json_key in serializer also. 

I am using 0.10.0.rc3

PR #1048, opened 2015-08-11
Let FlattenJson adapter decide it doesn't include meta

bf4:
Why should the superclass get to decide it wants to
exclude meta from one of its subclasses.  I say don't ask
if we're a FlattenJson adapter, tell the adapter to
include the meta and let FlattenJson no-op on it.

From discussion in https://github.com/rails-api/active_model_serializers/pull/1041

### DIFF
--- a/lib/active_model/serializer/adapter/flatten_json.rb
+++ b/lib/active_model/serializer/adapter/flatten_json.rb
@@ -6,6 +6,13 @@ module ActiveModel
           super
           @result
         end
+
+        private
+
+        # no-op: FlattenJson adapter does not include meta data, because it does not support root.
+        def include_meta(json)
+          json
+        end
       end
     end
   end
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -22,7 +22,7 @@ module ActiveModel
 
       def as_json(options = nil)
         hash = serializable_hash(options)
-        include_meta(hash) unless self.class == FlattenJson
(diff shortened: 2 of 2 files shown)


bacarini:
:+1:

bf4:
@joaomdmoura any objections to moving the include_meta unless FlattenJson from the Adapter class to the FlattenJson class?

lib/active_model/serializer/adapter/flatten_json.rb
```diff
@@ -6,6 +6,13 @@ def serializable_hash(options = {})
           super
           @result
         end
+
+        private
+
+        # no-op
```
joaomdmoura: What about a more specific comment, nothing big, just to help ppl jumping in the code.

joaomdmoura:
:+1:  LGTM, just waiting for the comment before merging.

lib/active_model/serializer/adapter/flatten_json.rb
bf4: `no-op: The FlattenJson adapter does not include meta data`? `no-op: The FlattenJson adapter only serializes attributes`?

joaomdmoura: `no-op: FlattenJson adapter does not include meta data, because it does not support root.` :smile:

bf4:
updated

joaomdmoura:
Great! I'm merging it.

PR #1049, opened 2015-08-11
Fix incorrect s/options = {}/options ||= {}

bf4:
Introduced in #965, surfaced in #1041

(Thanks @bacarini for finding this!)


bacarini:
@bf4  what is the difference between both examples below? Why not put `options = {}` as method parameters?

```ruby
   options ||= {}
```
```ruby
```

bf4:
See referenced PR #965

joaomdmoura:
Good catch folks!

PR #1050, opened 2015-08-12
Add top-level jsonapi member

bf4:
Ref: http://jsonapi.org/format/#document-jsonapi-object

- `jsonapi_include_toplevel_object`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
   in the response document.
   Default: `false`.
- Used when `jsonapi_include_toplevel_object` is `true`:
  - `jsonapi_version`: The latest version of the spec the API conforms to.
    Default: `'1.0'`.
  - `jsonapi_toplevel_meta`: Optional metadata. Not included if empty.
    Default: `{}`.

Example meta usage:

```ruby
       
              jsonapi_include_toplevel_object = true,
              jsonapi_toplevel_meta = {
                'copyright' => 'Copyright 2015 Example Corp.'
              }

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Features:
 - [#1158](https://github.com/rails-api/active_model_serializers/pull/1158) Add support for wildcards in `include` option (@beauby)
 - [#1127](https://github.com/rails-api/active_model_serializers/pull/1127) Add support for nested
     associations for JSON and Attributes adapters via the `include` option (@NullVoxPopuli, @beauby).
+- [#1050](https://github.com/rails-api/active_model_serializers/pull/1050) Add support for toplevel jsonapi member (@beauby, @bf4)
 
 Fixes:
 
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -9,3 +9,11 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## JSON API
 
 - `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
+- `jsonapi_include_toplevel_object`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
+   in the response document.
+   Default: `false`.
+- Used when `jsonapi_include_toplevel_object` is `true`:
+  - `jsonapi_version`: The latest version of the spec the API conforms to.
+    Default: `'1.0'`.
+  - `jsonapi_toplevel_meta`: Optional metadata. Not included if empty.
+    Default: `{}`.
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -6,6 +6,41 @@ module ActiveModel
         autoload :PaginationLinks
         autoload :FragmentCache
 
+        # TODO: if we like this abstraction and other API objects to it,
+        # then extract to its own file and require it.
+        module ApiObjects
+          module JsonApi
+            ActiveModel::Serializer.config.jsonapi_version = '1.0'
+            ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}
+            # Make JSON API top-level jsonapi member opt-in
+            # ref: http://jsonapi.org/format/#document-top-level
+            ActiveModel::Serializer.config.jsonapi_include_toplevel_object = false
+
+            module_function
+
+            def add!(hash)
+              hash.merge!(object) if include_object?
+            end
+
+            def include_object?
+              ActiveModel::Serializer.config.jsonapi_include_toplevel_object
+            end
+
+            # TODO: see if we can cache this
+            def object
+              object = {
+                jsonapi: {
+                  version: ActiveModel::Serializer.config.jsonapi_version,
+                  meta: ActiveModel::Serializer.config.jsonapi_toplevel_meta
+                }
+              }
+              object[:jsonapi].reject! { |_, v| v.blank? }
+
+              object
+            end
+          end
+        end
+
         def initialize(serializer, options = {})
           super
           @include_tree = IncludeTree.from_include_args(options[:include])
@@ -21,11 +56,16 @@ module ActiveModel
         def serializable_hash(options = nil)
           options ||= {}
 
-          if serializer.respond_to?(:each)
-            serializable_hash_for_collection(options)
-          else
-            serializable_hash_for_single_resource(options)
-          end
+          hash =
+            if serializer.respond_to?(:each)
+              serializable_hash_for_collection(options)
+            else
+              serializable_hash_for_single_resource(options)
+            end
+
+          ApiObjects::JsonApi.add!(hash)
+
+          hash
         end
 
         def fragment_cache(cached_hash, non_cached_hash)
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -4,6 +4,8 @@ module ActiveModel
       include ActiveSupport::Configurable
       extend ActiveSupport::Concern
 
+      # Configuration options may also be set in
+      # Serializers and Adapters
       included do |base|
         base.config.array_serializer = ActiveModel::Serializer::ArraySerializer
         base.config.adapter = :attributes
--- a/test/adapter/json_api/toplevel_jsonapi_test.rb
+++ b/test/adapter/json_api/toplevel_jsonapi_test.rb
@@ -1,0 +1,84 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        class TopLevelJsonApiTest < Minitest::Test
+          def setup
+            @author = Author.new(id: 1, name: 'Steve K.')
+            @author.bio = nil
+            @author.roles = []
+            @blog = Blog.new(id: 23, name: 'AMS Blog')
+            @post = Post.new(id: 42, title: 'New Post', body: 'Body')
+            @anonymous_post = Post.new(id: 43, title: 'Hello!!', body: 'Hello, world!!')
+            @comment = Comment.new(id: 1, body: 'ZOMG A COMMENT')
+            @post.comments = [@comment]
+            @post.blog = @blog
+            @anonymous_post.comments = []
+            @anonymous_post.blog = nil
+            @comment.post = @post
+            @comment.author = nil
+            @post.author = @author
+            @anonymous_post.author = nil
+            @blog = Blog.new(id: 1, name: 'My Blog!!')
+            @blog.writer = @author
+            @blog.articles = [@post, @anonymous_post]
+            @author.posts = []
+          end
+
+          def test_toplevel_jsonapi_defaults_to_false
+            assert_equal config.fetch(:jsonapi_include_toplevel_object), false
+          end
+
+          def test_disable_toplevel_jsonapi
+            with_config(jsonapi_include_toplevel_object: false) do
+              hash = serialize(@post)
+              assert_nil(hash[:jsonapi])
+            end
+          end
+
+          def test_enable_toplevel_jsonapi
+            with_config(jsonapi_include_toplevel_object: true) do
+              hash = serialize(@post)
+              refute_nil(hash[:jsonapi])
+            end
+          end
+
+          def test_default_toplevel_jsonapi_version
+            with_config(jsonapi_include_toplevel_object: true) do
+              hash = serialize(@post)
+              assert_equal('1.0', hash[:jsonapi][:version])
+            end
+          end
+
+          def test_toplevel_jsonapi_no_meta
+            with_config(jsonapi_include_toplevel_object: true) do
+              hash = serialize(@post)
+              assert_nil(hash[:jsonapi][:meta])
+            end
+          end
+
+          def test_toplevel_jsonapi_meta
+            new_config = {
+              jsonapi_include_toplevel_object: true,
+              jsonapi_toplevel_meta: {
+                'copyright' => 'Copyright 2015 Example Corp.'
+              }
+            }
+            with_config(new_config) do
+              hash = serialize(@post)
+              assert_equal(new_config[:jsonapi_toplevel_meta], hash.fetch(:jsonapi).fetch(:meta))
+            end
+          end
+
+          private
+
+          def serialize(resource, options = {})
+            serializable(resource, { adapter: :json_api }.merge!(options)).serializable_hash
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -1,5 +1,4 @@
 require 'thread_safe'
-require 'active_model/serializer/adapter'
 require 'active_model/serializer/array_serializer'
 require 'active_model/serializer/include_tree'
 require 'active_model/serializer/associations'
@@ -11,6 +10,7 @@ module ActiveModel
   class Serializer
     include Configuration
     include Associations
+    require 'active_model/serializer/adapter'
 
     # Matches
     #  "c:/git/emberjs/ember-crm-backend/app/serializers/lead_serializer.rb:1:in `<top (required)>'"
--- a/test/support/serialization_testing.rb
+++ b/test/support/serialization_testing.rb
@@ -1,4 +1,8 @@
 module SerializationTesting
+  def config
+    ActiveModel::Serializer.config
+  end
+
   private
 
   def generate_cached_serializer(obj)
@@ -18,6 +22,18 @@ module SerializationTesting
     ActiveModel::Serializer.config.adapter = old_adapter
   end
   alias_method :with_configured_adapter, :with_adapter
+
+  def with_config(hash)
+    old_config = config.dup
+    ActiveModel::Serializer.config.update(hash)
+    yield
+  ensure
+    ActiveModel::Serializer.config.replace(old_config)
+  end
+
+  def serializable(resource, options = {})
+    ActiveModel::SerializableResource.new(resource, options)
+  end
 end
 
 class Minitest::Test


joaomdmoura:
Yup! I'm interest on this @bf4

bf4:
This is a little harder than I thought due the somewhat recursive way adapters work.

joaomdmoura:
No worries, take your time, let me know if there is something I can do in order to help you.

bf4:
Well, I think I'm going to work on other stuff for a bit.  I was just looking at warnings, and also at how using `class Adapter` as both a namespace and subclass causes some circular requires, and if maybe we should separate `AbstractAdapter` out of `Adapter`, leaving just the `create` and `adapter_class` methods (put those added by the  https://github.com/rails-api/active_model_serializers/pull/1017 )

joaomdmoura:
Yep, please clean our warnings would be great! :smile:

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -54,7 +78,7 @@ def serializable_hash_for_single_resource(serializer, options)
           primary_data = primary_data_for(serializer, options)
           relationships = relationships_for(serializer)
           included = included_for(serializer)
-          hash = { data: primary_data }
+          hash[:data] = primary_data # FIXME: should be an [], not a {}
```
bf4: ugh

docs/general/configuration_options.md
```diff
@@ -9,3 +9,9 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## JSON API
 
 - `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
+- `jsonapi_toplevel_member`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
```
bf4: defined in JsonApi adapter

test/adapter/json_api/toplevel_jsonapi_test.rb
```diff
@@ -0,0 +1,84 @@
...
+            end
+          end
+
+          def test_toplevel_jsonapi_meta
+            new_config = {
+              jsonapi_toplevel_member: true,
+              jsonapi_toplevel_meta: {
+                'copyright': 'Copyright 2015 Example Corp.'
+              }
+            }
```
bf4: The value of meta must be a hash

CHANGELOG.md
```diff
@@ -11,4 +11,5 @@
   * remove root key option and split JSON adapter [@joaomdmoura]
   * adds FlattenJSON as default adapter [@joaomdmoura]
   * adds support for `pagination links` at top level of JsonApi adapter [@bacarini]
-  * adds extended format for `include` option to JSONAPI adapter [@beauby]
+  * adds extended format for `include` option to JsonApi adapter [@beauby]
+  * adds support for top level jsonapi member support [@beauby]
```
beauby: There's an extraneous "support" at the end. Plus it should read `[@beauby, @bf4]` :)

lib/active_model/serializer/adapter/json_api.rb
bf4: n/m, I guess this is legal

> Primary data MUST be either:

> - a single resource object, a single resource identifier object, or null, for requests that target single resources
> - an array of resource objects, an array of resource identifier objects, or an empty array ([]), for requests that target resource collections

beauby: It is, indeed.

beauby: More than legal, it is actually the right behavior when targeting a single resource.

bf4: learning..

bf4: rebased to remove comment, but this comment remains...

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,10 +1,36 @@
...
+          ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}
+          module JsonApi
+            module_function
+
+            def add!(document)
+              return document unless ActiveModel::Serializer.config.jsonapi_toplevel_member
+              object = {
+                jsonapi: {
+                  version: ActiveModel::Serializer.config.jsonapi_version,
+                  meta: ActiveModel::Serializer.config.jsonapi_toplevel_meta
```
beauby: Meta is optional as per the spec, so it should't be included by default.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,10 +1,36 @@
...
+                }
+              }
+              object[:jsonapi].reject! { |_, v| v.blank? }
```
beauby: Aaaand I hadn't read that line yet.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,10 +1,36 @@
...
+            end
+          end
+        end
         extend ActiveSupport::Autoload
         autoload :PaginationLinks
         autoload :FragmentCache
 
         def initialize(serializer, options = {})
           super
+          @hash = {}
```
beauby: Why the instance variable? I think it adds confusion.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,10 +1,36 @@
 class ActiveModel::Serializer::Adapter::JsonApi < ActiveModel::Serializer::Adapter
+        # JsonApiObject: namespace for functions that build
+        # JSON API objects
+        module JsonApiObject
```
beauby: This could be extracted to `adapter/json_api/json_api_object.rb`, what do you think?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,10 +1,36 @@
...
+          # Make JSON API top-level jsonapi member opt-in
+          # ref: http://jsonapi.org/format/#document-top-level
+          ActiveModel::Serializer.config.jsonapi_toplevel_member = false
+          ActiveModel::Serializer.config.jsonapi_version = '1.0'
+          ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}
+          module JsonApi
```
beauby: Why the `JsonApi` namespace?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,10 +1,36 @@
...
+          end
+        end
         extend ActiveSupport::Autoload
         autoload :PaginationLinks
         autoload :FragmentCache
 
         def initialize(serializer, options = {})
           super
+          @hash = {}
+          JsonApiObject::JsonApi.add!(hash)
```
beauby: I'd much rather have it as a conditional assignation than a magical `add!` call that may or may not add stuff to its argument. Moreover, I think it belongs to the `serializable_hash` method rather than the constructor. How about:
```ruby
hash[:jsonapi] = JsonApiObject.serializable_hash if ActiveModel::Serializer.config.jsonapi_toplevel_member
```

Plus I'm starting to wonder whether the config should be named `jsonapi_toplevel_member` or something more explicit like `include_jsonapi_object`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -51,10 +77,11 @@ def serializable_hash_for_collection(serializer, options)
         end
 
         def serializable_hash_for_single_resource(serializer, options)
+          hash[:data] = {}
           primary_data = primary_data_for(serializer, options)
           relationships = relationships_for(serializer)
           included = included_for(serializer)
-          hash = { data: primary_data }
+          hash[:data].merge!(primary_data)
```
beauby: Why the `merge`?

lib/active_model/serializer/adapter/json_api.rb
bf4: Yeah, just not sure if we'll go with this as a pattern. So, I'm incubating it here

lib/active_model/serializer/adapter/json_api.rb
bf4: It's the `JsonApi` Object in the JSON API

lib/active_model/serializer/adapter/json_api.rb
bf4: good point

lib/active_model/serializer/adapter/json_api.rb
bf4: Why not?

beauby: Because it is more confusing than an assignation.

lib/active_model/serializer/adapter/json_api.rb
beauby: Yeah, but it's already the `JsonApiObject` Object in AMS, so why double namespace?

lib/active_model/serializer/adapter/json_api.rb
bf4: But it's not an assignation... it's a mutation

beauby: C.f. my other comment, I think it should not be. I don't think there's any valid reason to start building the hash in the constructor as it is the sole responsibility of the `serializable_hash` method.

beauby: On a technical note, it's a mutation solely because you assign it to `{}` right before. Maybe a bit of context will be helpful: the main goal of the JsonApi refactor I did was to decouple methods as much as possible so that they don't modify their arguments and just return the value they're supposed to compute. Hence, there is no need to either initialize `hash[:data]` or `merge` anything into it.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -51,10 +77,11 @@ def serializable_hash_for_collection(serializer, options)
         end
 
         def serializable_hash_for_single_resource(serializer, options)
+          hash[:data] = {}
           primary_data = primary_data_for(serializer, options)
           relationships = relationships_for(serializer)
           included = included_for(serializer)
```
beauby: Those three methods have no side-effect on `@hash`.

lib/active_model/serializer/adapter/json_api.rb
bf4: good call. my intention was that I was considering having a whole bunch of `JsonApiObjects` like `JsonApi`, `Link`, `Error`, `Meta`, `Data`, etc.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -31,14 +48,26 @@ def fragment_cache(cached_hash, non_cached_hash)
           ActiveModel::Serializer::Adapter::JsonApi::FragmentCache.new.fragment_cache(root, cached_hash, non_cached_hash)
         end
 
+        def fieldset
```
beauby: This probably belongs to a different PR.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,19 +6,36 @@ class JsonApi < Adapter
...
+              }
+              object[:jsonapi].reject! { |_, v| v.blank? }
+              document.merge!(object)
+            end
           end
         end
 
         def serializable_hash(options = nil)
           options ||= {}
+          self.hash = {}
```
beauby: Why the instance variable/side-effects?

bf4: Do you mean why `self.hash = ` vs. `@hash =`?

lib/active_model/serializer/adapter/json_api.rb
bf4: Yeah, I suppose

lib/active_model/serializer/adapter/json_api.rb
beauby: No I mean, why work on a `self.hash`/`@hash` instead of building an object in the method and returning it?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -123,10 +154,13 @@ def relationship_value_for(serializer, options = {})
...
 
         def relationships_for(serializer)
-          Hash[serializer.associations.map { |association| [association.key, { data: relationship_value_for(association.serializer, association.options) }] }]
+          Hash[serializer.associations.map { |association|
+            [association.key, { data: relationship_value_for(association.serializer, association.options) }]
+          }]
         end
 
         def included_for(serializer)
+          included = ActiveModel::Serializer::Utils.include_args_to_hash(instance_options[:include])
```
beauby: This will recompute the hash each time `serializable_hash` gets called. Granted, currently it should only happen once, but there might be outside-Rails use-cases where it's not the case.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,19 +6,36 @@ class JsonApi < Adapter
...
+              object[:jsonapi].reject! { |_, v| v.blank? }
+              document.merge!(object)
+            end
           end
         end
 
         def serializable_hash(options = nil)
           options ||= {}
+          self.hash = {}
+          ApiObjects::JsonApi.include_member? && ApiObjects::JsonApi.add!(hash)
```
beauby: I'd rather have it like that:
```ruby
hash.merge!(ApiObjects::JsonApi.member) if ApiObjects::JsonApi.include_member?
```

bf4: I like reading left to right, for the most part, and rubocop actually recommended this

docs/general/configuration_options.md
```diff
@@ -9,3 +9,11 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## JSON API
 
 - `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
+- `jsonapi_include_toplevel_member`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
+   in the response document.
+   Default: `false`.
+- Used when `jsonapi_include_toplevel_member` is `true`:
+  - `jsonapi_version`: The latest version of the spec the API conforms to.
+    Default: `'1.0'`.
+  - `jsonapi_toplevel_meta`: Optional metadata. Not included if empty.
```
bf4: I do think there's a latent `json_api` config object this exposes

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,26 +6,56 @@ class JsonApi < Base
...
         autoload :FragmentCache
 
-        def initialize(serializer, options = {})
+        module ApiObjects
+          # Make JSON API top-level jsonapi member opt-in
+          # ref: http://jsonapi.org/format/#document-top-level
+          ActiveModel::Serializer.config.jsonapi_include_toplevel_member = false
+          ActiveModel::Serializer.config.jsonapi_version = '1.0'
+          ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}
+          module JsonApi
```
bf4: Probably is a latent jsonapi config object here, or at least `JsonApi.config.version` via `class JsonApi < Base; def config; ActiveModel::Serializer.config.jsonapi; end;`

lib/active_model/serializer/adapter/json_api.rb
beauby: I don't mind it being `ApiObjects::JsonApi.include_member? && hash.merge!(ApiObjects::JsonApi.member)`, I just dislike the unnecessary side-effect.

docs/general/configuration_options.md
beauby: Indeed there is.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,6 +6,32 @@ class JsonApi < Base
...
+            module_function
+
+            def include_member?
+              ActiveModel::Serializer.config.jsonapi_include_toplevel_member
+            end
+
+            def add!(document)
```
beauby: Why the side-effects :crying_cat_face:

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -21,11 +47,17 @@ def initialize(serializer, options = {})
         def serializable_hash(options = nil)
           options ||= {}
 
-          if serializer.respond_to?(:each)
-            serializable_hash_for_collection(options)
-          else
-            serializable_hash_for_single_resource(options)
-          end
+          hash =
+            case
```
beauby: Why the extra line for a `case` statement? I generally like it, but this really looks like an `if...else` situation to me.

lib/active_model/serializer/adapter/json_api.rb
bf4: what side-effects?

lib/active_model/serializer/adapter/json_api.rb
bf4: When I was a young dev I learned to dislike using the return value from and if/else, since it's not as comfortable as using the return value from a case/switch

bf4: Since if/else is usually used for control flow

bf4: alternatively, 

```ruby
if ...
  hash = ...
else 
  hash = ...
end
```

lib/active_model/serializer/adapter/json_api.rb
beauby: Having `add!` modify its argument instead of returning a value.

lib/active_model/serializer/adapter/json_api.rb
beauby: I think the StyleGuide advises against the latter, and actually encourages using the return value of `if/else`, which is more than common in functional languages.

lib/active_model/serializer/adapter/json_api.rb
bf4: But, isn't this question the same as why do I prefer `foo.merge!(bar)` over `foo.merge!(build_bar)`?

beauby: Hmm not really, to me, it's more a matter of `do_some_magic(foo)` vs `foo[:attribute] = compute_something`.

beauby: With the former, in case of bug, you have to check how exactly is `do_some_magic` modifying its argument, whereas with the latter, you can directly see whether it potentially has anything to do with it at all.

bf4: So, what do you want?

lib/active_model/serializer/adapter/json_api.rb
bf4: I don't really care what the style guide says, honestly, if we agree on something that makes sense to us.  It does mention 'leveraging' the return value of if/else and case statements.

Is this a blocker?

lib/active_model/serializer/adapter/json_api.rb
bf4: I think `do_some_magic(foo)` is disingenous since the method is `JsonApiObject.add!(foo)`, it's clear you're mutating foo to addd a JsonApiObject.

lib/active_model/serializer/adapter/json_api.rb
beauby: I feel quite strongly about it. I personally think we should leverage the functional constructs of ruby in order to have the clearest/"no surprise-est" code possible, and I think using the return value of an `if/else` allows for that.

lib/active_model/serializer/adapter/json_api.rb
beauby: True, but it is not clear what "adding" means, it could be setting `foo[:data][:jsonapi]` for instance.

bf4: Or, its job is to add and you don't need to worry about it ;)

lib/active_model/serializer/adapter/json_api.rb
beauby: Agreed, as mentioned in https://github.com/rails-api/active_model_serializers/issues/1097, we should somehow namespace the config options.

lib/active_model/serializer/adapter/json_api.rb
bf4: Would you mind just adding a commit to my branch?  You obviously have a pretty strong opinions I don't share, so it would be faster if you changed it

bf4: I really don't care that strongly since I see it as internal for now till we figure out how we'd like to use it, if at all

beauby: Will do immediately.

beauby: Done.

bf4: Since you have the (first and) last commit, do you might (rebasing against master and) getting it passing on ci?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -21,11 +46,16 @@ def initialize(serializer, options = {})
...
-            serializable_hash_for_single_resource(options)
-          end
+          hash =
+            if serializer.respond_to?(:each)
+              serializable_hash_for_collection(options)
+            else
+              serializable_hash_for_single_resource(options)
+            end
+
+          ApiObjects::JsonApi.include_member? && hash[:jsonapi] = ApiObjects::JsonApi.member
```
bf4: So, the one thing I was trying to avoid is `hash[:jsonapi] =` give the Adapter#serializable_hash the responsibility for knowing the jsonapi object key.  I was hoping the JsonApiObject would maintain all knowledge of building that object

bf4: that is, I was thinking that each ApiObject would know how to build the json api object it represents

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -21,11 +48,16 @@ def initialize(serializer, options = {})
...
-            serializable_hash_for_single_resource(options)
-          end
+          hash =
+            if serializer.respond_to?(:each)
+              serializable_hash_for_collection(options)
+            else
+              serializable_hash_for_single_resource(options)
+            end
+
+          ApiObjects::JsonApi.include_member? && hash.merge!(ApiObjects::JsonApi.member)
```
beauby: Do you prefer it like this?

lib/active_model/serializer/adapter/json_api.rb
bf4: Maybe we should do that in this PR before its too late?

lib/active_model/serializer/adapter/json_api.rb
bf4: I do, esp. if you're ok with it... I'm willing to be persuaded... the whole ApiObjects::JsonApi is just an idea I'm playing with.. not very useful if it's the only one, but might lead to good abstractions for other objects, like errors, paginations, meta, links, etc.

What do you think of renaming `member` to `object`?  It would align with serializer.object and make sense in context.

lib/active_model/serializer/adapter/json_api.rb
beauby: Yeah, it would make sense. Any specific way you'd like it to be?

lib/active_model/serializer/adapter/json_api.rb
beauby: I'm ok with renaming `member` to `object`, although both those names are a bit confusing.

lib/active_model/serializer/adapter/json_api/api_objects.rb
```diff
@@ -0,0 +1,38 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        module ApiObjects
+          # Make JSON API top-level jsonapi member opt-in
+          # ref: http://jsonapi.org/format/#document-top-level
+          ActiveModel::Serializer.config.jsonapi_include_toplevel_object = false
+          ActiveModel::Serializer.config.jsonapi_version = '1.0'
+          ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}
```
bf4: I could see a race condition here now that it's lazy loaded.

bf4: (btw, the reason I didn't originally create this file was because I wasn't sure if I liked the abstraction of `ApiObjects` and wanted it to incubate a bit)

lib/active_model/serializer/adapter/json_api/api_objects.rb
```diff
@@ -0,0 +1,38 @@
...
+          module JsonApi
+            module_function
+
+            def add!(hash)
+              hash.merge!(object) if include_object?
```
bf4: IMHO, this should either return the resulting hash, or nil, but not both

lib/active_model/serializer/configuration.rb
```diff
@@ -4,6 +4,8 @@ module Configuration
       include ActiveSupport::Configurable
       extend ActiveSupport::Concern
 
+      # Configuration options may also be set in
+      # Serializers and Adapters
```
bf4: No longer correct if we keep api_objects.rb around

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -21,11 +22,16 @@ def initialize(serializer, options = {})
...
-            serializable_hash_for_collection(options)
-          else
-            serializable_hash_for_single_resource(options)
-          end
+          hash =
+            if serializer.respond_to?(:each)
+              serializable_hash_for_collection(options)
+            else
+              serializable_hash_for_single_resource(options)
+            end
```
bf4: someday we'll reify these two latent objects..

beauby:
@bf4 Let's pair on this one when you have time and get it over with.

NullVoxPopuli:
:+1:

lib/active_model/serializer/adapter/json_api/api_objects.rb
```diff
@@ -0,0 +1,38 @@
...
+            end
+
+            def include_object?
+              ActiveModel::Serializer.config.jsonapi_include_toplevel_object
+            end
+
+            def object
+              object = {
```
bf4: change to `@object ||= `

lib/active_model/serializer/adapter/json_api/api_objects.rb
bf4: return hash

lib/active_model/serializer/adapter/json_api/api_objects.rb
bf4: move back in for now

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -21,11 +22,16 @@ def initialize(serializer, options = {})
...
+
+          ApiObjects::JsonApi.add!(hash)
```
bf4: do we prefer this after?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -5,6 +5,7 @@ class JsonApi < Base
         extend ActiveSupport::Autoload
         autoload :PaginationLinks
         autoload :FragmentCache
+        autoload :ApiObjects
```
bf4:
```ruby
def config
  ActiveModel::Serializer.config.jsonapi
end
```

then `ActiveModel::Serializer::Adapter::JsonApi.version = '1.0'`

bf4: then update usage

docs/general/configuration_options.md
```diff
@@ -9,3 +9,11 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## JSON API
 
 - `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
+- `jsonapi_include_toplevel_object`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
+   in the response document.
+   Default: `false`.
+- Used when `jsonapi_include_toplevel_object` is `true`:
+  - `jsonapi_version`: The latest version of the spec the API conforms to.
+    Default: `'1.0'`.
+  - `jsonapi_toplevel_meta`: Optional metadata. Not included if empty.
```
bf4: update config to encapsulate in json_api

bf4:
@beauby squashed to one commit, set you as the 'committer' `git log --pretty='%cn' -n1 HEAD` shows you (I'm the 'author')

didn't do the jsonapi.config as it was funky and not high value

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,6 +6,40 @@ class JsonApi < Base
...
         autoload :FragmentCache
 
+        # TODO: if we like this abstraction and other API objects to it,
+        # then extract to its own file and require it.
+        module ApiObjects
+          module JsonApi
+            ActiveModel::Serializer.config.jsonapi_version = '1.0'
+            ActiveModel::Serializer.config.jsonapi_toplevel_meta = {}
+            # Make JSON API top-level jsonapi member opt-in
+            # ref: http://jsonapi.org/format/#document-top-level
```
bf4: @beauby  :)

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,6 +6,40 @@ class JsonApi < Base
...
+            def add!(hash)
+              hash.merge!(object) if include_object?
+            end
+
+            def include_object?
+              ActiveModel::Serializer.config.jsonapi_include_toplevel_object
+            end
+
+            # TODO: see if we can cache this
+            def object
```
beauby: How about something like:
```ruby
def object
  return @object if @object
  object = {
    jsonapi: {
      version: ActiveModel::Serializer.config.jsonapi_version,
      meta: ActiveModel::Serializer.config.jsonapi_toplevel_meta
    }
  }
  object[:jsonapi].reject! { |_, v| v.blank? }

  @object = object
  @object
end
```

docs/general/configuration_options.md
bf4: rename to `jsonapi_toplevel_object_meta`?

bf4: top level meta wouldn't be set in a global config.. but name could be confusing

beauby: Agreed.

lib/active_model/serializer/adapter/json_api.rb
bf4: since we add meta in some of the tests, I'd have to write a way to clear the cache, which I don't want to do right now, could be in a test helper I guess.

bf4: which is why I left it as a todo :)

PR #1052, opened 2015-08-13
Use underscored json_root when serializing a collection

whatthewhat:
#### The Issue
Assuming we're using the `:json` adapter with an activerecord model `DataPoint` currently we get this when serializing a collection:

```
GET http://localhost:3000/data_points

=>
{
  "datapoints": [...]
}
```

Expected key would be `data_points`.

The behaviour was changed in [this](https://github.com/rails-api/active_model_serializers/commit/418721302bf494b01618fbde6f198201a8a4b3dd) commit and there are some great examples in [this comment](https://github.com/rails-api/active_model_serializers/commit/418721302bf494b01618fbde6f198201a8a4b3dd#commitcomment-12442906) by @bolshakov.

I've included a high level test in `test/adapter/json/collection_test.rb` to better illustrate the point, changes to this file can be reverted before merging.

Appreciate any feedback, will gladly rewrite this multiple times, etc.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -127,7 +127,7 @@ module ActiveModel
     end
 
     def json_key
-      @root || object.class.model_name.to_s.downcase
+      @root || object.class.model_name.to_s.underscore
     end
 
     def id
--- a/test/adapter/json/collection_test.rb
+++ b/test/adapter/json/collection_test.rb
@@ -17,26 +17,27 @@ module ActiveModel
             @first_post.blog = @blog
             @second_post.blog = nil
 
-            @serializer = ArraySerializer.new([@first_post, @second_post])
-            @adapter = ActiveModel::Serializer::Adapter::Json.new(@serializer)
             ActionController::Base.cache_store.clear
           end
 
           def test_with_serializer_option
             @blog.special_attribute = "Special"
             @blog.articles = [@first_post, @second_post]
-            @serializer = ArraySerializer.new([@blog], serializer: CustomBlogSerializer)
-            @adapter = ActiveModel::Serializer::Adapter::Json.new(@serializer)
+            serializer = ArraySerializer.new([@blog], serializer: CustomBlogSerializer)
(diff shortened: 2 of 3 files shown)


manic:
+1

ferraniki:
+1

dukex:
+1

joaomdmoura:
I'm willing to merge it but there an open PR that might impact it as well, and will also solve this issue, can you check it out to see if we can close this one in favor of the other? 
I''m looking forward to have it! :smile: 
link: https://github.com/rails-api/active_model_serializers/pull/1029

whatthewhat:
Hey @joaomdmoura thanks for pointing out the OR, I somehow have missed it :)
The solution does seem a lot more complete, but there is a slight issue: it does not touch the json_key, meaning that changing the `key_format` still leaves us with `"datapoints"` from my example.

And looking at [this](https://github.com/rails-api/active_model_serializers/pull/1029#issuecomment-130194632) comment currently this might be the intention of the PR:

>Those should be converted wherever they are in the hash. Anything else:

> - top level schema keys like data, etc
> - values of attributes

> should be left alone.

We might want to move the discussion to https://github.com/rails-api/active_model_serializers/pull/1029 or I'll update this PR to use `key_format` after 1029 is merged. WDYT?

joaomdmoura:
@whatthewhat Indeed that was the @jfelchner approach, but I'm also in favor of doing the the to the `root` as you can see on my [comment](https://github.com/rails-api/active_model_serializers/pull/1029#issuecomment-130498191) so it would impact  the `key_format` as well. 

I know that is a lot of ppl that needs it, so what I'm doing is, I'm merging this one so ppl can use `master`, and we can move the discussion to #1029, maybe even pair on it in other to make it faster :smile:, it will probably override this implementation.

So I'm merging it for now.

whatthewhat:
@joaomdmoura sounds great :+1: 

I'll try to take a closer look at #1029 during the week, for example `json_key` behaviour when `key_format` is set to `:unaltered` seems like an interesting problem :)

PR #1060, opened 2015-08-17
Update fragment cache to support namespaced objects

aaronlerch:
The code to name the cached and uncached serializers didn't take into account namespaces in object names. This would exhibit itself with an error like: `NameError: wrong constant name Spam::unrelatedLinkCachedSerializer`.

The use of `.capitalize` didn't take all the scenarios into account. Since the name being passed in is already from an existing class name (`serializer.object.class.name`) there is no need to capitalize the value, instead I just did a simple string replacement for `::` namespace delimiters into `_`s.


joaomdmoura:
:+1: LGTM, a great catch

PR #996, opened 2015-07-09
Add linter for serializable resource

bf4:
Based on:
https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/lint.rb
https://github.com/rails/rails/blob/4-2-stable/activemodel/test/cases/lint_test.rb

With reference to:
https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/json.rb
https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb

PR #1063, opened 2015-08-18
Lead by example: lint PORO model

bf4:


bf4:
@joaomdmoura This, arguably, was missing from https://github.com/rails-api/active_model_serializers/pull/996

test/fixtures/poro.rb
```diff
@@ -13,12 +13,8 @@ def cache_key
...
   end
 
-  def cache_key_with_digest
-    "#{cache_key}/#{FILE_DIGEST}"
-  end
-
-  def updated_at
-    @attributes[:updated_at] ||= DateTime.now.to_time
+  def serializable_hash(options = nil)
+    @attributes
```
bf4: Interestingly, this method was missing

test/fixtures/poro.rb
```diff
@@ -42,6 +41,14 @@ def method_missing(meth, *args)
       super
     end
   end
+
+  def cache_key_with_digest
+    "#{cache_key}/#{FILE_DIGEST}"
+  end
+
+  def updated_at
+    @attributes[:updated_at] ||= DateTime.now.to_time
```
bf4: This should really be `Time.now`.  We can be pretty confident we're using the Gregorian calendar, no?

joaomdmoura:
:+1:

test/fixtures/poro.rb
joaomdmoura: :+1:

test/fixtures/poro.rb
joaomdmoura: I don't mind keeping it :smile:

bf4: I didn't change it in this PR. :) OT: In 1.8 and 1.9 I think DateTime was much slower than Time.  Now, I think avoiding DateTime is mostly about it being the wrong tool and being weird with timezones sometimes...

PR #1065, opened 2015-08-21
Add relationship information to PORO models.

beauby:
In order to test deserializers (#950), test models should respond to associations getters depending on the type of association (`has_many => [], belongs_to => nil, has_one => nil`), event if the association is not explicitly set.

### DIFF
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -1,6 +1,12 @@
 class Model
   FILE_DIGEST = Digest::MD5.hexdigest(File.open(__FILE__).read)
 
+  @@associations = {}
+
+  def self.associations=(assoc)
+    @@associations = assoc
+  end
+
   def self.model_name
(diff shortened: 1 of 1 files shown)


joaomdmoura:
This is really helpful! I'll check it tomorrow in order to merge it! :+1: 
I start to think you might be right, we might need to move forward with an AM / AR dependency in a near future.

beauby:
So, as far as testing the deserializers is concerned, I think we should test separately the following:

1. sanitizing the parameters,
2. actually deserialize sanitized parameters to a format that ActiveModel understands,
3. ensuring the format from 2. is, does actually work well with ActiveModel,
4. integration tests to ensure the whole machine is working.

Currently, 2. and 3. are limited by the lack of access to a persistence layer in the test infrastructure (deserializers call `Model.find()`).

beauby:
This PR might not be relevant anymore, considering the dependence on AR can be avoided.

PR #1070, opened 2015-08-21
Deserializers

beauby:
Following @joaomdmoura 's work and the discussion on #950, here is an up-to-date version of the deserializers, with some tests. Still a WIP.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,10 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in active_model_serializers.gemspec
 gemspec
 
-gem "minitest"
+group :development do
+  gem "minitest"
+  gem "sqlite3"
+end
 
 version = ENV["RAILS_VERSION"] || "4.2"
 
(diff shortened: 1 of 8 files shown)


beauby:
Added whitelisting at runtime, so that people can restrict the attributes/relationships allowed as follows:
```ruby
def create_params
  UserSerializer.deserialize(params, [:username, :email, :password, :referer])
end

def update_params
  PostSerializer.deserialize(params, [:username])
end

def create
  @post = Post.create(create_params)
  # ...
end

def update
  @post = Post.find_by(...)
  @post.update_attributes(update_params)
  # ...
end
```

beauby:
Finally added a test of end-to-end deserialization (backed by ActiveRecord).

beauby:
For some reason the installation of `sqlite3` fails on the jruby test envs.

joaomdmoura:
@beauby Could you submit this to my deserialization branch specifically? Otherwise I won;t be able to merge it.

Issue #921, opened 2015-05-21
json serializer ignores nested associations


quaternion:
Example 0.10.x:
````ruby
class NodeSerializer < ActiveModel::Serializer
  attribute :id
end

class UserSerializer < ActiveModel::Serializer
  attribute :id
  attribute :name
  belongs_to :node # <--- ignored
end

class PostSerializer < ActiveModel::Serializer
  attribute :id
  belongs_to :user
end
````
PostSerializer produces json without a user node:
````
{ id: 1, user: { id: 1, name: 'Bob' } }
````
Is this a bug or a feature?

**Comment by groyoh** [comment]
If you're using the JSON adapter then it is the expected behavior. The JSON adapter only render the attributes of associations (see [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json.rb#L28))

**Comment by quaternion** [comment]
Yes, you are right. I use json adapter.

**Comment by groyoh** [comment]
If you need a workaround, you can do this:
```ruby
class UserSerializer < ActiveModel::Serializer
  attribute :id
  attribute :name
  attribute :node
  def node
    serializer = ActiveModel::Serializer.serializer_for(object.node).new(object.node)
    ActiveModel::Serializer::Adapter::Json.new(serializer).as_json
  end
end
```

chrisbranson:
I ran into this today, having previously used 0.9.x where nested associations were found. Is this 'expected behaviour' in that that's the way it is supposed to work, or would a PR be welcome? A small change gets this behaviour back - see my fork https://github.com/insphire/active_model_serializers/commit/80de7191ef6e79c79714ef1a011c46c22ddff210

joaomdmoura:
@quaternion ans @groyoh said, this is indeed the expected behaviour.
This approach takes into account a more lean implementation avoiding heavy responses and also trying to help people into build a "more RESTfull" API.
I'm no saying we will definitely keep it this way, but before move forward into a PR I would like to hear more opinions from other AMS members. cc @kurko @guilleiguaran 
I'm closing this for now. But lets keep the discussion :smile:

aaronlerch:
[...]

PR #789, opened 2015-01-27
Avoid infinate loops due to circular association referecnes.

ileitch:
This commit fixes a` SystemStackError` which can arise in a situation such as this:

I have not tried with master.

```ruby
class UserSerializer < AcitveModel::Serializer
  has_one :review
end

class ReviewSerializer < ActiveModel::Serializer
  has_one :user
end
```

Issue #211, opened 2013-02-20
Infinite loop in circular associations


heartsentwined:
`active_model_serializers` can handle self-referencing associations fine, e.g. `categories`, but it will run into an infinite loop if you try to sideload circular associations. An example is the twitter-esque `follower`-`followed` association pair.

```rb
class BaseSerializer < ActiveModel::Serializer
  embed :ids, include: true
end

class FollowerSerializer < BaseSerializer
  has_many :followeds
end

class FollowedSerializer < BaseSerializer
  has_many :followers
end
```

Let's say you start by requesting a `follower`. `active_model_serializers` will traverse its `followeds`, sideload the `followed` models by initiating a new `FollowedSerializer` for each, and if any `followed` points back to the original `follower` via its own `has_many :followers` association, we have an infinite loop.

The current workaround is, unfortunately, disable sideloading entirely for the offending associations.

```rb
class BaseSerializer < ActiveModel::Serializer
  embed :ids, include: true
end

class FollowerSerializer < BaseSerializer
  has_many :followeds, embed: :ids, include: false
end

class FollowedSerializer < BaseSerializer
  has_many :followers, embed: :ids, include: false
end
```

Perhaps keeping track of the model + primary key while loading associations (recursively) can solve this issue? This way the same model will only be serialized once.

caspyin:
Does #722 fix this same issue? It changes `lib/active_model/serializer/adapter/json_api.rb` where @colby-swandale changes `lib/active_model/serializer/associations.rb`. I'm not familiar enough with the inner workings of this project to know for sure.

#722 has already been pulled into master so if it does fix this issue, it would be nice to have it applied to 0.8.x and 0.9.x branches.

ggordon:
@caspyin #722 is specifically for the json-api adapter in master so it's not applicable to 0.8 or 0.9 branches.

caspyin:
@ggordon Ah ok, thanks for clarifying.

kurko:
@caspyin could you open a PR? I'd me happy to merge :)

caspyin:
@kurko You might look at @colby-swandale. He has a fork with a solution 2e791b6 for this already. @colby-swandale consider updating your fork and issuing a PR?

evolve2k:
Wow, we're almost at a two year anniversary on this issue, having been open back on 20 Feb 2013.
It's obviously affecting a lot of people, Im sure it's hairy, but would be awesome if it could be resolved.

[...]

PR #1073, opened 2015-08-21
Create NestedJson adapter.

kayhide:
I needed an adapter to serialize models with multi-depth relations for my project.
So I wrote it.

I think this might be helpful for some people.


joaomdmoura:
wow! Didn't reviewed it yet, but this is awesome, I'm definitely interest on this. I'll check tonight probably.

kayhide:
Hi,
I have checked the discussion on #921.

This adapter raises an error if nesting is too deep, typically in case of circular relations.
I think this is a desirable behavior for many cases.

If you want to change the limit depth, you can do it just giving an argument `limit_depth: xx` to the adapter, or setting default value `NestedJson.default_limit_depth = xx`.

bf4:
Possibly related:

- https://github.com/rails-api/active_model_serializers/issues/211
- https://github.com/rails-api/active_model_serializers/pull/789

lib/active_model/serializer/adapter/nested_json.rb
```diff
@@ -0,0 +1,57 @@
+module ActiveModel
+  class Serializer
+    class Adapter
+      class NestedJson < Json
+        cattr_accessor :default_limit_depth
+        @@default_limit_depth = 5
+
+        def serializable_hash options = {}
```
bf4: Yikes, it's a pretty big smell in the Json serializer that we're subclassing it but completely overriding this method, where everything important happens

also, might this belong in the Adapter superclass? It looks on first inspection like it would apply to the JsonApi adapter as well, no?

test/fixtures/poro.rb
```diff
@@ -266,3 +266,18 @@ def json_key
...
+
+PostNestedSerializer = Class.new(ActiveModel::Serializer) do
+  attributes :title, :body, :id
+  has_many :comments, serializer: CommentNestedSerializer
+end
+
+AuthorNestedSerializer = Class.new(ActiveModel::Serializer) do
+  attributes :id, :name
+  has_many :posts, serializer: PostNestedSerializer
+end
```
bf4: @joaomdmoura I think we should start following some convention for associating poros with tests. I think it's becoming append-only...

bf4:
Thanks!

lib/active_model/serializer/adapter/nested_json.rb
kayhide: That is for `#fragment_cache`.

The way we cache fragmentation is exactly the same to `Adapter::Json` ( like `Adapter::FlattenJson` as well).
And `Adapter::JsonApi` caches differently.

test/fixtures/poro.rb
kayhide: I have a suggestion.
In my projects, I am doing like this way:

```ruby
class Author < ActiveModel::Serializer
  attributes :id, :name

  class ForShow < self
    has_many :posts
  end
end

class Post < ActiveModel::Serializer
  attributes :id, :title, :body

  class ForShow < self
    has_many :comments
  end
end
```

and so on.

I define default serializers not to have any  relations.
And for specific cases, like showing details, define serializers as nested classes.

When I use them in a controller, I write like:
```ruby
def show
  ...
  render json: @post, serializer: PostSerializer::ForShow
end
```

Like this way, I can easily avoid unexpected circular relations.
Also scoping with default serializer classes make it clear which serializer is for which model.

bf4: works for me

joaomdmoura:
Hey, sorry for taking so long, was really busy, now I stopped to check it our for real.
First of all tks @kayhide and @bf4 for your hard work. I really appreciate this. 

I know that you needed multi-depth relations and make your own adapter was the fastest way, but IMO the best scenario would be to find a way to add it to json + flatten-json existing adapters. What are your thoughts about it?  :smile:

kayhide:
@joaomdmoura I agree with you.

OK, I will add these functionalities on this adaptor.

- taking care of root
- adding an option to trim off deep associations not to raise an error

After covering these, I think we can replace the existing json, flatten-json adapaters.

kayhide:
Hi, I updated the adapter and replaced existing Json/FlattenJson adapter.

This adapter accepts options:
`limit_depth` limits the depth of associations.

`check_depth_strategy` defines the behavior when nesting is too deep.
if `:fail` is set, it raises an error.
if `:trim` is set, it ignores deeper associations.

Setting `limit_depth = 1` and `check_depth_strategy = :trim` makes its behavior just the same to the original Json/FlattenJson adapter.

And default values are available setting like:
```ruby
Adapter::Json.default_limit_depth = 5
Adapter::Json.default_check_depth_strategy = :fail
```

For now, I set default values to follow original adapters behavior, and made them to pass all the existing tests untouched.
But I think people want it to handle nested associations to some depth as a default.

To do so, we need to fix the default values and rewrite the relevant tests.

bryanrite:
I really like this approach, defining all appropriate associations in each serializer and then being able to limit the depth in the controller is perfect.

bryanrite:
One bug I noticed will trying out this branch, is that attributes with the name `name` cause an error now. For example, if the `Comment` model had a "name" attribute and you were serializing it like: 

```ruby
def Api::V1::CommentSerializer < ActiveModel::Serializer
  attribute :id
  attribute :name
end
```

you get:

```ruby
undefined method `name' for #<Api::V1::CommentSerializer:0x007fdec7a23408>`
```

This is fixed by defining a method:

```ruby
def name
  object.name
end
```

This is the only attribute it happens on that I have found so far.

bryanrite:
And +1 for supporting a `limit_depth` of 0.

kayhide:
@bryanrite hi, thanks for your support :)

> attributes with the name name cause an error 

You can try `attributes` not `attribute`.
It's plural not singular.
Those are different methods.

As I read the code, singular version is for the case when the key and the accessor method are not the same.

beauby:
The error is related to #1094.

kayhide:
Updated to support limitless depth.
Giving `check_depth_strategy = :pass` makes it go deep without any limit.
In this case the `limit_depth` value is ignored.

kayhide:
@beauby ah, right. thanks for the comment.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -4,49 +4,86 @@ module ActiveModel
...
           end
-
-          { root => @result }
+          hash
         end
 
-        def fragment_cache(cached_hash, non_cached_hash)
-          Json::FragmentCache.new().fragment_cache(cached_hash, non_cached_hash)
+        def too_deep? depth
+          if depth > @limit_depth
```
bf4: nice!

NullVoxPopuli:
closed in favor of #1127 and #1158 
deep relationships are now managed via the include option passed in to the adapter

example:

`render json: @posts, include: [author, comments: author]`

or

`render json: @posts, include: 'author,comments.author'`

beauby:
Moreover, the "depth strategy" is available via wildcards:
```ruby
render json: @posts, include: '*.*.*'
```
This would include all nested relationships 3 levels down.

Issue #1080, opened 2015-08-24
Using Ember Data 2.0.0-beta1 with ActiveModel::Serializer::Adapter::JsonApi


mattboldt:
Since the new ember no longer has `DS.ActiveModelSerializer` or `DS.ActiveModelAdapter`, I'm looking to switch to their new `DS.JSONAPISerializer`. This seems to be the best way to go, if the previous two won't be supported in the future.

Ember is expecting the data from the api to be something like 
```
store.push({
  data: {
    id: '1', 
    type: 'user', 
    attributes: { 
      name: 'Pangratz' 
    }
  }
});
```

But from my rails app, with the `:json_api` option set to true, my api is returning `user` as `users` plural.

```
  data: {
    id: '1', 
    type: 'users', 
    attributes: { 
      name: 'Pangratz' 
    }
  }
```

Ember expects a singular model name, and as such cannot push data.

beauby:
Right, the JSONAPI spec does not take a stand on whether the `type` should be singular or plural. I think there should be a global config for `ActiveModel::Serializer` to pluralize or not the resources.

beauby:
I'm working on a PR to add this. Should be ready in a few minutes. Could you test it when it's out?

mattboldt:
Absolutely!

beauby:
@mattboldt Here it is. Simply set `ActiveModel::Serializer.config[:jsonapi_resource_type] = :singular` and it should work.

mattboldt:
Yes, this worked! Thanks so much.

PR #1081, opened 2015-08-24
Add configuration option to set resource type to singular/plural

beauby:
JSONAPI spec (http://jsonapi.org/format/#document-resource-object-identification) does not specify whether the `type` should be singular or plural. By default, AMS makes it plural, while some client-side libraries default to singular. This PR adds a configuration option to globally set it to plural or singular. (c.f. #1080)

### DIFF
--- a/test/adapter/json_api/resource_type_config_test.rb
+++ b/test/adapter/json_api/resource_type_config_test.rb
@@ -1,0 +1,59 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    class Adapter
+      class JsonApi
+        class ResourceTypeConfigTest < Minitest::Test
+          def setup
+            @author = Author.new(id: 1, name: 'Steve K.')
+            @author.bio = nil
+            @author.roles = []
+            @blog = Blog.new(id: 23, name: 'AMS Blog')
+            @post = Post.new(id: 42, title: 'New Post', body: 'Body')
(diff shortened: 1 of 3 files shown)


beauby:
The proposed config key is `jsonapi_resource_type` (I couldn't find anything better... any idea?), which takes values `:plural` (by default) or `:singular`. I'll update the docs if there is interest on this PR.

joaomdmoura:
Cool @beauby Thanks you so much, I'm merging this :smile:

joaomdmoura:
Oh, Could you please also write some docs to it onto our new docs and maybe mention it _shortly_ on README?

test/adapter/json_api/resource_type_config_test.rb
```diff
@@ -0,0 +1,59 @@
...
+              ActionController::Base.cache_store.clear
+              assert_equal('comments', adapter.serializable_hash[:data][:type])
+            end
+          end
+
+          def test_config_singular
+            with_jsonapi_resource_type :singular do
+              serializer = CommentSerializer.new(@comment)
+              adapter = ActiveModel::Serializer::Adapter::JsonApi.new(serializer)
+              ActionController::Base.cache_store.clear
```
bf4: SerializableResource? :(

joaomdmoura: Indeed, we could use `SerializableResource` through all those serializer + adapter in the tests, glad that yyou pointed that out@bf4. @beauby could you make a PR updating it?

Issue #938, opened 2015-06-04
undefined method `embed' 


tommyblue:
I'm using the 0.10.0.rc1 version because 0.9.x doesn't work, but with this definition of the serializer: 

```
class Api::V1::ItinerarySerializer < ActiveModel::Serializer
  attributes :id, :title

  embed :ids, include: true
  has_many :pois, serializer: Api::V1::PoiSerializer
end
```
 
I get this error:

```
NoMethodError (undefined method `embed' for Api::V1::ItinerarySerializer:Class):
  app/serializers/api/v1/itinerary_serializer.rb:4:in `<class:ItinerarySerializer>'
  app/serializers/api/v1/itinerary_serializer.rb:1:in `<top (required)>'
  app/controllers/api/v1/itineraries_controller.rb:29:in `update'
```

Embed is deprecated? How can I generate the "old" JSON including associated objects, like this:

```
{
  "itinerary": {
    "poi_ids": [1,4,5]
  },
  { "pois": [
    // POI serialized objects
  ]}
}
```

groyoh:
I'm afraid that is not currently possible with 0.10. @joaomdmoura is this a planned feature for 0.10?
In the meantime, you could workaround it by creating your own adapter based on the json adapter. If you need some help with that let me know.

tommyblue:
ok, thanks. 
Shouldn't this an important option? I have an habtm association and embedding object is a good way to avoid data replication inside the single json response or avoiding multiple requests (one for each id).
Isn't it?

joaomdmoura:
Hey @tommyblue, indeed, @groyoh is right!

Its possible to accomplish something close to it by using JSON-API, but would be a different implementation and not the '"old" JSON'.

btw why not use only:
```ruby
 has_many :pois, serializer: Api::V1::PoiSerializer
```

This will return the serialized objects, isn't waht you need? You can map the ids if you want to, on a virtual attribute:

```ruby
class Api::V1::ItinerarySerializer < ActiveModel::Serializer
  attributes :id, :title, : poi_ids
  has_many :pois, serializer: Api::V1::PoiSerializer

  def poi_ids
    object.pois.map {|p| p.id}
  end
end
```

tommyblue:
If I'm not wrong, your solution puts the serialized POI objects inside their Itinerary "fathers". This means that  there will be a lot of replicated POI objects inside different itineraries. This is not a problem with few objects but with a lot of objects is a lot of useless weigth into the JSON

[...]

Issue #899, opened 2015-05-04
Side loading doesn't work in 0.10.0.rc1 (non-JSONAPI)


blakemcintyre:
I've been trying to setup side loading using the standard adapter, (not JSONAPI), and have been unable to generate any embedded data. This works in 0.8.1 but the issue stated in https://github.com/rails-api/active_model_serializers/issues/377 has lead me to trying 0.10.x. When I tried using the JSONAPI adapter I was able to get included/linked data.

The following is the basic setup I've been testing with:
```ruby
class ArticleSerializer < ActiveModel::Serializer
  attributes :id, :title, :published_at

  has_many :authors, embed: :ids, serializer: UserSerializer

  url :article
end

class UserSerializer < ActiveModel::Serializer
  attributes :id, :name

  url :user
end


class Article < ActiveRecord::Base
  has_many :article_authors
  has_many :authors, through: :article_authors, class: User
end

class ArticleAuthor < ActiveRecord::Base
  belongs_to :article
  belongs_to :author, class: User
end

class User < ActiveRecord::Base
  has_many :article_authors, foreign_key: :author_id
  has_many :articles, through: :article_authors
end


class ArticlesController < ApplicationController
  def index
    render json: Article.last(3), root: :articles, include: 'authors'
  end
end
```

henrymazza:
+1

I’m using ember-data, is there a way to achieve the same format I had with 0.8.x? That is with `embed :ids` and associations sideloaded?

meltingice:
This is currently the only thing stopping me from upgrading to 0.10.x. In order to maintain the same API format and prevent breaking our client, I need to be able to sideload objects in the root.

henrymazza:
Now I hear that ember data is aiming to the same JSONAPI format that AMS 0.10.x is defaulting to. Plus it should ship a compatible beta version somewhere in June. 

Can someone confirm this? 

Enviado de um aparelho móvel.

meltingice:
For what it's worth, we aren't using the JSONAPI spec, but we still need to sideload data. I've been playing around with the JSON serializer in order to add the functionality back. I'll see if I can get a proof-of-concept together.

joaomdmoura:
Hey everybody, I'm just jumping in to let you know that I'm following this and will try to give it a closer look on the weekend. @meltingice would be awesome to have some prof-of-concept :smile: 
Thank y'all! :heart:

benwoodward:
+1

saneshark:
[...]

PR #1088, opened 2015-08-27
Sideloading associations

NullVoxPopuli:
Hey all, since this is a highly requested feature, I figured I'd take a stab at it since I need it for my own ember work.

Thoughts on the implementation?

The goal of this is to address needs of #938 and #899 

Usage

    #config/initializers/active_model_serializers.rb
    ActiveModel::Serializer.config.adapter = :json
    ActiveModel::Serializer.config.sideload_associations = true

And that's it. :-)
No need to have embed: :ids on all the associations.

### DIFF
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -6,6 +6,8 @@ module ActiveModel
       class Json < Adapter
         def serializable_hash(options = nil)
           options ||= {}
+          sideload = ActiveModel::Serializer.config.sideload_associations
+
           if serializer.respond_to?(:each)
             @result = serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
           else
@@ -18,7 +20,7 @@ module ActiveModel
             serializer.associations.each do |association|
               serializer = association.serializer
               opts = association.options
-
+              
               if serializer.respond_to?(:each)
                 array_serializer = serializer
                 @hash[association.key] = array_serializer.map do |item|
@@ -37,10 +39,35 @@ module ActiveModel
                   end
               end
             end
-            @result = @core.merge @hash
+
+            if sideload
+              association_ids = {}
+              @hash.map do |association_name, associated_models|
+                # TODO: use active support inflectors?
+                letters = association_name.to_s.split('')
+                letters = letters[0..letters.length - 2] if letters.last == 's'
+                singular_name = letters.join
+                id_name = (singular_name.singularize + "_ids").to_sym
+
+                # build id list
+                association_ids[id_name] ||= []
+
+                ids = Array.wrap(associated_models).map{ |model_hash|
+                  model_hash[:id]
+                }
+                association_ids[id_name] = ids
+              end
+              @result = @core.merge association_ids
+            else
+              @result = @core.merge @hash
+            end
           end
 
-          { root => @result }
+          if sideload
+            { root => @result }.merge(@hash || {})
+          else
+            { root => @result }
+          end
         end
 
         def fragment_cache(cached_hash, non_cached_hash)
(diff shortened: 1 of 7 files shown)


lib/active_model/serializer/adapter/json.rb
```diff
@@ -6,6 +6,8 @@ class Adapter
       class Json < Adapter
         def serializable_hash(options = nil)
           options ||= {}
+          sideload = ActiveModel::Serializer.config.sideload_associations
```
NullVoxPopuli: not entirely related to the PR, but should this method be recursive? it looks like for the associations, the hash is being manually built.

Criten:
:+1: This is working for me so far in my application! I had to tweak one thing though. I have a has_one and this sideloads it as a singleton object as apposed to an array. Before I'm pretty sure it still stuck it in an array.

NullVoxPopuli:
@Criten could you elaborate? I haven't used a has_one in my app yet, so I'm uncertain what the exact issue you found is / was

Criten:
Just edited my comment.

NullVoxPopuli:
interesting. Does has_one work in normal AMS, nested the data under `foo`?

Criten:
Yeah.. just checked. To be a bit more clear....

``` ruby
# 0.9.X
  has_one :bar
end
```

``` json
bar: {
  foo_id: 1
},
foos:
[{
  id: 1
}]
```

VS

``` ruby
  has_one :foo
```

``` json
bar: {
  foo_id: 1
},
foo:
{
  id: 1
}
```

Criten:
I think that's what its doing....

NullVoxPopuli:
kk, I'll see if I can come up with a test and fix it

NullVoxPopuli:
@Criten I added a test but was not able to replicate. :-(

Is there anyway you have time to monkey around with the sideload_test.rb tests and see if you can replicate in there?

(I did add two more tests, though, for has_one and belongs_to)

estum:
Heh, About a week ago I've started to make (my own implementation)[https://github.com/estum/active_model_serializers/commits/sideload] of the same thing. Currently it is documented but doesn't contain the proper specs, that is why I haven't yet made a PR. In general It works for me, but doesn't support custom names of associations and excludes not sideloaded reflections from a "relationships" object. Also, I've added a sideloading to the SerializedResource's initialization process and now doubt about this decision.

BenMorganIO:
:+1:

Criten:
I don't think this is working for me with nested has_many? Do you know if that works?

NullVoxPopuli:
Does regular AMS work with nested has many? When I was looking through the
code, I don't think I saw any sort of recursion on the serialization bit. I
could be wrong though.

I do know that tested has many doesn't work. But I also don't know if it's
pre existing.

Criten:
It definitely worked in version 0.9...

NullVoxPopuli:
I'll take a look at the tests on master when I can, and see if there is a
nested serialization gap. :-)

NullVoxPopuli:
@Criten it looks like nested has many is broken in master: https://github.com/NullVoxPopuli/active_model_serializers/commit/15147de89bcda5546739142dd64d8cf5410c8388

lib/active_model/serializer/adapter/json.rb
```diff
@@ -37,10 +39,35 @@ def serializable_hash(options = nil)
                   end
               end
             end
-            @result = @core.merge @hash
+
+            if sideload
+              association_ids = {}
+              @hash.map do |association_name, associated_models|
+                # TODO: use active support inflectors?
```
joaomdmoura: Yup! :+1:

joaomdmoura: I'd say yes, but I'm concerned with ppl using AMS outside Rails context.
I'd keep it like that and come back at it after Rails 5 release

NullVoxPopuli:
Closing, in favor of an implementation based off my refactored work (#1111)

PR #1089, opened 2015-08-28
Add ActiveModelSerializers.logger with default null device

bf4:
Based on a discussion in the amserializers slack
It's not used anywhere in this PR, but it could be...

### DIFF
--- a/lib/active_model/serializer/railtie.rb
+++ b/lib/active_model/serializer/railtie.rb
@@ -1,6 +1,12 @@
 require 'rails/railtie'
 module ActiveModel
   class Railtie < Rails::Railtie
+    initializer 'active_model_serializers.logger' do
+      ActiveSupport.on_load(:action_controller) do
+        ActiveModelSerializers.logger = ActionController::Base.logger
+      end
+    end
+
     initializer 'generators' do |app|
       app.load_generators
       require 'generators/serializer/resource_override'
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -1,4 +1,12 @@
+require 'logger'
+require 'active_model'
+require "active_support/railtie"
+require 'action_controller'
+require "action_controller/railtie"
 module ActiveModelSerializers
+  mattr_accessor :logger
+  self.logger = Rails.logger || Logger.new(IO::NULL)
+
   module_function
 
   def silence_warnings
@@ -10,9 +18,6 @@ module ActiveModelSerializers
   end
 
 end
-
-require 'active_model'
-require 'action_controller'
 
 require 'active_model/serializer'
 require 'active_model/serializable_resource'
--- a/test/logger_test.rb
+++ b/test/logger_test.rb
@@ -1,0 +1,19 @@
+require 'test_helper'
+
+class ActiveModelSerializers::LoggerTest < Minitest::Test
+
+  def test_logger_is_set_to_action_controller_logger_when_initializer_runs
+    assert_equal ActiveModelSerializers.logger, ActionController::Base.logger
+  end
+
+  def test_logger_can_be_set
+    original_logger = ActiveModelSerializers.logger
+    logger = Logger.new(STDOUT)
+
+    ActiveModelSerializers.logger = logger
+
+    assert_equal ActiveModelSerializers.logger, logger
+  ensure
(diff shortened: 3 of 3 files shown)


joaomdmoura:
I don't remember this @bf4 can you refresh me? :smile:

bf4:
ref: https://amserializers.slack.com/archives/general/p1440623581000150


> beauby [4:13 PM] 
What's the preferred way to log stuff from AMS?

> joaomdmoura [4:14 PM] 
I think do no logging right now, what is the specifics?

> beauby [4:15 PM] 
@joaomdmoura: Adding the warning when hitting a resource without a defined serializer

> joaomdmoura [4:15 PM] 
you can have some method the tries to use rails log, if it does not find it then use a fallback

> beauby [4:16 PM] 
rails uses ruby's Logger class, no?

> joaomdmoura [4:16 PM] 
I think so, but I’m not sure if it implements something useful on top of it

> joaomdmoura [4:17 PM]
I think it also enables you can change what to log and where, so would be nice to use it

> beauby [4:17 PM] 
Nah, I just checked it, the ActiveSupport extension doesn't add much

> joaomdmoura [4:17 PM] 
really?

> beauby [4:18 PM] 
Yeah, it does add a method to silence a given level

> joaomdmoura [4:18 PM] 
well if that is the case then we case use ruby one, just maek sure to share what you found on the PR so we can check it as well

> beauby [4:18 PM] 
but the levels are defined in ruby core anyways

> bf4 [4:18 PM] 
rails uses a buffered logger or whatever

> beauby [4:18 PM] 
http://guides.rubyonrails.org/active_support_core_extensions.html#extensions-to-logger

> bf4 [4:18 PM] 
just

```
  class Serializer
   WINDOZE  = /win(32|64)/ =~ RUBY_PLATFORM
  DEV_NULL = (WINDOZE ? 'nul'      : '/dev/null')
  # example code. do not use
  @@logger = Rails.logger || Logger.new(DEV_NULL)
  def self.logger
    @@logger
  end
```

> this is what I did in metric_fu https://github.com/metricfu/metric_fu/blob/194f401fc0a0fdef71354de5d133260813e791de/lib/metric_fu/logger.rb for a custom logger (edited)

> beauby [4:23 PM] 
thanks @bf4, I'll look into it

> bf4 [4:23 PM] 
what you'd really want to do is probably have it default to dev/null, but in the rails engine, have an initializer that sets it to Rails.logger or something custom

> beauby [4:24 PM] 
@bf4: Hmm why /dev/null instead of stderr?

> bf4 [4:25 PM] 
oh, because it's super annoying to have output you're not interested int, and STDOUT and STDERR do that.. logs/ams.log could grow too big.  I've never gotten built-in log-rotation to work

> bf4 [4:25 PM]
especially in tests

> beauby [4:27 PM] 
Hmm but if you log to /dev/null by default, how is anyone ever going to see it?

> bf4 [4:29 PM] 
it's opt-in unless it's integrated with Rails, in which case it's configurable

> bf4 [4:30 PM]
so, the most common use-cases are covered.

> bf4 [4:30 PM]
with Rails, you'd probably want to use the TaggedLogger if available so we can tag it as coming from AMS.  There's probably a good example of this somewhere out there

bf4:
Also, ref: https://github.com/rails-api/active_model_serializers/pull/1099#discussion_r39120690

- https://github.com/rails/rails/blob/3e55fd755ac9c1d1f84f8dadd42112e2ee592a3a/activejob/lib/active_job/logging.rb#L19-L36 
- https://github.com/rails/rails/blob/3e55fd755ac9c1d1f84f8dadd42112e2ee592a3a/activejob/lib/active_job/railtie.rb#L9-L10 

```ruby
# railtie
    initializer 'active_job.logger' do
      ActiveSupport.on_load(:active_job) { self.logger = ::Rails.logger }
    end

# https://github.com/rails/rails/blob/7a3ca69959e312a215d5e8144fca79a08654fd89/activejob/lib/active_job/base.rb


    ActiveSupport.run_load_hooks(:active_job, self)
```

beauby:
This would be cool, so that we could log stuff such as:
+ trying to `include` an undefined relationship (for instance when misspelling `author` with `authr` and such)
+ trying to serialize a related resource for which there exists no serializer (which currently outputs the whole jsonified object)

NullVoxPopuli:
:+1:

PR #1094, opened 2015-08-28
Having `attribute :name` in a serializer causes exceptions

whatthewhat:
Having `attribute :name` in a serializer breaks with `NoMethodError: undefined method 'name' for #<AuthorSerializer:0x007ffe5afe6810>`.
But `attributes :id, :name` works..

I've changed one of the test serializers to demonstrate the issue.

I'll try to find time later to investigate some more, unless there is not an obvious fix someone could point me to.

### DIFF
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -134,7 +134,8 @@ end
 
 AuthorSerializer = Class.new(ActiveModel::Serializer) do
   cache key:'writer', skip_digest: true
-  attributes :id, :name
+  attribute :id
+  attribute :name
 
   has_many :posts, embed: :ids
   has_many :roles, embed: :ids


beauby:
Could you state the version and adapter that lead to the error?

whatthewhat:
@beauby should be clear from the tests when (if) they fail.

The error reproduces on master, the serializer is super simple:

```
class CompanySerializer < ActiveModel::Serializer
  attribute :id
  attribute :name
end
```

beauby:
Sure, but which adapter? (json, flatten_json, json_api?)

whatthewhat:
@beauby oh, sorry :) json

whatthewhat:
Looking at the test failures seems like it's universal for all adapters

beauby:
Investigating!

beauby:
Found the cause, but no fix yet. 
In [attributes](https://github.com/rails-api/active_model_serializers/blob/64168cbecd6dfff99d81e4ab5444b36ffe219beb/lib/active_model/serializer.rb#L61) we call `method_defined?(attr)`, and in [attribute](https://github.com/rails-api/active_model_serializers/blob/64168cbecd6dfff99d81e4ab5444b36ffe219beb/lib/active_model/serializer.rb#L73) we call `respond_to?(key, false)`. I'm not sure about the rationale behind it, but replacing `respond_to?(key, false)` with `method_defined?(key)` messes with the ability of overriding the `id` attribute.
@joaomdmoura @bf4 ideas?

beauby:
Hmm actually it's no surprise overriding `id` fails in that case because of [this](https://github.com/rails-api/active_model_serializers/blob/64168cbecd6dfff99d81e4ab5444b36ffe219beb/lib/active_model/serializer.rb#L149). @joaomdmoura what's the reason for this `id` method?

kayhide:
`#method_defined?` is for a class and looks for a instance method.
`#respond_to?` is for a object and looks for a method of the receiver object.

So [this](https://github.com/rails-api/active_model_serializers/blob/64168cbecd6dfff99d81e4ab5444b36ffe219beb/lib/active_model/serializer.rb#L73) returns whether the class method exists or not.
This does not check instance methods.

In this case, we are interested in instance methods, so `#method_defined?` is good to do that.

beauby:
Thanks a lot for the clarification @kayhide!

PR #1096, opened 2015-08-28
Fix definition of serializer attributes with multiple calls to `attri…

### LINKED ISSUE #1091: Response with multiple lists
Hi,

is there a way to specify something like this:

render json: {projects: @projects, each_serializer: ProjectSerializer, clients: clients, each_serializer: ProjectSerializer}, status: :ok

beauby:
…bute` instead of one single call to `attributes`.

Fixes #1094.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -70,7 +70,7 @@ module ActiveModel
       ActiveModelSerializers.silence_warnings do
         define_method key do
           object.read_attribute_for_serialization(attr)
-        end unless respond_to?(key, false) || _fragmented.respond_to?(attr)
+        end unless (key != :id && method_defined?(key)) || _fragmented.respond_to?(attr)
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/active_model/serializer.rb
```diff
@@ -70,7 +70,7 @@ def self.attribute(attr, options = {})
       ActiveModelSerializers.silence_warnings do
         define_method key do
           object.read_attribute_for_serialization(attr)
-        end unless respond_to?(key, false) || _fragmented.respond_to?(attr)
+        end unless (key != :id && method_defined?(key)) || _fragmented.respond_to?(attr)
```
bf4: looks like this should be extracted to a helper..

bf4:
:+1:

whatthewhat:
:+1:

lib/active_model/serializer.rb
beauby: While we're at it, should `_fragmented.respond_to?(attr)` not be `_fragmented.respond_to?(key)`? I haven't delved into the caching yet.

joaomdmoura:
Nice @beauby, I'm merging it, Tks! :smile:

PR #967, opened 2015-06-24
Document serialization interface and what to test

bf4:
Extracted from #954

ref: https://github.com/rails-api/active_model_serializers/issues/878#issuecomment-96073030

PR #1099, opened 2015-08-29
Bring back assert_serializer

### LINKED ISSUE #1011: Testing Infrastrucutre
I can't tell where (`git log --` returns nothing for that file), but at some point after 0.9.3 the [Rails test case integration code](https://github.com/rails-api/active_model_serializers/blob/97c9c5084a1531e086848e4d75fc90145fba9b83/lib/action_controller/serialization_test_case.rb) was removed.

I can't find where the equivalent code is in the current revision, so this breaks controller unit test like 

```ruby
assert_serializer :photos 
```

when trying to update, even though that behavior has not changed. Am I just not seeing where this code was moved to?

maurogeorge:
Hi guys,

closes #1011

The `assert_serializer` test helper was removed at some point.

This patch brings back the `assert_serializer` test helper. This is the last
revision[1] that has the helper. The original helper was used as base.

[1]: https://github.com/rails-api/active_model_serializers/tree/610aeb2e9297fa31b8d561f0be9a4597f0258f8c


bf4:
nice

trek:
@maurogeorge looks like this needs a rebase.

maurogeorge:
@trek rebased. Thanks for poiting

README.md
```diff
@@ -335,6 +335,34 @@ class PostSerializer < ActiveModel::Serializer
 end
 ```
 
+## Test helpers
```
joaomdmoura: @maurogeorge could you move this to an article on our [new docs](https://github.com/rails-api/active_model_serializers/tree/master/docs) and maybe but a reference here in the README?
Maybe into the "How to" section explaining hot to assert serializer.

lib/active_model/serializable_resource.rb
```diff
@@ -74,6 +75,14 @@ def serializer?
       use_adapter? && !!(serializer)
     end
 
+    def instrument
```
joaomdmoura: I understand the reason to use `instrument` name here, but would be nicer if we could use something more meaningful mostly to help other ppl that might jumping into the code :smile:  maybe something like `notify_active_support` idk make yourself comfortable to come up if something better :grin:

bf4: would a logger be a useful place to integrate with active support notifications? https://github.com/rails-api/active_model_serializers/pull/1089

joaomdmoura:
@maurogeorge this looks nice, thank you for your contribution :smile: we made some comments on it, can you check it?

maurogeorge:
Hi @joaomdmoura thanks for the review. Updated with the changes :smile:

lib/active_model/serializable_resource.rb
maurogeorge: Hi @bf4 thanks for te review.
Could you please provide more info about it? What kind of integration you are thinking? Maybe you can update the #1089 with more info.

maurogeorge:
I think that the broken test on the AppVeyor is not related with this PR, this test is broken on the master too.

lib/active_model_serializers.rb
```diff
@@ -18,11 +18,13 @@ def silence_warnings
 require 'active_model/serializer/version'
 
 require 'action_controller/serialization'
+require 'action_controller/serialization_test_case'
 ActiveSupport.on_load(:action_controller) do
   include ::ActionController::Serialization
   ActionDispatch::Reloader.to_prepare do
     ActiveModel::Serializer.serializers_cache.clear
   end
+  ActionController::TestCase.send(:include, ::ActionController::SerializationAssertions)
```
bf4: I'd rather put something like this in `active_model_serializers/test.rb`, no? like

- https://github.com/rails/rails/blob/master/activejob/lib/active_job/test_case.rb

```ruby
require 'active_support/test_case'

module ActiveJob
  class TestCase < ActiveSupport::TestCase
    include ActiveJob::TestHelper
  end
end
```

- https://github.com/rails/rails/blob/3e36db4406beea32772b1db1e9a16cc1e8aea14c/activejob/lib/active_job.rb#L35-L36

```ruby
module ActiveJob

  autoload :TestCase
  autoload :TestHelper
end
```
then

```ruby
class ActiveJobTestCaseTest < ActiveJob::TestCase
```
- https://github.com/rails/rails/blob/master/activejob/lib/active_job/test_helper.rb

lib/active_model/serializable_resource.rb
```diff
@@ -18,6 +18,7 @@ def initialize(resource, options = {})
     # returns the contents of the block
     def self.serialize(resource, options = {})
       serializable_resource = SerializableResource.new(resource, options)
+      serializable_resource.notify_active_support
       if block_given?
         yield serializable_resource
       else
```
bf4: move code into `initialize` as we're getting rid of `serialize` https://github.com/rails-api/active_model_serializers/pull/1129

event_name should be `active_model_serializers.serialize`, no?

This should also probably go in a logger like in https://github.com/rails/rails/blob/3e55fd755ac9c1d1f84f8dadd42112e2ee592a3a/activejob/lib/active_job/logging.rb#L19-L36 / https://github.com/rails/rails/blob/3e55fd755ac9c1d1f84f8dadd42112e2ee592a3a/activejob/lib/active_job/railtie.rb#L9-L10 and perhaps https://github.com/rails-api/active_model_serializers/pull/1089

maurogeorge: Thanks for the review @bf4 :heart: 

About the event name I think it is right, I followed the convention `action.class_name` is the same that is used in the Active Job "perform_start.active_job" and in the [Rails guides](http://edgeguides.rubyonrails.org/active_support_instrumentation.html).

About the logger I think we can expose this in a logger too, but in the moment I am only using the instrumentation to make to use in the tests. But yes, we can do this, I think this belongs to other PR and not this. What you think?

lib/active_model_serializers.rb
maurogeorge: Thanks for the review :heart: 

I try something in this line, but got no sucess :disappointed:  since Active Job defines the `ActiveJob::TestCase` in our case we do not want to define our `TestCase` class but mixes the helper in the `ActionController::TestCase`.

I followed the same idea we have in the in the [`ActionController::TestCase`](https://github.com/rails/rails/blob/e16afe61abd78c55f80752ca020b90d59ae1940f/actionpack/lib/action_controller/test_case.rb#L298-L300). I renamed/move the assertions module to a better place.

Let me know what you think

maurogeorge:
@joaomdmoura @bf4 sorry for the delay I was involved with other things :disappointed: 

I update the code following the @bf4 last comments. I waiting for your review guys :heart:

trek:
Appears to be more merge conflicts :-/

maurogeorge:
@trek I solved the conflicts, thanks for the report.

lib/active_model_serializers.rb
```diff
@@ -44,6 +44,7 @@ def silence_warnings
 require 'active_model/serializer'
 require 'active_model/serializable_resource'
 require 'active_model/serializer/version'
+require 'active_model/serializer/assertions'
```
bf4: This shouldn't be eager-loaded here as it's a test-only concern

test/assertions_test.rb
```diff
@@ -0,0 +1,70 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    class AssertionsTest < ActionController::TestCase
+      class MyController < ActionController::Base
+        def render_using_serializer
```
bf4: no need to do all controller tests if the active code is in the serializable resource.  (Also, controller tests should be in the action_controller folder with the others https://github.com/rails-api/active_model_serializers/tree/master/test/action_controller )

lib/active_model/serializer/assertions.rb
```diff
@@ -0,0 +1,104 @@
...
+                               matches_regexp?(expectation)
+                             when NilClass
+                               matches_nil?
+                             else
+                               fail ArgumentError, 'assert_serializer only accepts a String, Symbol, Regexp, ActiveModel::Serializer, or nil'
+                             end
+        assert(matches_serializer, msg)
+      end
+
+      private
```
bf4: I'd like to go over some of this with you.  When do you have time?

lib/active_model/serializable_resource.rb
```diff
@@ -9,6 +9,7 @@ def initialize(resource, options = {})
       @resource = resource
       @adapter_opts, @serializer_opts =
         options.partition { |k, _| ADAPTER_OPTION_KEYS.include? k }.map { |h| Hash[h] }
+      notify_active_support
```
bf4: not sure if it's any simpler, but we already provide a logger, so you could set the logger to use a stringio and capture the output.

in any case, I'm unsure if this is the best place in this class for this method, but don't have a better idea yet

bf4:
fwiw, here's some RSpec stuff I wrote for rc1 https://github.com/rails-api/active_model_serializers/issues/878#issuecomment-96073030 that I've been meaning to work into ams in https://github.com/rails-api/active_model_serializers/pull/967

lib/active_model_serializers.rb
maurogeorge: Thanks for the feedback.

I think if we try to autoload `ActiveModel::Serializer::Assertions`, this will be loaded when we mix this on the `ActionController::TestCase`, in this same file, so this wil be loaded too soon.
Do you have some idea how we can lazy load this?

test/assertions_test.rb
maurogeorge: I followed the same Idea of the [serialization_test.](https://github.com/rails-api/active_model_serializers/blob/master/test/action_controller/serialization_test.rb). In fact the code I am testing here is in the `lib/active_model/serializer/assertions.rb` not sure if this goes in the `test` folder or in the `action_controller` I can move if you prefer :smile:

lib/active_model/serializable_resource.rb
maurogeorge: Hi, I take a look at the code and we have a logger but actually we do not log anything, I am right or miss something?

Since Rails provides [instrumentation](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) I think we can invert this logic of capturing log output, instead we can use the instrumentation to generate the events and subscribe to these events like in this test helper, this way we do not need to open and parse log files.

And as a bonus we can generate the logs using instrumentation with [`ActiveSupport::Notifications`](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) and the [`ActiveSupport::LogSubscriber`](http://api.rubyonrails.org/classes/ActiveSupport/LogSubscriber.html). 
Active Job does [something like that using `ActiveSupport::Notifications`](https://github.com/rails/rails/blob/d2ef471b849b28c6962e4ed784fbec42aae9f9ec/activejob/lib/active_job/logging.rb#L19-L38) and [`ActiveSupport::LogSubscriber`](https://github.com/rails/rails/blob/d2ef471b849b28c6962e4ed784fbec42aae9f9ec/activejob/lib/active_job/logging.rb#L54-L119).

Some references about intrumentation are [this post](http://www.railstips.org/blog/archives/2013/01/23/an-instrumented-library-in-30-lines/) and in the [Crafting Rails Application](https://pragprog.com/book/jvrails2/crafting-rails-4-applications) have a chapter about it.

This makes sense to you?

maurogeorge:
@bf4 thanks for the review :100: , I answered each directly.

Pretty cool the RSpec stuff.

In my opinion I think we can try to generate test helpers like this and [this](https://github.com/rails-api/active_model_serializers/pull/1270) this way we do not depend on RSpec, and the people that are using RSpec can mix this module and use with RSpec too, like we do for example with the [`ActiveSupport::Testing::TimeHelpers`](http://blog.pivotal.io/labs/labs/goodbye-timecop-hello-timehelpers). 

What you think?

bf4:
Reading comments via email now (no code) but sounds good

B mobile phone

maurogeorge:
@bf4 @joaomdmoura @beauby I update the code could you guys please review? Thanks

maurogeorge:
@bf4 @joaomdmoura @beauby I updated to keep all tests green. Thanks

trek:
@bf4 I don't suppose this could get merged or closed? Poor @maurogeorge has to keep rebasing to keep it current with master.

beauby:
This looks good to me. @bf4 @joaomdmoura?

lib/active_model_serializers/test/serializer.rb
```diff
@@ -0,0 +1,122 @@
...
+      #  assert_serializer %r{\APost.+\Z}
+      #
+      #  # assert that no serializer was rendered
+      #  assert_serializer nil
+      #
+      def assert_serializer(expectation, message = nil)
+        @assert_serializer.expectation = expectation
+        @assert_serializer.message = message
+        @assert_serializer.response = response
+        assert(@assert_serializer.matches?, @assert_serializer.message)
```
bf4: So, people using RSpec would need to define `assert(expected, actual)` and any library not using `ActionController::TestCase` would also need to include `ActiveModelSerializers::Test::Serializer`

something like:

```ruby
module AMSAssertSerializer
  extend ActiveSupport::Concern
  included do
    include ActiveModelSerializers::Test::Serializer
  end
  def assert(expected, actual)
    expect(actual).to eq(expected)
  end
end
RSpec.configuration.include AMSAssertSerializer, type: :controller
```

test/active_model_serializers/test/serializer_test.rb
```diff
@@ -0,0 +1,71 @@
+require 'test_helper'
+require 'rails-controller-testing' if Rails::VERSION::MAJOR >= 5
```
bf4: What's in here that needed `'rails-controller-testing'`?

bf4:
Looking good.  I had a few questions above and I'm ok to merge.

@trek @maurogeorge is this something you'd like to use? Just confirming because I kind of [agree with @bolshakov that it doesn't seem very useful](https://github.com/rails-api/active_model_serializers/issues/1011#issuecomment-127608121). But, given that I'm thinking/saying this 4 months after the PR was opened, if people want to use it, I'll merge it.

lib/active_model_serializers.rb
```diff
@@ -51,6 +51,7 @@ def silence_warnings
 require 'active_model/serializer'
 require 'active_model/serializable_resource'
 require 'active_model/serializer/version'
+require 'active_model_serializers/test/serializer'
```
bf4: IMHO, this should only be loaded in the test environment.  Sell me on always loading it, and why it won't cause issues in production (including the railtie `ActionController::TestCase.send(:include, ActiveModelSerializers::Test::Serializer`)

- ref to [previous comment lost in the diffs](https://github.com/rails-api/active_model_serializers/pull/1099#discussion_r42064156)

bf4: Oh, and https://github.com/rails-api/active_model_serializers/pull/1270#discussion_r47167126

lib/active_model_serializers/test/serializer.rb
maurogeorge: Thanks for the review.

I don't think we need to care with RSpec and other libs at the moment, [Action Mailer uses Minitest methods](https://github.com/rails/rails/blob/68af99c942bf00f3a53df72555d2d5e3b94978cf/actionmailer/lib/action_mailer/test_helper.rb#L37) and [Active Job too](https://github.com/rails/rails/blob/68af99c942bf00f3a53df72555d2d5e3b94978cf/activejob/lib/active_job/test_helper.rb#L71). 

I think we can follow the same rationale, we are providing Minitest test helpers.

test/active_model_serializers/test/serializer_test.rb
maurogeorge: [This test uses](https://github.com/maurogeorge/active_model_serializers/blob/patch-02/test/active_model_serializers/test/serializer_test.rb#L67), this test was based on the original codebase of the test helper.

bf4: @maurogeorge that's kind of hilarious because it was removed from rails due to it being too intimate a test https://github.com/rails/rails/issues/18950
We can remove that test and the dependency.

Do you have a link to the file you're basing this on? Would be nice to see.

maurogeorge: @bf4 Yeah I know about the Rails decision removing this test helpers.
But I am not sure if we can remove this test, I think this was added to test the notification was not overwrite, based on the description of the test.

This is the last version of the file https://github.com/rails-api/active_model_serializers/blob/610aeb2e9297fa31b8d561f0be9a4597f0258f8c/test/integration/action_controller/serialization_test_case_test.rb#L67 I found.

maurogeorge:
@trek @bf4 @beauby I updated the code, based on the last comments. 

The Travis is broken, but I don't think this is related with this patch, It seems that these tests already be [broken on master](https://travis-ci.org/rails-api/active_model_serializers/jobs/98193857).

bf4:
@maurogeorge I have a PR to fix master, but wanted some eyes on it https://github.com/rails-api/active_model_serializers/pull/1384

You're right, not related

test/active_model_serializers/test/serializer_test.rb
bf4: Ok, but we could subscribe to our own notifications here, right?

bf4: I can look at that in a followup PR if you'd prefer. I think that'd be nicer at this point :)

lib/active_model_serializers/test/serializer.rb
bf4: Yeah, I know.  :)

test/active_model_serializers/test/serializer_test.rb
maurogeorge: @bf4 I am ok with that, I think you can follow the same idea you did on this https://github.com/rails-api/active_model_serializers/pull/1322 squash all these commits into a single one, create a new PR and close this one, and @mention me so I can follow up :smile:

trek:
Yes, this is something we'd use in unit tests for controllers. We have separate unit tests for the serializers themselves (so ensure they match the a described JSON Schema), but are also currently reusing this strategy for controller tests, which seems a bit odd.

bf4:
@trek so you're saying you used but don't think it's a good idea? (ref what's I've done in the past)

maurogeorge:
Closing this PR, since the code was moved to the new one https://github.com/rails-api/active_model_serializers/pull/1390

trek:
@bf4 no, I'm saying schema validation in both unit tests for a serializer and as part of a unit test of a controller means you get twice the failures.

bf4:
@trek so you basically if someone is using [`assert_json_schema`](https://github.com/rails-api/active_model_serializers/pull/1270), there's no need for `assert_serializer`? Makes sense.  Do you think both option should exist i.e. be supported by AMS and its people?

trek:
@bf4 I think which one you use will depend on your testing strategy. We've moved on from wanting `assert_serializer` just due to the need to get coding. But, now we get multiple test failures if someone changes the schema accidentally.

Issue #762, opened 2014-12-19
has_one with embed: :id still loads the associated model


YoranBrondsema:
I have the following scenaro:

```
class Child < ActiveRecord::Base
  belongs_to :parent
end

class Parent < ActiveRecord::Base
  has_one :child
end
```

In `ChildSerializer`, if I do
```
class ChildSerializer < ActiveModel::Serializer
  has_one :parent, embed: :id
end
```

then it does a query to fetch the `Parent` model from the database, even though only the `id` attribute is needed (which equals to the foreign key `parent_id` on the `Child` model). So this database query is kind of redundant.

I can fix it by doing
```
class ChildSerializer < ActiveModel::Serializer
  attribute :parent_id
end
```

which gives me the same JSON without the extra query. But I don't like it that much because by using `attribute` I lose the semantics of the association. Readability is important.

I don't know if it's doable to detect such cases where a query on the associated model is not needed, and prevent it. Maybe someone who's more familiar with the AMS code can give his opinion, then I don't mind doing a PR that solves it.

YoranBrondsema:
@kurko I'm using 0.8.3. While I figure out how I can write a failing test, I created a sample app that demonstrates the behavior: https://github.com/YoranBrondsema/ams-test-case.

toreriklinnerud:
I'm seeing the same thing on 0.8.3. Is this likely fixed on newer versions or should I try to submit a fix?

joaomdmoura:
@toreriklinnerud I think it's fixed on master already, but we don't have a test covering that, maybe open a PR with this test would be a good start :smile:

PR #1100, opened 2015-08-29
belongs_to causes unnecessary db hits

ssendev:
failing test for #762 

also fails with `Adapter::Json` but i'm not sure what the expected hash would look like (embed ids appears to have no effect with json adapter blog name still appears in the resulting hash (was this changed in 0.10?)).

### DIFF
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -126,6 +126,26 @@ module ActiveModel
         assert expected_association_keys.include? :site
       end
 
+      class BelongsToTestPostSerializer < ActiveModel::Serializer
+        belongs_to :blog
+      end
+
+      def test_belongs_to_doesnt_load_record
+        post = Post.new
+        post.blog_id = 'blog'
+
+        class << post
+          def blog
+            fail 'should use blog_id'
+          end
+        end
+
+        actual = serializable(post, adapter: :json_api, serializer: BelongsToTestPostSerializer).as_json
+        expected = { data: { id: 'post', type: 'posts', relationships: { blog: { data: { id: 'blog', type: 'blogs' } } } } }
+
+        assert_equal expected, actual
+      end
+
       class InlineAssociationTestPostSerializer < ActiveModel::Serializer
         has_many :comments
         has_many :comments, key: :last_comments do


test/serializers/associations_test.rb
```diff
@@ -145,6 +145,24 @@ def test_associations_custom_keys
         assert expected_association_keys.include? :writer
         assert expected_association_keys.include? :site
       end
+
+      def test_belongs_to_dosnt_load_record
```
bf4: typo: doesnt

test/serializers/associations_test.rb
```diff
@@ -145,6 +145,24 @@ def test_associations_custom_keys
...
+        post = Post.new({ id: 997 })
+
+        def post.blog_id
+          996
+        end
+        def post.blog
+          fail "should use blog_id"
+        end
```
bf4: I think you could also 

```ruby
class << post
  def blog_id
    996
  end
  def blog
    fail "should use blog_id"
  end
end
```

if I recall my meta (crazy) programming...

test/serializers/associations_test.rb
```diff
@@ -145,6 +145,24 @@ def test_associations_custom_keys
...
+
+        post_serializer = Class.new(ActiveModel::Serializer) do
+          belongs_to :blog, embed: :ids
+        end
+
+        adapter = ActiveModel::Serializer::Adapter::JsonApi.new(post_serializer.new(post))
```
bf4: probably better as `ActiveModel::SerializableResource.new(post, each_serializer: post_serializer, adapter: :json_api).serializable_hash` (assuming I typed that correctly)

bf4:
:+1: Nice

test/serializers/associations_test.rb
ssendev: almost worked.

i took the code from [here](https://github.com/rails-api/active_model_serializers/blob/master/test/serializers/attribute_test.rb#L54-L55)

test/serializers/associations_test.rb
```diff
@@ -145,6 +145,26 @@ def test_associations_custom_keys
...
+        class << post
+          def blog_id
+            996
+          end
+          def blog
+            fail "should use blog_id"
+          end
+        end
+
+        post_serializer = Class.new(ActiveModel::Serializer) do
```
beauby: I would call the class `PostSerializer` instead of `post_serializer` (although I know you took the code from somewhere else, but I think this is more 'standard').
Also, I would write it like that:
```ruby
  belongs_to :blog, embed: :ids
end
```
as I find it less confusing, but here people might disagree.

ssendev: tried changing it but it seems like this notation is not accepted in a method.

beauby: Very true indeed, I overlooked that part.

test/serializers/associations_test.rb
```diff
@@ -145,6 +145,26 @@ def test_associations_custom_keys
...
+          belongs_to :blog, embed: :ids
```
beauby: The `embed: :ids` here is useless though, as by JSONAPI standard, the default is to include just Resource Identifiers (i.e. an object `{ data: { id: 1, type: 'post' } }`), and also, the `embed` option does not seem to exist anymore (as searching the codebase for `embed` yields only results in tests and comments.

beauby:
After giving it a thought, I am not sure we can avoid this "unnecessary db hit". What happens if we have `belongs_to :author` in the serializer, and `belongs_to :author, class_name: "User"` in the model? Then I fear AMS wouldn't know how to set the `type` property of the Resource Identifier.

ssendev:
`belongs_to :author, type: "users"` comes to mind. i mean for `attributes` there's `key`

test/serializers/associations_test.rb
bf4: You could

```ruby
post_serializer = Class.new(ActiveModel::Serializer)
Object.const_set(:PostSerializer, post_serializer)
PostSerializer.whatever
```

e.g. see https://github.com/rails-api/active_model_serializers/blob/18e5812c9310b2b0240950a4a0d92e2200aeb17c/test/serializers/adapter_for_test.rb#L93-L144

beauby:
`key` for attributes is something different: it exposes an existing attribute under a different name (in the serialized document).
I have a seemless fix for that extra db hit, I just need #1103 to be merged first.

test/serializers/associations_test.rb
ssendev: very interesting. but in the referenced test it's required since that's what's being tested. here it seems like it would just make the test more complex for "no" benefit. especially considering there already is a `PostSerializer` which would have to be restored.

beauby: Agreed

Le lundi 31 août 2015, ssendev <notifications@github.com> a écrit :

> In test/serializers/associations_test.rb
> <https://github.com/rails-api/active_model_serializers/pull/1100#discussion_r38342653>
> :
>
> > @@ -145,6 +145,26 @@ def test_associations_custom_keys
> >          assert expected_association_keys.include? :writer
> >          assert expected_association_keys.include? :site
> >        end
> > +
> > +      def test_belongs_to_doesnt_load_record
> > +        post = Post.new({ id: 997 })
> > +
> > +        class << post
> > +          def blog_id
> > +            996
> > +          end
> > +          def blog
> > +            fail "should use blog_id"
> > +          end
> > +        end
> > +
> > +        post_serializer = Class.new(ActiveModel::Serializer) do
>
> very interesting. but in the referenced test it's required since that's
> what's being tested. here it seems like it would just make the test more
> complex for "no" benefit. especially considering there already is a
> PostSerializer which would have to be restored.
>
> —
> <https://github.com/rails-api/active_model_serializers/pull/1100/files#r38342653>
> .
>


-- 
Lucas Hosseini
lucas.hosseini@gmail.com

joaomdmoura: I'm okay with it.j

test/serializers/associations_test.rb
```diff
@@ -145,6 +145,26 @@ def test_associations_custom_keys
...
+
+      def test_belongs_to_doesnt_load_record
+        post = Post.new({ id: 997 })
+
+        class << post
+          def blog_id
+            996
+          end
+          def blog
+            fail "should use blog_id"
```
joaomdmoura: haha nice trick! :+1:

joaomdmoura:
Thank you @ssendev this is indeed something that we should address, we should not do unnecessary queries or anything that might slow down the apps performance.

As @beauby this might be a little bit tricky because of the example:
`belongs_to :author, class_name: "User"`

What you suggest might be a way to make it work `belongs_to :author, type: "users"` but I think we can try to discuss and come up with a more elegant solution. @beauby can you open a `[WIP]` PR with your solution, meanwhile I'll check #1103

ssendev:
an alternative would be to specify a serializer and require `json_api_type` to be defined on it, or guess it from the serializers name . both solutions wouldn't be able to handle polymorphic relations but then there is probably an `author_type` column which could be consulted.

sandstrom:
This would be a godsend. Currently using AMS and need manual work-arounds to avoid these costly DB-lookups.

@ssendev has a great point, simply use reflection to inspect the relationships of the model. Specifying a type manually would be rare and only needed for polymorphic relationships.

I don't use ActiveRecord, so not 100% sure but I think `Author.reflect_on_association(:books).klass` works[1]. Mongoid, another popular ORM, also has good reflection support.

[1] https://github.com/rails/rails/blob/master/activerecord/lib/active_record/reflection.rb

bf4:
So sorry we lost track of this!

NullVoxPopuli:
needs rebase, I'd like to get this merged soonish

ssendev:
this fails with

```diff
-{:data=>{:id=>"post", :type=>"posts", :relationships=>{:blog=>{:data=>{:id=>"blog", :type=>"blogs"}}}}}
+{:data=>{:id=>"post", :type=>"posts", :relationships=>{:blog=>{:data=>nil}}}}
```

is this expected or am i missing something?

NullVoxPopuli:
@ssendev is that a conflict in the rebase? or is that the result of a test? (I guess I don't know what context that is in)

ssendev:
https://travis-ci.org/rails-api/active_model_serializers/jobs/132809729#L323-L324

sandstrom:
@ssendev I'd guess it's a test-failure. It says `--- expected` and `+++ actual`

NullVoxPopuli:
it is yeah.

my initial hunch is that something is off, because specifying a relationship with a data of nil seems weird. I'm going to check the json api spec, and see what is says about unspecified relationships. 

It could be that we want to just omit that relationship entry.

ssendev:
but the relationship is specified it just doesn't get serialized

NullVoxPopuli:
ah, yes, I was looking at the wrong test.

NullVoxPopuli:
I think I was confused earlier, and thought this was a fix for the problem. ha. I should not be on github when I'm tired.

NullVoxPopuli:
anywho, I'm looking at fixing this

NullVoxPopuli:
Closing in favor of #1750, as I added some notes in the code about what's happening where.

PR #1102, opened 2015-08-30
Remove remains of `embed` option.

beauby:
As stated in #1100, the `embed` association option is only mentioned in a comment and in tests, but nowhere else in the codebase.


joaomdmoura:
Nice, merging it :D

PR #1103, opened 2015-08-31
Move `id` and `json_api_type` methods from `Serializer` to `JsonApi`.

beauby:
Refactored `add_resource_relationship` along the way.


lib/active_model/serializer.rb
```diff
@@ -146,18 +146,6 @@ def json_key
...
-
-    def json_api_type
-      if config.jsonapi_resource_type == :plural
-        object.class.model_name.plural
-      else
-        object.class.model_name.singular
-      end
-    end
-
     def attributes(options = {})
```
beauby: These two methods clearly did not belong there, as they are JSONAPI specific.

lib/active_model/serializer.rb
```diff
@@ -146,18 +146,6 @@ def json_key
...
-    end
-
-    def json_api_type
-      if config.jsonapi_resource_type == :plural
-        object.class.model_name.plural
-      else
-        object.class.model_name.singular
-      end
-    end
-
```
beauby: These two methods clearly did not belong there, as they are JSONAPI specific.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -44,19 +44,15 @@ def fragment_cache(cached_hash, non_cached_hash)
...
-          if serializer && serializer.object
-            resource[:relationships][name][:data] = { type: serializer.json_api_type, id: serializer.id.to_s }
-          end
+        def resource_identifier(serializer)
+          type = if ActiveModel::Serializer.config.jsonapi_resource_type == :plural
+                   serializer.object.class.model_name.plural
+                 else
+                   serializer.object.class.model_name.singular
+                 end
+          id = serializer.object.id.to_s
```
bf4: This might be easier to read as a case statement

```ruby
type = 
  case
  when  ActiveModel::Serializer.config.jsonapi_resource_type == :plural
    serializer.object.class.model_name.plural
  else
    serializer.object.class.model_name.singular
  end
```

maybe `resource_type` and `resource_id` as the local variable names?  

This method seems like there's a latent object in here that transforms a single resource into a serializable hash

Nice on the naming  http://jsonapi.org/format/#document-resource-identifier-objects :)

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -100,15 +96,12 @@ def attributes_for_serializer(serializer, options)
...
 
           cache_check(serializer) do
             attributes = serializer.attributes(options)
+            attributes.delete(:id)
 
-            result = {
-              id: attributes.delete(:id).to_s,
-              type: attributes.delete(:json_api_type)
-            }
+            result = resource_identifier(serializer)
```
bf4: I missed why you're throwing out `seriarlizer.attributes(options)[:id]`?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -136,22 +129,24 @@ def check_assoc(assoc)
...
+                      serializer.map { |s| resource_identifier(s) }
+                    else
+                      if opts[:virtual_value]
+                        opts[:virtual_value]
+                      elsif serializer && serializer.object
+                        resource_identifier(serializer)
+                      else
+                        nil
+                      end
+                    end
```
bf4: yeah, definitely a latent object in here for serializer and serializer collections...  it would also make naming blocks easier e.g. `serializers.each { |serializer| }`

bf4: (That's OT, sorry... just thinking out loud)

lib/active_model/serializer/adapter/json_api.rb
beauby: Because we do not want it to be part of the `attributes` property of the resource, but on the top-level alongside `type`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -100,15 +96,12 @@ def attributes_for_serializer(serializer, options)
 
         def resource_object_for(serializer, options)
           options[:fields] = @fieldset && @fieldset.fields_for(serializer)
-          options[:required_fields] = [:id, :json_api_type]
```
bf4: So, we don't need required fields because they're the `{ id: '', type: ''}` in the resource identifier object method?

lib/active_model/serializer/adapter/json_api.rb
beauby: I agree for naming blocks but here there are already two local variables named `serializer` (and in an other PR I'd actually like to use the outside one, so any suggestion on how to handle the naming situation?).

test/serializers/attributes_test.rb
```diff
@@ -23,12 +23,6 @@ def test_attributes_with_fields_option
                      @profile_serializer.attributes(fields: [:name]))
       end
 
-      def test_required_fields
-        assert_equal({name: 'Name 1', description: 'Description 1'},
-                     @profile_serializer.attributes(fields: [:name, :description], required_fields: [:name]))
-
-      end
-
```
bf4: This makes it look like you could set `:name` as a required_field.  Is that still possible?

lib/active_model/serializer/adapter/json_api.rb
beauby: Exactly.

lib/active_model/serializer/adapter/json_api.rb
bf4: No, I get that.. but you're not using it, but it's there... that's what I mean by throwing it out... I'm sure there's a detail I'm missing in the refactor, but figured it'd be good to ask from ignorance :)

test/serializers/attributes_test.rb
beauby: What you imply is true, and the answer is no, it is not possible anymore. However, this option is not used anywhere inside of AMS, and I am quite unsure as to why one would use it outside considering its only use was to add attributes without declaring them as attributes. I'm convinced it was created as a hack for the sole purpose of handling the `json_api_type` and `id`.

lib/active_model/serializer/adapter/json_api.rb
beauby: Agreed for `resource_type` and `resource_id`.
Could you elaborate about this "latent object"?

lib/active_model/serializer/adapter/json_api.rb
bf4: I mean, is the serializer.attributes(options)[:id] the same as serializer.object.id

beauby: No, good catch. This change actually breaks `id` overriding.

beauby: It is fixed now.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -133,29 +131,29 @@ def check_assoc(assoc)
...
-
-            attrs[:relationships] ||= {}
-
-            if serializer.respond_to?(:each)
-              add_relationships(attrs, key, serializer)
+        def resource_relationship_value(serializer, options = {})
+          if serializer.respond_to?(:each)
+            serializer.map { |s| resource_identifier(s) }
+          else
+            if options[:virtual_value]
```
beauby: After giving it a thought, why would anyone *ever* use a virtual value for a relationship with json_api? The only value this can take is `nil` or a Resource Identifier/array of Resource Identifiers, which means the virtual attribute has to be either of those, which means the use case is something like:
```ruby
PostSerializer < ActiveModel::Serializer
  attributes :id, :name
  belongs_to :author

  def author
    { type: 'authors', id: '3' }
  end
end
```

which in turn implies that one can't really include the corresponding resource.

Hence, I vote for removing the virtual values for relationships with json_api adapter.

lib/active_model/serializer.rb
joaomdmoura: :+1: I'm happy we finally did it!

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -44,21 +44,29 @@ def fragment_cache(cached_hash, non_cached_hash)
...
         end
 
-        def add_relationship(resource, name, serializer, val=nil)
-          resource[:relationships] ||= {}
-          resource[:relationships][name] = { data: val }
-
-          if serializer && serializer.object
-            resource[:relationships][name][:data] = { type: serializer.json_api_type, id: serializer.id.to_s }
+        def resource_identifier_id(serializer)
+          if serializer.respond_to?('id')
```
joaomdmoura: We are also adding possibility to override it then?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -44,21 +44,29 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+            serializer.id.to_s
+          else
+            serializer.object.id.to_s
```
joaomdmoura: Nice touch converting it to string in order to follow json-api conventions

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -88,28 +96,18 @@ def add_included(resource_name, serializers, parent = nil)
 
         def attributes_for_serializer(serializer, options)
           if serializer.respond_to?(:each)
-            result = []
-            serializer.each do |object|
-              result << resource_object_for(object, options)
-            end
+            serializer.map { |s| resource_object_for(s, options) }
```

beauby: I'm working on the final bit to improve handling of the "included"
relationships. After that, it should be pretty clean/understandable/easy to
work with.

Le mardi 1 septembre 2015, João Moura <notifications@github.com> a écrit :

> In lib/active_model/serializer/adapter/json_api.rb
> <https://github.com/rails-api/active_model_serializers/pull/1103#discussion_r38415770>
> :
>
> > @@ -88,28 +96,18 @@ def add_included(resource_name, serializers, parent = nil)
> >
> >          def attributes_for_serializer(serializer, options)
> >            if serializer.respond_to?(:each)
> > -            result = []
> > -            serializer.each do |object|
> > -              result << resource_object_for(object, options)
> > -            end
> > +            serializer.map { |s| resource_object_for(s, options) }
>
> [image: :+1:]
>
> —
> <https://github.com/rails-api/active_model_serializers/pull/1103/files#r38415770>
> .
>


-- 
Lucas Hosseini
lucas.hosseini@gmail.com

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: Maybe we could do something similar to JsonAdapter, that if found a instance of an model object it will query that object and use it as the relationship. I know that it might be an edge case, but would be nice to cover, taking into account the level o abstraction we are working with when trying to cover everyone using AMS :grin:

test/serializers/attributes_test.rb
joaomdmoura: Despite of felling okay removing this, it just remembers me how badly we need the filter method hahah because handle with different set `fields` using the same serializer is a need.
Btw as I commented [here](https://github.com/rails-api/active_model_serializers/pull/1103/files#r38414212), it seems we are adding some new behavior, would nice to have it tested :smile:

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: I also see you some symbols on the other places and here a string, I'd go with a symbol as well

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -133,29 +131,29 @@ def check_assoc(assoc)
...
-            serializer = association.serializer
-            opts = association.options
-
-            attrs[:relationships] ||= {}
-
-            if serializer.respond_to?(:each)
-              add_relationships(attrs, key, serializer)
+        def resource_relationship_value(serializer, options = {})
+          if serializer.respond_to?(:each)
+            serializer.map { |s| resource_identifier(s) }
```
joaomdmoura: I'm not sure if we have a test covering this to check if we are following json-api conventions, but would be nice to cover this one:
>...
> Resource linkage MUST be represented as one of the following:
>...
> - an empty array ([]) for empty to-many relationships.

beauby: Agreed. Will issue a PR for the corresponding test in a bit.

test/serializers/attributes_test.rb
bf4: I'd just prefer to keep breaking changes out of refactoring pr...

beauby: As explained, this is not a breaking change, as this feature was not
actually a frature but merely a hack that happened to be exposed.

Le mercredi 2 septembre 2015, Benjamin Fleischer <notifications@github.com>
a écrit :

> In test/serializers/attributes_test.rb
> <https://github.com/rails-api/active_model_serializers/pull/1103#discussion_r38569520>
> :
>
> > @@ -23,12 +23,6 @@ def test_attributes_with_fields_option
> >                       @profile_serializer.attributes(fields: [:name]))
> >        end
> >
> > -      def test_required_fields
> > -        assert_equal({name: 'Name 1', description: 'Description 1'},
> > -                     @profile_serializer.attributes(fields: [:name, :description], required_fields: [:name]))
> > -
> > -      end
> > -
>
> I'd just prefer to keep breaking changes out of refactoring pr...
>
> —
> <https://github.com/rails-api/active_model_serializers/pull/1103/files#r38569520>
> .
>


-- 
Lucas Hosseini
lucas.hosseini@gmail.com

beauby:
Rebased.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -9,7 +9,13 @@ def initialize(serializer, options = {})
           super
           @hash = { data: [] }
 
-          if fields = options.delete(:fields)
+          @options[:include] ||= []
+          if @options[:include].is_a?(String)
```
joaomdmoura: Can you explain more about this. It's not related to `type` and `id`, and I also don't understood the `String` verification

joaomdmoura:
How do you feel about split the `add_resource_relationship` into a different PR? Both seems pretty solid, just want to make it easier to review :smile:

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -31,142 +37,128 @@ def serializable_hash(options = nil)
...
+            @hash[:data] = primary_data
+            @hash[:data][:relationships] = relationships if relationships.any?
+            @hash[:included] = included if included.any?
           end
           @hash
         end
 
         def fragment_cache(cached_hash, non_cached_hash)
           root = false if @options.include?(:include)
-          JsonApi::FragmentCache.new().fragment_cache(root, cached_hash, non_cached_hash)
```

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -31,142 +37,128 @@ def serializable_hash(options = nil)
...
 
-        def add_included(resource_name, serializers, parent = nil)
-          unless serializers.respond_to?(:each)
-            return unless serializers.object
-            serializers = Array(serializers)
-          end
-          resource_path = [parent, resource_name].compact.join('.')
-          if include_assoc?(resource_path)
-            @hash[:included] ||= []
+        def resource_identifier_for(serializer)
```

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -31,142 +37,128 @@ def serializable_hash(options = nil)
...
 
-          serializers.each do |serializer|
-            serializer.associations.each do |association|
-              serializer = association.serializer
-
-              add_included(association.key, serializer, resource_path) if serializer
-            end if include_nested_assoc? resource_path
+        def primary_data_for(serializer, options)
+          if serializer.respond_to?(:each)
+            serializer.map { |s| resource_object_for(s, options) }
```

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: nvm, understood :grin:

joaomdmoura:
LGTM, I'm merging it

joaomdmoura:
Oh, could you please @beauby open a PR with a new 'how to' article explaning how to override `id` and `type`?

beauby:
@joaomdmoura With this PR, `type` cannot be overridden (but a small change as in #1122 allows for that as well). Type can be overridden as one would expect it, by defining an `id` method on the serializer class. Maybe a little note in the existing doc would make more sense than a whole article.

joaomdmoura:
The article size don't bother me and indeed it can even be a note, just want to make it easier to be found :smile:  but I'll check it on #1122

PR #1104, opened 2015-08-31
Cannot override `id` with JsonApi.

beauby:
Currently, `id` cannot be overriden with the JsonApi adapter (although it can be with other adapters).
Here is a test to prove it/make sure the situation gets fixed.


beauby:
Oops, this is actually not true 😅

joaomdmoura:
:grin: Now it definitely can

PR #1090, opened 2015-08-28
Clarify AMS dependencies

bf4:

PR #1105, opened 2015-08-31
Add ActiveRecord-backed fixtures.

beauby:
They are namespaced inside `ARModels` to avoid conficts with PORO models (that we might want to namespace as well).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,5 +25,11 @@ else
   gem 'actionpack', gem_version
 end
 
+group :test do
+  gem 'activerecord'
+  gem 'sqlite3', platform: :ruby
+  gem 'activerecord-jdbcsqlite3-adapter', platform: :jruby
+end
+
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
--- a/test/fixtures/active_record.rb
+++ b/test/fixtures/active_record.rb
@@ -37,7 +37,6 @@ module ARModels
 
   class PostSerializer < ActiveModel::Serializer
     attributes :id, :title, :body
-    params :title, :body
 
     has_many :comments
     belongs_to :author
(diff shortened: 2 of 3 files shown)


test/fixtures/active_record.rb
```diff
@@ -0,0 +1,58 @@
...
+    has_many :posts
+  end
+
+  class PostSerializer < ActiveModel::Serializer
+    attributes :id, :title, :body
+    params :title, :body
+
+    has_many :comments
+    belongs_to :author
+    url :comments
```
bf4: Didn't someone suggest a convention of something like

```ruby
class Post < AR::Base
  # stuff
  class PostSerializer < AM::Serializer
    # stuff
  end
end
```

bf4:
:+1: except master doesn't have activerecord nor sqlite, right?

beauby:
You're obviously right!

bf4:
It sort of does by accident in dev... which is why I made https://github.com/rails-api/active_model_serializers/pull/1090

beauby:
Fixed the deps.

Gemfile
```diff
@@ -25,5 +25,11 @@ else
   gem 'actionpack', gem_version
 end
 
+group :test do
+  gem 'activerecord'
+  gem 'sqlite3', platform: :ruby
+  gem 'activerecord-jdbcsqlite3-adapter', platform: :jruby
+end
```
bf4: you're probably right that all the rails stuff above should be in a `gem :test` group..

test/fixtures/active_record.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    belongs_to :author
+  end
+
+  class AuthorSerializer < ActiveModel::Serializer
+    attributes :id, :name
+
+    has_many :posts
+  end
+end
```
bf4: would you be interested in something like https://github.com/rails-api/active_model_serializers/blob/master/test/poro_test.rb that just checks they're legit?

beauby: Yup, makes sense, I'll add it straight away.

bf4: ideally, I'd also want to test 'em all like i commented in https://github.com/rails-api/active_model_serializers/issues/878#issuecomment-96073030  (this is a pr i mean to make)

joaomdmoura:
Cool @beauby this will be really helpful into new feature cc/ @NullVoxPopuli

PR #962, opened 2015-06-22
Rendering array objects that doesn't have serializers

joaomdmoura:
As reported on #772 and #945.
There was a crash when rendering arrays that don't consist of "serializables" classes.
This PR is a first implementation to quickly fix it.

**P.S.** It won't fix #877 yet. 
**P.S.** #954 will need to be updated based on this once merged, around [here](https://github.com/rails-api/active_model_serializers/pull/954/files#diff-07e5a5aaa325f8dc2bcf944d12f0db83R32)

Issue #540, opened 2014-04-02
Limiting stack depth association embeds


bradynpoulsen:
I have been unable to find a way to limit embedding associations when they refer each other.
## Example

Lets say I have a `Project` and a `Client` where the client `has_many :projects`. 

app/serializers/project_serializer.rb

``` ruby
class ProjectSerializer < ActiveModel::Serializer
  embed :ids, include: true

  attributes :id, :name
  has_one :client
end
```

app/serializers/client_serializer.rb

``` ruby
class ClientSerializer < ActiveModel::Serializer
  embed :ids, include: true

  attributes :id, :name
  has_many :projects
end
```

But because it continues to include all of the clients and projects until all of the embeds have been processed.

So if I request `/projects/3.json`, I might get the result:

``` json
{
  "clients": [
    {"id": 1, "name": "HelloWorld, Inc", "project_ids":[1,3]}
  ],
  "projects": [
    {"id": 1, "name": "My First Project", "client_id": 1},
    {"id": 3, "name": "My Cool Idea", "client_id": 1}
  ],
  "project": {"id": 3, "name": "My Cool Idea", "client_id": 1}
}
```

when all I really wanted was just the `clients` and `project` entries because of redundant data transfer.
## Possible Solution

Of course there are ways to fix my purpose, but I wish there were some function we could define to define the scope of embeds, ie:

``` ruby
class ProjectSerializer < ActiveModel::Serializer
  def embed_level(level)
    embed :ids, include: true if level < 2
  end
end
```

PR #1107, opened 2015-08-31
[WIP] Nested Sideloading for the JSON adapter

NullVoxPopuli:
Hey all,

I took another stab at side loading (other PR is at #1088).

This time, it's all in a different adapter.

I'm 100% sure I've completely ignored how the cache works - but!, the logic is there.. so that's good... For the sake of prototyping, the `@associations` variable is a cache local to the specific adapter.

Any prelim feedback is gladly welcomed.

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -47,4 +47,5 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'timecop', '>= 0.7'
+
 end
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -4,6 +4,7 @@ module ActiveModel
       extend ActiveSupport::Autoload
       require 'active_model/serializer/adapter/json'
       require 'active_model/serializer/adapter/json_api'
+      require 'active_model/serializer/adapter/json_sideload'
       autoload :FlattenJson
       autoload :Null
       autoload :FragmentCache
@@ -86,7 +87,11 @@ module ActiveModel
       end
 
       def root
-        serializer.json_key.to_sym if serializer.json_key
+        if serializer.json_key
+          serializer.json_key.to_sym
+        else
+          serializer.root
+        end
       end
 
       def include_meta(json)
--- a/lib/active_model/serializer/adapter/json_sideload.rb
+++ b/lib/active_model/serializer/adapter/json_sideload.rb
@@ -1,0 +1,187 @@
+module ActiveModel
+  class Serializer
+    class Adapter
+      class JsonSideload < Json
+
+        # the list of what has already been serialized will be kept here to
+        # help avoid infinite recursion
+        #
+        # this should be a list of association_name to a list of objects
+        attr_accessor :serialized
+
+        # When we are sideloading associations, we can more easily track what
+        # has been serialized, so that we avoid infinite
+        # recursion / serialization.
+        def initialize(*args)
+          @serialized = {}
+
+          super(*args)
+        end
+
+        def serializable_hash(options = nil)
+          options ||= {}
+
+          serialize_hash(options)
+
+          singularize_lone_objects
+
+          @serialized
+        end
+
+
+        def serialize_hash(options)
+          result = nil
+
+          if serializer.respond_to?(:each)
+            #TODO: Is this ever hit?
+            # binding.pry
+            ap "wwwwwwwwwwwwwwwwwwwwww"
+            result = serialize_array(serializer, options)
+          else
+            # skip if we are already serialized
+            key_name = serializer.object.class.name.tableize
+            existing_of_kind = @serialized[key_name]
+            exists = existing_of_kind ? existing_of_kind.select{|a| a[:id] == s.object.id } : false
+            return if exists
+
+            # we aren't an array! woo!
+
+            result = serialized_attributes_of(serializer, options)
+            # add to the list of the serialized
+
+
+            # now, go over our associations, and add them to the master
+            # serialized hash
+            serializer.associations.each do |association|
+              serializer = association.serializer
+              opts = association.options
+
+              # make sure the association key exists in the master
+              # serialized hash
+              @serialized[association.key] ||= []
+
+              if serializer.respond_to?(:each)
+                array = serialize_array(serializer, opts)
+                association_list = @serialized[association.key]
+
+
+                array.each do |item|
+                  if not association_list.include?(item)
+                    association_list << item
+                  end
+                end
+                @serialized[association.key] = association_list
+
+                # add the ids to the result
+                result[ids_name_for(association.key)] = array.map{|a| a[:id] }
+              else
+                hash = (
+                  if serializer && serializer.object
+                    serialized_attributes_of(serializer, options)
+                  elsif opts[:virtual_value]
+                    opts[:virtual_value]
+                  end
+                )
+
+                add(association.key, hash)
+
+                # add the id to the result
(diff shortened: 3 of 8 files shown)


active_model_serializers.gemspec
```diff
@@ -47,4 +47,9 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'timecop', '>= 0.7'
+
+  # development dependencies, cause... how can you not use these? :-)
+  spec.add_development_dependency "pry-byebug"
+  spec.add_development_dependency "awesome_print"
```
bf4: So, two very opinionated comments

1. I try to only include dev deps in the gemspec if they are required to run tests.  It sort of doesn't matter because gem test is dead, but still
2. https://github.com/rails-api/active_model_serializers/pull/1017#discussion_r37602184 would let you use your own local gems

lib/active_model/serializer/adapter.rb
```diff
@@ -4,6 +4,7 @@ class Adapter
       extend ActiveSupport::Autoload
       require 'active_model/serializer/adapter/json'
       require 'active_model/serializer/adapter/json_api'
+      require 'active_model/serializer/adapter/json_sideload'
```
bf4: @joaomdmoura  this is killing my registerable adapters pr :( https://github.com/rails-api/active_model_serializers/pull/1017

lib/active_model/serializer/adapter.rb
```diff
@@ -86,7 +87,11 @@ def meta_key
       end
 
       def root
-        serializer.json_key.to_sym if serializer.json_key
+        if serializer.json_key
+          serializer.json_key.to_sym
+        else
+          serializer.root
```
bf4: to_sym?

lib/active_model/serializer/adapter/json_sideload.rb
```diff
@@ -0,0 +1,185 @@
...
+
+
+        def serialize_hash(options)
+          result = nil
+
+          if serializer.respond_to?(:each)
+            #TODO: Is this ever hit?
+            # binding.pry
+            ap "wwwwwwwwwwwwwwwwwwwwww"
+            result = serialize_array(serializer, options)
```
bf4: ha ha

lib/active_model/serializer/adapter/json_sideload.rb
```diff
@@ -0,0 +1,185 @@
...
+
+        def serializable_hash(options = nil)
+          options ||= {}
+
+          serialize_hash(options)
+
+          singularize_lone_associations
+
+          @associations
+        end
```
bf4: I'm ok with this as a way to get it out the door in a kind of feature toggle, but it should probably be generalized so that either json or jsonapi could use it, no?

active_model_serializers.gemspec
NullVoxPopuli: Thanks, I'll apply that method of local debugging tools!

lib/active_model/serializer/adapter.rb
NullVoxPopuli: :-(

lib/active_model/serializer/adapter.rb
NullVoxPopuli: `to_sym` is what was there previously, I added the `else serializer.root` ?

lib/active_model/serializer/adapter/json_sideload.rb
NullVoxPopuli: :-)

but seriously, what's an example of how that block gets invoked?

lib/active_model/serializer/adapter/json_sideload.rb
NullVoxPopuli: I was using `ActiveModel::Serializer.config.sideload_associations = true` in my other branch.

I realized, the downside to this approach is that it will recursively build every relationship (and a relationship's relationships) from a serializer - so -- I don't know if there is a way we want to specify against that? or.. maybe that's not a bad thing? idk. I'd need to use this in a real app to see how it behaves.

lib/active_model/serializer/adapter.rb
bf4: Sorry, `serializer.root` is what I meant.. (`serializer.json_key` can't be nil.to_sym because it's the consequence to the truthy json_key

NullVoxPopuli: I guess, I should have added a test for this, cause I ran in to a situation where json_key was nil. hmm. I'll look in to this

lib/active_model/serializer/adapter/json_sideload.rb
bf4: when the resource is a collection it uses the array serializer with has `each`, e.g. `render json: User.all` or `render json: [user, post]` would both go in the `each` block.  I'm thinking though, in https://github.com/rails-api/active_model_serializers/pull/1004#discussion_r38288295 it might make sense to have a 'collection' adapter and 'resource' adapter to clean this up and separate concerns.. though it introduces a bit more complexity, so we'd need to add pieces of it whenever it's more useful than not :)

NullVoxPopuli: I think it would be very beneficial to be able to have customizable collection adapters and serializers.

I know back when I was doing a versioned API, the collections were formatted different, and that would have been pretty helpful. :-)

lib/active_model/serializer/adapter/json_sideload.rb
```diff
@@ -0,0 +1,206 @@
...
+
+        def serializable_hash(options = nil)
+          options ||= {}
+
+          # begin recursive serialization
+          # TODO: this will evaluate ALL specified relationships
+          # throughout the entire tree.
+          #
+          # Do we want a way to limit this?
+          serialize_hash(options)
```
bf4: yes!. max depth of 3 or 4 should be good, imho. 

Related issues
- https://github.com/rails-api/active_model_serializers/pull/1073/files#r38325279 (definitely see this!)
- https://github.com/rails-api/active_model_serializers/issues/211#issuecomment-32900106
- https://github.com/rails-api/active_model_serializers/issues/540
- https://github.com/rails-api/active_model_serializers/pull/722
- https://github.com/rails-api/active_model_serializers/pull/962
- https://github.com/rails-api/active_model_serializers/commit/2e791b6d5b0d40ab4ad427ba0e02f84b78bb8345

I know rubygems has a good recursive resolution lib ([molinillo](https://github.com/rubygems/rubygems/blob/e3088dd686f0fa4f935ac34298eea9e4f563d13b/lib/rubygems/resolver/molinillo/lib/molinillo/resolver.rb))

NullVoxPopuli: what if, for a particular serializer's associations, depth could be defined by using the `include` keyword?

something like this:

```ruby
class BlogSerializer < ...
  has_many :posts, include: [:author, comments: :authors]
```

that way, we'd only get exactly what we expect in the json?

 - 1 blog
 - however many authors
 - however many comments

lib/active_model/serializer/adapter.rb
NullVoxPopuli: how should additional adapters be required?

NullVoxPopuli:
So, based on my other PR, #1111, I think it may be possible to maybe have some sort of sideloading wrapper, assuming that some of the new methods in #1111 could be also defined in the other adapters - sort if using an interface-ish pattern.

And then sideloading would just override the serialization of an association / array? perhaps?

NullVoxPopuli:
Out of date. 
Sideloading depends on which upcoming refactors to the json adapter

PR #1108, opened 2015-08-31
Better lint

bf4:
Extracted from
https://github.com/rails-api/active_model_serializers/pull/1004/files#diff-56455571c4ba7a2b4c640b9e8168f522R40

```ruby
      actual_arity = resource.method(:read_attribute_for_serialization).arity
      if defined?(::Rubinius)
        #  1 for def read_attribute_for_serialization(name); end
        # -2 for alias :read_attribute_for_serialization :send for rbx because :shrug:
        assert_includes [1, -2], actual_arity, "expected #{actual_arity.inspect} to be 1 or -2"
      else
        # using absolute value since arity is:
        #  1 for def read_attribute_for_serialization(name); end
        # -1 for alias :read_attribute_for_serialization :send
        assert_includes [1, -1], actual_arity, "expected #{actual_arity.inspect} to be 1 or -1"
      end
```

Correct cache_key lint for ActiveRecord 4.1+

https://github.com/rails/rails/blob/4-0-stable/activerecord/lib/active_record/integration.rb
`def cache_key`

https://github.com/rails/rails/blob/4-1-stable/activerecord/lib/active_record/integration.rb
`def cache_key(*timestamp_names)`

### DIFF
--- a/lib/active_model/serializer/lint.rb
+++ b/lib/active_model/serializer/lint.rb
@@ -36,7 +36,17 @@ module ActiveModel::Serializer::Lint
     # Typically, it is implemented by including ActiveModel::Serialization.
     def test_read_attribute_for_serialization
       assert_respond_to resource, :read_attribute_for_serialization, "The resource should respond to read_attribute_for_serialization"
-      assert_equal resource.method(:read_attribute_for_serialization).arity, 1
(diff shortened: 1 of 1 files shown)


joaomdmoura:
Nice, thank you @bf4, merging it

bf4:
wee!

PR #1111, opened 2015-08-31
refactor to make way for easy feature adding

NullVoxPopuli:
Big methods can be hard to work with.

Since there are upcoming changes to the json adapter, I figured I'd make things a little easier to digest (especially re: recursion).

### DIFF
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -11,51 +11,73 @@ module ActiveModel
           else
             @hash = {}
 
-            @core = serialized_attributes_of(serializer, options)
+            @core = resource_object_for(serializer, options)
 
-            serializer.associations.each do |association|
-              serializer = association.serializer
-              opts = association.options
+            add_resource_relationships(serializer)
 
-              if serializer.respond_to?(:each)
-                @hash[association.key] = serialize_array(serializer, opts)
-              else
-                @hash[association.key] = serialized_or_virtual_of(serializer, opts)
-              end
-            end
             @result = @core.merge @hash
           end
 
           { root => @result }
         end
 
+        def fragment_cache(cached_hash, non_cached_hash)
+          Json::FragmentCache.new().fragment_cache(cached_hash, non_cached_hash)
+        end
+
+        private
+
+        # iterate through the associations on the serializer,
+        # adding them to @hash as needed (as singular or plural)
+        def add_resource_relationships(serializer)
+          serializer.associations.each do |association|
+            serializer = association.serializer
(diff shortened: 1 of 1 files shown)


lib/active_model/serializer/adapter/json.rb
```diff
@@ -43,6 +29,31 @@ def serializable_hash(options = nil)
           { root => @result }
         end
 
+        def serialize_array_without_root(serializer, options)
+          serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
+        end
+
+        # TODO: what is a virtual value?
```
bf4: it's an internal thing that uses that value instead of a serialized attribute.  e.g. there's no serializer for Ruby arrays, so it would just use the array as the virtual value. see https://github.com/rails-api/active_model_serializers/pull/962

bf4:
:100: Some thoughts:

- I think all the new methods should be private so that we can change them with impunity, and would need to `opt-in` any we want to be part of the public interface
- Would be good to take a look at the json api adapter for any common themes, but not required esp if all the new methods are private

thoughts?

NullVoxPopuli:
- I think some of the methods could be private
 - There is def some overlap in functionality of some methods (at a high level) between these refactored json methods and the json_api methods. 

I'll see what I can do to mimic json_api's naming and flow.

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: ah ok, cool. 

I'll add a comment about that to help future contributors.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +25,61 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+
+        # iterate through the associations on the serializer,
+        # adding them to @hash as needed (as singular or plural)
+        def add_resource_relationships(serializer)
+          serializer.associations.each do |association|
+            serializer = association.serializer
+            opts = association.options
+
+            if serializer.respond_to?(:each)
+              add_relationships(association.key, serializer, opts)
```
beauby: I know the terminology comes from the `json_api` adapter, but I think `add_to_many_relationship` and `add_to_one_relationship` are more intuitive name than the confusingly similar `add_relationship`/`add_relationships`.

NullVoxPopuli: I agree, should json_api's naming change as well? - I am not as familiar with json_api though

joaomdmoura:
This is really nice. I'm closing it because it was already ported on #1114

Issue #1113, opened 2015-09-01
[RFC]Thoughts on an explicit way to define nested relationships?


NullVoxPopuli:
I've been working on sideloading (#1107) for the json adapter, and needed recursive relationship building of the json object.

The problem with doing a strictly recursive approach (independent of sideloading or not), is that if you have enough references on your objects to other objects, a single request (of any kind) is basically going to return the whole database. 
Example: `Post` `belongs_to` an `Author` which `has_many` `Post`s, and each `Post` `has_many` `Comment`s, and each `Comment` `belongs_to` an `Author`.... it and would just keep going. The code in #1107 at least doesn't allow for things to be serialized multiple times (like if a post has the same author, that author only gets serialized once).

So, a potential solution that I'm working on prototyping right now is something like this:

```ruby
PostSerializer < #...
  has_many :comments, include: :author
end
```

So, instead of recursively serializing the comments, we'd serialize them like we do today, but because we have the author key in the include hash, we'd also serialize each of the coments' authors.

```ruby
BlogSerializer < #...
  has_many :posts, include: { author: [:bio], comments: [:author]] }
end
```
resulting json may look like:

```
{
  blog: {
    posts: [
      {
        author: {
          bio: {...}
        },
        comments: [
          {  author: {...} }
        ]
      }
    ]
  }
}
```
or

```ruby
AuthorSerializer < #...
  has_many :blogs, include: { posts: [:author, comments: :author] }
end
```

resulting json may look like:

```
{
  author: {
    blogs: [
      {
        posts: [
          {
            author: {..}
            comments: [
              {
                author: {..}
              }
            ]
          }
        ]
      }
    ]
  }
}
```

this gets in to a problem of potentially having redundant data in the resulting json, so caching would be super important here (which I think is implemented by default? or does a cache key need to be specified in the controller? if not, should there be a default? - or would this kind of cacheing be separate, and more like fancy memoization?)

Note, I found that there was a previous attempt at solving this problem here: #952, but I think it's not clear in #952 how deep the nested relationships go / it could imply that the same recursive problem exists.

iMacTia:
Hi @NullVoxPopuli, just to answer your last question (I'm the author of #952) the nesting can actually bring the recursive problem, but that's intended. It is actually a very similar feature to your side-loading, as you're actually asking to AMS to side-load all associations without explicitly declaring them. That's allow you to keep your code easier to maintain (if you add a new relationship later, that will be taken automatically).
[...]

PR #1114, opened 2015-09-01
Support for Nested Relationships for the Json Adapter

NullVoxPopuli:
This PR is based off of #1111, cause I needed the refactoring.

Discussion of idea here: #1113

Basically:

```ruby
BlogSerializer < #...
  has_many :posts, include: { author: [:bio], comments: [:author]] }
end
```
resulting json may look like:

```
{
  blog: {
    posts: [
      {
        author: {
          bio: {...}
        },
        comments: [
          {  author: {...} }
        ]
      }
    ]
  }
}
```

See the included tests for more examples.


lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +26,151 @@ def fragment_cache(cached_hash, non_cached_hash)
           Json::FragmentCache.new().fragment_cache(cached_hash, non_cached_hash)
         end
 
+        private
```
joaomdmoura: :+1:

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +26,151 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+        def serialize_association(association)
+          serialized_or_virtual_of(association.serializer, association.options)
+        end
+
+        # iterate through the associations on the serializer,
+        # adding them to the parent as needed (as singular or plural)
+        #
+        # restrict_to is a list of symbols that governs what
+        # associations on the passed in seralizer to include
+        def add_resource_relationships(parent, serializer, restrict_to = [])
```
joaomdmoura: `restrict_to` could be renamed to something else that could express it better.
We are facing it as restricting because our mindset on #1113 , but you are actually listing what you want, this is a non-blocking comment, just some details that might help when new ppl jump into this :smile:

NullVoxPopuli: fair point! 

how do you feel about `included_child/sub_serializations` ?

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +26,151 @@ def fragment_cache(cached_hash, non_cached_hash)
           Json::FragmentCache.new().fragment_cache(cached_hash, non_cached_hash)
         end
 
+        private
+
+        def serialize_association(association)
```
joaomdmoura: This method seems to not be used anywhere.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +26,151 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+
+        def serialize_array(serializer, options)
+          serializer.map do |item|
+            serialized_item = resource_object_for(item)
+            serialized_item = yield(serialized_item, item) if block_given?
+            serialized_item
+          end
+        end
+
+        def resource_object_for(serializer)
```
joaomdmoura: Nice abstraction! :+1:

NullVoxPopuli: thanks!

lib/active_model/serializer/adapter/json.rb
joaomdmoura: `nested_associations` maybe? Pretty straightforward, but I'm okay with `included_child` as well.

NullVoxPopuli: I think I like `nested_associations` better. I'm already using `nested_relationships` in `add_relationship` what do you think about that? `association` vs `relationship`?

joaomdmoura: We have the `Association` and it architecture on AMS, so I'd go for association if this it related to it, otherwise we can consider `relationship`

NullVoxPopuli: association it is. stand by

test/adapter/json/nested_relationships_test.rb
```diff
@@ -0,0 +1,203 @@
...
+            @blog = Blog.new(id: 1, name: "My Blog!!")
+            @post.blog = @blog
+            @author.posts = [@post]
+          end
+
+          def test_complex_nested_has_many
+            @first_comment.author = @author
+            @second_comment.author = @author
+
+            serializer = ComplexNestedAuthorSerializer.new(@author)
```
bf4: loving the name

test/adapter/json/nested_relationships_test.rb
```diff
@@ -0,0 +1,203 @@
...
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
```
bf4: should be `serializable_resource = ActiveModel::SerializableResource.new(@author, serializer: ComplexNestedAuthorSerializer, adapter: :json)`, then `serializable_resource.serializable_hash`, no?

NullVoxPopuli: I didn't know you could do that. huh

NullVoxPopuli: should I change it? 
using serializableResource may introduce unneeded dependencies on other functionality

lib/active_model/serializer/adapter/json.rb
```diff
@@ -7,37 +7,16 @@ class Json < Adapter
...
           if serializer.respond_to?(:each)
-            @result = serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
+            @result = serialize_array_without_root(serializer, options)
           else
-            @hash = {}
+            @result = resource_object_for(serializer)
+            @result = add_resource_relationships(
+              @result,
+              serializer,
+              serializer.options[:include])
```
joaomdmoura: I'm curious about this, because here you are on the serializer context, but the included option is an option by association, I thikn this could miss behave if you had multiple associations declaraitons on the serializer using include option, you probably want the include from `serializer.association.<first/whatever/looping>.options`

NullVoxPopuli: your right, so because in `add_resource_relationship` we *then* iterate over the associations and use the association options, this third parameter here isn't needed at all.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +26,147 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+        # iterate through the associations on the serializer,
+        # adding them to the parent as needed (as singular or plural)
+        #
+        # nested_associations is a list of symbols that governs what
+        # associations on the passed in seralizer to include
+        def add_resource_relationships(parent, serializer, nested_associations = [])
+          # have the include array normalized
+          nested_associations = include_array_to_hash(nested_associations)
+
+          included_associations = if nested_associations.present?
```
joaomdmoura: so `nested_associations` should be an normalized array with the nested objects you need.

Eg.
`BlogSerializer` with `has_many :posts include: [:author]`

so `nested_associations` will tell you that you need `author` but inside this if you check if it is inside `serializer.associations`, that will have only `posts`.

Initially I thought this was to check if it an already rendered relationship and avoid render it again. but it seems it is used [here](https://github.com/rails-api/active_model_serializers/pull/1114/files#diff-7b562c84aa219fb9dbfb8112b95df4d9R50) to loop the associations.

What I presume is the the code inside this if might never be executed unless you have something like (probably):

Eg.
`BlogSerializer` with `has_many :posts include: [:author.posts]`

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +23,147 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+            serializer.associations.select{ |association|
+              # nested_associations is a hash of:
+              #   key => nested association to include
+              nested_associations.keys.include?(association.name)
+            }
+          else
+            serializer.associations
+          end
+
+          included_associations.each do |association|
```
joaomdmoura: I think you can loop straight through `serializer.associations` here and  delete the whole `included_associations` block before it because of [what I mentioned](https://github.com/rails-api/active_model_serializers/pull/1114/files#r38724815)

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: The if body actually decides what nested association to render.

Cause, lets say a post has a ton of associations, but you only want the author.
by the time you get to rendering the posts, `[:author]` will be passed in as `nested_associations` and the select filters for `:author` among all of a post's associations.

(The tests also cover this scenario)

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +23,147 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+        def add_resource_relationships(parent, serializer, nested_associations = [])
+          # have the include array normalized
+          nested_associations = include_array_to_hash(nested_associations)
+
+          included_associations = if nested_associations.present?
+            serializer.associations.select{ |association|
+              # nested_associations is a hash of:
+              #   key => nested association to include
+              nested_associations.keys.include?(association.name)
+            }
```
bf4: Should probably be `nested_associations.has_key?(association.name)`

test/adapter/json/nested_relationships_test.rb
bf4: It's the way the controller does it.. so if that's what you're mimicing...

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: I dig it.

test/adapter/json/nested_relationships_test.rb
bf4: i.e. given a resource with serializer and adapter options, seralize it...

test/adapter/json/nested_relationships_test.rb
```diff
@@ -0,0 +1,203 @@
...
+  has_many :posts, include: [:comments]
+end
+
+ComplexNestedAuthorSerializer = Class.new(ActiveModel::Serializer) do
+  attributes :id, :name
+
+  # it would normally be silly to have this in production code, cause a post's
+  # author in this case is always going to be your root object
+  has_many :posts, include: [:author, comments: [:author]]
+end
```
joaomdmoura: Would be nice to cover scenarios using multiple association declarations as well :+1:

test/adapter/json/nested_relationships_test.rb
bf4: it also highlights the current inability to use adapters outside of the AMS namespace :)

test/adapter/json/nested_relationships_test.rb
```diff
@@ -0,0 +1,203 @@
+require 'test_helper'
+
+NestedPostSerializer = Class.new(ActiveModel::Serializer) do
+  attributes :id, :title
+
+  has_many :comments, include: :author
+end
```
bf4: I know this is the pattern in the poros.. but can we just do regular Ruby class declarations here, unless there's a good reason and we can comment that?

test/adapter/json/nested_relationships_test.rb
NullVoxPopuli: personally, I feel that that what the controller does should be in separate tests. So that way if something breaks, if your tests are all separated by responsibility, it'll be easier to track down the cause.

test/adapter/json/nested_relationships_test.rb
NullVoxPopuli: I'll add a test for that real quick, just to be sure

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +23,147 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+            end
+
+            serialized_item
+          end
+        end
+
+
+        def serialize_array_without_root(serializer, options)
+          serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
+        end
```
bf4: I don't think the serializer should be calling the adapter... but I guess you didn't add that

bf4: the json_api adapter https://github.com/rails-api/active_model_serializers/blob/6784866a2d4a57e5b1c571afcbf76362708b11ab/lib/active_model/serializer/adapter/json_api.rb#L22-L23 does

```ruby
            serializer.each do |s|
              result = self.class.new(s, @options.merge(fieldset: @fieldset)).serializable_hash(options)
              @hash[:data] << result[:data]
```

joaomdmoura:
We made a nice and big review here, this is complex but solid, something we aim to have on 0.10.x.
I think it's almost there, let's check the existing comments and then get back to it.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -47,6 +23,147 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+          # but this also allows
+          #  include: :author syntax
+          include_array = Array[*include_array].compact
+
+          result = {}
+
+          hashes = include_array.select{|a| a.is_a?(Hash)}
+          non_hashes = include_array - hashes
+
+          hashes += non_hashes.map{ |association_name| { association_name => [] } }
```
bf4: better to mutate `hashes` than generate a new object, I think

test/adapter/json/nested_relationships_test.rb
NullVoxPopuli: I'd prefer normal class definitions.
I don't know why `Class.new(...)` is used though. :-\

I was just following existing patterns.

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: that doesn't look very nice to look at. :-\

aren't there json_api refactors in the prs?

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: how come?

bf4: this seems more like a function than a (private) property of the JSON Adapter.. What would you think about adding a 

```ruby
module ActiveModel::Serializer::Utils
  module_function
  
        def include_array_to_hash(include_array)
          # still don't trust input
          # but this also allows
          #  include: :author syntax
          include_array = Array[*include_array].compact

          result = {}

          hashes = include_array.select{|a| a.is_a?(Hash)}
          non_hashes = include_array - hashes

          hashes += non_hashes.map{ |association_name| { association_name => [] } }

          # now merge all the hashes
          hashes.each{|hash| result.merge!(hash) }

          result
        end
end
```

and then `Utils.include_array_to_hash` or mix it in

NullVoxPopuli: I like that. Calling it as a class metho would be my preference.

joaomdmoura:
Just waiting to merge #1017 before going back to this

NullVoxPopuli:
woo!

joaomdmoura:
related links: 
#921
#1073

NullVoxPopuli:
Well, I accidentally committed a merge instead of a rebase. tried to checkout my last commit, and rebase that then force push to this branch... didn't work out so well. haha

NullVoxPopuli:
Closed in favor of #1124 (Cleaner git history)

joaomdmoura:
:+1:

PR #1116, opened 2015-09-04
Prefer setters over ivars

bf4:


lib/active_model/serializer.rb
```diff
@@ -124,17 +127,37 @@ def self.root_name
...
+    attr_reader :options
 
-    def initialize(object, options = {})
-      @object = object
-      @options = options
-      @root = options[:root]
-      @meta = options[:meta]
-      @meta_key = options[:meta_key]
-      @scope = options[:scope]
+    def initialize(object, opts = {})
```
beauby: Any reason for the `options` → `opts` change?

bf4: Good eye.  Because there is an `options` method set by `@options` method and `options` arguments passed to method that are not the same thing, so I thought to distinguish.  Methods blow up more more nicely than ivars, too :)

bf4: though in the scope of this initializer, they may still be called options, now that I think about it

lib/active_model/serializer.rb
```diff
@@ -158,15 +181,15 @@ def json_api_type
       end
     end
 
-    def attributes(options = {})
+    def attributes(opts = {})
       attributes =
-        if options[:fields]
-          self.class._attributes & options[:fields]
+        if opts[:fields]
```
bf4: @beauby but here, these `opts` are not the same as `options`

bf4:
Closed in favor of https://github.com/rails-api/active_model_serializers/pull/1166

PR #1117, opened 2015-09-04
FlattenJson adapter no longer inherits Json adapter, renamed to Attributes

bf4:
- FlattenJson adapter no longer inherits Json adapter
- Rename FlattenJson to Attributes (allow plural adapter names)
- Consistently refer to the 'JSON API' and the 'JsonApi' adapter

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ AMS does this through two components: **serializers** and **adapters**.
 Serializers describe _which_ attributes and relationships should be serialized.
 Adapters describe _how_ attributes and relationships should be serialized.
 
-By default AMS will use the **Flatten Json Adapter**. But we strongly advise you to use **JsonApi Adapter** that follows 1.0 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
+By default AMS will use the **Attributes Adapter**. But we strongly advise you to use **JsonApi Adapter** that follows 1.0 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
 Check how to change the adapter in the sections bellow.
 
 # RELEASE CANDIDATE, PLEASE READ
@@ -66,7 +66,7 @@ ActiveModel::Serializer.config.adapter = :json_api
 You won't need to implement an adapter unless you wish to use a new format or
 media type with AMS.
 
-If you want to have a root key on your responses you should use the Json adapter, instead of the default FlattenJson:
+If you want to have a root key on your responses you should use the Json adapter, instead of the default Attributes:
 
 ```ruby
 ActiveModel::Serializer.config.adapter = :json
@@ -137,7 +137,7 @@ render json: @post, meta: { total: 10 }, meta_key: "custom_meta"
 ```
 
 `meta` will only be included in your response if you are using an Adapter that supports `root`,
-as JsonAPI and Json adapters, the default adapter (FlattenJson) doesn't have `root`.
+as JsonAPI and Json adapters, the default adapter (Attributes) doesn't have `root`.
 
 ### Using a serializer without `render`
 
@@ -190,7 +190,7 @@ end
 
 ### Built in Adapters
 
-#### FlattenJSON
+#### Attributes
 
 It's the default adapter, it generates a json response without a root key.
 Doesn't follow any specifc convention.
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -112,12 +112,14 @@ module ActiveModel
         end
       end
 
+      private
+
       def meta
         serializer.meta if serializer.respond_to?(:meta)
       end
 
       def meta_key
-        serializer.meta_key || 'meta'
+        serializer.meta_key || 'meta'.freeze
       end
 
       def root
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -5,11 +5,11 @@ module ActiveModel
       ADAPTER_MAP = {}
       private_constant :ADAPTER_MAP if defined?(private_constant)
       extend ActiveSupport::Autoload
+      autoload :Attributes
+      autoload :Null
       autoload :FragmentCache
       autoload :Json
       autoload :JsonApi
-      autoload :Null
-      autoload :FlattenJson
       autoload :CachedSerializer
 
       def self.create(resource, options = {})
@@ -74,7 +74,8 @@ module ActiveModel
         # @api private
         def find_by_name(adapter_name)
           adapter_name = adapter_name.to_s.classify.tr('API', 'Api')
-          ActiveModel::Serializer::Adapter.const_get(adapter_name.to_sym) or # rubocop:disable Style/AndOr
+          "ActiveModel::Serializer::Adapter::#{adapter_name}".safe_constantize ||
+            "ActiveModel::Serializer::Adapter::#{adapter_name.pluralize}".safe_constantize or # rubocop:disable Style/AndOr
             fail UnknownAdapterError
         end
         private :find_by_name
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -1,8 +1,8 @@
-class ActiveModel::Serializer::Adapter::FlattenJson < ActiveModel::Serializer::Adapter::Json
+class ActiveModel::Serializer::Adapter::Attributes < ActiveModel::Serializer::Adapter
         def serializable_hash(options = nil)
           options ||= {}
           if serializer.respond_to?(:each)
-            result = serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
+            result = serializer.map { |s| Attributes.new(s).serializable_hash(options) }
           else
             hash = {}
 
@@ -43,7 +43,7 @@ class ActiveModel::Serializer::Adapter::FlattenJson < ActiveModel::Serializer::A
 
         private
 
-        # no-op: FlattenJson adapter does not include meta data, because it does not support root.
+        # no-op: Attributes adapter does not include meta data, because it does not support root.
         def include_meta(json)
           json
         end
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -4,7 +4,7 @@ class ActiveModel::Serializer::Adapter::Json < ActiveModel::Serializer::Adapter
 
         def serializable_hash(options = nil)
           options ||= {}
-          { root => FlattenJson.new(serializer).serializable_hash(options) }
+          { root => Attributes.new(serializer).serializable_hash(options) }
         end
 
         private
--- a/lib/active_model/serializer/adapter/json.rb
+++ b/lib/active_model/serializer/adapter/json.rb
@@ -4,44 +4,12 @@ class ActiveModel::Serializer::Adapter::Json < ActiveModel::Serializer::Adapter
 
         def serializable_hash(options = nil)
           options ||= {}
-          if serializer.respond_to?(:each)
-            result = serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
-          else
-            hash = {}
-
-            core = cache_check(serializer) do
-              serializer.attributes(options)
-            end
-
-            serializer.associations.each do |association|
-              serializer = association.serializer
-              association_options = association.options
-
-              if serializer.respond_to?(:each)
-                array_serializer = serializer
-                hash[association.key] = array_serializer.map do |item|
-                  cache_check(item) do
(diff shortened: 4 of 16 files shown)


beauby:
:+1:

bf4:
@beauby Thanks. I followed a somewhat non-linear path to get there, which is why I extracted CacheCheck along the way, but decided is was fine :)

README.md
```diff
@@ -10,7 +10,7 @@ AMS does this through two components: **serializers** and **adapters**.
 Serializers describe _which_ attributes and relationships should be serialized.
 Adapters describe _how_ attributes and relationships should be serialized.
 
-By default AMS will use the **Flatten Json Adapter**. But we strongly advise you to use **JsonApi Adapter** that follows 1.0 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
+By default AMS will use the **Attributes Adapter**. But we strongly advise you to use **JsonApi Adapter** that follows 1.0 of the format specified in [jsonapi.org/format](http://jsonapi.org/format).
 Check how to change the adapter in the sections bellow.
```
bf4: One thing that I *don't* like about calling the adapter `attributes` is that it is harder to grep for, esp if we want to rename it again. But, we liked the name in the Slack channel, so until something better..

README.md
```diff
@@ -10,7 +10,7 @@ AMS does this through two components: **serializers** and **adapters**.
 Serializers describe _which_ attributes and relationships should be serialized.
 Adapters describe _how_ attributes and relationships should be serialized.
```
bf4: reminder to my self to note that one cannot use a Serializer alone to get a hash or JSON.  It is the job of the adapter to compose the attributes-to-be-serialized of each object into the hash or JSON document.

lib/active_model/serializer/adapter.rb
```diff
@@ -2,12 +2,15 @@ module ActiveModel
...
       extend ActiveSupport::Autoload
+      autoload :CacheCheck
       require 'active_model/serializer/adapter/json'
       require 'active_model/serializer/adapter/json_api'
-      autoload :FlattenJson
+      autoload :Attributes
       autoload :Null
       autoload :FragmentCache
 
+      include CacheCheck
```
joaomdmoura: We are using `autoload` and including it in the sequence, shouldn't we go with `require` instead then?

README.md
joaomdmoura: yup, I'm not sure about the name as well, but have no suggestions so far

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -0,0 +1,56 @@
+module ActiveModel
+  class Serializer
+    class Adapter
+      class Attributes < Adapter
```
joaomdmoura: The duplicated code it bad :cry: but I know we need it in order to remove the inheritance.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -4,45 +4,12 @@ module ActiveModel
...
-                    opts[:virtual_value]
-                  end
-              end
-            end
-            @result = @core.merge @hash
-          end
-
-          { root => @result }
+        def serializable_hash(options = {})
+          { root => Adapter::Attributes.new(serializer).serializable_hash(options) }
```
joaomdmoura: Nice :+1:

joaomdmoura:
Nice work here @bf4, really fast! Can you check the comments and rebase it please ? :smile:

lib/active_model/serializer/adapter.rb
bf4: good point :)

README.md
bf4: `AttributeSerialization`?

lib/active_model/serializer/adapter/json.rb
beauby: Yup, that's how it should be!

bf4:
cleaned up

lib/active_model/serializer/adapter.rb
bf4: fixed

NullVoxPopuli:
needs rebase

NullVoxPopuli:
Why not inherit from the json adapter?

bf4:
> 
Can't. Would need to undef methods
>

NullVoxPopuli:
why?

this is just so clean:

          super.each_value.first
        end

my concern is that duplicating code is prone to errors and repeated bugs

bf4:
@NullVoxPopuli asked:

> Why not inherit from the json adapter?

Because, if anything, FlattenJson/Attributes is the super class. It's a more general case.

> I knew someone who preferred inheritance over composition. He was a *super* guy.

```diff
+        def serializable_hash(options = {})
+          { root => Adapter::Attributes.new(serializer).serializable_hash(options) }
```

Nice eh?

Aside: adapters and serializers should both have their own `meta`, the adapter's going at the top level. `serializer.meta` should be part of `Attributes`. http://jsonapi.org/format/#document-resource-objects

NullVoxPopuli:
> Because, if anything, FlattenJson/Attributes is the super class. It's a more general case.

Maybe the common functionality could be pulled out in to a module then? or a class that both inherit from?

I know inheritance isn't the answer for everything, but I do hate duplicate code.

also, I hate combining that sort of stuff in to one line.. 
I would have done:

```ruby
  adapter = Adapter::attributes.new(serializer)
  serialized = adapter.serializable_hash(options)

  { root => serialized }

```

much cleaner :-) and easier to debug.

But anyway, with the upcoming `Adapter::Base` changes, I'm hoping to eventually move JSON stuff in there as well.

I think there is a way to use `json api` and `json` such that they pass blocks to 'Base' for there specific different formats

bf4:
> Maybe the common functionality could be pulled out in to a module then? or a class that both inherit from?

Modules are inheritance. I'm not even sure how that would work without having two subclasses that include such a module if we want to keep the abstract methods in Adapter.

> I know inheritance isn't the answer for everything, but I do hate duplicate code.

Where's the duplicate code. I'm not seeing it :)

bf4:
Rebased

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -60,12 +60,6 @@ def resource_identifier_type_for(serializer)
           end
         end
 
-        def add_relationships(resource, name, serializers)
-          resource[:relationships] ||= {}
-          resource[:relationships][name] ||= { data: [] }
-          resource[:relationships][name][:data] += serializers.map { |serializer| { type: serializer.json_api_type, id: serializer.id.to_s } }
-        end
-
```
bf4: Was accidentally re-introduced in a rebased PR

lib/active_model/serializer/adapter.rb
```diff
@@ -103,39 +107,8 @@ def fragment_cache(*args)
...
-        parts << @klass._cache_digest unless @klass._cache_options && @klass._cache_options[:skip_digest]
-        parts.join('/')
-      end
-
-      def object_cache_key
-        object_time_safe = @cached_serializer.object.updated_at
-        object_time_safe = object_time_safe.strftime('%Y%m%d%H%M%S%9N') if object_time_safe.respond_to?(:strftime)
-        (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{object_time_safe}" : @cached_serializer.object.cache_key
+      ActiveModelSerializers.silence_warnings do
+        attr_reader :options
```
bf4: I'm thinking we should rename this to `instance_options` so that we can replace the `@options` with `instance_options` and don't need to change any method arguments from `options` to e.g. `opts`.. but it seems.. but not in this PR, mostly because I might not have time to do it right

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -0,0 +1,49 @@
...
+          if serializer.respond_to?(:each)
+            serializer.map { |s| Attributes.new(s).serializable_hash(options) }
+          else
+            hash = {}
+
+            core = cache_check(serializer) do
+              serializer.attributes(options)
+            end
+
+            serializer.associations.each do |association|
```
NullVoxPopuli: I see that all this came from the json adapter.
I think I'm getting PRs mixed in my head. bah.

it'll be interesting to see how this plays in to the nested associations stuff.

lib/active_model/serializer/adapter.rb
```diff
@@ -69,7 +72,8 @@ def lookup(adapter)
         # @api private
         def find_by_name(adapter_name)
           adapter_name = adapter_name.to_s.classify.tr('API', 'Api')
-          ActiveModel::Serializer::Adapter.const_get(adapter_name.to_sym) or # rubocop:disable Style/AndOr
+          ActiveModel::Serializer::Adapter.const_get(adapter_name.to_sym) ||
+            ActiveModel::Serializer::Adapter.const_get(adapter_name.pluralize.to_sym) or # rubocop:disable Style/AndOr
```
bf4: the const_get breaks with the pluralize.. so either back to `safe_constantize` or singularize all adapter names when registering and looking up.. sigh

bf4: I mean, not with the pluralize.. with having it twice

NullVoxPopuli: what's rubocop got against this bit here?

beauby: Rubocop advises against using low-precedence boolean operators (`or`, `and`, etc.).

NullVoxPopuli: can we either obey rubocop, or change the rubocop setting? 
disabling static analysis warnings via comments (i think) looks tacky.

beauby: I'd rewrite this as
```ruby
adapter = ActiveModel::Serializer::Adapter.const_get(adapter_name.to_sym) ||
  ActiveModel::Serializer::Adapter.const_get(adapter_name.pluralize.to_sym)
fail UnknownAdapterError unless adapter
```

NullVoxPopuli: looks good :-)

README.md
NullVoxPopuli: `Fields`, `Properties`?? idk. ha

NullVoxPopuli:
needs rebase.

But.. I think I am ok with these changes -- I like the separation of responsibility for attributes. Hopefully this will help cleanup #1127 as well.

NullVoxPopuli:
:+1:

jfelchner:
Hey guys, I just tried rebasing #1029 and ran into this. Can someone explain to me what benefit was provided by doing this as opposed to either:

a) Leaving it as it was.  After all `FlattenJson` was a more specific `Json` adapter and therefore inheritance made sense IMO.
b) Creating a module that both adapters inherit from with the common functionality.

In all three scenarios (including the changes in this PR) there is no duplicated code anywhere.

Now effectively any adapter-specific options/defaults need to be explicitly parsed before passing them into the `Attributes` adapter.  Whereas if we'd used inheritance everything would Just Work because it would be using the appropriate values for whatever class it was instantiated as.

I favor composition the majority of the time, but I don't think this one is appropriate.

I personally think we should instead break the `serializable_hash` method down into about a dozen smaller methods, and keep them on the `Json` adapter.  If the `Json` adapter's `root` is `nil` then there's no root (and no allowed `meta`).  Then the `Attributes` adapter can be completely removed.

As a concrete example, this is causing major problems for me rebasing #1029 and I don't like the solution nearly as much as I did before.  IMO it's much less elegant.

bf4:
@jfelchner 

> I personally think we should instead break the [Adapter#]serializable_hash method down into about a dozen smaller methods, 

Totally agree

> and keep them on the Json adapter. 

I get what you're saying, but the Json adapter is, I think, the wrong place

> If the Json adapter's root is nil then there's no root (and no allowed meta). Then the Attributes adapter can be completely removed.

True. I agree there's got to be a better division of reponsibilities, but is the Json adapter the same as Attributes, but only with a root and meta data?  That's certainly how the code looks now, but will it continue to be true?

For the moment, it's simpler, I think, for the Json adapter to use the Attributes adapter than to worry about which adapter should inherit which.  I'm also not convinced that mixing in an 'AttributesSerialization' module to both the Attributes Adapter and the Json Adapter improves on this.

Here's how I'm thinking of the pieces now:

- A model/resource has attributes that are 'readable for serialization'.
  - (We should really document that `read_attribute_for_serialization(name)` could be used for any user-controlled transformations in the model...
- A Serializer for the model that expresses the desired attributes and any transformations of the model's attributes
- The Attributes adapter more or less passes through the serializer attributes into the serializable hash, except that it also handles collections.
- The other adapters use the serialization (via the Attributes adapter) of the models and any other input or options to build the response document aka serializable_hash

(or something like that)

PR #1118, opened 2015-09-06
[WIP] Fix extra db hit with belongs_to association

beauby:
As mentioned in #1100, `belongs_to` associations currently load the record instead of using the `association_name_id` attribute when available (i.e. with ActiveRecord).
The explicit calls have been fixed, but the way things are, the records are actually loaded when building the associations anyways. A satisfying fix would be to lazify that part.

This PR is built on top of #1103 (which was initially intended to provide an easy way to fix #1100).

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -54,27 +54,61 @@ module ActiveModel
 
         private
 
-        def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+        def formatted_type(type)
+          type = if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
rebase needed.

@beauby, `resource_identifier_type_for` is a useful method - I should see how the json adapter works / doesn't under the same scenario of `belongs_to :author, class_name: User.name`

beauby:
I'll rebase this eventually but currently it requires more work during the phase when the serializer actually gets built in order to fix the issue.

sheldonnbbaker:
Has there been any recent activity on this or #1100? I'm currently getting bit pretty hard by this issue. :)

NullVoxPopuli:
@beauby can you rebase this? we recently had another issue open about this. #1552

beauby:
This doesn't fix it as the problem lies upstream + needs discussion because the real fix implies tighter coupling with AR. Closing this for now.

PR #1120, opened 2015-09-07
Add windows platform to loading sqlite3 to make testing suite pass, which caused by #1105

Eric-Guo:


Eric-Guo:
I do several time, seems ruby 200 fail is random, eIther x86/x64, guess a problem of Windows, not relative with AWS

joaomdmoura:
Nice @Eric-Guo tks! I'm merging it.

Gemfile
```diff
@@ -40,5 +40,5 @@ end
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 group :development, :test do
-  gem 'rubocop', '~> 0.33.0', require: false
+  gem 'rubocop', '~> 0.34.0', require: false
```
bf4: Why change to the Rubocop version?

Gemfile
```diff
@@ -27,7 +27,7 @@ end
 
 group :test do
   gem 'activerecord'
-  gem 'sqlite3', platform: :ruby
+  gem 'sqlite3', platform: [:ruby, :mingw, :x64_mingw, :mswin]
```
beauby: Are `mingw` and `x64_mingw` not included in `mswin`?

Gemfile
joaomdmoura: @Eric-Guo

Eric-Guo: @bf4 @joaomdmoura because rubocop release this version and seems no [break change](https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#change-log) between 0.33 and 0.34.

:smile:

Gemfile
Eric-Guo: They are using different compiler, (mswin using Visual C++), (mingw using GCC) but should be included in all mswin, so writing mswin should enough here, although I don't known why Rails [still using all of them](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/Gemfile#L56), guess the history reason of rubygems, but not check yet.

Gemfile
joaomdmoura: nice :smile:

PR #1121, opened 2015-09-07
Refactor `add_links` in JSONAPI adapter.

beauby:
+ Moved the `paginated?` method in the `ArraySerializer`
+ Refactored `add_links` into `links_for` in order to be consistent with the refactor of the adapter (#1103)

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -29,7 +29,10 @@ class ActiveModel::Serializer::Adapter::JsonApi < ActiveModel::Serializer::Adapt
               end
             end
 
-            add_links(options)
+            if serializer.paginated?
+              @hash[:links] ||= {}
+              @hash[:links].update(links_for(serializer, options))
+            end
           else
             primary_data = primary_data_for(serializer, options)
             relationships = relationships_for(serializer)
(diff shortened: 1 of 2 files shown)


lib/active_model/serializer/array_serializer.rb
```diff
@@ -29,6 +29,12 @@ def json_key
         key = root || @serializers.first.try(:json_key) || object.try(:name).try(:underscore)
         key.try(:pluralize)
       end
+
+      def paginated?
```
bacarini: :+1:

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -19,35 +17,50 @@ def initialize(serializer, options = {})
...
+            result = self.class.new(s, @options.merge(fieldset: @fieldset)).serializable_hash(options)
+            hash[:data] << result[:data]
+
+            if result[:included]
+              hash[:included] ||= []
+              hash[:included] |= result[:included]
+            end
+          end
+
+          if serializer.paginated?
```
bacarini: this check could be inside `links_for` method. Today we have just pagination link, but for the future other links need be included as well.

something like that
```
hash[:links]  = links_for(serializer, options)
```

```
def links_for(serializer, options)
  return {} unless serializer.paginated?
  JsonApi::PaginationLinks.new(serializer.object, options[:context]).serializable_hash(options)
end
```

Is that make sense?

beauby: The thing is, as with `relationships` and `include`, we do not want to add the corresponding json key if not requested.

bacarini: :+1:

joaomdmoura:
Nice, I'm merging this one :smile:

PR #1122, opened 2015-09-07
[WIP] Add config option to globally format JSONAPI type.

beauby:
This is a response to #974. The idea is that client-side libraries have different conventions on the format of the `type` attribute, which this PR offers to set globally as a lambda.
I'm interested in opinions on the best way to offer a good experience with this. A few leads:
+ Having a few available lambdas (like `pluralize_dasherize`, `singularize_camelize`, etc)
+ Offering the possibility to chain some provided "formatters", possibly with a mix of lambdas as well (like `[ :pluralize, :underscore, -> (x) { ... } ]`

This is also related to #1029 (although the solution here is to globally handle the formatting, where #1029 offers to set it at the serializer level).

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -59,7 +59,7 @@ module ActiveModel
             serializer.type
           else
             type = serializer.object.class.model_name.singular
-            ActiveModel::Serializer.config.jsonapi_type_transformer.call(type)
+            ActiveModel::Serializer.config.jsonapi_type_formatter.call(type)
           end
         end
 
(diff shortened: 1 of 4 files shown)


beauby:
This PR actually also provides the possibility to override the type at the serializer level, simply by defining the `type` method on the serializer.

bf4:
I'm thinking this should wait for deserliazation, since any deviation from snake_case would make a rails-to-rails json-api not work (and ember, fwiw, already supports this typo of mapping). esp see https://github.com/ember-data/active-model-adapter/blob/210c2302edfe2c2a3c375668411ffa81ef77dfe3/addon/active-model-adapter.js which comes with ember-rails https://github.com/emberjs/ember-rails

bf4:
ref: https://github.com/rails/rails/pull/20389

PR #1028, opened 2015-07-31
Add supports for :links and :data options to has_many associations for jsonapi adapter

lsylvester:
Adds option to `has_many` relationships have links (`links: true`) and turn of data (`data: false`) for jsonapi.

Replaces #840 
Fixes #834

PR #1123, opened 2015-09-07
Remove url options

bacarini:
Removing url options because It does not work any way.
Thus, there are others PRs (#1028 and #1018) at the moment to include url(links) as well.


beauby:
:+1:

joaomdmoura:
Tks @bacarini I'm merging it.

Issue #1128, opened 2015-09-08
JSON API nicer include syntax


willcosgrove:
I have a lot of nested associations in my app, and was getting tired of writing really long strings/arrays for the include option.  So I wrote a little helper that turns an include that looks like this:

```ruby
render json: @project_group, include: [projects: [project_tasks: [:part, :specification_revisions, :technique, :estimated_by, :comments, task: [:area]]]]
```

into an include value that looks like that

```ruby
render json: @project_group, include: ["projects", "projects.project_tasks", "projects.project_tasks.task", "projects.project_tasks.task.area", "projects.project_tasks.part", "projects.project_tasks.specification_revisions", "projects.project_tasks.technique", "projects.project_tasks.estimated_by", "projects.project_tasks.comments"]
```

I tried to follow the format of what you can pass in to `ActiveRecord::Relation#includes` method.

It's a little messy but something along these lines:

```ruby
  def sanitize_includes(includes, nest: [])
    case includes
    when Array
      includes.flat_map { |include|
        case include
        when Hash
          include.map { |key, value|
            [
              (nest + [key]).join('.'),
              sanitize_includes(value, nest: nest.push(key)),
            ]
          }
        else
          (nest + [include]).join('.')
        end
      }.flatten
    else
      includes
    end
  end
```

Does this sound helpful to anyone else?  I wouldn't mind submitting a PR if it sounds like people are interested.

NullVoxPopuli:
I can work on this once my json adapter implementation is merged

NullVoxPopuli:
See #1113 and #1127

beauby:
I like the idea of a better syntax for `include`, especially since nested resources are by essence a tree structure. However, I think it would make more sense to actually transform the text syntax (the current one) into a tree (i.e. an array of hash/symbols), than the opposite. Will open a PR for JSONAPI.

NullVoxPopuli:
But why use the string syntax? what does it get us?

beauby:
@NullVoxPopuli I would like to get rid of the string syntax but 
  1. it might be a concern to people who already use it
  2. for some simple cases (like no nesting), people might arguably prefer `'author, comments'` to `[ :author, :comments ]`

beauby:
Also, the string syntax makes it easy to sideload a "grand child" resource without sideloading the "child resource", which I don't really see a use case of, but still.

NullVoxPopuli:
[...]

PR #1127, opened 2015-09-07
Support nested associations for Json and Attributes adapters + Refactor Attributes adapter

NullVoxPopuli:
Originally #1114 (I messed up the history because I merged instead of rebased)

This PR is based off of #1111, cause I needed the refactoring.

Discussion of idea here: #1113

Relates to:
#921
#1073

Basically:

```ruby
BlogSerializer < #...
  has_many :posts, include: { author: [:bio], comments: [:author]] }
end
```
resulting json may look like:

```
{
  blog: {
    posts: [
      {
        author: {
          bio: {...}
        },
        comments: [
          {  author: {...} }
        ]
      }
    ]
  }
}
```

See the included tests for more examples.

PR #1131, opened 2015-09-10
Nicer syntax for jsonapi `include`

beauby:
This PR ~~removes~~ *keeps* the old string format for includes (`include: 'posts, posts.author, posts.comments, posts.comments.author, post.comments.likes, friends`) and introduces a new format ~~instead~~ *as well* (`include: [ :friends, posts: [:author, comments: [:author, :likes]]]`), as discussed in #1128.

The "user-friendly" syntax is transformed into a big hash internally (like `{ friends: {}, posts: { author: {}, comments: { author: {}, likes: {} } } }`), which is much easier to work with. The helper method that does the translation could be extracted and put in some kind of `Utils` module (WIP by @NullVoxPopuli, #1127).

The tests, which use a mix of the old string syntax and an hybrid "array of strings" syntax ~~obviously fail for now~~ *are now fixed*.

Notice that one test was actually _changed_, since the previous code allowed for including a resource nested 2 levels deep without including the resource in between, which breaks full linkage (see note http://jsonapi.org/format/#fetching-includes).

### DIFF
--- a/docs/general/adapters.md
+++ b/docs/general/adapters.md
@@ -30,6 +30,8 @@ resources in the `"included"` member when the resource names are included in the
   render @posts, include: 'authors,comments'
 ```
 
+The format of the `include` option can be either a String composed of a comma-separated list of [relationship paths](http://jsonapi.org/format/#fetching-includes), an Array of Symbols and Hashes, or a mix of both.
+
 ## Choosing an adapter
 
 If you want to use a specify a default adapter, such as JsonApi, you can change this in an initializer:
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -7,11 +7,7 @@ class ActiveModel::Serializer::Adapter::JsonApi < ActiveModel::Serializer::Adapt
           super
           @hash = { data: [] }
 
-          @options[:include] ||= []
-          if @options[:include].is_a?(String)
-            @options[:include] = @options[:include].split(',')
-          end
-
+          @included = ActiveModel::Serializer::Utils.include_args_to_hash(@options[:include])
           fields = options.delete(:fields)
           if fields
             @fieldset = ActiveModel::Serializer::Fieldset.new(fields, serializer.json_key)
@@ -117,46 +113,36 @@ class ActiveModel::Serializer::Adapter::JsonApi < ActiveModel::Serializer::Adapt
         end
 
         def included_for(serializer)
-          serializer.associations.flat_map { |assoc| _included_for(assoc.key, assoc.serializer) }.uniq
+          included = @included.flat_map do |inc|
+            association = serializer.associations.find { |assoc| assoc.key == inc.first }
+            _included_for(association.serializer, inc.second) if association
+          end
+
+          included.uniq
         end
 
-        def _included_for(resource_name, serializer, parent = nil)
+        def _included_for(serializer, includes)
           if serializer.respond_to?(:each)
-            serializer.flat_map { |s| _included_for(resource_name, s, parent) }.uniq
+            serializer.flat_map { |s| _included_for(s, includes) }.uniq
           else
             return [] unless serializer && serializer.object
-            result = []
-            resource_path = [parent, resource_name].compact.join('.')
 
-            if include_assoc?(resource_path)
-              primary_data = primary_data_for(serializer, @options)
-              relationships = relationships_for(serializer)
-              primary_data[:relationships] = relationships if relationships.any?
-              result.push(primary_data)
-            end
+            primary_data = primary_data_for(serializer, @options)
+            relationships = relationships_for(serializer)
+            primary_data[:relationships] = relationships if relationships.any?
 
-            if include_nested_assoc?(resource_path)
-              non_empty_associations = serializer.associations.select(&:serializer)
+            included = [primary_data]
 
-              non_empty_associations.each do |association|
-                result.concat(_included_for(association.key, association.serializer, resource_path))
-                result.uniq!
+            includes.each do |inc|
+              association = serializer.associations.find { |assoc| assoc.key == inc.first }
+              if association
+                included.concat(_included_for(association.serializer, inc.second))
+                included.uniq!
               end
             end
-            result
+
+            included
           end
-        end
-
-        def include_assoc?(assoc)
-          check_assoc("#{assoc}$")
-        end
-
-        def include_nested_assoc?(assoc)
-          check_assoc("#{assoc}.")
-        end
-
-        def check_assoc(assoc)
-          @options[:include].any? { |s| s.match(/^#{assoc.gsub('.', '\.')}/) }
         end
 
         def add_links(options)
--- a/lib/active_model/serializer/utils.rb
+++ b/lib/active_model/serializer/utils.rb
@@ -1,0 +1,35 @@
+module ActiveModel::Serializer::Utils
+  module_function
+
+  # Translates a comma separated list of dot separated paths (JSONAPI format) into a Hash.
+  # Example: `'posts.author, posts.comments.upvotes, posts.comments.author'` would become `{ posts: { author: {}, comments: { author: {}, upvotes: {} } } }`.
+  #
+  # @param [String] included
+  # @return [Hash] a Hash representing the same tree structure
+  def include_string_to_hash(included)
+    included.delete(' ').split(',').inject({}) do |hash, path|
+      hash.deep_merge!(path.split('.').reverse_each.inject({}) { |a, e| { e.to_sym => a } })
+    end
+  end
+
+  # Translates the arguments passed to the include option into a Hash. The format can be either
+  # a String (see #include_string_to_hash), an Array of Symbols and Hashes, or a mix of both.
+  # Example: `posts: [:author, comments: [:author, :upvotes]]` would become `{ posts: { author: {}, comments: { author: {}, upvotes: {} } } }`.
+  #
+  # @param [Symbol, Hash, Array, String] included
+  # @return [Hash] a Hash representing the same tree structure
+  def include_args_to_hash(included)
+    case included
+    when Symbol
+      { included => {} }
+    when Hash
+      included.each_with_object({}) { |(key, value), hash| hash[key] = include_args_to_hash(value) }
+    when Array
+      included.inject({}) { |a, e| a.merge!(include_args_to_hash(e)) }
+    when String
+      include_string_to_hash(included)
+    else
+      {}
+    end
+  end
+end
--- a/test/action_controller/json_api/linked_test.rb
+++ b/test/action_controller/json_api/linked_test.rb
@@ -43,29 +43,29 @@ module ActionController
 
           def render_resource_with_include
             setup_post
-            render json: @post, include: 'author', adapter: :json_api
+            render json: @post, include: [:author], adapter: :json_api
           end
 
           def render_resource_with_nested_include
             setup_post
-            render json: @post, include: 'comments.author', adapter: :json_api
+            render json: @post, include: [comments: [:author]], adapter: :json_api
           end
 
           def render_resource_with_nested_has_many_include
             setup_post
-            render json: @post, include: ['author', 'author.roles'], adapter: :json_api
+            render json: @post, include: 'author.roles', adapter: :json_api
           end
 
           def render_resource_with_missing_nested_has_many_include
             setup_post
             @post.author = @author2 # author2 has no roles.
-            render json: @post, include: 'author,author.roles', adapter: :json_api
+            render json: @post, include: [author: [:roles]], adapter: :json_api
           end
 
           def render_collection_with_missing_nested_has_many_include
             setup_post
             @post.author = @author2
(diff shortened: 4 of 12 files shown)


willcosgrove:
Excited for this!  Will this continue to work with the JSONAPI spec for an include string?  It would still probably be good to be able to accept a url parameter for the include, and pass that directly to the serializer.

beauby:
@willcosgrove Yeah, I just brought back the string format as well, as I had not thought about that use case when I decided to get rid of it.

lib/active_model/serializer/adapter.rb
```diff
@@ -19,6 +19,23 @@ def self.adapter_class(adapter)
...
+        case included
+        when Symbol
+          { included => {} }
+        when Hash
+          Hash[included.map { |key, value| [key, include_option_to_hash(value)] }]
+        when Array
+          included.inject({}) { |a, e| a.merge(include_option_to_hash(e)) }
+        when String
+          included.delete(' ').split(',').inject({}) do |hash, path|
+            hash.deep_merge(path.split('.').reverse_each.inject({}) { |a, e| { e.to_sym => a } })
```
estum: It allocates many extra hashes. Why not to use a `deep_merge!`?

beauby: Fixed. Thanks for spotting that!

test/adapter/include_option_to_hash_test.rb
```diff
@@ -0,0 +1,67 @@
...
+      def test_simple_array
+        assert_equal(
+          { author: {}, comments: {} },
+          Adapter.include_option_to_hash([:comments, :author]))
+      end
+
+      def test_nested_array
+        assert_equal(
+          { posts: { author: {}, comments: { author: {} } }, comments: {} },
+          Adapter.include_option_to_hash([:comments, posts: [:author, comments: [:author]]]))
```
NullVoxPopuli: for more complex data, having the method call as part of the assert_equal can make it more difficult to compare the expected with the input, and result.

for the sake of debugging, I like to do

    expected = ...
    parameter_stuff = ...
    actual = something(parameter_stuff)

that way, if something is going weird, you can easily throw a `binding.pry` in there without getting lost in minitest/rspec.

lib/active_model/serializer/adapter.rb
```diff
@@ -19,6 +19,23 @@ def self.adapter_class(adapter)
...
+      def self.include_option_to_hash(included)
+        case included
+        when Symbol
+          { included => {} }
+        when Hash
+          Hash[included.map { |key, value| [key, include_option_to_hash(value)] }]
+        when Array
+          included.inject({}) { |a, e| a.merge!(include_option_to_hash(e)) }
+        when String
+          included.delete(' ').split(',').inject({}) do |hash, path|
```
NullVoxPopuli: this should be broken in to a helper method and explained

joaomdmoura:
Would be nice to have the commits with failing test squashed :grin: I'm triggering travis again.

lib/active_model/serializer/adapter.rb
joaomdmoura: Would be nice to get nested association merged in order to put this on `Utils`

test/adapter/include_option_to_hash_test.rb
joaomdmoura: Yup, I usually do the same

test/action_controller/json_api/linked_test.rb
```diff
@@ -43,29 +43,29 @@ def render_resource_without_include
 
           def render_resource_with_include
             setup_post
-            render json: @post, include: 'author', adapter: :json_api
+            render json: @post, include: [:author], adapter: :json_api
```
joaomdmoura: These are the only tests involving `action_controller` + `include` would be nice to have some of them or new ones covering the string behavior in to make sure we con't break it in the future.

beauby:
All comments taken into account + squashed.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -1,178 +1,164 @@
...
-          pagination_links = ActiveModel::Serializer::Adapter::JsonApi::PaginationLinks.new(resources, options[:context]).serializable_hash(options)
-          links.update(pagination_links)
-        end
-
-        def paginated?(collection)
-          collection.respond_to?(:current_page) &&
-            collection.respond_to?(:total_pages) &&
-            collection.respond_to?(:size)
-        end
+  extend ActiveSupport::Autoload
```
bf4: indent!!!

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,24 @@
...
+      hash.deep_merge!(path.split('.').reverse_each.inject({}) { |a, e| { e.to_sym => a } })
+    end
+  end
+
+  def include_args_to_hash(included)
+    case included
+    when Symbol
+      { included => {} }
+    when Hash
+      Hash[included.map { |key, value| [key, include_args_to_hash(value)] }]
```
bf4: just curious why you didn't go with reduce here...

test/action_controller/json_api/linked_test.rb
```diff
@@ -43,29 +43,29 @@ def render_resource_without_include
...
           end
 
           def render_resource_with_nested_include
             setup_post
-            render json: @post, include: 'comments.author', adapter: :json_api
+            render json: @post, include: [comments: [:author]], adapter: :json_api
```
bf4: did these tests break?  if so, that's worrysome

test/utils_test.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        def include_options_to_hash(included)
+          if included.is_a?(Symbol)
+            { included => {} }
+          elsif included.is_a?(Hash)
+            Hash[included.map { |k, v| [k, include_options_to_hash(v)] }]
+          elsif included.is_a?(Array)
+            included.inject({}) { |a, e| a.merge(include_options_to_hash(e)) }
+          end
+        end
```
bf4: is this supposed to be here?

bf4:
Overall, looks good, should have docs added, if it's not too much

lib/active_model/serializer/utils.rb
beauby: Something like `included.reduce { |a, e| a.merge!(e.first => include_args_to_hash(e.second) }`? Idk, it felt more natural to iterate over the `key, value` pairs and combine the result as a Hash, but I wouldn't mind changing it.

test/action_controller/json_api/linked_test.rb
beauby: They broke when I removed the string syntax, and stopped breaking when I added it back.

test/utils_test.rb
beauby: Hmm no, I probably messed up my rebasing...

lib/active_model/serializer/utils.rb
bf4: oh, instead of `{ |hash, (k, v)|` I get that. You could also `hash = {}`; included.each do |k,v| hash[k] = include_args_to_hash(v) end`

lib/active_model/serializer/adapter/json_api.rb
beauby: The indentation of the whole file was wrong.

bf4: it's supposed to be so that the diffs are easy to read..  e.g. so we can https://github.com/rails-api/active_model_serializers/pull/1138

beauby: Fair enough, will revert it.

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,24 @@
...
+      hash.deep_merge!(path.split('.').reverse_each.inject({}) { |a, e| { e.to_sym => a } })
+    end
+  end
+
+  def include_args_to_hash(included)
+    case included
+    when Symbol
+      { included => {} }
+    when Hash
+      included.each_with_object({}) { |(key, value), hash| hash[key] = include_args_to_hash(value) }
```
beauby: @bf4: How about this?

test/action_controller/json_api/linked_test.rb
beauby: I reverted some to the string syntax.

lib/active_model/serializer/utils.rb
NullVoxPopuli: this aligns well with https://github.com/rails-api/active_model_serializers/pull/1127/files#diff-2e4bf15928c11c2f4eff9285340b602d

comments would be nice though. 

Have you tried seeing what happens if you somehow pass an invalid string as an include option?

beauby: It would pretty much do what it's supposed to. What do you have in mind?

NullVoxPopuli: Possibly just adding tests for if someone specified 'authr' instead of 'author' stuff like that.

strings can be a little difficult to catch errors in sometimes, due to silent failing.

beauby: Oh I see. Currently, the way it works, it should silently ignore any included relationship that was not defined on the serializer. Moreover, I believe there already are some tests on the adapter for trying to include a missing relationship.

NullVoxPopuli: Do you think there should be a logger output for when that happens? maybe not in this PR, but in another?

beauby: Yeah it would make sense. Also, there should be a logger output when attempting to serialize a related relationship for which there exists no serializer.

test/adapter/json_api/linked_test.rb
```diff
@@ -153,11 +153,11 @@ def test_include_multiple_posts_and_linked
             serializer = BioSerializer.new @bio1
             adapter = ActiveModel::Serializer::Adapter::JsonApi.new(
               serializer,
-              include: ['author', 'author.posts']
+              include: [author: [:posts]]
```
NullVoxPopuli: should we still have tests for the string equivalent of all these? just to be thorough?

beauby: I left some of the tests with the string syntax. We could add more, although I think the right place to test this is in the `include_args_to_hash` tests.

NullVoxPopuli: agreed

test/utils/include_args_to_hash_test.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        def test_single_string
+          input = 'author'
+          expected = { author: {} }
+          actual = ActiveModel::Serializer::Utils.include_args_to_hash(input)
+          assert_equal(expected, actual)
+        end
+
+        def test_multiple_strings
+          input = 'author,comments'
```
NullVoxPopuli: can you add some tests for invalid input? such as `authr`?

NullVoxPopuli: or 'posts.comments.author.invalid'

test/adapter/json_api/linked_test.rb
beauby: Oops, I meant in the adapter tests when I said `include_args_to_hash`!

test/utils/include_args_to_hash_test.rb
beauby: This would produce the expected output at this level, since this is just building a hash.

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,24 @@
+module ActiveModel::Serializer::Utils
+  module_function
+
+  def include_string_to_hash(included)
+    included.delete(' ').split(',').inject({}) do |hash, path|
+      hash.deep_merge!(path.split('.').reverse_each.inject({}) { |a, e| { e.to_sym => a } })
+    end
+  end
+
+  def include_args_to_hash(included)
```
NullVoxPopuli: could you add some docs to this? 
you could take from my PR if you want, there just needs to be something explaining what this does. :)

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,24 @@
+module ActiveModel::Serializer::Utils
+  module_function
+
+  def include_string_to_hash(included)
```
NullVoxPopuli: could you add some docs to this? :-)

test/adapter/json_api/linked_test.rb
NullVoxPopuli: I gotchyou :-)

docs/general/adapters.md
```diff
@@ -30,6 +30,8 @@ resources in the `"included"` member when the resource names are included in the
   render @posts, include: 'authors,comments'
 ```
 
+The format of the `include` option can be either a String composed of a comma-separated list of [http://jsonapi.org/format/#fetching-includes](relationship paths), an Array of Symbols and Hashes, or a mix of both.
```
bacarini: @beauby  I think this link is wrong, isn't?

I believe the right way is something like that:
```
[relationship paths](http://jsonapi.org/format/#fetching-includes)
```

beauby: Good catch, thanks @bacarini!

PR #1132, opened 2015-09-10
Get rid of unnecessary instance variables, and implied dependencies.

beauby:
In the meantime before #1117 gets merged, here is a PR to cleanup the use of unnecessary instance variables, and the nasty effect of not setting `@result` in `Json` adapter breaking `FlattenJson` adapter.


bf4:
:+1:

NullVoxPopuli:
:+1:

joaomdmoura:
Nice!
:+1:

PR #1133, opened 2015-09-10
Add basic support for xml

beauby:
It's all in the title: do `render xml: resource` instead of `render json: resource`, and you get xml (the actual format will depend on the adapter used).

### DIFF
--- a/test/action_controller/serialization_test.rb
+++ b/test/action_controller/serialization_test.rb
@@ -136,6 +136,13 @@ module ActionController
           render json: like
         end
 
+        def render_object_xml
+          profile = Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+          with_adapter :flatten_json do
+            render xml: profile
+          end
+        end
+
         private
 
         def generate_cached_serializer(obj)
@@ -419,6 +426,14 @@ module ActionController
         assert_equal '', (capture(:stderr) {
           controller.get_serializer(Profile.new)
         })
+      end
+
+      def test_render_object_xml
+        get :render_object_xml
+
+        hash = Hash.from_xml(@response.body).each_value.first
+        assert_equal('application/xml', @response.content_type)
+        assert_equal({ 'name' => 'Name 1', 'description' => 'Description 1' }, hash)
       end
     end
   end
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -6,6 +6,12 @@ module ActionController
 
     include ActionController::Renderers
 
+    RENDERER_METHODS = [
+      :_render_option_json,
+      :_render_with_renderer_json,
+      :_render_option_xml,
+      :_render_with_renderer_xml
+    ]
     # Deprecated
     ADAPTER_OPTION_KEYS = ActiveModel::SerializableResource::ADAPTER_OPTION_KEYS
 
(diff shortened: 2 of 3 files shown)


bf4:
Looks fine, missing as_xml and delegation to adapter in serializable resource, but i hope we're not promissong too mich by adding it... All rails models already support xml... 

I have some thougts on design for supporting sifferent mime types
Also.See my serializers pr in Rails


Who wants to maintain it in ams?

bf4:
Also, AMS has json code all over..

beauby:
@bf4 Does it? I don't see much code in AMS that is really json specific.

test/action_controller/serialization_test.rb
```diff
@@ -420,6 +427,14 @@ def use_adapter?
           controller.get_serializer(Profile.new)
         })
       end
+
+      def test_render_object_xml
+        get :render_object_xml
+
+        hash = Hash.from_xml(@response.body).each_value.first
```
bf4: So, ActiveModel::Serializers::XML does this.. https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/xml.rb#L232-L234

test/action_controller/serialization_test.rb
```diff
@@ -136,6 +136,13 @@ def render_fragment_changed_object_with_relationship
           render json: like
         end
 
+        def render_object_xml
+          profile = Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+          with_adapter :flatten_json do
+            render xml: profile
```
bf4: This would work nicely, name-wise, with https://github.com/rails-api/active_model_serializers/pull/1117/

bf4:
@beauby What do you think of, instead of adding this code to AMS, writing it up in docs, maybe still test it?

e.g.

```ruby
# docs/other_mime_type.md

See the Rails built-in xml Renderer and how to override it
- https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_controller/metal/renderers.rb#L66-L128

Or

Just override the xml renderer method Rails defines

# docs/xml_mime_example.rb
# In your ApplicationController or a mixin,
# Rails 4.0 - 4.1, use `:_render_option_xml`
# Rails 4.2 use `:_render_with_renderer_xml`
# Rails 5 TBD


def _render_with_renderer_xml(xml, options)
  options[:adapter] ||= :flatten_json # or :attributes, hopefully
  serializable_resource = get_serializer(xml, options)
  if serializable_resource == xml
    super
  else
    super(serializable_resource, options).serializable_hash(options).to_xml
  end
end
```

beauby:
@bf4 I do not have any strong feeling either way, to be honest. I made the PR because it seemed like a "quick win" for the project, but I personally have no use for it, so I'm fine if we decide to make it a note instead!

bf4:
Can we close this?

NullVoxPopuli:
I think so

NullVoxPopuli:
Unless we actually want to support xml. IMO, it needs to die. (Xml)

beauby:
I was keeping this issue open until we make a line in the docs about it.

Le vendredi 2 octobre 2015, L. Preston Sego III <notifications@github.com>
a écrit :

> Unless we actually want to support xml. IMO, it needs to die. (Xml)
>
> —
> <https://github.com/rails-api/active_model_serializers/pull/1133#issuecomment-144972393>
> .
>


-- 
Lucas Hosseini
lucas.hosseini@gmail.com

PR #1137, opened 2015-09-11
Rename Serializer subclasses to Serialization

bf4:
ref:
- https://github.com/joaomdmoura/active_model_serializers/commit/0d2ef9a3c0efc0552f7a22aaf8dd80b51e39617c
- https://github.com/rails-api/active_model_serializers/pull/950#discussion_r36134401

- [ ] Fix the build
- [ ] Discuss naming of files and superclasses, esp with respect to https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -157,7 +157,7 @@ module ActiveModel
 
     def self.get_serializer_for(klass)
       serializers_cache.fetch_or_store(klass) do
-        serializer_class_name = "#{klass.name}Serializer"
+        serializer_class_name = "#{klass.name}Serialization"
         serializer_class = serializer_class_name.safe_constantize
 
         if serializer_class
--- a/test/serializers/association_macros_test.rb
+++ b/test/serializers/association_macros_test.rb
@@ -3,15 +3,15 @@ require 'test_helper'
 module ActiveModel
   class Serializer
     class AssociationMacrosTest < Minitest::Test
-      AuthorSummarySerializer = Class.new
-      class AssociationsTestSerializer < Serializer
-        belongs_to :author, serializer: AuthorSummarySerializer
+      AuthorSummarySerialization = Class.new
+      class AssociationsTestSerialization < ActiveModel::Serializer
+        belongs_to :author, serializer: AuthorSummarySerialization
         has_many :comments
         has_one :category
       end
 
       def before_setup
-        @reflections = AssociationsTestSerializer._reflections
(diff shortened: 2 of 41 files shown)


lib/active_model/serializer.rb
```diff
@@ -156,7 +156,7 @@ def self.digest_caller_file(caller_line)
 
     def self.get_serializer_for(klass)
       serializers_cache.fetch_or_store(klass) do
-        serializer_class_name = "#{klass.name}Serializer"
+        serializer_class_name = "#{klass.name}Serialization"
```
bf4: hmm

test/serializers/association_macros_test.rb
```diff
@@ -3,15 +3,15 @@
 module ActiveModel
   class Serializer
     class AssociationMacrosTest < Minitest::Test
-      AuthorSummarySerializer = Class.new
-      class AssociationsTestSerializer < Serializer
-        belongs_to :author, serializer: AuthorSummarySerializer
+      AuthorSummarySerialization = Class.new
+      class AssociationsTestSerialization < Serialization
```
bf4: so, I guess this should be `s/Serialization/ActiveModel::Serializer`

beauby:
I am not too sure about this change personally. I like that in a rails project I have my controllers in `app/controllers/` and my serializers in `app/serializers`. It would be weird to have "serializations" in `app/serializers` and even weirder to have them in `app/serializations`.
I agree the name "serializer" is not perfect, but at this point I'm afraid we're kinda stuck with it.

NullVoxPopuli:
I don't like the name change. Serializer leads us to believe that the class is something that can serialize. Serialization makes it sound like it is data that is already serialized, imo.

beauby:
@NullVoxPopuli Thing is, our "serializers" do not actually "serialize" anything, they merely present the objects in an adapter-friendly way.

NullVoxPopuli:
So then I guess `Hasherizer` may be more appropriate? :-)

beauby:
Or `SerializationPresenter`, but in any case, the project is called `ActiveModel::Serializers` so we kinda have to stick with `Serializers` in my opinion.

NullVoxPopuli:
how strongly does `rails` feel about "`Serialization`"? it looks like they just have the one module https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb

bf4:
This has been rejected...

PR #1138, opened 2015-09-11
Introduce Adapter::Base

bf4:
Why? 

- using a class as a namespace that you also inherit from is complicated and circular at time i.e. buggy (see https://github.com/rails-api/active_model_serializers/pull/1177)
- The class methods on Adapter aren't necessarily related to the instance methods, they're more Adapter functions
- named `Base` because it's a Rails-ism
- It helps to isolate and highlight what the Adapter interface actually is

TODO
- [ ] Add to changelog
- [ ] document breaking change in Adapter from class -> module and base class from Adapter -> Adapter::Base

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ### 0.10.0
 
+Breaking changes:
+  * Adapters now inherit Adapter::Base. 'Adapter' is now a module, no longer a class. [@bf4], #1138
+    * using a class as a namespace that you also inherit from is complicated and circular at time i.e.
+      buggy (see https://github.com/rails-api/active_model_serializers/pull/1177)
+    * The class methods on Adapter aren't necessarily related to the instance methods, they're more
+        Adapter functions
+    * named `Base` because it's a Rails-ism
+    * It helps to isolate and highlight what the Adapter interface actually is
+
+Features:
   * adds adapters pattern
   * adds support for `meta` and `meta_key` [@kurko]
   * adds method to override association [@kurko]
@@ -12,3 +22,7 @@
   * adds FlattenJSON as default adapter [@joaomdmoura]
   * adds support for `pagination links` at top level of JsonApi adapter [@bacarini]
   * adds extended format for `include` option to JsonApi adapter [@beauby]
+
+Fixes:
+
+Misc:
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -1,26 +1,30 @@
 module ActiveModel
   class Serializer
-    class Adapter
+    module Adapter
       UnknownAdapterError = Class.new(ArgumentError)
       ADAPTER_MAP = {}
       private_constant :ADAPTER_MAP if defined?(private_constant)
       require 'active_model/serializer/adapter/fragment_cache'
       require 'active_model/serializer/adapter/cached_serializer'
 
-      def self.create(resource, options = {})
-        override = options.delete(:adapter)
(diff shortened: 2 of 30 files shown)


lib/active_model/serializer/adapter.rb
```diff
@@ -1,30 +1,33 @@
 module ActiveModel
   class Serializer
-    class Adapter
+    module Adapter
```
bf4: possible source of error

lib/active_model/serializer/adapter/flatten_json.rb
```diff
@@ -1,4 +1,7 @@
-class ActiveModel::Serializer::Adapter::FlattenJson < ActiveModel::Serializer::Adapter::Json
```
bf4: possible source of lookup failure

lib/active_model/serializer/adapter.rb
bf4: fixed

lib/active_model/serializer/adapter.rb
```diff
@@ -1,26 +1,28 @@
...
 
-      def self.create(resource, options = {})
-        override = options.delete(:adapter)
-        klass = override ? adapter_class(override) : ActiveModel::Serializer.adapter
-        klass.new(resource, options)
-      end
+      class << self # All methods are class functions
+        def new(*args)
+          fail ArgumentError, [args, caller[0]].inspect
+        end
```
bf4: Just in case.. but perhaps I could write a better message..  is really meant to be a helpful deprecation warning

NullVoxPopuli:
Are you gonna add to changlog and document the breaking change in this PR?

NullVoxPopuli:
otherwise :+1:

NullVoxPopuli:
ping @bf4

lib/active_model/serializer/adapter/base.rb
```diff
@@ -0,0 +1,58 @@
...
+        def meta
+          serializer.meta if serializer.respond_to?(:meta)
+        end
+
+        def meta_key
+          serializer.meta_key || 'meta'.freeze
+        end
+
+        def root
+          serializer.json_key.to_sym if serializer.json_key
```
beauby: What's the rationale behind calling it `root` in the adapter and `json_key` in the serializer?

lib/active_model/serializer/adapter/base.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+
+        def include_meta(json)
```
beauby: It's probably not within the realm of this PR, but I think we should stay away from side effects as much as possible when unnecessary.

CHANGELOG.md
```diff
@@ -12,3 +22,7 @@
   * adds FlattenJSON as default adapter [@joaomdmoura]
```
beauby: There should be a note somewhere stating that the FlattenJSON adapter changed name.

lib/active_model/serializer/adapter/base.rb
bf4: yup

lib/active_model/serializer/adapter/base.rb
bf4: I think just evoluation...

CHANGELOG.md
bf4: Yeah, that's why I wanted a label for 'add to changelog' for all the times we haven't done it

beauby:
LGTM, merging.

PR #1142, opened 2015-09-13
Updating wording on cache expiry in README

leighhalliday:
Just thought the wording would be a little better for when cache is expired than how it was. Small copy tweak.


NullVoxPopuli:
:+1:

Issue #1097, opened 2015-08-28
Namespacing config options


beauby:
As more and more config options become available, some of which being relevant only to certain adapters, it would make sense to namespace them. I'm opening this issue to have a place to discuss it.

joaomdmoura:
Thank you for opening this @beauby, indeed there are quite some new config option on the 0.10.x version.
Could you point out on the issue what are the ones that would be affected?

PR #1147, opened 2015-09-14
Add support for jsonapi top level member

beauby:
C.f. http://jsonapi.org/format/#document-top-level

Supports : 
+ a global boolean config option `jsonapi_toplevel_member`, that decides whether a top level `jsonapi` member will be included in each serialized resource, which defaults to `false`,
+ a global option `jsonapi_version`, which defaults to `1.0`,
+ an adapter option `jsonapi_toplevel_meta`.

This PR is based on #1121 for convenience.

### DIFF
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -9,3 +9,5 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## JSON API
 
 - `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
+- `jsonapi_toplevel_member`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object) in the response document. Default: `false`.
+- `jsonapi_version`: The latest version of the spec the API conforms to. Used when `jsonapi_toplevel_member` is `true`. Default: `'1.0'`.
--- a/lib/active_model/serializable_resource.rb
+++ b/lib/active_model/serializable_resource.rb
@@ -1,7 +1,8 @@
 require 'set'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter,
+                                   :jsonapi_toplevel_meta])
 
     def initialize(resource, options = {})
       @resource = resource
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -8,6 +8,8 @@ module ActiveModel
         base.config.array_serializer = ActiveModel::Serializer::ArraySerializer
         base.config.adapter = :flatten_json
         base.config.jsonapi_resource_type = :plural
+        base.config.jsonapi_toplevel_member = false
+        base.config.jsonapi_version = '1.0'
       end
     end
   end
--- a/test/adapter/json_api/toplevel_jsonapi_test.rb
+++ b/test/adapter/json_api/toplevel_jsonapi_test.rb
@@ -1,0 +1,86 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    class Adapter
+      class JsonApi
+        class TopLevelJsonApiTest < Minitest::Test
+          def setup
+            @author = Author.new(id: 1, name: 'Steve K.')
+            @author.bio = nil
+            @author.roles = []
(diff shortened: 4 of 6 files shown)


docs/general/configuration_options.md
```diff
@@ -9,3 +9,5 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## JSON API
 
 - `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
+- `jsonapi_toplevel_member`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object) in the response document. Default: `false`.
+- `jsonapi_version`: The latest version of the spec the API conforms to. Used when `jsonapi_toplevel_member` is `true`. Default: `'1.0'`.
```
joaomdmoura: Nice :+1:

joaomdmoura:
This is cool, I'm not sure about the adapter options, feels like edge cases for me that might add complexity to maintain.
It also makes me thing we indeed need to add namespace to the  config options :grin: It will get mess quickly if we keep it this way and adding options.

beauby:
Hmm namespacing config options... I remember someone very wise mentioning that in https://github.com/rails-api/active_model_serializers/issues/1097 :p
I don't mind removing the option for now.

joaomdmoura:
So lets remove the adapter options? So can add it if someone supports/needs it :grin:
We can handle the namespace config on other PR :smile:

beauby:
Done :+1:

NullVoxPopuli:
I'm good with this. :+1:

test/adapter/json_api/toplevel_jsonapi_test.rb
```diff
@@ -0,0 +1,86 @@
...
+            with_adapter :json_api do
+              with_config(:jsonapi_toplevel_member, true) do
+                hash = ActiveModel::SerializableResource.new(@post).serializable_hash
+                assert_nil(hash[:jsonapi][:meta])
+              end
+            end
+          end
+
+          def test_toplevel_jsonapi_meta
+            with_adapter :json_api do
```
bf4: no need for `with_adapter`.  Just pass in `adapter: :json_api`

lib/active_model/serializable_resource.rb
```diff
@@ -1,7 +1,8 @@
 require 'set'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter,
+                                   :jsonapi_toplevel_meta])
```
bf4: I think this should be a global config.  I can't see any reason someone would want to configure any of this from the controller.  Maybe if you're serving different versions of the JSON API standard, but that hasn't happened yet.  There's a PR in the JSON API to add revision at various levels.. But I think that would still be handled via a global `display resource revision` `display document revisions` etc. and the Serializer knows how to get the revision for a given resource.

> A JSON API document MAY include information about its implementation under a top level jsonapi member. 
>  If present, the value of the jsonapi member MUST be an object (a "jsonapi object").
> The jsonapi object MAY contain a version member whose value is a string indicating the highest JSON API version supported. 
>  This object MAY also contain a meta member, whose value is a meta object that contains non-standard meta-information.

Ref: https://github.com/json-api/json-api/pull/824

bf4:
ref: https://github.com/rails-api/active_model_serializers/pull/1050/files#diff-ae1e38e96a8bf28ae37918a65c448c73R7

lib/active_model/serializer/configuration.rb
```diff
@@ -8,6 +8,8 @@ module Configuration
         base.config.array_serializer = ActiveModel::Serializer::ArraySerializer
         base.config.adapter = :flatten_json
         base.config.jsonapi_resource_type = :plural
+        base.config.jsonapi_toplevel_member = false
+        base.config.jsonapi_version = '1.0'
```
bf4: should be in the JsonApi adapter IMHO

beauby: Yeah, that makes sense.

bf4:
Alternative implementation at https://github.com/rails-api/active_model_serializers/pull/1050 which was rebased to include this work

lib/active_model/serializable_resource.rb
beauby: My thought was that people might (although ti wouldn't be great) want to handle two APIs with one app, maybe for migrating to a newer version of the standard or something.

NullVoxPopuli:
needs rebase :-(

beauby:
Closing in favor of #1050.

PR #1149, opened 2015-09-14
Update README with nested included association example.

mattmueller:
Spent some time scratching my head today until I found out this was possible.

Hopefully including the example in the docs will save folks some time in the future when they come across the need to include nested associations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,12 +204,12 @@ Doesn't follow any specifc convention.
 This adapter follows 1.0 of the format specified in
 [jsonapi.org/format](http://jsonapi.org/format). It will include the associated
 resources in the `"included"` member when the resource names are included in the
-`include` option.
+`include` option. Including nested associated resources is also supported.
 
 ```ruby
-  render @posts, include: ['authors', 'comments']
+  render @posts, include: ['author', 'comments', 'comments.author']
   # or
-  render @posts, include: 'authors,comments'
+  render @posts, include: 'author,comments,comments.author'
 ```
 
 ## Installation


NullVoxPopuli:
looks good! :+1:

mattmueller:
:+1:

bacarini:
It is missing that doc on https://github.com/rails-api/active_model_serializers/blob/master/docs/general/adapters.md

README.md
```diff
@@ -204,12 +204,12 @@ Doesn't follow any specifc convention.
 This adapter follows 1.0 of the format specified in
 [jsonapi.org/format](http://jsonapi.org/format). It will include the associated
 resources in the `"included"` member when the resource names are included in the
-`include` option.
+`include` option. Including nested associated resources is also supported.
 
 ```ruby
-  render @posts, include: ['authors', 'comments']
+  render @posts, include: ['author', 'comments', 'comments.author']
```
beauby: It would be nice to include as well a hash example, like `[ :author, comments: [:author]]`.

beauby: It would also be nice to have written somewhere that including `comments.author` will automatically include the `comments` (i.e. all resources on a nested path will be included). At least, it is the case with the JsonApi adapter (the spec requires full linkage), but it would probably be good to enforce it on all adapters.

NullVoxPopuli: that's how it works on the json adapter, too. I thought it would be implied.

beauby: The docs should at least mention it I think.

Issue #1153, opened 2015-09-15
JSON adapter with empty array does not maintain root key


meltingice:
If you attempt to serialize an empty array with the JSON adapter and ArraySerializer, the root key becomes an empty string. This issue is present in 0.10.0.rc2 and the latest master branch commit at this time (`17c353826056bc3eac3e6d6a2fd081af175bb6ad`).

**Example Code**
``` ruby
class Api::V1::ArticlesController < Api::V1::BaseController
  def index
    @articles = []
    render json: @articles, each_serializer: Api::V1::ArticleSerializer
  end
end
```

**Expected**
``` json
{
  "articles": []
}
```

**Actual**
``` json
{
  "": []
}
```

beauby:
Thing is, when you have an empty array, there is no real way for AMS to infer the type of resources you would have had in it (at least not based on the resources themselves since, well there aren't any). Two possible workarounds:
+ explicitly specify the `root` option
+ make a custom collection serializer that inherits from `ArraySerializer` that sets the `root` for you.

NullVoxPopuli:
AMS could attempt to safe_constantize the model part of  `{model}Serializer` and derive the root key from that?

meltingice:
From a design standpoint, is there ever really an instance where you wouldn't have an array of homogenous object types? At least, for the JSON adapter? I don't see why AMS couldn't deduce the root key based on the serializer. The workarounds are fine, but it was definitely a surprise to me how it reacted in this instance. Doesn't seem like a good expected result.
--
Ryan LeFevre (@meltingice)
HODINKEE, Inc.
Sr. Software Engineer

    _____________________________

NullVoxPopuli:
@beauby @joaomdmoura how do you feel about inferring the root name in this scenario? Can you think of any down sides?

beauby:
I think we should do it but only if it cannot be inferred from the resources themselves (mainly because there are none), otherwise it will bring pain when serializing with a custom serializer (like `PostPreviewSerializer`).

NullVoxPopuli:
yeah, that's what I was thinking.

NullVoxPopuli:
Imma get a PR for this

NullVoxPopuli:
Addressed in #1156

bf4:
@meltingice are we discussing `Array.new` or `Article.none`?  

[...]

PR #1156, opened 2015-09-16
Serializing an empty array should maintain the root key

NullVoxPopuli:
Relates to: #1153

### DIFF
--- a/lib/active_model/serializer/array_serializer.rb
+++ b/lib/active_model/serializer/array_serializer.rb
@@ -5,9 +5,10 @@ module ActiveModel
       include Enumerable
       delegate :each, to: :@serializers
 
-      attr_reader :object, :root, :meta, :meta_key
+      attr_reader :object, :root, :meta, :meta_key, :options
 
       def initialize(resources, options = {})
+        @options = options
         @root = options[:root]
         @object = resources
         @serializers = resources.map do |resource|
@@ -27,7 +28,16 @@ module ActiveModel
 
       def json_key
         key = root || @serializers.first.try(:json_key) || object.try(:name).try(:underscore)
+        key ||= inferred_json_key
         key.try(:pluralize)
+      end
+
+      def inferred_json_key
+        serializer = options[:serializer]
+        return unless serializer
+        potential_model_name = serializer.name.sub('Serializer', '')
+        klass = potential_model_name.safe_constantize
+        klass.name.underscore if klass
       end
 
       def paginated?
--- a/test/action_controller/explicit_serializer_test.rb
+++ b/test/action_controller/explicit_serializer_test.rb
@@ -62,9 +62,23 @@ module ActionController
 
           render json: place, each_serializer: PlaceSerializer
         end
+
+        def render_empty_array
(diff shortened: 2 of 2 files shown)


lib/active_model/serializer/array_serializer.rb
```diff
@@ -27,9 +28,18 @@ def initialize(resources, options = {})
...
       def json_key
         key = root || @serializers.first.try(:json_key) || object.try(:name).try(:underscore)
+        key ||= inferred_json_key
         key.try(:pluralize)
       end
 
+      def inferred_json_key
+        serializer = options[:serializer]
+        return unless serializer
+        potential_model_name = serializer.name.sub('Serializer', '')
```
bf4: `serializer.name.remove('Serializer')` :shrug:

bf4:
Let's hold off on merging this till we clarify which arguments, `serializer` vs. `each_serializer` we should be passing to manually specify the serializer for a non-compliant resource (i.e. would fail the Lint).

See my comment in the OP's issue: https://github.com/rails-api/active_model_serializers/issues/1153#issuecomment-140624893

NullVoxPopuli:
@bf4  for an empty object (Array, or Hash) - would it matter which serializer vs each_serializer we're using? there isn't really anything to serialize aside from the root key.

it would either be

```
{
  root: {}
}
```
or
```
{
  roots: []
}
```

bf4:
So, AMS doesn't handle primitives, and when they find their way it, they are passed through as is. ( Specifically, when you're not rendering an object with a defined serializer.) 

How options are parsed

1. `serializable_resource = ActiveModel::SerializableResource.new(resource, options)`
1. `if serializable_resource.serializer?`
  - Where `serializer?` is `use_adapter? && !!(serializer)`
    - Where `use_adapter?`: 'True when no explicit adapter given, or explicit appear is truthy (non-nil); False when explicit adapter is falsy (nil or false)'
    - Where `serializer`:
      1. from explicit `:serializer` option, else
      2. implicitly from resource `ActiveModel::Serializer.serializer_for(resource)`
   - The `:serializer` option is removed from the serializer_opts hash
   - If the `:each_serializer` option is present, it is removed from the serializer_opts hash and set as the `:serializer` option
1. The serializer and adapter are created as
  1. `serializer_instance = serializer.new(resource, serializer_opts)`
  2. `adapter_instance = ActiveModel::Serializer::Adapter.create(serializer_instance, adapter_opts)`
1. If the `serializer_instance` was a `ArraySerializer` and the `:serializer` serializer_opts is present, then that serializer is passed into each resource: https://github.com/rails-api/active_model_serializers/blob/a54d237e2828fe6bab1ea5dfe6360d4ecc8214cd/lib/active_model/serializer/array_serializer.rb#L14-L16

The upshot of which is that
- for a collection, the collection serializer is the `:serializer` option and `:each_serializer` is used as the serializer for each resource in the collection
- for a single resource, the `:serializer` option is the resource serializer

NullVoxPopuli:
that is a fantastic explanation!

thanks -- I'll close this PR.

PR #1157, opened 2015-09-16
Add support for nested serializers.

beauby:
Following discussion on #1113.
This adds support for defining serializers inside other serializers, so that when serializing a nested resource, the most nested serializer is used in priority.

### DIFF
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -7,31 +7,6 @@ module ActiveModel
           @include_tree = IncludeTree.from_include_args(options[:include] || '*')
         end
 
-        def serializable_hash_for_collection(options)
-          serializer.map { |s| Attributes.new(s, instance_options).serializable_hash(options) }
-        end
-
-        def serializable_hash_for_single_resource(options)
-          hash = {}
-
-          core = cache_check(serializer) do
-            serializer.attributes(options)
-          end
-
-          serializer.associations(@include_tree).each do |association|
-            hash[association.key] =
-              if association.options[:virtual_value]
-                association.options[:virtual_value]
-              elsif association.serializer && association.serializer.object
-                Attributes.new(association.serializer, association.options.merge(include: @include_tree[association.key]))
-                  .serializable_hash(options)
-              end
-          end
-          result = core.merge hash
-
-          result
-        end
-
         def serializable_hash(options = nil)
           options ||= {}
 
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -7,6 +7,31 @@ module ActiveModel
           @include_tree = IncludeTree.from_include_args(options[:include] || '*')
         end
 
+        def serializable_hash_for_collection(options)
+          serializer.map { |s| Attributes.new(s, instance_options).serializable_hash(options) }
+        end
+
+        def serializable_hash_for_single_resource(options)
+          hash = {}
+
+          core = cache_check(serializer) do
+            serializer.attributes(options)
+          end
+
+          serializer.associations(@include_tree).each do |association|
+            hash[association.key] =
+              if association.options[:virtual_value]
+                association.options[:virtual_value]
+              elsif association.serializer && association.serializer.object
+                Attributes.new(association.serializer, association.options.merge(include: @include_tree[association.key]))
+                  .serializable_hash(options)
+              end
+          end
+          result = core.merge hash
+
+          result
+        end
+
         def serializable_hash(options = nil)
           options ||= {}
 
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -93,7 +93,7 @@ module ActiveModel
         Enumerator.new do |y|
           self.class._reflections.each do |reflection|
             next unless include_tree.key?(reflection.name)
-            y.yield reflection.build_association(self, instance_options)
+            y.yield reflection.build_association(self, instance_options.merge(_parent: self.class))
           end
         end
       end
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -93,7 +93,7 @@ module ActiveModel
         Enumerator.new do |y|
           self.class._reflections.each do |reflection|
             next unless include_tree.key?(reflection.name)
-            y.yield reflection.build_association(self, instance_options.merge(_parent: self.class))
+            y.yield reflection.build_association(self, instance_options)
           end
         end
       end
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -42,13 +42,13 @@ module ActiveModel
       def build_association(subject, parent_serializer_options)
         association_value = subject.send(name)
         reflection_options = options.dup
-        serializer_class = ActiveModel::Serializer.serializer_for(association_value, reflection_options)
+        serializer_class = subject.class.serializer_for(association_value, reflection_options)
 
         if serializer_class
           begin
             serializer = serializer_class.new(
               association_value,
-              serializer_options(parent_serializer_options, reflection_options)
+              serializer_options(parent_serializer_options, reflection_options, subject)
             )
           rescue ActiveModel::Serializer::ArraySerializer::NoSerializerError
             reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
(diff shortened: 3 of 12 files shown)


NullVoxPopuli:
ya have a lot of chained PRs :-p

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -9,40 +9,12 @@ def initialize(serializer, options = {})
```
beauby: This whole file is not required for this PR.

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,18 @@
+module ActiveModel::Serializer::Utils
+  module_function
+
+  def nested_lookup_paths(klass)
```
NullVoxPopuli: can you add some yardoc?

beauby: Done :+1:

test/fixtures/poro.rb
```diff
@@ -258,4 +260,34 @@ def maker
...
   attributes :id
 end
+
+class TweetSerializer < ActiveModel::Serializer
+  attributes :id, :body, :date
+
+  belongs_to :author
+  class AuthorSerializer < ActiveModel::Serializer
+    attributes :id
+  end
```
bf4: How is this different from `belongs_to: :author, serializer: TweetAuthorSerializer`?

lib/active_model/serializer/associations.rb
```diff
@@ -93,7 +93,7 @@ def associations(include_tree = DEFAULT_INCLUDE_TREE)
         Enumerator.new do |y|
           self.class._reflections.each do |reflection|
             next unless include_tree.key?(reflection.name)
-            y.yield reflection.build_association(self, instance_options)
+            y.yield reflection.build_association(self, instance_options.merge(_parent: self.class))
```
bf4: Why :`_parent` and not `:serializer`?

test/fixtures/poro.rb
beauby: One thing is that the namespacing makes it so that you don't have to use funky adapter names that you then have to specify manually.

lib/active_model/serializer/associations.rb
beauby: Good catch, that was a rebase mistake. It is actually unnecessary to supply it here. However, the name is `_parent` because it is passed down to a child serializer later on. The underscore is there because we don't want users to toy with this option.

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,25 @@
+module ActiveModel::Serializer::Utils
+  module_function
+
+  # Gives the existing lookup paths that are suffixes of the fully qualified path of `klass`.
+  # @param [Object] klass
+  # @return [Array<Symbol>]
+  def nested_lookup_paths(klass)
+    paths = klass.to_s.split('::').reverse.reduce([]) { |a, e| a + [[e] + Array(a.last)] }.reverse
```
bf4: Would you mind helping me read this code?

1. `TweetSerializer::AuthorSerializer.to_s.split('::').reverse` #=> ['AuthorSerializer', 'TweetSerializer']`
2. `_.reduce([]) {|result, element| result + [[element] + Array(result.last)] }.reverse`
  1. `paths = []`
    1. `paths = [] + [['AuthorSerializer'] + Array([])  ] #=> [['AuthorSerializer']]`
    1. `paths = [['AuthorSerializer']] + [ ['TweetSerializer'] + Array(['AuthorSerializer']) ] #=> [ ["AuthorSerializer"], ["TweetSerializer", "AuthorSerializer"] ]`
  1. `paths.reverse #=> ["TweetSerializer", "AuthorSerializer"], ["AuthorSerializer"]]`
3. ` ["TweetSerializer", "AuthorSerializer"], ["AuthorSerializer"]].map! {|path| path.join('::') } #=> ["TweetSerializer::AuthorSerializer", "AuthorSerializer"]`
4. `[TweetSerializer::AuthorSerializer", "AuthorSerializer"].select! {|path| Object.const_defined?(path, false) }.map! {|path| Object.const_get(path) }.push(Object) #=> [`TweetSerializer::AuthorSerializer, AuthorSerializer, Object]`

Is that right?

beauby: It is indeed. I admit that part was slightly cryptic.

bf4: So that as the test says

`Utils.nested_lookup_paths(TweetSerializer::ShareSerializer::AuthorSerializer) #=> [::TweetSerializer::ShareSerializer::AuthorSerializer,
                      ::ShareSerializer::AuthorSerializer,
                      ::AuthorSerializer,
                      ::Object]`

beauby: Yessir.

bf4: And 
`Utils.nested_lookup([::TweetSerializer, ::Object], 'ShareSerializer') #=> ::TweetSerializer::ShareSerializer` ?

beauby: I believe so.

bf4: So that

```ruby
          @share1 = Share.new
TweetSerializer.serializer_for(@share1) #=> TweetSerializer::ShareSerializer
ActiveModel::Serializer.serializer_for(@share1) #=> ShareSerializer
```

beauby: Exactly.

lib/active_model/serializer/reflection.rb
```diff
@@ -62,11 +62,12 @@ def build_association(subject, parent_serializer_options)
 
       private
 
-      def serializer_options(parent_serializer_options, reflection_options)
+      def serializer_options(parent_serializer_options, reflection_options, subject)
         serializer = reflection_options.fetch(:serializer, nil)
```
bf4: why is subject called parent?

beauby: Because, as any parent, it is its own subject, but its children's parent. The `serializer_options` is the hash that we build for passing down to the associated (i.e. child) serializer.

lib/active_model/serializer/utils.rb
bf4: So that you can build associations on tweet.author using `serializer: TweetSerializer::AuthorSerializer`?

beauby: Yes, exactly, _if available_, and safely fallback to any less specialized serializer that can handle an `Author`.

test/fixtures/poro.rb
bf4: So, this PR is more a syntactic sugar than a new feature?

NullVoxPopuli: that's sorta how I see it as well. But for me, the biggest benefit would be to greatly reduce the number of top level serializers that actually have VERY specific purposes

beauby: It is, indeed syntactic sugar. However, I could see in the future having, say the Json adapter default to including all nested relationships (default to `**` instead of `*` currently), in a kind of safe way since the user would provide all the subsequent serializers to make the graph explicit (maybe with a warning when there is an undefined serializer and possibly when using a fallback serializer).

lib/active_model/serializer/reflection.rb
beauby: If you feel strongly about this, and think of a better name, I don't mind changing it.

lib/active_model/serializer/utils.rb
```diff
@@ -0,0 +1,31 @@
...
+    paths.map! { |path| "#{path.join('::')}" }
+    paths.map! do |path|
```
NullVoxPopuli: @beauby can you add comments after each line, explaining what it does -- so that if debugging of this needs to happen in the future, we don't need to go through the exercise you and @bf4 went through?

beauby: I'll comment the first line, which might be pretty cryptic. The rest is pretty much self-explanatory, no?

beauby:
The tests currently fail because ruby <= 2.0.0 does not allow namespaced constants to be checked via `namespace.const_defined?`.

NullVoxPopuli:
how long is < ruby 2.0 supported for?

beauby:
@NullVoxPopuli Official support for 1.9.3 [has ended in Feb 2015](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/) (so we could probably drop support for 1.9.3 as well...). However, this issue still arises with ruby 2.0.0... I believe it was fixed in 2.0.1.

beauby:
Closing this in favor of #1193.

joaomdmoura:
Nice! #1193 looks awesome and way more mature

PR #1158, opened 2015-09-16
Add support for wildcards in nested includes

beauby:
Adds support for the following:
```ruby
render json: posts, include: 'author, comments.*'
# includes the author, the comments, and all of the comments' related resources

render json: posts, include: [:author, comments: '*']
# same, with the array format

render json: posts, include: 'author, comments.**'
# includes the author, the comments, and all of the comments' related resources, and their related resources, etc.
```
Note that the wildcard (`*`) only expands one level, so you can chain them if you want more and keep control over the depth of the inclusion (`'comments.*.*'` for instance), or you can specify a double wildcard (`**`) if you want infinite expansion.


lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -119,10 +119,26 @@ def relationships_for(serializer)
           Hash[serializer.associations.map { |association| [association.key, { data: relationship_value_for(association.serializer, association.options) }] }]
         end
 
+        def expand_includes(includes, associations)
```
NullVoxPopuli: can you write some yardoc about what this method does, and some example inputs and outputs?

test/action_controller/json_api/linked_test.rb
```diff
@@ -51,9 +51,9 @@ def render_resource_with_nested_include
             render json: @post, include: [comments: [:author]], adapter: :json_api
           end
 
-          def render_resource_with_nested_has_many_include
+          def render_resource_with_nested_has_many_include_wildcard
             setup_post
-            render json: @post, include: 'author.roles', adapter: :json_api
+            render json: @post, include: 'author.*', adapter: :json_api
           end
```
NullVoxPopuli: should there be an additional test instead of changing an existing one?

lib/active_model/serializer/adapter/json_api.rb
beauby: Will do!

test/action_controller/json_api/linked_test.rb
beauby: There should be many more tests, indeed.

beauby:
Now also works with (Flatten)Json adapters.

lib/active_model/serializer.rb
```diff
@@ -167,5 +167,25 @@ def self.get_serializer_for(klass)
         end
       end
     end
+
+    # Transforms an inclusion hash into an array of corresponding existing associations,
+    # and corresponding inclusion hashes.
+    # @param [Hash] includes
+    # @return [Array] an array of pairs [association, include_hash] for matching associations
+    def expand_includes(includes)
```
NullVoxPopuli: Should this format be different from the other format?
I'd prefer it to be the same, hash of hashes

beauby: This one has a different purpose. I'm not happy with the name of this method, because what it really does is lookup each included association and returns the corresponding `Association` and include hash. It just happens to expand stuff because at this stage we can't avoid it anymore.

beauby: Having this function just return a hash would force an other lookup of the associations down the road.

NullVoxPopuli: gotchya

lib/active_model/serializer/adapter/json.rb
```diff
@@ -2,10 +2,15 @@ class ActiveModel::Serializer::Adapter::Json < ActiveModel::Serializer::Adapter
         extend ActiveSupport::Autoload
         autoload :FragmentCache
 
+        def initialize(serializer, options = {})
```
NullVoxPopuli: I'm worried about continued changes to this adapter - because #1127 is so big

lib/active_model/serializer/adapter/json.rb
```diff
@@ -2,10 +2,15 @@ class ActiveModel::Serializer::Adapter::Json < ActiveModel::Serializer::Adapter
...
+          super
+          @included = ActiveModel::Serializer::Utils.include_args_to_hash(@options[:include] || '*')
+        end
+
         def serializable_hash(options = nil)
           options ||= {}
           if serializer.respond_to?(:each)
-            result = serializer.map { |s| FlattenJson.new(s).serializable_hash(options) }
+            result = serializer.map { |s| FlattenJson.new(s, include: @included).serializable_hash(options) }
```
bf4: could it go in serializable hash?

```ruby
included = ActiveModel::Serializer::Utils.include_args_to_hash(options[:include] || '*') # shouldn't include_args_to_hash handle the fallback to '*'?
  ... FlattenJson.new(s, include: included)...
else
  included_associations = serializer.expand_includes(included) # why is this the serializer's job?
  ...
end
```

bf4: looks like it could also be shared by json and jsonapi adapters

beauby: The name `expand_includes` was poorly chosen. What it does is take an "include hash" and spits out an array of pairs `[Association, RelatedIncludeHash]`, that's why it's the serializer's job in my opinion.

beauby: As for the fallback, I'm not too sure about delegating it to `include_args_to_hash` because we might want a different fallback depending on the adapter (`'', '*', '**').

beauby: I didn't get your comment about going into `serializable_hash` though. Are you refering to the `include: @included`?

lib/active_model/serializer/adapter/json.rb
```diff
@@ -13,27 +18,16 @@ def serializable_hash(options = nil)
...
-                    item.attributes(opts)
-                  end
+            included_associations = serializer.expand_includes(@included)
+            included_associations.each do |association, assoc_includes|
+              hash[association.key] =
+                if association.options[:virtual_value]
+                  association.options[:virtual_value]
+                elsif association.serializer && association.serializer.object
+                  FlattenJson.new(association.serializer,
+                                  association.options.merge(include: assoc_includes))
```
beauby: This line gives a rubocop warning about aligning method parameters that span multiple lines, although I believe the indentation is right here.

lib/active_model/serializer/adapter/json.rb
beauby: ping @bf4

test/action_controller/json_api/linked_test.rb
beauby: Actually, I don't believe there should be more integration tests for this PR. We should just thoroughly test the `expand_includes` method.

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: Have the first param on the next line, indented two spaces from the start of the F?

beauby: According to https://github.com/bbatsov/ruby-style-guide#no-double-indent it should be ok like that.

bf4: could be a bug in rubocop. I don't think it validates itself against the style guide and diverges at times

bf4: but there's definitley too much going on in that block

lib/active_model/serializer/adapter/json.rb
bf4: I believe I was, but it's not clear, is it?

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -2,6 +2,11 @@ module ActiveModel
   class Serializer
     module Adapter
       class Attributes < Base
+        def initialize(serializer, options = {})
+          super
+          @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
```
beauby: By default, include all the toplevel specified association (current behavior).

lib/active_model/serializer/adapter/json.rb
```diff
@@ -7,7 +7,7 @@ class Json < Base
 
         def serializable_hash(options = nil)
           options ||= {}
-          { root => Attributes.new(serializer).serializable_hash(options) }
+          { root => Attributes.new(serializer, instance_options).serializable_hash(options) }
```
beauby: Why was `instance_options` not passed along before?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -19,10 +20,11 @@ def initialize(serializer, options = {})
 
         def serializable_hash(options = nil)
           options ||= {}
+
           if serializer.respond_to?(:each)
-            serializable_hash_for_collection(serializer, options)
```
beauby: Not required for this PR, but passing the serializer reference along is useless here as it is available via an `attr_reader`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -19,10 +20,11 @@ def initialize(serializer, options = {})
...
+            serializable_hash_for_collection(options)
           else
-            serializable_hash_for_single_resource(serializer, options)
```
beauby: Not required for this PR, but passing the serializer reference along is useless here as it is available via an `attr_reader`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -34,10 +36,10 @@ def fragment_cache(cached_hash, non_cached_hash)
         private
 
         ActiveModel.silence_warnings do
-          attr_reader :included, :fieldset
+          attr_reader :fieldset
         end
 
-        def serializable_hash_for_collection(serializer, options)
```
beauby: Not required for this PR, but passing the serializer reference along is useless here as it is available via an `attr_reader`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -57,10 +59,10 @@ def serializable_hash_for_collection(serializer, options)
           hash
         end
 
-        def serializable_hash_for_single_resource(serializer, options)
```
beauby: Not required for this PR, but passing the serializer reference along is useless here as it is available via an `attr_reader`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -57,10 +59,10 @@ def serializable_hash_for_collection(serializer, options)
...
+        def serializable_hash_for_single_resource(options)
           primary_data = primary_data_for(serializer, options)
           relationships = relationships_for(serializer)
-          included = included_for(serializer)
+          included = included_for
```
beauby: Not required for this PR, but passing the serializer reference along is useless here as it is available via an `attr_reader`.
Here, `included_for` could be renamed to `included`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -123,19 +125,20 @@ def relationship_value_for(serializer, options = {})
         end
 
         def relationships_for(serializer)
-          Hash[serializer.associations.map { |association| [association.key, { data: relationship_value_for(association.serializer, association.options) }] }]
+          serializer.associations.each_with_object({}) do |association, hash|
+            hash[association.key] = { data: relationship_value_for(association.serializer, association.options) }
+          end
```
beauby: Not required for this PR, but it feels nicer this way.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -123,19 +125,20 @@ def relationship_value_for(serializer, options = {})
...
         end
 
-        def included_for(serializer)
-          included.flat_map { |inc|
-            association = serializer.associations.find { |assoc| assoc.key == inc.first }
-            _included_for(association.serializer, inc.second) if association
+        def included_for
+          serializer.associations(@include_tree).flat_map { |association|
```
beauby: This is where the magic happens. `Serializer#associations` now takes an `IncludeTree` as an optional parameter in order to filter out unwanted associations.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -57,10 +59,10 @@ def serializable_hash_for_collection(serializer, options)
           hash
         end
 
-        def serializable_hash_for_single_resource(serializer, options)
+        def serializable_hash_for_single_resource(options)
           primary_data = primary_data_for(serializer, options)
           relationships = relationships_for(serializer)
-          included = included_for(serializer)
+          included = included_resources(@include_tree)
```
beauby: I know there's a trend towards avoiding instance variables, but I think they make the situation really clear, compared to private accessors.

lib/active_model/serializer/adapter/attributes.rb
NullVoxPopuli: should the default value be move to the method?

beauby: No, because we want the flexibility to have different default values per adapter. For instance, the JsonApi adapter defaults to none (since the *relationship objects* themselves will be serialized anyways, we don't want to enforce the *related resources* to be included in the answer unless specified by the user).

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -13,7 +18,7 @@ def serializable_hash(options = nil)
               serializer.attributes(options)
             end
 
-            serializer.associations.each do |association|
+            serializer.associations(@include_tree).each do |association|
```
NullVoxPopuli: I like this

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: Do we have documented the difference between the two kinds of options and what is available? I know this is kinda beating a dead horse, but it seems to be more and more relevant

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -8,7 +8,8 @@ class JsonApi < Base
 
         def initialize(serializer, options = {})
           super
-          @included = ActiveModel::Serializer::Utils.include_args_to_hash(instance_options[:include])
+          @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include])
```
NullVoxPopuli: do we need to have the fully qualified namespace?

since we are already under the `Serializer` namespace, wouldn't just `IncludeTree` work?

lib/active_model/serializer/adapter/json.rb
beauby: Not yet, but here, even with virtually no information whatsoever, not passing along the `instance_options` means "no options can ever be set on the Json adapter".

lib/active_model/serializer/adapter/json_api.rb
NullVoxPopuli: because of the new associations @include stuff :-)

beauby: What do you mean?

lib/active_model/serializer/adapter/json.rb
bf4: I'm not sure why it wasn't passed along before. There are some differences in which options you'd want to send from a collection serializer.  In any case, best not to change that here, as it deserves its own discussion/tests/etct.

The difference

- `instance_options` are the options passed to the initializer
- `options`, if present, are an argument to the method

Previously, instance_options were just `@options`, which I found confusing when `options` was a local and `@options` was set at initialization.  To preserve method argument consistency, when I removed the instance vars, I chose `instance_options` as the name.

beauby: I get that, but what about my last comment?

lib/active_model/serializer/include_tree.rb
```diff
@@ -0,0 +1,77 @@
...
+    # Description
+
+    class IncludeTree
+      class << self
+        # Builds an IncludeTree from a comma separated list of dot separated paths (JSON API format).
+        # @example `'posts.author, posts.comments.upvotes, posts.comments.author'`
+        #
+        # @param [String] included
+        # @return [IncludeTree]
+        #
```
NullVoxPopuli: yay documentation!!!

but I think `@return` is supposed to be the last line of the comment?

lib/active_model/serializer/adapter/attributes.rb
bf4: Does this need to be in the initializer?  Does it need to be eager evaluated? Also, you could

```ruby
def initialize(*)
  super
  # use whatever the super method did
```

lib/active_model/serializer/include_tree.rb
```diff
@@ -0,0 +1,77 @@
...
+            hash.deep_merge!(include_tree)
+          end
+        end
+
+        def include_args_to_hash(included)
+          case included
+          when Symbol
+            { included => {} }
+          when Hash
+            included.each_with_object({}) { |(key, value), hash|
```
NullVoxPopuli: do end for multiline block?

lib/active_model/serializer/include_tree.rb
beauby: Not sure, there are many places in the current code base where the format is as above. I don't mind changing it, once we make a consistent choice for the future.

lib/active_model/serializer/include_tree.rb
```diff
@@ -0,0 +1,77 @@
...
+      def initialize(hash = {})
+        @hash = hash
+      end
+
+      def key?(key)
+        @hash.key?(key) || @hash.key?(:*) || @hash.key?(:**)
+      end
+
+      def [](key)
+        # TODO(beauby): Adopt a lazy caching strategy for generating subtrees.
```
NullVoxPopuli: :+1:

lib/active_model/serializer/include_tree.rb
beauby: So I'm not too sure since it's not technically a multiline block (there's only one line in the block), and this format is present in the codebase already. I don't mind changing it, but let's make it a lasting decision then.

lib/active_model/serializer.rb
```diff
@@ -5,7 +5,7 @@
 require 'active_model/serializer/configuration'
 require 'active_model/serializer/fieldset'
 require 'active_model/serializer/lint'
-require 'active_model/serializer/utils'
+require 'active_model/serializer/include_tree'
```
bf4: Not a blocker for this PR, but I don't find `IncludeTree` an immediately obvious  name->responsibility.  Maybe `IncludedAttributesTree`, which makes it sound like it should be in the `serializer` folder, maybe

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -2,6 +2,11 @@ module ActiveModel
   class Serializer
     module Adapter
       class Attributes < Base
+        def initialize(serializer, options = {})
+          super
+          @include_tree = IncludeTree.from_include_args(options[:include] || '*')
```
NullVoxPopuli: should just options[:include] be passed, and then in the method, do `include || '*'`

lib/active_model/serializer.rb
beauby: `IncludedAttributesTree` clearly does not work as this is currently used only for relationships. Moreover, it's general enough so it could virtually be used for representing any arborescent structure with wildcards. I don't mind changing the name, though.

lib/active_model/serializer/adapter/json.rb
NullVoxPopuli: @bf4 thanks for the explanation -- now if only comments about that were in the code :-)

lib/active_model/serializer/adapter/json_api.rb
bf4: I'd kind of like to simplify defining `def serializable_hash` in subclasses so that we can either `def serializable_hash(*); super; whatever` or `def _serializable_hash(options)` a method called by `serializable_hash`, or alternatively, even just use `read_attribute_for_serialization`

NullVoxPopuli: scratch that, I thought I was looking at a different adapter.

lib/active_model/serializer/adapter/attributes.rb
beauby: Could you be more specific? If you're referring to doing `@include_tree = IncludeTree.from_include_args(options[:include]) || '*'`, then it wouldn't work as 1. `#from_include_args` always returns an empty `IncludeTree` for "shady" inputs, and 2. even if that was not the case, we would still have to parse the `'*'`.

lib/active_model/serializer/adapter/json_api.rb
beauby: @bf4: Not sure I understand.

lib/active_model/serializer/associations.rb
```diff
@@ -82,14 +82,17 @@ def associate(reflection)
...
       end
 
+      # @param [IncludeTree] include_tree (optional)
       # @return [Enumerator<Association>]
       #
-      def associations
+      def associations(include_tree = nil)
         return unless object
 
+        include_tree ||= ActiveModel::Serializer::IncludeTree.from_string('*')
```
bf4: Style:  I'd rather 

```ruby
def associations(include_tree = self.class.default_include_tree)
...
end

def self.default_include_tree
  @default_include_tree ||= ActiveModel::Serializer::IncludeTree.from_string('*')
  @default_include_tree.dup
end
```

beauby: :+1: Will update immediately.

lib/active_model/serializer/associations.rb
```diff
@@ -82,14 +82,17 @@ def associate(reflection)
...
+
         Enumerator.new do |y|
           self.class._reflections.each do |reflection|
-            y.yield reflection.build_association(self, instance_options)
+            y.yield reflection.build_association(self, instance_options) if include_tree.key?(reflection.name)
```
bf4: Having a conditional at the end of a long line is hard to read (i.e. easy to miss).  Better:

```ruby
each do |reflection|
  next unless include_tree.key?(reflection.name)
  association = reflection.build_association(self, instance_options)
  y.yield association
...
end
```

lib/active_model/serializer/include_tree.rb
NullVoxPopuli: yeah, idk. according to the styleguide, this should all be on one line -- but it's so long.  I wonder if this should be extracted to a method, like you did with the string

lib/active_model/serializer/include_tree.rb
NullVoxPopuli: does this mean the PR merging should wait on this being done?

beauby: No, it does not. Just that it's something we will probably want to do in a subsequent PR.

lib/active_model/serializer/include_tree.rb
bf4: Long `class << self` blocks are hard to read.  Better to just use `def self.method_name`.

I only did it in Adapter because I was `class << Adapter` which is easier to read than `def Adapter.whatever` and a little less surprising, but also I was intending to get rid of it eventually.

lib/active_model/serializer.rb
NullVoxPopuli: `IncludedAssociationsTree`?

lib/active_model/serializer/include_tree.rb
bf4: Two private class methods all taking the `included` argument? Seems a latent object to me.

lib/active_model/serializer/associations.rb
beauby: Actually, I'm not sure it's the best solution, since it would expose the `self.default_include_tree` method to the public. Maybe make it a constant member of the class? We don't event need to `dup` actually.

lib/active_model/serializer/adapter/attributes.rb
NullVoxPopuli: maybe like this:

`IncludeTree.from_include_args(options[:include])`
```ruby
def from_include_args(included)
  new(include_args_to_hash(included || '*'))
end
```

lib/active_model/serializer/include_tree.rb
beauby: I tend to agree, although it's a pattern that is wildly used within rails. What do you think? I don't mind changing it.

lib/active_model/serializer/include_tree.rb
beauby: `included` are different things in those contexts.

lib/active_model/serializer/include_tree.rb
```diff
@@ -0,0 +1,77 @@
...
+        case
+        when @hash.key?(key)
+          self.class.new(@hash[key])
+        when @hash.key?(:*)
+          self.class.new(@hash[:*])
+        when @hash.key?(:**)
+          self.class.new(:** => {})
```
bf4: A `case` missing an `else` is a bug waiting to happen. In part, it doesn't tell me whether you intend a no-op or hadn't thought of what the behavior should be.

beauby: According to the Ruby Style Guide, one should avoid redundant else-clauses.

test/action_controller/json_api/linked_test.rb
bf4: > We should just thoroughly test the expand_includes method.

@beauby Yes, moar `unit` tests

lib/active_model/serializer/include_tree.rb
NullVoxPopuli: cool

lib/active_model/serializer/adapter/attributes.rb
beauby: Doesn't work for the reasons specified somewhere else: we need the flexibility of different fallbacks depending on the adapter.

test/action_controller/json_api/linked_test.rb
beauby: That method does not exist anymore.

bf4:
I think this PR would benefit from including documentation of usage as well as caveats.  I see this feature as a major bug report attractor, so would be great to just point to docs which describe what it can't be guaranteed to do well, or performantly.

lib/active_model/serializer/include_tree.rb
bf4: Link? I don't see it, nor do I think it's a style issue

lib/active_model/serializer/adapter/attributes.rb
NullVoxPopuli: gotchya.  What if the class-level method to get the include associations tree per-adapter? adapters could just override that method if they need to, and we could have a generic one in base?

beauby: Not sure I understand. Could you be more specific?

lib/active_model/serializer/include_tree.rb
NullVoxPopuli: is the intent to return nil for the else case?

if so, a method yardoc would be great :-)

beauby: My mistake, it was just my linter. Will update immediately.

lib/active_model/serializer/adapter/attributes.rb
NullVoxPopuli: maybe something like this?

```ruby
def self.included_associations(options)
  IncludeTree.from_include_args(options[:include] || '*')
end
```

and then `@include_tree = included_associations(options)` ?

idk

beauby:
My intent was to build it first for internal usage (the wildcards themselves make the nested serializers much easier to write), and document it in a subsequent PR.

lib/active_model/serializer/adapter/attributes.rb
beauby: Not sure it adds much value at this stage, but I am not opposed to it.

lib/active_model/serializer/include_tree.rb
beauby: Fixed.

lib/active_model/serializer/adapter/json_api.rb
bf4: my concern with ivars is that they're more buggy. If I type `@includ_tree` I get some method called on nil error which I have to track down.  If I call `includ_tree`, it blows up at load.  So, I tend not to make them outside of the initializers or getters and setters.  I get that there's probably some performance hit

lib/active_model/serializer/adapter/attributes.rb
bf4: sorta railsy

lib/active_model/serializer/include_tree.rb
bf4: naming?

lib/active_model/serializer/adapter/json_api.rb
NullVoxPopuli: that is true - :+1: @bf4

beauby: My concern is not that much towards performance than clarity: it's easy to define a local variable with the same name as a getter/setter. Plus `@var` makes it really clear where to look for when something is wrong with `@var`. I guess it's a tradeoff.

bf4:
@beauby Ok with me if you're doing it in steps like that.  I haven't had the issue this PR is solving, so I'm not a 'domain' expert on what's best for AMS in this regard

lib/active_model/serializer/include_tree.rb
beauby: They could basically be renamed `include_string` and `include_object` (the latter could be almost anything). Would that seem better to you?

lib/active_model/serializer.rb
```diff
@@ -1,11 +1,11 @@
 require 'thread_safe'
 require 'active_model/serializer/adapter'
 require 'active_model/serializer/array_serializer'
+require 'active_model/serializer/include_tree'
 require 'active_model/serializer/associations'
```
beauby: It has to be before `associations`, because we use it in there.

lib/active_model/serializer/adapter/attributes.rb
beauby: Let's keep it like that for now, and refactor it later if it makes stuff easier?

lib/active_model/serializer/adapter/json.rb
beauby: @bf4: what's your take on
> [...] here, even with virtually no information whatsoever, not passing along the instance_options means "no options can ever be set on the Json adapter".

bf4: I think it's out of scope for this PR

beauby: Sounds fair, will revert immediately. But we have to make sure we take care of it someday.

lib/active_model/serializer/adapter/json_api.rb
bf4: Basically just avoid the

```ruby
           options ||= {}
```

boilerplate

lib/active_model/serializer/include_tree.rb
bf4: I'd prefer `def self.method_name` unless you feel it's clearer or better otherwise

lib/active_model/serializer/include_tree.rb
bf4: Sure.  Mind you, I'm kind of tired, so anything that makes me think less sounds good :)

lib/active_model/serializer/adapter/json_api.rb
beauby: Oh I see. I'm all up for that, although I don't think it falls within the scope of this PR. Moreover, I think we should have `serializable_hash` defined in `Base`, and have serializers override `serializable_hash_for_(collection|single_resource)`.

lib/active_model/serializer/adapter/json.rb
bf4: Wanna create an issue and references this discussion?

lib/active_model/serializer/include_tree.rb
beauby: I don't, I'll move to `self.`.

lib/active_model/serializer/adapter/json_api.rb
bf4: pretty much the same as what I'm thinking.  Yes, outside of scope

lib/active_model/serializer/include_tree.rb
beauby: Well, no, actually, I do, because it allows for all class methods (private and public) to be kept in the same place.

lib/active_model/serializer/associations.rb
```diff
@@ -82,13 +84,15 @@ def associate(reflection)
         end
       end
 
+      # @param [IncludeTree] include_tree (optional)
```
bf4: if you don't mind, s/optional/defaults to DEFAULT_INCLUDE_TREE when not set/ or something like that?

beauby: Yup, sounds great, updating immediately.

lib/active_model/serializer/include_tree.rb
bf4: So, here's how I see this working:

```ruby
     module IncludeTreeHelper #or some such name
       module_function

       def include_string_to_hash(included)
          included.delete(' ').split(',').reduce({}) do |hash, path|
            include_tree = path.split('.').reverse_each.reduce({}) { |a, e| { e.to_sym => a } }
            hash.deep_merge!(include_tree)
          end
       end

        def include_args_to_hash(included)
          case included
          when Symbol
            { included => {} }
          when Hash
            included.each_with_object({}) { |(key, value), hash|
              hash[key] = include_args_to_hash(value)
            }
          when Array
            included.reduce({}) { |a, e| a.merge!(include_args_to_hash(e)) }
          when String
            include_string_to_hash(included)
          else
            {}
          end
        end

     end
      def self.from_string(included)
        new(InlcudeTreeHelper.include_string_to_hash(included))
      end

      def self.from_include_args(included)
        new(IncludeTreeHelper.include_args_to_hash(included))
      end
```

bf4: i.e. two class methods that make use of two functions

beauby: Looks good to me, updating immediately.

bf4: Also, some of these styles will need to be autocorrected if we merge https://github.com/rails-api/active_model_serializers/pull/1183

bf4: I don't think `Helper` is a good name, but I wanted to just get the idea across :)

beauby: Yeah, I thought of `Parsing` instead, what do you think?

bf4: or Transformation.. but since it's internal so far, I'm not too concerned

lib/active_model/serializer/include_tree.rb
```diff
@@ -0,0 +1,77 @@
...
+        else
+          nil
```
bf4: @beauby You're saying Rubocop didn't like this?

lib/active_model/serializer/include_tree.rb
```diff
@@ -0,0 +1,77 @@
+module ActiveModel
+  class Serializer
+    class IncludeTree
+      module Parsing
```
bf4: This is the same as `ActiveModel::Serializer::Utils`, right?

lib/active_model/serializer/include_tree.rb
beauby: I am. https://travis-ci.org/rails-api/active_model_serializers/jobs/81323442

lib/active_model/serializer/include_tree.rb
beauby: Yes.

joaomdmoura:
Nice!  :+1:

PR #1159, opened 2015-09-16
Improve jsonapi include

beauby:
Building the `included` array in place, in order to save some temporary allocations and lots of `uniq!` calls. Plus, this prevents infinite include loops.

This is based on #1158.


NullVoxPopuli:
Review pending merge of #1158 
- looks like most of the changes here are from that PR.

beauby:
Those changes have been incorporated into #1158.

PR #1165, opened 2015-09-17
WIP: Serializer doesn't behave well with an attribute called `options`

thiagogabriel:
I have an attribute called options being serialized, but it causes a
`stack level too deep` on my controller and it started on 0.10.0.rc3.

Using AMS test suit I checked it happens because of `options` attribute.

The expected is:
`{:name=>"Name 1", :description=>"Description 1", :options=>'one'}`

but with @adapter.serializable_hash(options) I get:
`{:name=>"Name 1", :description=>"Description 1", :options=>{}}`

and with @serializable_resource.serializable_hash(options) I get:
`{:name=>"Name 1", :description=>"Description 1"}`

### DIFF
--- a/test/serializable_resource_test.rb
+++ b/test/serializable_resource_test.rb
@@ -3,8 +3,8 @@ require 'test_helper'
 module ActiveModel
   class SerializableResourceTest < Minitest::Test
     def setup
-      @resource = Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1' })
-      @serializer = ProfileSerializer.new(@resource)
+      @resource = Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1', options: 'one' })
+      @serializer = ProfileWithOptionsSerializer.new(@resource)
       @adapter = ActiveModel::Serializer::Adapter.create(@serializer)
       @serializable_resource = ActiveModel::SerializableResource.new(@resource)
     end
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -68,6 +68,10 @@ class ProfilePreviewSerializer < ActiveModel::Serializer
   attributes :name
 end
 
+class ProfileWithOptionsSerializer < ActiveModel::Serializer
+  attributes :name, :description, :options
+end
+
 Post     = Class.new(Model)
 Like     = Class.new(Model)
 Author   = Class.new(Model)


thiagogabriel:
@joaomdmoura is there any temporary recommendation (workaround) for this case different than rename my `options` attribute?

thiagogabriel:
Update:
The problem is not exactly with `SerializableResource`, but was introduced on rc3.

bf4:
I think I know what this is.  There's some confusion in the code between options locals vs. instance variables/methods.. Want me to take a stab at it? Thanks for the tests!

thiagogabriel:
@bf4 Sure. Thanks :+1:
Maybe it is better to move or include new tests to the proper place, since the bug is not exactly on `SerializableResource`.

bf4:
And you said this broke between rc2 and rc3?  I haven't isolated it as quickly as I thought

test/serializable_resource_test.rb
```diff
@@ -3,8 +3,8 @@
 module ActiveModel
   class SerializableResourceTest < Minitest::Test
     def setup
-      @resource = Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1' })
-      @serializer = ProfileSerializer.new(@resource)
+      @resource = Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1', options: 'one' })
+      @serializer = ProfileWithOptionsSerializer.new(@resource)
       @adapter = ActiveModel::Serializer::Adapter.create(@serializer)
       @serializable_resource = ActiveModel::SerializableResource.new(@resource)
```
bf4: The PR passes if we correct 

```diff
+   @serializer = ProfileWithOptionsSerializer.new(@resource)
    @adapter = ActiveModel::Serializer::Adapter.create(@serializer)		       
-   @serializable_resource = ActiveModel::SerializableResource.new(@resource)
+ 	@serializable_resource = ActiveModel::SerializableResource.new(@resource, serializer: ProfileWithOptionsSerializer)
```

So, I'm inclined to think this test doesn't reproduce your issue. :(

thiagogabriel:
@bf4 I changed where my problem is happening but it broke many tests. Can you take a look on the new broken tests? Replacing `try!` with `try` fixes all the broken tests but the `options` values are ignored.

test/serializable_resource_test.rb
bf4: There's a chance this PR might fix it https://github.com/rails-api/active_model_serializers/pull/1166

lib/active_model/serializer.rb
```diff
@@ -136,7 +136,7 @@ def attributes(options = {})
 
       attributes.each_with_object({}) do |name, hash|
         unless self.class._fragmented
-          hash[name] = send(name)
+          hash[name] = @object.try!(name) || public_send(name)
         else
```
bf4: such meta (why `try!` over `try`?)

bf4:
@thiagogabriel see https://github.com/rails-api/active_model_serializers/pull/1165#discussion_r39708239 for the reason for the test failure

Also, `attributes` and `attribute` both define methods on the Serializer that delegate to the object, so there should be no need to check the object for the method

lib/active_model/serializer.rb
thiagogabriel: Because `try` checks for `@object.respond_to?` which returns false.

thiagogabriel:
@bf4 I'll take a better look on you PR #1165 tomorrow.

bf4:
@thiagogabriel cool, and see if we can get a good failing test.. you might need to bring over more of your business logic, as well, as your Ruby version, Rails version, AMS version, backtrace, etc.

thiagogabriel:
@bf4 I'm checking master now. Sorry for my delay.

test/serializable_resource_test.rb
thiagogabriel: You are right about my wrong test. The test passed but the assertions are both wrong
```
[1] pry(#<ActiveModel::SerializableResourceTest>)> @adapter.serializable_hash(options)
=> {:name=>"Name 1", :description=>"Description 1", :options=>{}}
[2] pry(#<ActiveModel::SerializableResourceTest>)> @serializable_resource.serializable_hash(options)
=> {:name=>"Name 1", :description=>"Description 1", :options=>{}}
```

It is fixed on master, probably #1166 
```
@adapter.serializable_hash(options)
=> {:name=>"Name 1", :description=>"Description 1", :options=>"one"}
[2] pry(#<ActiveModel::SerializableResourceTest>)> @serializable_resource.serializable_hash(options)
=> {:name=>"Name 1", :description=>"Description 1", :options=>"one"}
```

PR #1166, opened 2015-09-17
Prefer methods over instance variables

bf4:
Just a refactor. 

Also, extracted latent Adapter::CachedSerializer class. See comments

Basically, wherever `options` was a method argument in a class, I renamed the `@options` ivar to `instance_options` and added an attr_reader.  I think it's a good practice to prefer getters/setters over ivars as it gives better failures and encapsulation. 

Done in response to https://github.com/rails-api/active_model_serializers/pull/1165 which describes a failure where the `options` attribute is confused with `options`.

While changing the options ivars, I just changed all of them.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -42,16 +42,16 @@ module ActiveModel
     end
 
     def self.inherited(base)
-      base._attributes = self._attributes.try(:dup) || []
-      base._attributes_keys = self._attributes_keys.try(:dup) || {}
+      base._attributes = _attributes.try(:dup) || []
+      base._attributes_keys = _attributes_keys.try(:dup) || {}
       base._cache_digest = digest_caller_file(caller.first)
       super
     end
 
     def self.attributes(*attrs)
       attrs = attrs.first if attrs.first.class == Array
-      @_attributes.concat attrs
-      @_attributes.uniq!
+      _attributes.concat(attrs)
+      _attributes.uniq!
 
       attrs.each do |attr|
         define_method attr do
@@ -62,8 +62,8 @@ module ActiveModel
 
     def self.attribute(attr, options = {})
       key = options.fetch(:key, attr)
-      @_attributes_keys[attr] = { key: key } if key != attr
-      @_attributes << key unless @_attributes.include?(key)
+      self._attributes_keys[attr] = { key: key } if key != attr
+      self._attributes << key unless _attributes.include?(key)
 
       ActiveModelSerializers.silence_warnings do
         define_method key do
@@ -73,16 +73,16 @@ module ActiveModel
     end
 
     def self.fragmented(serializer)
-      @_fragmented = serializer
+      self._fragmented = serializer
     end
 
     # Enables a serializer to be automatically cached
     def self.cache(options = {})
-      @_cache = ActionController::Base.cache_store if Rails.configuration.action_controller.perform_caching
-      @_cache_key = options.delete(:key)
-      @_cache_only = options.delete(:only)
-      @_cache_except = options.delete(:except)
-      @_cache_options = (options.empty?) ? nil : options
+      self._cache = ActionController::Base.cache_store if Rails.configuration.action_controller.perform_caching
+      self._cache_key = options.delete(:key)
+      self._cache_only = options.delete(:only)
+      self._cache_except = options.delete(:except)
+      self._cache_options = (options.empty?) ? nil : options
     end
 
     def self.serializer_for(resource, options = {})
@@ -91,7 +91,7 @@ module ActiveModel
       elsif resource.respond_to?(:to_ary)
         config.array_serializer
       else
-        options.fetch(:serializer, get_serializer_for(resource.class))
+        options.fetch(:serializer) { get_serializer_for(resource.class) }
       end
     end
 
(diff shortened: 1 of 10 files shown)


lib/active_model/serializer.rb
```diff
@@ -91,7 +91,7 @@ def self.serializer_for(resource, options = {})
       elsif resource.respond_to?(:to_ary)
         config.array_serializer
       else
-        options.fetch(:serializer, get_serializer_for(resource.class))
+        options.fetch(:serializer) { get_serializer_for(resource.class) }
```
bf4: not required for PR, just lazy evals the method when the fetch fails

lib/active_model/serializer.rb
```diff
@@ -104,17 +104,40 @@ def self.root_name
...
+        serializer_class = serializer_class_name.safe_constantize
+
+        if serializer_class
+          serializer_class
+        elsif klass.superclass
+          get_serializer_for(klass.superclass)
+        end
+      end
+    end
+
```
bf4: not required from PR, just grouped the class methods together

lib/active_model/serializer.rb
```diff
@@ -143,29 +166,10 @@ def attributes(options = {})
...
-    end
-
-    def self.digest_caller_file(caller_line)
-      serializer_file_path = caller_line[CALLER_FILE]
-      serializer_file_contents = IO.read(serializer_file_path)
-      Digest::MD5.hexdigest(serializer_file_contents)
-    end
-
-    attr_reader :options
+    private # rubocop:disable Lint/UselessAccessModifier
```
bf4: :shrug:

lib/active_model/serializer/adapter.rb
```diff
@@ -101,43 +102,12 @@ def fragment_cache(*args)
...
-        @cached_serializer = serializer
-        @klass             = @cached_serializer.class
-        if is_cached?
-          @klass._cache.fetch(cache_key, @klass._cache_options) do
-            yield
-          end
-        elsif is_fragment_cached?
-          FragmentCache.new(self, @cached_serializer, @options).fetch
-        else
+        CachedSerializer.new(serializer).cache_check(self) do
```
bf4: Extraction of latent class not required for PR.  All the other methods are essentially private methods in the service of `cache_check` that only work if cache_check is first called to set the ivars

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -0,0 +1,45 @@
+module ActiveModel
+  class Serializer
+    class Adapter
+      class CachedSerializer
+        def initialize(serializer)
+          @cached_serializer = serializer
+          @klass             = @cached_serializer.class
+        end
```
bf4: retained ivars from adapter for ease of 1to1 review of extraction

test/fixtures/poro.rb
```diff
@@ -60,7 +60,7 @@ class ProfileSerializer < ActiveModel::Serializer
   attributes :name, :description
 
   def arguments_passed_in?
-    options[:my_options] == :accessible
+    instance_options[:my_options] == :accessible
```
bf4: I thought this was kind of interesting

lib/active_model/serializer.rb
```diff
@@ -104,17 +104,40 @@ def self.root_name
...
     def initialize(object, options = {})
-      @object = object
-      @options = options
-      @root = options[:root]
-      @meta = options[:meta]
-      @meta_key = options[:meta_key]
-      @scope = options[:scope]
-
-      scope_name = options[:scope_name]
+      self.object = object
```
NullVoxPopuli: should we be preferring this type of setting / getting for instance attributes?, rather than `@instance_variable`

lib/active_model/serializer.rb
NullVoxPopuli: I think this is how class method are supposed to be declared private: http://stackoverflow.com/a/4953043/356849

lib/active_model/serializer/adapter.rb
```diff
@@ -80,11 +81,11 @@ def self.inherited(subclass)
         ActiveModel::Serializer::Adapter.register(subclass.to_s.demodulize, subclass)
       end
 
-      attr_reader :serializer
+      attr_reader :serializer, :instance_options
 
       def initialize(serializer, options = {})
         @serializer = serializer
-        @options = options
+        @instance_options = options
```
NullVoxPopuli: should this be `self.instance_options` like the other `self` accesses? 
or is the `self` usage only for `_`'ed attributes?

lib/active_model/serializer/adapter.rb
NullVoxPopuli: nice

NullVoxPopuli:
I'm mostly just confused on when `self` should be used

lib/active_model/serializer/adapter.rb
bf4: there's no accessor here, just a reader..

bf4:
@NullVoxPopuli here's the rules for `self` in Ruby:

- Is there a method to get the attribute? Don't use self. Ruby will look for the method unless you've defined a local above it (don't do that)
- Are you setting an attribute? Use self to ensure you're setting the attribute, and not creating a local variable

lib/active_model/serializer.rb
bf4: > why prefer self over just the attribute? 

Not sure what you mean by that.

see my answer https://github.com/rails-api/active_model_serializers/pull/1166#issuecomment-141119303

lib/active_model/serializer.rb
bf4: not a class method :)

NullVoxPopuli: oh I guess I'm a smidge dysfunctional in the morning. ha

why the rubocop error?

NullVoxPopuli: you could get rid of private?

bf4: And yet

```ruby
end
```

which is, I think, why rubocop thinks it's an unnecessary private

lib/active_model/serializer.rb
```diff
@@ -50,8 +50,8 @@ def self.inherited(base)
 
     def self.attributes(*attrs)
       attrs = attrs.first if attrs.first.class == Array
-      @_attributes.concat attrs
-      @_attributes.uniq!
+      self._attributes.concat(attrs)
+      self._attributes.uniq!
```
bf4: So, @beauby you're right that these don't need self

NullVoxPopuli:
:+1:

PR #1167, opened 2015-09-17
Delegate Serializer.attributes to Serializer.attribute

bf4:


lib/active_model/serializer.rb
```diff
@@ -50,20 +50,16 @@ def self.inherited(base)
...
     def self.attributes(*attrs)
       attrs = attrs.first if attrs.first.class == Array
-      @_attributes.concat attrs
-      @_attributes.uniq!
 
       attrs.each do |attr|
-        define_method attr do
-          object && object.read_attribute_for_serialization(attr)
-        end unless method_defined?(attr) || _fragmented.respond_to?(attr)
+        attribute(attr)
```
beauby: :+1:

lib/active_model/serializer.rb
```diff
@@ -50,20 +50,16 @@ def self.inherited(base)
...
       end
     end
 
     def self.attribute(attr, options = {})
       key = options.fetch(:key, attr)
-      @_attributes_keys[attr] = { key: key } if key != attr
-      @_attributes << key unless @_attributes.include?(key)
+      self._attributes_keys[attr] = { key: key } if key != attr
```
beauby: `self.` is redundant here

lib/active_model/serializer.rb
NullVoxPopuli: much cleaner!

lib/active_model/serializer.rb
beauby: c.f. https://github.com/bbatsov/ruby-style-guide#no-self-unless-required

NullVoxPopuli: so, because `_attributes` is a hash, does that mean it's not ok, because it's being accessed and the `attr` value is being set?

or is it ok, because *something* is getting set using `_attributes`?

lib/active_model/serializer.rb
```diff
@@ -50,20 +50,16 @@ def self.inherited(base)
...
+      self._attributes << key unless self._attributes.include?(key)
```
NullVoxPopuli: I think this second use of _attributes violates the reference @beauby linked.... but I think it makes sense here.... so... :-\

lib/active_model/serializer.rb
bf4: I've removed all the unnecessary selfs, such as this. If you think from a lookup point of view, `_attributes` is a method that returns something that I think call `:[]=` on

lib/active_model/serializer.rb
bf4: fixed

lib/active_model/serializer.rb
NullVoxPopuli: that's what I was thinking. Just trying to help define our conventions :-)

lib/active_model/serializer.rb
NullVoxPopuli: what was fixed? (maybe I'm just missing it?)
`self._attributes << key unless self._attributes.include?(key)`
but, is this --------------------------> ^ self supposed to be here?

NullVoxPopuli:
needs rebase

lib/active_model/serializer.rb
bf4: n/m wrong pr.. juggling...

bf4: fixed fureal

PR #1169, opened 2015-09-17
move model_name into serializer, so it can get overwritten by serializer, similar to 'root'

mreinsch:
I'm actually no quite sure why json_api uses `serializer.object.class.model_name` directly, instead of `serializer.json_key`. Then `resource_identifier_type_for` could be implemented like this:

      if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
        serializer.json_key
      else
        ActiveSupport::Inflector.pluralize(serializer.json_key)
      end
    end

The `json_key` is further used in json api adapter, line 11: https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json_api.rb#L11 - shouldn't this be unified?

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -122,8 +122,12 @@ module ActiveModel
       end
     end
 
+    def object_model_name
+      object.class.model_name
+    end
+
     def json_key
-      @root || object.class.model_name.to_s.underscore
+      @root || object_model_name.to_s.underscore
     end
 
     def attributes(options = {})
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -62,10 +62,11 @@ class ActiveModel::Serializer::Adapter::JsonApi < ActiveModel::Serializer::Adapt
         end
 
         def resource_identifier_type_for(serializer)
+          model_name = serializer.object_model_name
           if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+            model_name.singular
           else
-            serializer.object.class.model_name.plural
+            model_name.plural
           end
         end
 


beauby:
There has been a lot of refactoring going on with the JsonApi adapter, and your concern about `serializer.object.class.model_name` is legitimate. The general situation about handling the `type` attribute is not satisfying for the moment because we should have a way to 1. change the format globally (the first step was the plural/singular config, but we may want to support stuff like dasherized keys instead of underscored, and currently the namespaces don't play very well with the `type` attribute), 2. override the type in a serializer (which your PR offers, on top of fixing an inconsistency). However, I think an overridden `type` attribute should not be subject to inflection. This is arguable, but I believe users who actively decide to modify the type should know what they are doing.

mreinsch:
@beauby , regarding point 2) that an overridden type attribute should not be subject of inflection: I think the adapter design makes this a bit more difficult, and I can imagine situations where you'd want plural for JSON API, but singular for some legacy API.

lib/active_model/serializer.rb
```diff
@@ -122,8 +122,12 @@ def initialize(object, options = {})
       end
     end
 
+    def object_model_name
+      object.class.model_name
+    end
+
```
bf4: I like the encapsulation.. would like to think a bit about adding another method, and also the name of this method.. or should we just use `serializer.json_key` and let you overwrite that?

mreinsch: using the 'serializer.json_key' is what I suggested in the description of the pull request as an alternative (I had this code already done before I realised this). Using 'serializer.json_key' would probably make more sense, even though the json api adapter then needs to pluralise it.

jfelchner:
I worked on #1029 a bit last night and should have a new draft out today which will work better without diving into sub-hashes.

It'll also allow the user to define arbitrary (and separate) transformers for the `root`, the `attributes`, the `types` and the `relationships`.

I agree with @beauby that if the user is overriding something that it should not be subject to inflection.

I think the correct call here is to have two more methods on the serializer which return `singular_model_name` and `plural_model_name`.  These would do the inflection by default.  They could be overridden by the user to do whatever they wanted and then the adapter would just call `singular_model_name`/`plural_model_name`.

That would give the user the most flexibility.  For the majority case, they'd just override `model_name` and let the default inflectors do their job.  For the custom case, they could override one or both of the `*_model_name` methods.

jfelchner:
As it currently stands the serializer is too coupled to the adapter and needs to be broken apart, but I think the solution above will work for now.

lib/active_model/serializer.rb
jfelchner: @bf4 IMO I don't think the serializer should have any methods on it relating to a specific adapter.

mreinsch:
@jfelchner sounds good, looking forward to seeing this in master :)

Issue #1170, opened 2015-09-17
problem with loading ActiveModel::Serializer::Associations


shicholas:
here's my stacktrace:

```
/Users/nick/.gem/ruby/2.2.3/gems/active_model_serializers-0.10.0.rc3/lib/active_model/serializer.rb:15:in `<class:Serializer>': uninitialized constant ActiveModel::Serializer::Associations (NameError)
        from /Users/nick/.gem/ruby/2.2.3/gems/active_model_serializers-0.10.0.rc3/lib/active_model/serializer.rb:4:in `<module:ActiveModel>'
        from /Users/nick/.gem/ruby/2.2.3/gems/active_model_serializers-0.10.0.rc3/lib/active_model/serializer.rb:3:in `<top (required)>'
        from /Users/nick/.gem/ruby/2.2.3/gems/active_model_serializers-0.10.0.rc3/lib/active_model_serializers.rb:21:in `<top (required)>'
```

It seemed to work fine when I required the associations file before ``active_model_serializers.``

FWIW I'm developing using AMS in a rails engine. I don't think that should matter, but maybe it does.

NullVoxPopuli:
is this just when booting your app?

shicholas:
yes

shicholas:
* booting the app with the engine where I'm using AMS .10.rc3 in.

NullVoxPopuli:
is your engine open source, or is there a way you can make an example engine demonstrating the problem?

shicholas:
It's not but I can try to reproduce w/ an open-source project.

bf4:
Looks like it may be a race condition.  Which Ruby are you on?

https://github.com/rails-api/active_model_serializers/blob/3f0794bd390521616c1270d1350b95ad31e0731c/lib/active_model/serializer.rb#L7-L15

should probably

```diff
-  autoload :Configuration
-    autoload :Associations
+  require 'ams/configuration'
+  require 'ams/associations'
    include Configuration
    include Associations
```

you can `bundle open active_model_serializers` try that, if it works, `bundle gem pristine active_model_serializers` and submit it as a PR :)

shicholas:
ruby 2.2.3. and submitted a PR. Adding a require with the include did the trick.

bf4:
yay!

NullVoxPopuli:
@shicholas is this resolved?

shicholas:
Yup, ha

shicholas:
*thanks

PR #1171, opened 2015-09-17
add require statements to top of file

shicholas:
Based on
https://github.com/rails-api/active_model_serializers/issues/1170#issuecomment-141184047

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -1,4 +1,6 @@
 require 'thread_safe'
+require_relative 'serializer/configuration'
+require_relative 'serializer/associations'
 
 module ActiveModel
   class Serializer
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -1,16 +1,14 @@
 require 'thread_safe'
-require 'active_model/serializer/configuration'
+require 'active_model/serializer/adapter'
+require 'active_model/serializer/array_serializer'
 require 'active_model/serializer/associations'
+require 'active_model/serializer/configuration'
+require 'active_model/serializer/fieldset'
+require 'active_model/serializer/lint'
+require 'active_model/serializer/utils'
 
 module ActiveModel
   class Serializer
-    extend ActiveSupport::Autoload
-
-    autoload :ArraySerializer
-    autoload :Adapter
-    autoload :Lint
-    autoload :Fieldset
-    autoload :Utils
     include Configuration
     include Associations
 
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -1,6 +1,6 @@
 require 'thread_safe'
-require_relative 'serializer/configuration'
-require_relative 'serializer/associations'
+require 'active_model/serializer/configuration'
+require 'active_model/serializer/associations'
 
 module ActiveModel
   class Serializer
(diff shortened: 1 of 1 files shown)


lib/active_model/serializer.rb
```diff
@@ -1,14 +1,14 @@
 require 'thread_safe'
+require_relative 'serializer/configuration'
+require_relative 'serializer/associations'
```
bf4: I'd rather use the full require `active_model/serializer/configuration` without further discussion

bf4:
Awesome, thanks!

lib/active_model/serializer.rb

shicholas: np I'll update this.

NullVoxPopuli:
Is there any sort of test you can add for this @shicholas, so your #1170 doesn't happen in the future?

shicholas:
good question, I'm kind of stuck but I can't think of a test that won't introduce a ton of test dependencies (e.g. a ``test/dummy`` rails app that is present in rails engines). Any thoughts? I'm happy to implement it.

shicholas:
perhaps we can get rid of all ``autoload`` like @bf4 mentioned here: http://www.benjaminfleischer.com/2013/07/18/ruby-requires-confusion/.

NullVoxPopuli:
perhaps. I could get on board with that, even if there is a small performance and memory penalty in certain cases.
@beauby, @joaomdmoura thoughts on that?

bf4:
> perhaps we can get rid of all autoload like @bf4 mentioned here: http://www.benjaminfleischer.com/2013/07/18/ruby-requires-confusion/.

Nice, link!  I should note that Rubies 2.0+ are threadsafe with autoload ref: https://github.com/headius/thread_safe/issues/15#issuecomment-28936120

Here's some links on require vs. require_relative performance in rspec

- http://rubyforge.org/pipermail/rspec-users/2011-November/020760.html
- https://github.com/rspec/rspec-support/pull/45
- https://github.com/rspec/rspec-core/pull/1348
- https://github.com/rspec/rspec-core/pull/502

bf4:
I'm :+1: on getting rid of the autoload, but I'm okay with lazy loading

jfelchner:
I would :heart: getting rid of the autoloads.  Also `require_relative` gives almost [no performance improvement](https://github.com/rspec/rspec-core/commit/af5f60aba5a54bb33be0af196ab0a727319dc15c) to `require` so I think we should just go with `require` as `require_relative` has its own problems.

shicholas:
I pushed an update to use all require statements. Thanks for the links, @bf4, I tried playing around w/ some autoload stuff to no avail (could be user error :)).

NullVoxPopuli:
Well, I'd say if we are going to do away with autoload, and since the issue was with autoload, and that coming up up with a proper test would be really weird / tricky / hacky, I think this in good to go as in. Good work!

shicholas:
I'm having a tough time deciphering the build error messages; it looks like ``simplecov``?

NullVoxPopuli:
Our appveyer build breaks 80% of the time. We're working on it. Lol

shicholas:
sorry, I realized that the test failures were my fault b/c I only removed the ``autoload`` statements in ``lib/active_model/serializer.rb`` and not elsewhere. This is now updated and should now pass.

bf4:
Note of caution, we're not actually using Kernel.autoload but ActiveSupport::Autoload, which *cannot* co-exist with require.  Anyhow once we're using only requires, it won't matter.

Also, requires should *always* be one direction, never circular

NullVoxPopuli:
:+1:

PR #1178, opened 2015-09-18
env CAPTURE_STDERR=false lets devs see hard failures

bf4:
I noticed that if I had runtime errors, I wouldn't see the stacktrace
or exception message.

Turns out, this is something about the at_exit resolution of io
capturing between capture_errors and Minitest.

I couldn't figure out how to modify the io for Minitest without a lot of
code, so this is a safety valve for devs :) :(

MiniTest 4

- MiniTest::Unit.output = my_thing

Minitest 5

- Make a plugin can makes a reporter that holds onto a special io

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -38,8 +38,14 @@ else
   end
 end
 
-require 'capture_warnings'
-CaptureWarnings.new(_fail_build = true).execute!
+# If there's no failure info, try disabling capturing stderr:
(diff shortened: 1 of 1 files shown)


bf4:
@jfelchner :)

jfelchner:
:+1: :grinning:

NullVoxPopuli:
:+1:

Issue #1184, opened 2015-09-21
Options not passed from Json to Attributes adapters


beauby:
Currently, the Json adapter simply calls the Attributes adapter, and spits out the result inside a hash. However, as [discussed](https://github.com/rails-api/active_model_serializers/pull/1158#discussion_r39937199), the former does not pass the options to the latter. 
I'm not really aware of the situation of Json options, but I think this should be fixed.

beauby:
Solved by #1185.

PR #1185, opened 2015-09-21
Fix options passing in Json and Attributes adapters

beauby:
As mentioned in #1184, there are issues with the way options are passed between adapters. This PR adds a failing test and provides fixes.

### DIFF
--- a/test/adapter/json/collection_test.rb
+++ b/test/adapter/json/collection_test.rb
@@ -75,6 +75,16 @@ module ActiveModel
 
             assert_equal 1, adapter.serializable_hash[:virtual_values].length
           end
+
+          def test_include_option
+            serializer = ArraySerializer.new([@first_post, @second_post])
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer, include: '')
+            actual = adapter.serializable_hash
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
:+1:

test/adapter/json/collection_test.rb
```diff
@@ -75,6 +75,16 @@ def test_root_is_underscored
 
             assert_equal 1, adapter.serializable_hash[:virtual_values].length
           end
+
+          def test_include_option
+            serializer = ArraySerializer.new([@first_post, @second_post])
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer, include: '')
```
bf4: what did it do before?

beauby: It disregarded the `include` option.

beauby: It did so every time for the Json adapter, and only for collections for the Attribute adapter.

bf4: Ok, but when it failed, how was the output different? (lazy me)

beauby: It did include stuff, although it was explicitly told not to.

PR #1186, opened 2015-09-21
Enforce case requires else; allow else nil

bf4:


lib/active_model/serializer/include_tree.rb
```diff
@@ -68,6 +68,8 @@ def [](key)
           self.class.new(@hash[:*])
         when @hash.key?(:**)
           self.class.new(:** => {})
+        else
+          nil
```
bf4: Missing this else would now fail Rubocop

beauby: Great!

beauby:
LGTM, will merge once the tests are green.

bf4:
green

Issue #1188, opened 2015-09-22
Inheritance of serializer


givigier:
```ruby
class ApplicationSerializer < ActiveModel::Serializer
  attributes :id
  def id
    "test"
  end
end
```
```ruby
class CustomerSerializer < ApplicationSerializer
end
```
I'm trying to use something like this on branch master AMS but it does not work. When I use CustomerSerializer, it does include the attribute :id. Is it possible to do this on AMS?

bf4:
Try defining `read_attribute_for_serialization(attr)` on your model and handle id there

Otherwise, we'll need some more information of 
- what you're trying to serializer and 
- with what data
- what you expected
- what you saw
- what you tried
- ruby flavor and version, rails version

bf4:
Also, if you could help improve https://github.com/rails-api/active_model_serializers/blob/master/CONTRIBUTING.md#how-can-i-help that would be great, too :)

(and https://github.com/rails-api/active_model_serializers#getting-help )

givigier:
Thanks @bf4. I could ask this on Stackoverflow. Could you close this issue please?

beauby:
@givigier Closing for now, please post the link of the StackOverflow question here if you do open one.

bf4:
(And link from your question here, as well.)

givigier:
http://stackoverflow.com/questions/32722921/inheritance-of-serializer

givigier:
I guess that it's a bug. I'll try to explain exactly what I did.
I have two serializers, first is ApplicationSerializer and the second is CompanySerializer that inherits of ApplicationSerializer.
```ruby
class ApplicationSerializer < ActiveModel::Serializer
  def id
    object.slug
  end
end
```

```ruby
class V1::CompanySerializer < ApplicationSerializer
  attributes :id, :email
end
```
When I run  on ``branch: '0-10-stable'`` I got:

```json
{"id"=>"company-1", "email"=>"name3@email.com"}
```
"company-1" is the slug of company with email name3@email.com

When I run on ``branch: 'master'`` I got:

```json
{"id"=>"1", "email"=>"name3@email.com"}
```
In this case, slug is not setted as id. I also does not work when I try to use concerns like this:

```ruby
module SerializerSluggable
  extend ActiveSupport::Concern

  def id
    object.slug
  end
end
```
```ruby
require 'concerns/serializer_slugglable'
class V1::CompanySerializer <  ActiveModel::Serializer
  include SerializerSluggable

[...]

PR #1195, opened 2015-09-23
Fix id override

beauby:
Currently, the `id` attribute is [handled in a special way](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L65), which leads to the impossibility to override it *before* the `attribute :id` call.
I believe this special treatment is just legacy code. @bf4, @joaomdmoura?

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -62,7 +62,7 @@ module ActiveModel
       ActiveModelSerializers.silence_warnings do
         define_method key do
           object.read_attribute_for_serialization(attr)
-        end unless (key != :id && method_defined?(key)) || _fragmented.respond_to?(attr)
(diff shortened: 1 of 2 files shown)


beauby:
Ref #1188.

beauby:
Actually, the `id` check was there because there use to be an `id` method defined on the `Serializer` class [as follows](https://github.com/rails-api/active_model_serializers/commit/8482abfac716e9281ddbed4782869d2bff1dee9a#diff-5624f9cf35be4dc254412d979fd293c4L149):
```ruby
def id
  object.id if object
end
```
which was there to make sure the JsonApi adapter could get the idea. Now the logic has been (rightfully) moved to the JsonApi adapter, so the default `id` method and the check are not necessary anymore.

NullVoxPopuli:
:+1:

PR #1193, opened 2015-09-23
Inline nested serializers

beauby:
Allows people to do stuff like:
```ruby
class PostSerializer < ActiveModel::Serializer
  attributes :id, :name, :body
  has_many :comments do
    attributes :id, :body
    belongs_to :author do
      attributes :name
  end
end
```

I'd really love for people to submit tests to try and break the implementation.

ref #1190 #1157 #1113

Issue #916, opened 2015-05-19
Associations with namespaced models not working


dpmccabe:
I have two models that come from a Rails engine, `Spree::Product` and `Spree::Variant`, and I can't figure out how to correctly specify an association between them in my serializers. Here is my current setup:

##### app/serializers/spree/product_serializer.rb

```ruby
module Spree
  class ProductSerializer < ActiveModel::Serializer
    attributes :id, :name
    has_many :variants, serializer: Spree::VariantSerializer
  end
end
```

##### app/serializers/spree/variant_serializer.rb

```ruby
module Spree
  class VariantSerializer < ActiveModel::Serializer
    attributes :id, :price
    belongs_to :product, serializer: Spree::ProductSerializer
  end
end
```

A serialization of products loads the attributes correctly, but the variants don't appear (not even an empty array). If I change `has_many :variants, serializer: Spree::VariantSerializer` to simply `has_many :variants` then I get the variants, but it includes all the default attributes instead of using `VariantSerializer`.

I've tried all sorts of combinations of file placement, class declarations, and associations to no avail.

saneshark:
May be related to #879 (pending discussion)

groyoh:
@dpmccabe I'm not sure if I understood your need, but basically you want the serializer to automatically find a matching association serializer in the same module without specifying any additional option, right? If that's the case, then #879 won't solve this. It is mainly useful when you deal with polymorphic relationship and/or serializer that are defined in a different module. You would still have to specify `namespace: Spree` in your case.

groyoh:
I don't think that it is related to #879. It might be a bug.
@dpmccabe could you provide a snippet of how you serialize the products?

dpmccabe:
The pull request referenced in #879 fixed this for me. Here are the working association method calls:

`has_many :variants, serializer: VariantSerializer, namespace: Spree`

`belongs_to :product, serializer: ProductSerializer, namespace: Spree`

groyoh:
@dpmccabe I'm glad that it fixed it for you. BTW, you don't need to specify `serializer: VariantSerializer`, this should do it:
```ruby
has_many :variants, namespace: Spree
belongs_to :product, namespace: Spree
```
Could you still provide a snippet of the serialization call please? There might still be a bug. What I would like to know is if you render it this way:
```ruby
render json: products
```
or
```ruby
serializer = ActiveModel::Serializer::ArraySerializer.new(products)
[...]

Issue #886, opened 2015-04-27
Handle serializer namespacing more robustly


eprothro:
Continuing from #144 at the request of @joaomdmoura.

@MarkMurphy and others suggest automatically handling lookup context similar to ActionView. This solves two distinct issues:

1. We want namespaced controllers to use namespaced serializers, without explicitly specifying a serializer or namespace in the controller
2. We want namespaced serializers to use namespaced serializers for association serialization, without explicitly specifying a serializer or namespace with the association

More info from Mark's proposal, copied:

---
If the `get_serializer` method [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/action_controller/serialization.rb#L11) functioned in the same manner as rails does template lookups for a controller's views, then versioning would be as easy as namespacing your controllers.

For example:

```
# app/controllers/api/v1/users_controller.rb
module API::V1
  class UsersController < APIController
    def show
       @user = User.find(params[:id])
       render json: @user
    end
  end
end
```
Based on the above controllers' namespace, AMS would look for a User serializer at the following locations in order of priority:

app/serializers/api/v1/user_serializer.rb
app/serializers/api/user_serializer.rb
app/serializers/user_serializer.rb

So when api version 2 comes around you'll have the following:

```
# app/controllers/api/v2/users_controller.rb
module API::V2
  class UsersController < APIController
    def show
       @user = User.find(params[:id])
       render json: @user
    end
  end
end
```

Based on the above controllers' namespace, AMS would look for a User serializer at the following locations in order of priority:

app/serializers/api/v2/user_serializer.rb
app/serializers/api/user_serializer.rb
app/serializers/user_serializer.rb

Routes would look something like this:

````
# config/routes.rb
constraints subdomain: 'api' do
  namespace :api, path: nil, except: [:new, :edit], defaults: { format: 'json' } do
    # API V1
    constraints API::VersionConstraint.new(version: 1) do
      scope module: 'v1' do
        resources :users
      end
    end

    # API V2
    constraints API::VersionConstraint.new(version: 2, default: true) do
      scope module: 'v2' do
        resources :users
      end
    end
  end
end
```
---

skalee:
I strongly disagree with this proposal.

Serializers are meant to present objects as JSONs. This includes but is not limited to rendering responses in controllers. Wider namespaces support is required.

[...]

PR #1196, opened 2015-09-24
Namespaced serializer lookup for associations

beauby:
Given the following situation:
```ruby
module MyNamespace
  class MyModel < ActiveRecord::Base
    has_many :my_related_models
  end
  class MyRelatedModel < ActiveRecord::Base
    ...
  end
end

module MyNamespace
  class MyModelSerializer < ActiveModel::Serializer
    has_many :my_related_models
    ...
  end
  class MyRelatedModelSerializer < ActiveModel::Serializer
    ...
  end
end
```
AMS currently has troubles inferring the right serializer class for `MyRelatedModel` (it should be `MyNamespace::MyRelatedModelSerializer`, but it only looks up `MyRelatedModelSerializer`). This PR fixes that by keeping trace of the "root" serializer, that is the first non-`ArraySerializer` serializer in the chain of nested serializers, and looking up for a serializer in that "root serializer"'s namespace. Especially cool nor polymorphic associations in the namespaced scenario. 

Solves stuff such as #886, #916, #879 but without having to manually specify the namespace.

Based on #1193 because it sort of paved the way.

[TODO] Add some tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ Features:
 - [#1172](https://github.com/rails-api/active_model_serializers/pull/1172) Better serializer registration, get more than just the first module (@bf4)
 - [#1158](https://github.com/rails-api/active_model_serializers/pull/1158) Add support for wildcards in `include` option (@beauby)
 - [#1127](https://github.com/rails-api/active_model_serializers/pull/1127) Add support for nested
-    associations for JSON and Attributes adapters via the `include` option (@NullVoxPopuli, @beauby).
+    associations for JSON and Attributes adapters via the `include` option (@NullVoxPopuli, @beauby)
+- [#1193](https://github.com/rails-api/active_model_serializers/pull/1193) Add support for inline nested serializers (@beauby)
 
 Fixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Features:
 - [#1127](https://github.com/rails-api/active_model_serializers/pull/1127) Add support for nested
     associations for JSON and Attributes adapters via the `include` option (@NullVoxPopuli, @beauby)
 - [#1193](https://github.com/rails-api/active_model_serializers/pull/1193) Add support for inline nested serializers (@beauby)
+- [#1196](https://github.com/rails-api/active_model_serializers/pull/1196) Add support for namespaced serializers (@beauby)
 
 Fixes:
 
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -25,7 +25,8 @@ module ActionController
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
         options[:adapter] = false
       end
-      serializable_resource = ActiveModel::SerializableResource.new(resource, options)
+      serializable_options = options.merge(controller: self)
+      serializable_resource = ActiveModel::SerializableResource.new(resource, serializable_options)
       if serializable_resource.serializer?
         serializable_resource.serialization_scope ||= serialization_scope
         serializable_resource.serialization_scope_name = _serialization_scope
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -87,7 +87,7 @@ module ActiveModel
       elsif options.key?(:serializer)
         options[:serializer]
       else
-        get_serializer_for(resource.class, options[:parent_serializer])
+        get_serializer_for(resource.class, options[:parent_serializer], options[:controller])
       end
     end
 
@@ -110,7 +110,7 @@ module ActiveModel
       Digest::MD5.hexdigest(serializer_file_contents)
     end
 
-    def self.get_serializer_for(klass, parent_serializer)
+    def self.get_serializer_for(klass, parent_serializer, controller)
       serializers_cache.fetch_or_store(klass) do
         serializer_class_name = "#{klass.name}Serializer"
 
@@ -120,6 +120,9 @@ module ActiveModel
         if parent_serializer
           namespaced_serializer_class_name = "#{parent_serializer.root_serializer.class.name.deconstantize}::#{serializer_class_name}"
           serializer_class ||= namespaced_serializer_class_name.safe_constantize
+        elsif controller
+          controller_namespaced_serializer_class_name = "#{controller.class.name.deconstantize}::#{serializer_class_name}"
+          serializer_class ||= controller_namespaced_serializer_class_name.safe_constantize
         end
 
         serializer_class ||= serializer_class_name.safe_constantize
@@ -127,7 +130,7 @@ module ActiveModel
         if serializer_class
           serializer_class
         elsif klass.superclass
-          get_serializer_for(klass.superclass, parent_serializer)
+          get_serializer_for(klass.superclass, parent_serializer, controller)
         end
       end
     end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -84,8 +84,10 @@ module ActiveModel
         resource.serializer_class
       elsif resource.respond_to?(:to_ary)
         config.array_serializer
+      elsif options.key?(:serializer)
+        options[:serializer]
       else
-        options.fetch(:serializer) { get_serializer_for(resource.class) }
+        get_serializer_for(resource.class, options[:parent_serializer])
       end
     end
 
@@ -108,22 +110,29 @@ module ActiveModel
       Digest::MD5.hexdigest(serializer_file_contents)
     end
 
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(klass, parent_serializer)
       serializers_cache.fetch_or_store(klass) do
         serializer_class_name = "#{klass.name}Serializer"
-        global_serializer_class = serializer_class_name.safe_constantize
-        nested_serializer_class = "#{self}::#{serializer_class_name}".safe_constantize
-        serializer_class = nested_serializer_class || global_serializer_class
+
+        nested_serializer_class_name = "#{self}::#{serializer_class_name}"
+        serializer_class = nested_serializer_class_name.safe_constantize
+
+        if parent_serializer
+          namespaced_serializer_class_name = "#{parent_serializer.root_serializer.class.name.deconstantize}::#{serializer_class_name}"
+          serializer_class ||= namespaced_serializer_class_name.safe_constantize
+        end
+
+        serializer_class ||= serializer_class_name.safe_constantize
 
         if serializer_class
           serializer_class
         elsif klass.superclass
-          get_serializer_for(klass.superclass)
+          get_serializer_for(klass.superclass, parent_serializer)
         end
       end
     end
 
-    attr_accessor :object, :root, :meta, :meta_key, :scope
+    attr_accessor :object, :root, :meta, :meta_key, :scope, :parent_serializer, :root_serializer
 
     def initialize(object, options = {})
       self.object = object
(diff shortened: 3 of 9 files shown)


lib/active_model/serializer.rb
```diff
@@ -108,20 +110,29 @@ def self.digest_caller_file(caller_line)
       Digest::MD5.hexdigest(serializer_file_contents)
     end
 
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(klass, parent_serializer)
```
NullVoxPopuli: can you add some yardoc to various parts of this method? explaining what each section does, and what scenarios will be applicable to each block of code?

NullVoxPopuli:
I like this overall, but here is a scenario, which I'd like your opinion on:

rather than `MyNamespace::MyModel` correlating to `MyNamespace::MyModelSerializer`, If someone wanted to use a different set of serializers for different versions of their API, like desiring `API::V1::MyModelsController` to use `API::V1::MyModelSerializer`, do you think it would be too complicated to support the namespace lookup on both the model and the controller? (always falling back to `MyModelSerializer`, of course)

Back when I was at my previous company, the lack of controller namespacing was the one thing I didn't like about AMS. ha :-) (there was a requirement to have a versioned API, which it turned out ended up not being all that important :-\ )

But anyway, just wanted your thoughts on controller namespacing as well - but maybe that should be in an RFC

beauby:
@NullVoxPopuli Just added the controller namespace lookup.

lib/active_model/serializer.rb
```diff
@@ -108,20 +110,32 @@ def self.digest_caller_file(caller_line)
...
+
+        nested_serializer_class_name = "#{self}::#{serializer_class_name}"
+        serializer_class = nested_serializer_class_name.safe_constantize
+
+        if parent_serializer
+          namespaced_serializer_class_name = "#{parent_serializer.root_serializer.class.name.deconstantize}::#{serializer_class_name}"
+          serializer_class ||= namespaced_serializer_class_name.safe_constantize
+        elsif controller
+          controller_namespaced_serializer_class_name = "#{controller.class.name.deconstantize}::#{serializer_class_name}"
+          serializer_class ||= controller_namespaced_serializer_class_name.safe_constantize
```
NullVoxPopuli: what if this serializer isn't found? like, what if someone wants to use a straight up `PostSerializer` in `/blog/id/posts`?

tests for this would be great :-)

beauby: That's why I use `||=`. What this piece does is look for a `RelatedSerializer` serializer in:
 + `CurrentSerializer::RelatedSerializer` (which would be `ActiveModel::Serializer::RelatedSerializer` if it is not actually a related serializer, but that would fail so it's ok)
 + revert to `My::Name::Space::RelatedSerializer` if called from `My::Name::Space::CurrentSerializer`
 + or revert to `My::Controller::Name::Space::RelatedSerializer` if called from `My::Controller::Name::Space::SomeController`
 + finally, revert to `::RelatedSerializer` if none of the above exist

beauby: This part could clearly be improved though. Suggestions?

NullVoxPopuli: oh ok, I thought it was going the other way (if RelatedSerializer isn't found, try the namespaced version)

in any case, tests for this new logic would be stellar.

lib/active_model/serializer.rb
```diff
@@ -108,20 +110,50 @@ def self.digest_caller_file(caller_line)
...
+    # @param [ActiveModel::Serializer] parent_serializer The instance of the parent serializer, if any
+    # @param [ActionController] controller The instance of the calling controller, if any
+    #
+    # @return [Array<String>] The lookup chain
+    #
+    def self.serializer_lookup_chain_for(serializer_class_name, parent_serializer, controller)
+      chain = []
+
+      # Look for a serializer nested inside the current serializer first, if inside a user-defined serializer
+      chain.push("#{self}::#{serializer_class_name}") if self.class != ActiveModel::Serializer
```
beauby: That kind of check makes me believe it might be time to make `serializer_for` an instance method?

NullVoxPopuli:
thanks for the docs!

beauby:
Tests fail because of the lack of lazy enumerators in ruby 1.9.3.

lib/action_controller/serialization.rb
```diff
@@ -25,7 +25,8 @@ def get_serializer(resource, options = {})
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
         options[:adapter] = false
       end
-      serializable_resource = ActiveModel::SerializableResource.new(resource, options)
+      serializable_options = options.merge(controller: self)
```
bf4: why not mutate?

bf4: also why change options when not serializing

bf4: and do we need the whole controller? I'd rather not and looks like you just want the class name

beauby: Currently we just need the class name but there's virtually 0 overhead to getting the instance, so why not?

bf4: If you make it easy to use the controller directly, people will, and the scope will creep and and new ams will gain responsibilities it shouldn't gave and cause pain and bugs to users and maintainers 

Failure of Interface segregation, not keepig it narrow

B mobile phone

lib/active_model/serializer.rb
```diff
@@ -84,8 +84,10 @@ def self.serializer_for(resource, options = {})
         resource.serializer_class
       elsif resource.respond_to?(:to_ary)
         config.array_serializer
+      elsif options.key?(:serializer)
+        options[:serializer]
       else
-        options.fetch(:serializer) { get_serializer_for(resource.class) }
+        get_serializer_for(resource.class, options[:parent_serializer], options[:controller])
```
bf4: this is getting pretty hairy

lib/active_model/serializer/associations.rb
```diff
@@ -61,26 +62,41 @@ def belongs_to(name, options = {})
...
           self._reflections = _reflections.dup
 
           define_method reflection.name do
             object.send reflection.name
           end unless method_defined?(reflection.name)
 
-          self._reflections << reflection
+          _reflections << reflection
+
+          define_nested_serializer(reflection.name.to_s.singularize, &block) if block_given?
```
bf4: if the block is optional it shouldn't be a param. maybe this (chain of) method is doing too much?

lib/active_model/serializer/associations.rb
```diff
@@ -61,26 +62,41 @@ def belongs_to(name, options = {})
...
+
+        # Define a nested serializer
+        # @param [String] resource_name The name of the association
+        # @return [void]
+        #
+        # @api private
+        #
+        def define_nested_serializer(resource_name, &block)
+          serializer_name = "#{resource_name.camelize}Serializer"
+          serializer = Class.new(ActiveModel::Serializer)
```
bf4: frankly, all this meta programming scares me.  Is it really necessary?

lib/action_controller/serialization.rb
beauby: In my opinion, the real issue lies in the fact that we miss a layer of separation between the serializer definition and the serializer instances.

lib/active_model/serializer.rb
beauby: Yeah, I think the whole `serializer_for` calling `get_serializer_for` thing is pretty bad design already. I'm really open to suggestions to make the whole serializer lookup cleaner.

lib/active_model/serializer/associations.rb
beauby: Why should it not be a param if it's optional? I'm open to suggestions for doing this differently.

lib/active_model/serializer/associations.rb
beauby: What other solution do you have in mind?

itsmechlark:
+1

lib/active_model/serializer.rb
bf4: It's been around for a while, so I'd rather deprecate it

lib/active_model/serializer/associations.rb
bf4: usually you'd only use `&block` if you need an object, otherwise, you'd just

```ruby
def foo
  if block_given?
   puts yield
  else
   puts 'no block'
  end
end
```

CHANGELOG.md
```diff
@@ -16,7 +16,9 @@ Features:
 - [#1172](https://github.com/rails-api/active_model_serializers/pull/1172) Better serializer registration, get more than just the first module (@bf4)
 - [#1158](https://github.com/rails-api/active_model_serializers/pull/1158) Add support for wildcards in `include` option (@beauby)
 - [#1127](https://github.com/rails-api/active_model_serializers/pull/1127) Add support for nested
-    associations for JSON and Attributes adapters via the `include` option (@NullVoxPopuli, @beauby).
+    associations for JSON and Attributes adapters via the `include` option (@NullVoxPopuli, @beauby)
+- [#1193](https://github.com/rails-api/active_model_serializers/pull/1193) Add support for inline nested serializers (@beauby)
```
bf4: 1193?

test/adapter/json/nested_serializers_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module ActiveModel
+  class Serializer
+    module Adapter
+      class Json
+        class NestedSerializersTest < Minitest::Test
+          def setup
+            @tweet = Tweet.new(id: 1, body: 'Tweet 1', date: 'Jan 15')
+            @share1 = Share.new(id: 1, platform: 'facebook', date: 'Jan 16')
+            @author = Author.new(id: 1, name: 'Lucas H.')
```
bf4: Please define all of these here. I can't tell what you're testing.. doesn't look nested, y'know?

CHANGELOG.md
beauby: This PR is based on #1193, that's why.

test/adapter/json/nested_serializers_test.rb
beauby: True, that's why I opened #1206. I really think serializers/resources should be defined in each test file somehow.

bf4:
Example test case

```ruby
module ActiveModel
  class Serializer
    class SerializerForTest < Minitest::Test
      class Tweet < Model; end
      class Share < Model; end
      class TweetSerializer < ActiveModel::Serializer
        attributes :id, :body, :date
        has_many :shares
        class ShareSerializer < ActiveModel::Serializer
          attributes :id, :platform, :date
        end
      end
      class ShareSerializer < ActiveModel::Serializer
        attributes :id, :platform, :date
      end

      def test_serializer_for_namespaced_serializers
        tweet = Tweet.new
        share = Share.new
        # called from the has_many :shares on TweetSerializer 
        # when passed a tweet with shares, which returns an ArraySerializer
        # which uses the TweetSerializer::ShareSerializer on each share
        serializer = TweetSerializer.serializer_for(share)
        assert_equal(TweetSerializer::ShareSerializer, serializer)
      end

      def TweetSerializer.serializer_for(resource)
        resource_serializer = "#{resource.class}Serializer"
        serializer = constants.find {|constant| constant.to_s == resource_serializer}
        if serializer
          const_get(:"#{serializer}")
        else
          nil
        end
      end

    end
  end
end
```

drafted, not tested

bf4:
Closed by https://github.com/rails-api/active_model_serializers/pull/1029 ?

lib/active_model/serializer/associations.rb
beauby: `&block` needs to be a param (i.e. named block) if you want to pass it as an argument to a method call. `method(yield)` does not work, nor does `method do yield end` (the `self` object gets messed up).

bf4: I get that, i just mean design wise.. Also could check if block since you have the object

B mobile phone

beauby:
Closing this in favor of #1225/#1226.

bf4:
ref: https://github.com/rails-api/active_model_serializers/pull/1436

PR #1201, opened 2015-09-25
Add documentation about wildcard includes

### LINKED ISSUE #1200: JSON API - Include documentation
Looking here:

https://github.com/rails-api/active_model_serializers#json-api

it neglects to mention the `*` and `**` options which I found in `include_tree.rb`. Are these intentionally undocumented? Because `**`, I think, is exactly what I want in order to include one layer of everything. It seems like `**` isn't mentioned anywhere in the JSON-API spec, so is this intentionally undocumented in case support is removed? Or was this an oversight?

hut8:
Resolves #1200

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,26 @@ resources in the `"included"` member when the resource names are included in the
   render @posts, include: 'author,comments,comments.author'
 ```
 
+In addition, two types of wildcards may be used. `*` includes one level of associations, and `**` includes all recursively. These can be combined with other paths.
+
+```ruby
+  render @posts, include: '**' # or '*' for a single layer
+```
+
+The following would render posts and include the author, the author's comments, and every resource referenced by the author's comments (recursively). It could be combined, like above, with other paths in any combination desired.
+
+```ruby
+  render @posts, include: 'author.comments.**'
+```
+
+The JSON API [specifies](http://jsonapi.org/format/#fetching-includes) that the user may supply a parameter specifying which related resources are to be included:
+
+```ruby
+  render @posts, include: params[:include]
+```
+
+This raises some security concerns since the user could pass in `include=**`, so filter the values for `include` appropriately if you decide to support this JSON API feature.
+
 ## Installation
 
 Add this line to your application's Gemfile:


README.md
```diff
@@ -213,6 +213,20 @@ resources in the `"included"` member when the resource names are included in the
...
+  render @posts, include: '**' # or '*' for a single layer
+```
+
+The JSON API [specifies](http://jsonapi.org/format/#fetching-includes) that the user may supply a parameter specifying which related resources are to be included:
+
+```ruby
+  render @posts, include: params[:include]
+```
+
+This raises some security concerns since the user could pass in `include=**`, so filter the values for `include` appropriately if you decide to support this JSON API feature.
```
bf4: :+1:

beauby: Nice to mention this.

README.md
```diff
@@ -213,6 +213,20 @@ resources in the `"included"` member when the resource names are included in the
   render @posts, include: 'author,comments,comments.author'
 ```
 
+In addition, two types of wildcards may be used. `*` includes one level of associations, and `**` includes all recursively.
+
+```ruby
+  render @posts, include: '**' # or '*' for a single layer
```
beauby: Maybe it would be nice to include an example mixing both paths and wildcards, as wildcards can be used pretty much anywhere (stuff like `author.comments.*.*` (you could even do stuff such as `author.comments.*.*.posts`, although I can't really think of a use case which such homogeneous resources that it would make sense, maybe some polymorphic association stuff), and `author.comments.**`).

beauby:
@hut8 Thanks for the PR, we really need this. Please see my comments.

README.md
beauby: Cool. Could you add an additional example using nested routes, like `author.comments.posts` and `author.comments.*` so that a) people know the wildcards can be used anywhere, and b) you can add a note saying that when you include `author.comments`, it includes `authors` as well (which makes sense, but might not be clear at first glance).

hut8: I will do this tomorrow morning.

hut8:
@beauby Better?

README.md
```diff
@@ -213,6 +213,26 @@ resources in the `"included"` member when the resource names are included in the
   render @posts, include: 'author,comments,comments.author'
 ```
 
+In addition, two types of wildcards may be used. `*` includes one level of associations, and `**` includes all recursively. These can be combined with other paths.
+
+```ruby
+  render @posts, include: '**' # or '*' for a single layer
+```
+
+The following would render posts and include every resource referenced by the posts' authors' comments (recursively). It could be combined, like above, with other paths in any combination desired.
```
beauby: How about 
> include the author, the author's comments, and every resource referenced by the authors' comments (recursively)

beauby: LGTM other than this. Sorry for being a bit a bit tedious on this PR but I can foresee many issues caused by people misunderstanding this functionality.

hut8:
@beauby Not tedious at all! It's important to make these things clear and this is a pretty awesome gem. I'm glad to be able to help.

README.md
```diff
@@ -213,6 +213,26 @@ resources in the `"included"` member when the resource names are included in the
   render @posts, include: 'author,comments,comments.author'
 ```
 
+In addition, two types of wildcards may be used. `*` includes one level of associations, and `**` includes all recursively. These can be combined with other paths.
+
+```ruby
+  render @posts, include: '**' # or '*' for a single layer
+```
+
+The following would render posts and include the author, the author's comments, and every resource by the author's comments (recursively). It could be combined, like above, with other paths in any combination desired.
```
beauby: There seems to be a missing word here:
> and every ??? resource by the author's comments (recursively)

beauby:
See my last remark, then let's squash and merge!

hut8:
Seemas though I accidentally a word. Fixed. Squashed and pushed.

beauby:
LGTM, merging. Thanks a lot @hut8!

PR #1208, opened 2015-09-28
Added doc command available

julianosaless:
Added table about command available

### DIFF
--- a/docs/general/getting_started.md
+++ b/docs/general/getting_started.md
@@ -15,6 +15,12 @@ And then execute:
 ```
 $ bundle
 ```
+### Available Commands
+
+|command                               |Description                                                 |
+|--------------------------------------|------------------------------------------------------------|
+| rails g resource + `your model`      | Create a new serializer is to generate a new resource.     |
+| rails g serializer + `your model`    | Generate a serializer in  your new model.                  |
 
 ## Creating a Serializer
 


NullVoxPopuli:
:+1:

docs/general/getting_started.md
```diff
@@ -15,6 +15,12 @@ And then execute:
 ```
 $ bundle
 ```
+### Available Commands
+
+|command                               |Description                                                 |
+|--------------------------------------|------------------------------------------------------------|
+| rails g resource + `your model`      | Create a new serializer is to generate a new resource.     |
```
beauby: This should probably be a note that, by default, a serializer will be generated for every new resource.

docs/general/getting_started.md
```diff
@@ -15,6 +15,12 @@ And then execute:
 ```
 $ bundle
 ```
+### Available Commands
```
beauby: I like that we add some rails-specific notes, but I thing they should rather be at the end of the file than at the top.

docs/general/getting_started.md
julianosaless: @beauby  i totally agree with you, i will change  this line;

docs/general/getting_started.md
julianosaless: @beauby  What you think about this point  below?

when new user are reading the documentation , I think him needs is all the commands available after the that him  can do with these command.

beauby: Thing is, not all new users are rails users. Plus, most rails developers rarely use generators. I don't have a definitive opinion on this though, let's ask @joaomdmoura.

julianosaless: @joaomdmoura  any position about  my pr :)

beauby:
Actually @julianosaless, after looking a bit more into this PR, it seems the information you added is already present in the next paragraph: https://github.com/julianosaless/active_model_serializers/blob/add-doc-available/docs/general/getting_started.md#creating-a-serializer

bf4:
Closing as it appears to now be addressed by the docs in `docs/general/getting_started.md` and `lib/generators/rails/USAGE`. Sorry for lost time @julianosaless

PR #1213, opened 2015-09-29
`type` directive for serializer to control type field with json-api adapter

youroff:
This pull-requests implements `type` directive for serializer declaration. Which sets type attribute while using json-api adapter. So it's possible to use ProfileSerializer with User model and set type field to 'profile', like this:
```
class ProfileSerializer < ActiveModel::Serializer
  attributes :name
  type 'profile'
end
```

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -46,6 +46,10 @@ module ActiveModel
       super
     end
 
+    def self.type(type)
+      self._type = type
+    end
+
     def self.attributes(*attrs)
       attrs = attrs.first if attrs.first.class == Array
 
@@ -122,6 +126,7 @@ module ActiveModel
     end
 
     attr_accessor :object, :root, :meta, :meta_key, :scope
+    class_attribute :_type, instance_writer: false
 
     def initialize(object, options = {})
       self.object = object
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -71,6 +71,7 @@ module ActiveModel
         end
 
         def resource_identifier_type_for(serializer)
+          return serializer._type if serializer._type
           if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
             serializer.object.class.model_name.singular
           else
--- a/test/adapter/json_api/resource_type_config_test.rb
+++ b/test/adapter/json_api/resource_type_config_test.rb
@@ -5,6 +5,11 @@ module ActiveModel
     module Adapter
       class JsonApi
         class ResourceTypeConfigTest < Minitest::Test
+          class ProfileTypeSerializer < ActiveModel::Serializer
+            attributes :name
+            type 'profile'
+          end
+
           def setup
             @author = Author.new(id: 1, name: 'Steve K.')
             @author.bio = nil
@@ -36,21 +41,28 @@ module ActiveModel
           end
 
           def test_config_plural
-            with_adapter :json_api do
-              with_jsonapi_resource_type :plural do
-                hash = ActiveModel::SerializableResource.new(@comment).serializable_hash
-                assert_equal('comments', hash[:data][:type])
-              end
+            with_jsonapi_resource_type :plural do
+              hash = serializable(@comment, adapter: :json_api).serializable_hash
+              assert_equal('comments', hash[:data][:type])
             end
           end
 
           def test_config_singular
-            with_adapter :json_api do
-              with_jsonapi_resource_type :singular do
-                hash = ActiveModel::SerializableResource.new(@comment).serializable_hash
-                assert_equal('comment', hash[:data][:type])
-              end
+            with_jsonapi_resource_type :singular do
+              hash = serializable(@comment, adapter: :json_api).serializable_hash
+              assert_equal('comment', hash[:data][:type])
             end
+          end
+
+          def test_explicit_type_value
+            hash = serializable(@author, serializer: ProfileTypeSerializer, adapter: :json_api).serializable_hash
+            assert_equal('profile', hash.fetch(:data).fetch(:type))
+          end
+
+          private
+
+          def serializable(resource, options = {})
+            ActiveModel::SerializableResource.new(resource, options)
           end
         end
       end


NullVoxPopuli:
First off, thanks for the PR!

My only concern is that `type` is often the name of a field used in single table inheritance for storing the class name of an object. 

So, I guess that leaves two options:
 - change the name of the method from type to something else?
 - or write a test to simulate the potential problem (probably just rendering a model with a type field, and see if it still gets included in the resulting hash) (this would be my preference)  If the test passes, you're good, and I'll give ya the :+1:  :-)

youroff:
There is a test which actually made me to change an approach. So now there's only class method test, which sets _test class variable. And this is used on the class level again, so it'll not interfer with type attribute.
Here's that test:
https://github.com/rails-api/active_model_serializers/blob/master/test/serializers/attribute_test.rb#L55

youroff:
But I think it would be way smarter to have setting which class name to use (Serializer or Model) AND type directive to set it manually. If you like this idea, I'll add this feature.

NullVoxPopuli:
oh good! thanks for looking for that. :-)

you have my :+1: 

@beauby seems to be the JsonApi Adapter expert though, so I'll let him make the final call.

youroff:
I added functionality to derive resource type from serializer class, however it broke one test, which I'm not sure is really correct. Depends on standard which I'm not sure in. So the question is how we deal with namespaced models/serializers. Method `underscore` converts `::` into `/`, however expected behavior is make it underscore as well. Opinions?

beauby:
Hi @youroff, you are right, I added the possibility to override the type in a pending PR that got lost somewhere. I am totally up for this PR, I'll review it right away.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+          unless serializer.class._type.nil?
+            serializer.class._type
           else
```
beauby: Avoid `unless` with an `else` clause. Rewrite as an `if` with the positive clause first (https://github.com/bbatsov/ruby-style-guide#no-else-with-unless)

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
...
-            serializer.object.class.model_name.plural
+            if ActiveModel::Serializer.config.type_source == :model
+              type = serializer.object.class.name.underscore
+            else
+              type = serializer.class.name.underscore.split('_serializer').first
```
beauby: This seems a bit overcomplicated. How about `serializer.class.name.chomp('Serializer').underscore`?

youroff:
Done, but for namespace case we have to decide what to change: test or code. I'm not sure what json-api says about namespaced models.

lib/active_model/serializer/adapter/json_api.rb
youroff: Ok, wasn't sure what could serve more elegant here.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+          unless serializer.class._type.nil?
+            serializer.class._type
           else
-            serializer.object.class.model_name.plural
+            if ActiveModel::Serializer.config.type_source == :model
```
beauby: I think this should be extracted to another PR. I'm generally not in favor of adding too many config options, and with the `type` method you introduced, there's virtually no use for this one, except for allowing wildly different workflows for users, which, as well, I'm not too keen on.

lib/active_model/serializer/configuration.rb
```diff
@@ -8,6 +8,7 @@ module Configuration
         base.config.array_serializer = ActiveModel::Serializer::ArraySerializer
         base.config.adapter = :attributes
         base.config.jsonapi_resource_type = :plural
+        base.config.type_source = :model
```
beauby: See my [other remark](https://github.com/rails-api/active_model_serializers/pull/1213/files#r40739506).

beauby:
Thing is, JSON API [explicitly](http://cantremembertheissue.jsonapi.org/) does not say anything about namespaced models. There are PRs in the pipe for better handling of namespaced models, but my personal feeling is that the default should be to remove the namespace for the `type` attribute. We should probably open an RFC to discuss it though.

youroff:
I agree. So I remove namespace for this time to make tests pass?

lib/active_model/serializer/configuration.rb
youroff: I'm quite lost in comments... which remark exactly?

beauby:
Namespaces shouldn't be a problem here, are they?

lib/active_model/serializer/configuration.rb
beauby: The one I linked ;) https://github.com/rails-api/active_model_serializers/pull/1213/files#r40739506

youroff:
They caused one test to fail. Last commit fixes it. If we agree to remove namespaces.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+          if serializer.class._type.nil?
+            if ActiveModel::Serializer.config.type_source == :model
+              type = serializer.object.class.name.underscore
```
beauby: We shouldn't modify the behavior here (before it read the type from `object.class.model_name` (which is AR-specific, so we may change it later, but not now) to `object.class.name`).

beauby:
What caused the test to fail was that you got the name from the object's class name instead of the AR model name.

youroff:
Right, but it doesn't really matter. Anyway we should decide either we leave namespace with separation of some form (`/` or `_`) or we get rid of it.

lib/active_model/serializer/adapter/json_api.rb
youroff: The problem is that weird conversion of namespace comes from `plural` or `singular` methods. I Thought it would be reasonable to rely on general pluralization applied to string. But I can roll back to previous behavior of course.

youroff:
http://jsonapi.org/format/#document-member-names
At least it says that `/` is not allowed.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+          if serializer.class._type.nil?
+            if ActiveModel::Serializer.config.type_source == :model
+              type = serializer.object.class.name.demodulize.underscore
```
beauby: I'd rather we stick with `model_name`, because it might very well be that there are 0 differences, but I have no certainty of that, and would rather change it in a subsequent PR if it makes sense. I get that it makes the pluralizing situation trickier, but it will be refactored soon anyways.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
+          if serializer.class._type.nil?
+            if ActiveModel::Serializer.config.type_source == :model
```
beauby: Let's loose the `config.type_source` for this PR, maybe extract it to another PR, but I personally favor the `type` directive way.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,10 +71,16 @@ def serializable_hash_for_single_resource(options)
...
+              type = serializer.object.class.name.demodulize.underscore
+            else
+              type = serializer.class.name.demodulize.chomp('Serializer').underscore
```
beauby: Also, I'm totally in favor of removing the namespace, but I think it should be done in a separate PR.

youroff:
So I roll everything back to `type` directive? I agree to get type from model_name, but why can't we leave type_source? With default setting (:model) it's transparent and `type` will override it anyways (if needed). I mean it's the feature that I would use in my project, so I'm really interested to keep it.

beauby:
It's not that we can't leave `type_source`, it's just that PRs should be kept so one semantic change as much as possible, so that people can review them more easily, and the history remains clean.

youroff:
Btw, what `plural` and `singular` on `ActiveModel::Name` does is exactly the same as we did:
https://github.com/rails/rails/blob/0edb6465971f7d937fce2bf0a8e1e2b540d56e0a/activemodel/lib/active_model/naming.rb#L193-L195

youroff:
Ok, so I'll just split this to two PR's?

beauby:
Does it? https://github.com/rails/rails/blob/0edb6465971f7d937fce2bf0a8e1e2b540d56e0a/activemodel/lib/active_model/naming.rb#L235

beauby:
That would be nice, yes.

youroff:
Well, in general case it'll return ActiveModel::Name populated with class name of the model, but there are some edge cases, that I was not aware of (and I still don't understand the use-case). So probably we need tests for that weird case to make sure that model_name return exactly what we expect.

youroff:
If I get it right, it expects method `use_relative_model_naming?` defined on some of parent classes and destroys other segments. So if we talk about removing namespace as general behavior, then we can easily use class name, instead of `model_name`.

beauby:
We should split your work into 3 PRs IMO, one for the `type` directive, one for the `type_model` config, and one for removing the namespace.
My point for `model_name` vs `name` is that we should not modify something unless there's a good reason for it, so that we do not risk changing behavior for no reason.

youroff:
Ok, I'll do that soon.
For model name, I think if we remove namespace as a PR, we have to get rid of model_name anyway, because otherwise we'll call it like `object.class.model_name.name.demodulize.underscore` which is definitely longer way to the same point as `object.class.name.demodulize.underscore`

beauby:
Sure, let's include that change in the PR that removes the namespace.

beauby:
Thanks for your work and your implication btw, it is highly appreciated!

youroff:
I have one concern, though. If I remove namespaces in separate commit, it'll be impossible to merge automatically. So namespace removal PR should be made over master, that will already have type_source merged.

youroff:
Just submitted a PR for namespaces. `type_source` one can be safely made only over this one, so this should be merged first to avoid mess. 
https://github.com/rails-api/active_model_serializers/pull/1216

lib/active_model/serializer.rb
```diff
@@ -46,6 +47,10 @@ def self.inherited(base)
       super
     end
 
+    def self.type(type)
+      self._type = type
+    end
```
bf4: Awesome! :100: Thanks for helping keep the public interface narrow!

youroff: I'm not sure it's narrow. Underscored class-level accessor will be accessible anyways.

test/adapter/json_api/resource_type_config_test.rb
```diff
@@ -52,6 +52,13 @@ def test_config_singular
               end
             end
           end
+
+          def test_config_custom
```
beauby: Why is it named "config"?

lib/active_model/serializer.rb
bf4: well, basically we want to avoid creating internal instance methods that collide with attribute names.. less important in class variables, but still :)

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,6 +71,7 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
+          return serializer.class._type unless serializer.class._type.nil?
```
bf4: Might be nice to use `class_attribute :_type, instance_writer: false` so we can just `_type` here

test/fixtures/poro.rb
```diff
@@ -68,6 +68,11 @@ class ProfilePreviewSerializer < ActiveModel::Serializer
   attributes :name
 end
 
+class ProfileTypeSerializer < ActiveModel::Serializer
+  attributes :name
+  type 'profile'
+end
```
bf4: let's define this inside the test.  this file is betting kind of instance

test/adapter/json_api/resource_type_config_test.rb
bf4: `test_explicit_type`

test/fixtures/poro.rb
youroff: Haha... for me whole this minitest set is already insane. I agree, so I define it inside of file with test?

lib/active_model/serializer/adapter/json_api.rb
youroff: Ok

test/adapter/json_api/resource_type_config_test.rb
youroff: Just copying name convention from sibling tests, no idea why they have config :)

lib/active_model/serializer.rb
youroff: Yeah, right, I see that instance should be as clean as possible to let users use their names for attributes.

youroff:
Made all corrections mentioned

lib/active_model/serializer.rb
```diff
@@ -37,6 +37,7 @@ class << self
       attr_accessor :_cache_except
       attr_accessor :_cache_options
       attr_accessor :_cache_digest
+      attr_accessor :_type
```
bf4: I wish there were a better name for this, or a way to signal it is for json_api... `resource_type_name` ? no need to do anything

bf4: n/m, it can/should be removed since `class_attribute :_type, instance_writer: false` is added below

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,6 +71,7 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
+          return serializer._type unless serializer._type.nil?
```
bf4: simpler `return serializer._type if serializer._type` no?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,6 +71,7 @@ def serializable_hash_for_single_resource(options)
...
           if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
             serializer.object.class.model_name.singular
```
bf4: I know you didn't write this, but urgh

NullVoxPopuli: :+1:

test/adapter/json_api/resource_type_config_test.rb
```diff
@@ -35,6 +35,11 @@ def with_jsonapi_resource_type type
             ActiveModel::Serializer.config.jsonapi_resource_type = old_type
           end
 
+          ProfileTypeSerializer = Class.new(ActiveModel::Serializer) do
+            attributes :name
+            type 'profile'
+          end
+
```
bf4: Let's do regular classes plz :)  I don't know what the previous ones were like this.

```ruby
class ResourceTypeConfigTest < Minitest::Test
  class ProfileTypeSerializer < ActiveModel::Serializer
    attributes :name
    type 'profile'
   end
```

test/adapter/json_api/resource_type_config_test.rb
```diff
@@ -52,6 +57,13 @@ def test_config_singular
               end
             end
           end
+
+          def test_explicit_type_value
+            with_adapter :json_api do
+              hash = ActiveModel::SerializableResource.new(@author, serializer: ProfileTypeSerializer).serializable_hash
+              assert_equal('profile', hash[:data][:type])
```
bf4: `with_adapter` is a [global thing that needs to go away](https://github.com/rails-api/active_model_serializers/blob/master/test/support/serialization_testing.rb#L8-L12).  if you don't mind, let's

```ruby
serialized_resource = ActiveModel::SerializableResource.new(@author, serializer: ProfileTypeSerializer, adapter: :json_api).serializable_hash
assert_equal('profile', serialized_resource.fetch(:data).fetch(:type))
```

(the fetch is just to get a better exception if serialized_resource isn't expected)

if you want to [take advantage of a future test simplification](https://github.com/rails-api/active_model_serializers/pull/1050/files#diff-02cd2f48e8583af0b097a44913eb31baR33), add the private method `serializable` to this test

```ruby
def serializable(resource, options = {})
    ActiveModel::SerializableResource.new(resource, options)
  end
```

so then you have `hash = serializable(@author, serializer: ..., adapter: :json_api).serializable_hash`

bf4:
A few small things and this is ready to merge. :100:

lib/active_model/serializer/adapter/json_api.rb
beauby: :+1: @bf

test/adapter/json_api/resource_type_config_test.rb
beauby: :+1:

lib/active_model/serializer/adapter/json_api.rb
youroff: How about using `.present?` here? As long as we rely on `ActiveSupport` and empty type field doesn't make too much sense... Using variable directly in conditional would work, but I'm trying to avoid such things. Just in case. :)

lib/active_model/serializer.rb
bf4: - [ ] plz remove this accessor as the class_attribute below duplicates the behavior

lib/active_model/serializer.rb
```diff
@@ -122,6 +127,7 @@ def self.get_serializer_for(klass)
     end
 
     attr_accessor :object, :root, :meta, :meta_key, :scope
+    class_attribute :_type, instance_writer: false
```
bf4: - [ ] this can go up near the other class methods

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,6 +71,7 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
+          return serializer._type if serializer._type.present?
```
bf4: Since the default is nil, anything else is user entered.  This basically just guards against an empty string, but since they want that, let'em :)

I'd prefer `return serializer._type if serializer._type`... since it doesn't have to go all the way up the method call stack... https://github.com/rails/rails/blob/4-0-stable/activesupport/lib/active_support/core_ext/object/blank.rb  

but if you really like present?, not a blocker.  I incline to avoid monkeypatching

youroff: I'm not crazy about `present?` at all and totally ok with having just _type. I just find implicit type-conversion confusing and try to avoid that when possible.

bf4:
Looks good, last thing to do is rebase it to squash it down to one commit.  Lemme know if you want any tips on doing that..

youroff:
If you have some short recipe how to do that, would be nice. Otherwise, I'll google. ;)

youroff:
So what I did:
```
git checkout master
git pull
git checkout type_setting
git rebase master
```
As long as nothing changed since I started this PR, nothing happened... Should I merge to master locally or what?

bf4:
This is pretty good
- http://help.exercism.io/git-workflow.html
- A lot more awesome, but a bit more to learn http://rakeroutes.com/blog/deliberate-git/
- What I would do is `git rebase -i 305ec8dc50a16e462943cdffb4ebe346637c61a7^` turn all the `pick` into `f`, except the first, which I'd change to `r` and rewrite the commit message to `Add Serializer 'type' directive to control type field, for use by the JsonApi adapter`, `git fetch master; git rebase origin/master; git push -f`

bf4:
you missed updating your local master to the remote on origin :) see above for how I squash.. or if you want we can pair or I can do it.. up to you

youroff:
Does it look ok? I expected previous commits to disappear, is that ok that they are still there?

bf4:
beautiful!  GitHub still knows about the other commits... they're just not part of the PR. It git, you need to work hard to remove things :)

test/adapter/json_api/resource_type_config_test.rb
```diff
@@ -36,22 +41,29 @@ def with_jsonapi_resource_type type
...
-              end
+            with_jsonapi_resource_type :singular do
+              hash = serializable(@comment, adapter: :json_api).serializable_hash
+              assert_equal('comment', hash[:data][:type])
             end
           end
+
+          def test_explicit_type_value
+            hash = serializable(@author, serializer: ProfileTypeSerializer, adapter: :json_api).serializable_hash
+            assert_equal('profile', hash.fetch(:data).fetch(:type))
```
beauby: Why using `fetch` here instead of brackets?

bf4: Better error on failure

B mobile phone

beauby: Then why not use it [there](https://github.com/rails-api/active_model_serializers/pull/1213/files#diff-2e5487b4844a235b4d646ae346b76499R53)?

bf4: Not at computer now, could be oversight, not important in any case

B mobile phone

beauby: Agreed it's not important, but best to remain consistent.

joaomdmoura: > remain consistent

please

PR #1216, opened 2015-09-30
Namespace separator fix for type field in jsonapi adapter

youroff:
This PR removes namespace from automatically generated `type` field.
Discussion's going on here:
http://discuss.jsonapi.org/t/convention-for-namespacing-type/99

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -71,11 +71,8 @@ module ActiveModel
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
-          else
-            serializer.object.class.model_name.plural
-          end
+          type = serializer.object.class.name.demodulize.underscore
+          ActiveModel::Serializer.config.jsonapi_resource_type == :singular ? type : type.pluralize
         end
 
         def resource_identifier_id_for(serializer)
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -71,7 +71,7 @@ module ActiveModel
         end
 
         def resource_identifier_type_for(serializer)
-          type = serializer.object.class.name.demodulize.underscore
+          type = serializer.object.class.name.underscore.gsub('/', '--')
           ActiveModel::Serializer.config.jsonapi_resource_type == :singular ? type : type.pluralize
         end
 
--- a/test/adapter/json_api/linked_test.rb
+++ b/test/adapter/json_api/linked_test.rb
@@ -203,7 +203,7 @@ module ActiveModel
             assert_equal expected, alt_adapter.serializable_hash[:included]
           end
 
-          def test_underscore_model_namespace_for_linked_resource_type
+          def test_remove_model_namespace_for_linked_resource_type
             spammy_post = Post.new(id: 123)
             spammy_post.related = [Spam::UnrelatedLink.new(id: 456)]
             serializer = SpammyPostSerializer.new(spammy_post)
@@ -212,7 +212,7 @@ module ActiveModel
             expected = {
               related: {
                 data: [{
-                  type: 'spam_unrelated_links',
+                  type: 'unrelated_links',
                   id: '456'
                 }]
               }
--- a/test/adapter/json_api/linked_test.rb
+++ b/test/adapter/json_api/linked_test.rb
@@ -212,7 +212,7 @@ module ActiveModel
             expected = {
               related: {
                 data: [{
-                  type: 'unrelated_links',
+                  type: 'spam--unrelated_links',
                   id: '456'
                 }]
               }


test/adapter/json_api/linked_test.rb
```diff
@@ -212,7 +212,7 @@ def test_underscore_model_namespace_for_linked_resource_type
             expected = {
               related: {
                 data: [{
-                  type: 'spam_unrelated_links',
+                  type: 'unrelated_links',
```
bf4: why did the expectation change here?

aside, the type should probably be `spam:unrelated_links' or `spam-unrelated_links`

https://github.com/json-api/json-api/issues/850 (the issue I made that the jsonapi forum references )

(aside, why is there a `related=` writer on Post that is important?)

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,11 +71,8 @@ def serializable_hash_for_single_resource(options)
         end
 
         def resource_identifier_type_for(serializer)
-          if ActiveModel::Serializer.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
-          else
-            serializer.object.class.model_name.plural
-          end
+          type = serializer.object.class.name.demodulize.underscore
```
bf4: For the moment, I would argue this is incorrect.  e.g `User::Mailer`, `Admin::Mailer` would both resolve to mailer. and see https://github.com/rails-api/active_model_serializers/pull/1029#discussion_r37211368

youroff:
The idea was to reject namespace by default:
- There's no standard way to encode namespace yet.
- The way how it was done before is confusing: `Namespace::SomeModel` was converted to `namespace_some_model` which is the same as it would be for the model `NamespaceSomeModel`.

So my idea was to cut namespace by default, but type directive will be there to set it explicitly.

youroff:
For #1029 I would stick with `/` separator for namespace, but it's against json-api standard (hopefully it can be changed to comply that usecase). But `_` or `-` wouldn't work as it's standard separator for words in multiword classes. Which will bring ambiguity.

bf4:
@youroff agree about `_` and `-`.  how about `spam:unrelated_links` ?  Or even `spam::unrelated_links`. it's up to us, and we can create the ruby convention right here.

However, per spec recommendation https://github.com/json-api/json-api/issues/850#issuecomment-140972621 it would be `spam--unrelated_links` which I'm ok with

beauby:
After giving it a thought, I tend to agree with @bf4. A user that uses namespaces should know what they are doing, and overriding the type with `type "MyType"` seems like a good option for those willing to get rid of the namespace.

youroff:
Honestly, `/` looks more reasonable for me, cause it's common sign for paths (which namespaces in fact are for the class hierarchy), so maybe it's worth proposing for json-api team. But right now I can do `--` to move it farther, if everyone agrees with this convention.

BTW, not related to this one, but I discovered recently while playing with ember, that by default it generates attribute names dasherized, and not underscored for payload. Not sure if it's explicitly stated in standard, but the point is that when we use ams with ember the keys are transferred differently encoded for inbound and outbound packets. And the payload for server doesn't seem to be ready for rails because of that, so I got a thought — are there deserializers for AR? Could AMS solve this issue be providing some wrapper for strong parameters?

beauby:
@youroff Let's comply with JSON API for now. If they change the spec, we can always change AMS then.

Considering deserializers, it's a WIP, it's kinda almost ready, but we need to discuss a few things before rebasing/pushing it.

bf4:
See linked issue, ember is 'wrong'

B mobile phone

youroff:
You mean that it doesn't convert `/`?

bf4:
> You mean that it doesn't convert /?

No, that ember uses `/` which is invalid

youroff:
Does ember support namespaced models or it uses it somehow else?

youroff:
Btw, what if we add setting for separator? So that anyone will be able to change it to `/` or anything else for their setup.

bf4:
legit.. new global config

bf4:
Looks like we lost sight of this. Is a good PR. @youroff Would you like to rebase and we'll finish it up?

youroff:
Sorry, missed that on holidays... Sure, will take a look at it soon. What's the final decision on how we deal with namespaces?
And btw, I remember talking about global settings for deriving type name from seralizer name instead of model name. What do you think about that one? Is it worth of implementation?

bf4:
`type` is really the resource type, so should be derived from the object, I think.

`--` is fine until we think of something better.  I've been meaning to make a PR to ember to fix their impl :)

youroff:
Well, in our app we use the same model in the different ways. For example, there is a User -> Membership -> Dealership relation. And Membership reflects into two different Ember models: one ProfileMembership — is a light version to be attachable to profile (and loadable with). So it's something used throughout the site to adjust current user's experience, and another Memebership is manageable by dealership Admin. It contains various additional fields, which are used on the backend. Currently I use type directive to set type explicitly inside of different serializers and having option to derive type name from the name of serializer would make the code a bit cleaner. So I believe it's not that unique a use case and having an option (with taking type from the model as a default) wouldn't hurt anyone, what do you think? And it's pretty easy to implement, I believe I've done some sketch of it already.

So for namespace, we'll have a setting for separator with '--' as default, right?

Btw, what's the way of using namespaces on the side of Ember? If you have some example in mind, could you point out?

youroff:
Here's more generic example... Let's say we have Post model. It may have title, intro, body, lots of metadata. Though in a news feed we only need title and intro. So that would be PostPreview on ember's end. And when user wants to read whole article it would load Post model. Sure we could implement fields parameter, but the approach seems cleaner. Also, not sure about now, but when ember was about version 1-something, it would take record from the store by default, so once loaded reduced record, it wouldn't reload it when called with findRecord.

NullVoxPopuli:
I think we need to wait for the JSON API spec to decide on what to do about namespaces, yeah?

bf4:
@nullvoxpopuli 

>we need to wait for the JSON API spec to decide on what to do about namespaces, yeah?

In 1.0 it is up to us. See linked issue I made re 'kebab' case in the json api repo

remear:
Can we make a decision on this?

youroff:
Actually, I would commit some time to fix that if we have a decission. Just ran into the need of namespace separator compatible with Ember. @bf4?

remear:
I'm late to this conversation, but I'm fine with either `spam--unrelated_links` or `spam__unrelated_links`.

youroff:
I think all agree that we need some global setting for json-api adapter and the last question is what would be the default.

youroff:
Just created a PR for this feature. Codebase changed a lot, so I just made it from scratch. @bf4 could you review?

NullVoxPopuli:
Closing in favor of #16009

PR #1217, opened 2015-09-30
Change default rake task to run test and rubocop

maurogeorge:
The rubocop only runs in the CI this way a contributor probably will see a
rubocop offense only in the CI.

Running the rubocop in the default rake task we have more chance that a offense
be get in the local machine.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -36,4 +36,4 @@ Rake::TestTask.new do |t|
   t.verbose = true
 end
 
-task :default => :test
+task default: [:test, :rubocop]
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,6 @@ install:
   - bundle install --retry=3
 
 test_script:
-  - bundle exec rake
+  - bundle exec rake test
 
 build: off
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ install:
 
 script:
   - env CAPTURE_STDERR=false bundle exec rake
-  - bundle exec rake rubocop
 
 env:
   - "RAILS_VERSION=4.0"


Rakefile
```diff
@@ -36,4 +36,4 @@ Rake::TestTask.new do |t|
   t.verbose = true
 end
 
-task :default => :test
+task default: [:test, :rubocop]
```
bf4: I get where you're coming from, CI may fail due to rubocop...

I'd rather recommend adding a pre-push hook that runs rubocop... or documenting it

Perhaps we could add a `task ci: [:test, :rubocop]`...

NullVoxPopuli: I'm actually ok with this change, cause rubocop on my machines runs pretty fast.
If the resistance is that sometimes you just want to run tests, ya can always explicitly run `rake test`

NullVoxPopuli:
something got messed up with the rubocop run on appveyor :-\

Rakefile
maurogeorge: @bf4 I think a task `ci` do not solves the problem, since the default command when we clone a project and want to check if everything is ok is `$ rake`. And as @NullVoxPopuli said we still can run only test if we want.

What your concern about it? The status code is preserved so, if the tests fail the rubocop does not run and the CI will fail.

bf4: I suppose i can get behind 'rake test' as an alternative....

Thoughts about making autocorrct easy? Make rake does rubocop:auto_correct and rake ci just rubocop?

B mobile phone

NullVoxPopuli: I'm in favor of an autocorrect rake task.

maurogeorge: The problem this patch try to resolve is: 

> Newcomers make a PR and got a error about rubocop only in the travis, how we can show that this project uses rubocop?

The proposed solution is

> Add the rubocop on the default rake task too, since this is natural to people run `$ rake` to check the suite and other things are ok

I think the autocorrect task is not related with this PR.
If the contributor want autocorrect him can use the optional param directly on the rubocop.

Make sense to you guys?

NullVoxPopuli: yeah, def shouldn't be part of this pr.
would be an easy second pr though.

NullVoxPopuli:
:+1: on this, but appveyor seems to have forgotten about the rubocop.yml :-\

maurogeorge:
@NullVoxPopuli I update the code to not run the rubocop on the appveyor, as you can [see here](https://ci.appveyor.com/project/joaomdmoura/active-model-serializers/build/439/job/pqiqku642gtq947e) in the last merged PR, the appveyor does not run the rubocop, I bring back this behavior, running only the tests.

NullVoxPopuli:
I'm good to merge this. @bf4 ?

appveyor.yml
```diff
@@ -21,6 +21,6 @@ install:
   - bundle install --retry=3
 
 test_script:
-  - bundle exec rake
+  - bundle exec rake test
```
bf4: I have to say I don't like this.. not obvious that `rake test` is a way to exclude rubinius

NullVoxPopuli: ah, I didn't catch that.

bf4:
@maurogeorge lemme know if you wanna pair and finish this

bf4:
lulz https://travis-ci.org/rails-api/active_model_serializers/jobs/83169335

```bash

Offenses:
lib/active_model/serializer/adapter/json_api.rb:18:13: C: Style/EmptyLinesAroundAccessModifier: Keep a blank line before and after module_function.
            module_function
            ^^^^^^^^^^^^^^^
93 files inspected, 1 offense detected
RuboCop failed!
```

maurogeorge:
Thanks for the merge :heart:

PR #1224, opened 2015-10-02
Separate default rake from rake ci

bf4:
Follows https://github.com/rails-api/active_model_serializers/pull/1217

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,18 +11,26 @@ begin
 rescue LoadError
 else
   Rake::Task[:rubocop].clear if Rake::Task.task_defined?(:rubocop)
-  if !defined?(::Rubinius)
+  require 'rbconfig'
+  # https://github.com/bundler/bundler/blob/1b3eb2465a/lib/bundler/constants.rb#L2
+  windows_platforms = /(msdos|mswin|djgpp|mingw)/
+  if RbConfig::CONFIG['host_os'] =~ windows_platforms
+    desc 'No-op rubocop on Windows-- unsupported platform'
+    task :rubocop do
+      puts 'Skipping rubocop on Windows'
+    end
+  elsif defined?(::Rubinius)
+    desc 'No-op rubocop to avoid rbx segfault'
+    task :rubocop do
+      puts 'Skipping rubocop on rbx due to segfault'
+      puts 'https://github.com/rubinius/rubinius/issues/3499'
+    end
+  else
     Rake::Task[:rubocop].clear if Rake::Task.task_defined?(:rubocop)
     desc 'Execute rubocop'
     RuboCop::RakeTask.new(:rubocop) do |task|
       task.options = ['--rails', '--display-cop-names', '--display-style-guide']
       task.fail_on_error = true
-    end
-  else
-    desc 'No-op rubocop to avoid rbx segfault'
-    task :rubocop do
-      puts 'Skipping rubocop on rbx due to segfault'
-      puts 'https://github.com/rubinius/rubinius/issues/3499'
     end
   end
 end
@@ -37,3 +45,6 @@ Rake::TestTask.new do |t|
 end
 
 task default: [:test, :rubocop]
+
+desc 'CI test task'
+task :ci => [:default]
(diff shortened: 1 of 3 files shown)


bf4:
@rails-api/ams Any thoughts?

NullVoxPopuli:
I'm good with it. Straight forward

Rakefile
```diff
@@ -36,4 +36,22 @@ Rake::TestTask.new do |t|
   t.verbose = true
 end
 
-task default: [:test, :rubocop]
+task default: [:test, 'rubocop:auto_correct']
```
maurogeorge: `rubocop:auto_correct` will change the code directly? Or we have a prompt before the changes? If we have a prompt, I am :+1:

NullVoxPopuli: I don't think `rubocop -a` prompts. :-/

Rakefile
```diff
@@ -36,4 +36,22 @@ Rake::TestTask.new do |t|
...
+desc 'CI test task'
+task :ci => [:test, 'ci:rubocop']
+namespace :ci do
+  require 'rbconfig'
+  # https://github.com/bundler/bundler/blob/1b3eb2465a64d938490110e84b5ca97a51253d14/lib/bundler/constants.rb#L2
+  windows_platforms = /(msdos|mswin|djgpp|mingw)/
+  if RbConfig::CONFIG['host_os'] !~ windows_platforms
```
maurogeorge: Do you think it is possible to move this Windows check to the before conditional  https://github.com/rails-api/active_model_serializers/pull/1224/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR16 ? I think this task is related to windows at all and not CI Windows only. Makes sense?

Rakefile
maurogeorge: :( So In my opinion :-1: to this, I think it is a little weird you run `$ rake` and get your code modifed after that.

beauby: I agree with @maurogeorge, the `rubocop:auto_correct` should not be part of default.

NullVoxPopuli: why's that?

the auto correction that it does is just easy style things (you still need to do the more complicated ones yourself)

Rakefile
bf4: Oh, good call. that could go in the rubocop task definition for sure

Rakefile
bf4: So, I always run with auto_correct locally since if there's a problem, I would just run it next.  then, I review the changes and either accept them, filter them, or change some config.  I don't see any advantage to running rubocop by default in dev but not auto-correcting.

So, I feel pretty strongly that if we're going to force devs to opt out of rubocop when testing locally, to at least let the rubocop help.  Other repos just remind you what the dev process is, e.g. https://github.com/rubygems/rubygems.org/blob/master/CONTRIBUTING.md#running-rubocop

NullVoxPopuli: I feel similarly, and will probably Implement this on my projects.

bf4:
@rails-api/ams any objections here?

NullVoxPopuli:
Nope :+1:

Rakefile
```diff
@@ -37,3 +45,6 @@ Rake::TestTask.new do |t|
 end
 
 task default: [:test, :rubocop]
+
+desc 'CI test task'
+task :ci => [:default]
```
bf4: It's good to have a ci task, but you're right, not necessary

B mobile phone

Rakefile
```diff
@@ -11,19 +11,27 @@ begin
...
-  if !defined?(::Rubinius)
-    Rake::Task[:rubocop].clear if Rake::Task.task_defined?(:rubocop)
-    desc 'Execute rubocop'
-    RuboCop::RakeTask.new(:rubocop) do |task|
-      task.options = ['--rails', '--display-cop-names', '--display-style-guide']
-      task.fail_on_error = true
+  require 'rbconfig'
+  # https://github.com/bundler/bundler/blob/1b3eb2465a/lib/bundler/constants.rb#L2
+  windows_platforms = /(msdos|mswin|djgpp|mingw)/
+  if RbConfig::CONFIG['host_os'] =~ windows_platforms
```
NullVoxPopuli: I think this alone is reason enough to merge this PR.

Any objections?

PR #1225, opened 2015-10-02
Add support for nested serializers

beauby:
This is half of #1193. This PR brings lookup for serializer classes in different namespaces (namely that of the parent serializer, that of the resource), so that the following is possible:

```ruby
class PostSerializer < AMS
  has_many :comments
  class CommentSerializer < AMS
    ...
  end
end
```
(this is a way to make single-purpose nested serializers contained in one class instead of polluting the global namespace and having to call them `PostCommentSerializer`).

Cases not handled yet:
 + looking up for a serializer in the root namespace of a chain of serializers (post has_many comments belongs_to author: when a `Post` is serialized with a `Namespace::PostSerializer`, we might want to look for `Namespace::AuthorSerializer`),
 + looking up for a serializer in the controller's namespace (when `Namespace::PostController` serializes a `Post`, we might want to look for `Namespace::PostSerializer`).

### DIFF
--- a/docs/general/getting_started.md
+++ b/docs/general/getting_started.md
@@ -57,6 +57,37 @@ class CommentSerializer < ActiveModel::Serializer
 end
 ```
 
+### Namespaced Models
+
+When serializing a model inside a namespace, such as `Api::V1::Post`, AMS will expect the corresponding serializer to be inside the same namespace (namely `Api::V1::PostSerializer`).
+
+### Model Associations and Nested Serializers
+
+When declaring a serializer for a model with associations, such as:
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  has_many :comments
+end
+```
+AMS will look for `PostSerializer::CommentSerializer` in priority, and fall back to `::CommentSerializer` in case the former does not exist. This allows for more control over the way a model gets serialized as an association of an other model.
+
+For example, in the following situation:
+
+```ruby
+class CommentSerializer < ActiveModel::Serializer
+  attributes :body, :date, :nb_likes
+end
+
+class PostSerializer < ActiveModel::Serializer
+  has_many :comments
+  class CommentSerializer < ActiveModel::Serializer
+    attributes :body_short
+  end
+end
+```
+
+AMS will use `PostSerializer::CommentSerializer` (thus including only the `:body_short` attribute) when serializing a `Comment` as part of a `Post`, but use `::CommentSerializer` when serializing a `Comment` directly (thus including `:body, :date, :nb_likes`).
+
 ## Rails Integration
 
 AMS will automatically integrate with you Rails app, you won't need to update your controller, this is a example of how it will look like:
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -108,10 +108,25 @@ module ActiveModel
       Digest::MD5.hexdigest(serializer_file_contents)
     end
 
+    # @api private
+    def self.serializer_lookup_chain_for(klass)
+      chain = []
+
+      resource_class_name = klass.name.demodulize
+      resource_namespace = klass.name.deconstantize
+      serializer_class_name = "#{resource_class_name}Serializer"
+
+      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
+      chain.push("#{resource_namespace}::#{serializer_class_name}")
+
+      chain
+    end
+
+    # @api private
     def self.get_serializer_for(klass)
       serializers_cache.fetch_or_store(klass) do
-        serializer_class_name = "#{klass.name}Serializer"
-        serializer_class = serializer_class_name.safe_constantize
+        # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
+        serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x }
 
         if serializer_class
           serializer_class
--- a/lib/active_model/serializer/array_serializer.rb
+++ b/lib/active_model/serializer/array_serializer.rb
@@ -11,9 +11,8 @@ module ActiveModel
         @root = options[:root]
         @object = resources
         @serializers = resources.map do |resource|
-          serializer_class = options.fetch(:serializer) do
-            ActiveModel::Serializer.serializer_for(resource)
-          end
+          serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)
+          serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
 
           if serializer_class.nil?
             fail NoSerializerError, "No serializer found for resource: #{resource.inspect}"
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -42,13 +42,13 @@ module ActiveModel
       def build_association(subject, parent_serializer_options)
         association_value = subject.send(name)
         reflection_options = options.dup
-        serializer_class = ActiveModel::Serializer.serializer_for(association_value, reflection_options)
+        serializer_class = subject.class.serializer_for(association_value, reflection_options)
 
         if serializer_class
           begin
             serializer = serializer_class.new(
               association_value,
-              serializer_options(parent_serializer_options, reflection_options)
+              serializer_options(subject, parent_serializer_options, reflection_options)
             )
           rescue ActiveModel::Serializer::ArraySerializer::NoSerializerError
             reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
@@ -62,11 +62,12 @@ module ActiveModel
 
       private
 
-      def serializer_options(parent_serializer_options, reflection_options)
+      def serializer_options(subject, parent_serializer_options, reflection_options)
         serializer = reflection_options.fetch(:serializer, nil)
 
         serializer_options = parent_serializer_options.except(:serializer)
         serializer_options[:serializer] = serializer if serializer
+        serializer_options[:serializer_context_class] = subject.class
         serializer_options
       end
     end
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -125,6 +125,88 @@ module ActiveModel
         assert expected_association_keys.include? :writer
         assert expected_association_keys.include? :site
       end
+
+      class NamespacedResourcesTest < Minitest::Test
+        class ResourceNamespace
+          Post    = Class.new(::Model)
+          Comment = Class.new(::Model)
+          Author  = Class.new(::Model)
+          Description = Class.new(::Model)
+          class PostSerializer < ActiveModel::Serializer
+            has_many :comments
+            belongs_to :author
+            has_one :description
+          end
+          CommentSerializer     = Class.new(ActiveModel::Serializer)
+          AuthorSerializer      = Class.new(ActiveModel::Serializer)
+          DescriptionSerializer = Class.new(ActiveModel::Serializer)
+        end
+
+        def setup
+          @comment = ResourceNamespace::Comment.new
+          @author = ResourceNamespace::Author.new
+          @description = ResourceNamespace::Description.new
+          @post = ResourceNamespace::Post.new(comments: [@comment],
+                                              author: @author,
+                                              description: @description)
+          @post_serializer = ResourceNamespace::PostSerializer.new(@post)
+        end
+
+        def test_associations_namespaced_resources
+          @post_serializer.associations.each do |association|
+            case association.key
+            when :comments
+              assert_instance_of(ResourceNamespace::CommentSerializer, association.serializer.first)
+            when :author
+              assert_instance_of(ResourceNamespace::AuthorSerializer, association.serializer)
+            when :description
+              assert_instance_of(ResourceNamespace::DescriptionSerializer, association.serializer)
+            else
+              flunk "Unknown association: #{key}"
+            end
+          end
+        end
+      end
+
+      class NestedSerializersTest < Minitest::Test
+        Post    = Class.new(::Model)
+        Comment = Class.new(::Model)
+        Author  = Class.new(::Model)
+        Description = Class.new(::Model)
+        class PostSerializer < ActiveModel::Serializer
+          has_many :comments
+          CommentSerializer = Class.new(ActiveModel::Serializer)
+          belongs_to :author
+          AuthorSerializer = Class.new(ActiveModel::Serializer)
+          has_one :description
+          DescriptionSerializer = Class.new(ActiveModel::Serializer)
+        end
+
+        def setup
+          @comment = Comment.new
+          @author = Author.new
+          @description = Description.new
+          @post = Post.new(comments: [@comment],
+                           author: @author,
+                           description: @description)
+          @post_serializer = PostSerializer.new(@post)
+        end
+
+        def test_associations_namespaced_resources
+          @post_serializer.associations.each do |association|
+            case association.key
+            when :comments
+              assert_instance_of(PostSerializer::CommentSerializer, association.serializer.first)
+            when :author
+              assert_instance_of(PostSerializer::AuthorSerializer, association.serializer)
+            when :description
+              assert_instance_of(PostSerializer::DescriptionSerializer, association.serializer)
+            else
+              flunk "Unknown association: #{key}"
+            end
+          end
+        end
+      end
     end
   end
 end
--- a/test/serializers/serializer_for_test.rb
+++ b/test/serializers/serializer_for_test.rb
@@ -27,8 +27,19 @@ module ActiveModel
       end
 
       class SerializerTest < Minitest::Test
+        module ResourceNamespace
+          Post    = Class.new(::Model)
+          Comment = Class.new(::Model)
+
+          class PostSerializer < ActiveModel::Serializer
+            class CommentSerializer < ActiveModel::Serializer
+            end
+          end
+        end
+
         class MyProfile < Profile
         end
+
         class CustomProfile
           def serializer_class; ProfileSerializer; end
         end
@@ -58,6 +69,18 @@ module ActiveModel
         def test_serializer_custom_serializer
           serializer = ActiveModel::Serializer.serializer_for(@custom_profile)
           assert_equal ProfileSerializer, serializer
+        end
+
+        def test_serializer_for_namespaced_resource
+          post = ResourceNamespace::Post.new
+          serializer = ActiveModel::Serializer.serializer_for(post)
+          assert_equal(ResourceNamespace::PostSerializer, serializer)
+        end
+
+        def test_serializer_for_nested_resource
+          comment = ResourceNamespace::Comment.new
+          serializer = ResourceNamespace::PostSerializer.serializer_for(comment)
+          assert_equal(ResourceNamespace::PostSerializer::CommentSerializer, serializer)
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Breaking changes:
 
 Features:
 
+- [#1225](https://github.com/rails-api/active_model_serializers/pull/1125) Better serializer lookup, use nested serializer when it exists (@beauby)
 - [#1172](https://github.com/rails-api/active_model_serializers/pull/1172) Better serializer registration, get more than just the first module (@bf4)
 - [#1158](https://github.com/rails-api/active_model_serializers/pull/1158) Add support for wildcards in `include` option (@beauby)
 - [#1127](https://github.com/rails-api/active_model_serializers/pull/1127) Add support for nested


lib/active_model/serializer.rb
```diff
@@ -112,10 +112,25 @@ def self.digest_caller_file(caller_line)
       Digest::MD5.hexdigest(serializer_file_contents)
     end
 
+    # TODO(beauby): Cache this.
+    def self.serializer_lookup_chain_for(klass)
```
NullVoxPopuli: I like this extraction.

lib/active_model/serializer.rb
```diff
@@ -112,10 +112,25 @@ def self.digest_caller_file(caller_line)
       Digest::MD5.hexdigest(serializer_file_contents)
     end
 
+    # TODO(beauby): Cache this.
```
NullVoxPopuli: soonish? :-)

NullVoxPopuli: though, is this really that expensive of a method?

beauby: I'd like to do it in a subsequent PR in order to keep this one to a minimum.

lib/active_model/serializer.rb
NullVoxPopuli: some documentation on how this works (when to use, expected input/output, etc) would be fantabulous

lib/active_model/serializer.rb
beauby: Actually, caching here should not even be needed as the final serializer class will be cached.

test/serializers/serializer_for_test.rb
```diff
@@ -27,6 +27,16 @@ def test_overwritten_serializer_for_array
       end
 
       class SerializerTest < Minitest::Test
+        module ResourceNamespace
+          class Post < ::Model
```
NullVoxPopuli: is it bad form to do one line classes when the body is empty?

lib/active_model/serializer.rb
beauby: Looks pretty much self-explanatory to me. What kind of comment would you add that would not be direct paraphrasing of the code?

lib/active_model/serializer.rb
NullVoxPopuli: :-)

test/serializers/serializer_for_test.rb
beauby: No, it's actually the recommended way, I'll update this.

lib/active_model/serializer.rb
NullVoxPopuli: me too, but I would like to prefer to over document than to under document. Like, look at how rails  documents everything with plenty of examples: https://github.com/rails/rails/blob/master/activemodel/lib/active_model/errors.rb

NullVoxPopuli: (I guess that is more or less direct paraphrasing in some cases though) :-\

beauby: Yeah but `serializer_lookup_chain_for` is part of the private API.

NullVoxPopuli: fair point

test/serializers/serializer_for_test.rb
```diff
@@ -59,6 +70,18 @@ def test_serializer_custom_serializer
...
+        def test_serializer_for_namespaced_resource
+          post = ResourceNamespace::Post.new
+          serializer = ActiveModel::Serializer.serializer_for(post)
+          assert_equal(ResourceNamespace::PostSerializer, serializer)
+        end
+
+        def test_serializer_for_nested_resource
+          comment = ResourceNamespace::Comment.new
+          serializer = ResourceNamespace::PostSerializer.serializer_for(comment)
+          assert_equal(ResourceNamespace::PostSerializer::CommentSerializer, serializer)
```
bf4: what's missing from the tests is when `ResourceNamespace::PostSerializer.serializer_for` is called.  the tests don't confirm this is hooked up anywhere

lib/active_model/serializer/reflection.rb
```diff
@@ -42,13 +42,13 @@ class Serializer
       def build_association(subject, parent_serializer_options)
         association_value = subject.send(name)
         reflection_options = options.dup
-        serializer_class = ActiveModel::Serializer.serializer_for(association_value, reflection_options)
+        serializer_class = subject.class.serializer_for(association_value, reflection_options)
```
bf4: :+1:

lib/active_model/serializer/reflection.rb
```diff
@@ -42,13 +42,13 @@ class Serializer
...
 
         if serializer_class
           begin
             serializer = serializer_class.new(
               association_value,
-              serializer_options(parent_serializer_options, reflection_options)
+              serializer_options(parent_serializer_options, reflection_options).merge(_parent_serializer: subject.class)
```
bf4: this is dup'd above, so it's safe to mutate.  should probably do that where it's dup'd

lib/active_model/serializer/array_serializer.rb
```diff
@@ -10,9 +10,10 @@ class ArraySerializer
       def initialize(resources, options = {})
         @root = options[:root]
         @object = resources
+        parent_serializer = options.fetch(:_parent_serializer, ActiveModel::Serializer)
```
bf4:
```ruby
serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)
serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
```

lib/active_model/serializer/reflection.rb
beauby: True, I wanted to keep this PR as small as possible just to show the idea first. I could do that inside `serializer_options` (it would require an extra `subject` argument though).

lib/active_model/serializer/array_serializer.rb
bf4: I'm thinking `context` requires less thinking than `parent`

test/serializers/serializer_for_test.rb
beauby: I don't understand. `ResourceNamespace::PostSerializer.serializer_for` is called just above?

lib/active_model/serializer/array_serializer.rb
beauby: Agreed, will modify accordingly.

lib/active_model/serializer/reflection.rb
bf4:
```
serializer_context_class = subject.class
reflection_options = options.dup.merge!(serializer_context_class: serializer_context_class)
serializer_class = serializer_context_class.serializer_for(association_value, reflection_options)
# etc, no changes needed below
```

lib/active_model/serializer.rb
```diff
@@ -112,10 +112,24 @@ def self.digest_caller_file(caller_line)
...
+
+      chain
+    end
+
     def self.get_serializer_for(klass)
       serializers_cache.fetch_or_store(klass) do
-        serializer_class_name = "#{klass.name}Serializer"
-        serializer_class = serializer_class_name.safe_constantize
+        # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
+        serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x }
```
bf4: I'm going to block this PR till the code is more readable

behavior looks great!

test/serializers/serializer_for_test.rb
bf4: What I see is:

```ruby
ActiveModel::Serializer.serializer_for(post) #=> ResourceNamespace::PostSerializer
ResourceNamespace::PostSerializer.serializer_for(comment) #=> ResourceNamespace::PostSerializer::CommentSerializer
```

which tests serialization from an AM::S subclass, but not that the subclass is ever used.. maybe my comment just means there should be another test somewhere else

lib/active_model/serializer/reflection.rb
bf4: see https://github.com/rails-api/active_model_serializers/pull/1225#discussion_r41045787

bf4:
Nice and slim! good encapsulation!

test/serializers/serializer_for_test.rb
bf4: I didn't know there was any consensus on `class Foo; end` vs. `class Foo\n end`

lib/active_model/serializer.rb
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1225/files#r41045890

lib/active_model/serializer.rb
this line really should be its own multi-line method.

test/serializers/serializer_for_test.rb
NullVoxPopuli: I personally prefer single line things when there is no block content.

beauby: https://github.com/bbatsov/ruby-style-guide#single-line-classes

test/serializers/serializer_for_test.rb
beauby: Got it.

beauby: We need to add a test to `test/serializers/associations_test.rb` to make sure the corresponding serializer gets used. This will be doable in a decent way once the test infrastructure refactor has been done.

lib/active_model/serializer.rb
beauby: This line simply does try every constant in the lookup chain until one that exists is found. How would you make it more readable?

beauby: This line simply does try every constant in the lookup chain until one that exists is found. How would you make it more readable?

lib/active_model/serializer/reflection.rb
beauby: `serializer_context_class`/`parent_serializer` (however we may call it) is not a reflection option but a serializer option. I agree that there is room for improvement though.

beauby: Modified the code to set the `serializer_context_class` option inside the `serializer_options` method.

lib/active_model/serializer.rb
beauby: @NullVoxPopuli Plus, there already are `serializer_for` and `get_serializer_for`, so if we extract this line into its own method, what would it be? `get_actual_serializer_for`?

NullVoxPopuli: Serializer from lookup chain?

test/serializers/serializer_for_test.rb
beauby: Added tests.

test/serializers/serializer_for_test.rb
bf4: I strongly disagree in the context of a class this is neither a semaphore nor struct nor exception object.

beauby: Any specific reason why?

lib/active_model/serializer.rb
bf4: y'all are discussing the method name, I was referring to the code

beauby: For the code, see [my remark](https://github.com/rails-api/active_model_serializers/pull/1225#discussion_r41082841).

beauby: I could be convinced replacing `.find { |x| x }` with `.find { |x| !x.nil? }` might be better, but other than that I don't really see what's not readable about that line.

bf4: ok, to be clearer, I'm including the source of the new method as well

NullVoxPopuli: I think chaining methods together is a code smell, and it'd be better to extract to a method to also increase testability, especially for stubbing purposes. (I do a lot of this pattern with all my rspec projects)

beauby: @NullVoxPopuli Chaining method calls is kinda the base of functional programming.

lib/active_model/serializer.rb
```diff
@@ -112,10 +112,24 @@ def self.digest_caller_file(caller_line)
...
+      resource_class_name = klass.name.demodulize
+      resource_namespace = klass.name.deconstantize
+      serializer_class_name = "#{resource_class_name}Serializer"
+
+      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
+      chain.push("#{resource_namespace}::#{serializer_class_name}")
+      chain.push(serializer_class_name)
+
+      chain
+    end
```
beauby: @bf4 So if you were referring to this part of code saying it was not easily understandable, then I agree. Suggestions? I could add comments, but I don't see how to make the code itself clearer.

lib/active_model/serializer.rb
NullVoxPopuli: I just mean that chaining method calls in a method who's responsibility is outside the scope of the method chain is kinda smelly.

I'm fine with a separate method having the method chain, as long as that is that method's only responsibility

beauby: I don't really see the problem here. This methods' purpose is to find the serializer class, which this line actually does.

test/serializers/associations_test.rb
```diff
@@ -125,6 +125,88 @@ def test_associations_custom_keys
...
+          @post_serializer.associations.each do |association|
+            case association.key
+            when :comments
+              assert_instance_of(ResourceNamespace::CommentSerializer, association.serializer.first)
+            when :author
+              assert_instance_of(ResourceNamespace::AuthorSerializer, association.serializer)
+            when :description
+              assert_instance_of(ResourceNamespace::DescriptionSerializer, association.serializer)
+            else
+              flunk "Unknown association: #{key}"
```
bf4: never seen that, hmm

test/serializers/associations_test.rb
```diff
@@ -125,6 +125,88 @@ def test_associations_custom_keys
...
+      end
+
+      class NestedSerializersTest < Minitest::Test
+        Post    = Class.new(::Model)
+        Comment = Class.new(::Model)
+        Author  = Class.new(::Model)
+        Description = Class.new(::Model)
+        class PostSerializer < ActiveModel::Serializer
+          has_many :comments
+          CommentSerializer = Class.new(ActiveModel::Serializer)
```
bf4: does the order matter?

test/serializers/associations_test.rb
beauby: It was all around so I sticked with the current convention.

test/serializers/associations_test.rb
beauby: No it does not, but it felt natural to group stuff by "feature".

NullVoxPopuli:
can you rebase?

beauby:
Squashed.

lib/active_model/serializer.rb
bf4: It makes me think too much. I'd rather put it all in one complicated place

```ruby
      chain = []

      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
      chain.push("#{resource_namespace}::#{serializer_class_name}")
      chain.push(serializer_class_name)

        if serializer_class
          serializer_class
        elsif klass.superclass
          get_serializer_for(klass.superclass)
        end
      end
    end
```

or

```diff
    def self.get_serializer_for(klass)
      serializers_cache.fetch_or_store(klass) do
-      chain = []

      resource_class_name = klass.name.demodulize
      resource_namespace = klass.name.deconstantize
      serializer_class_name = "#{resource_class_name}Serializer"

 -     chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
+      serializer_class = "#{name}::#{serializer_class_name}".safe_constantize
_      chain.push("#{resource_namespace}::#{serializer_class_name}")
+      serializer_class ||= "#{resource_namespace}::#{serializer_class_name}".safe_constantize

-      chain.push(serializer_class_name)
+      serializer_class ||= serializer_class_name.safe_constantize

        if serializer_class
          serializer_class
        elsif klass.superclass
          get_serializer_for(klass.superclass)
        end
      end
    end
```

bf4: and maybe annotate it with an example of what's going on, like in Adapter.get

beauby: I'm definitely against a sequence of `serializer_class ||=`. What's wrong with explicitly create a lookup chain (which is what we would do implicitly anyways), and then use it to lookup the serializer?

bf4: It's harder to understand in the current state, and it's just another utility method on the Serializer

beauby: I guess it's a personal opinion, because to me, separating the concerns, using common patterns and having short methods is easier to understand than large methods of ad-hoc code.

bf4: In general I agree.  Just in this case I think it's easier to read inline, and probably be extracted into a utility as we refactor lookup stuff.  The nice thing about it being inline is there's nothing else to support until we want to. :)

beauby: How about we declare `serializer_lookup_chain_for` as part of the private API and get both benefits?

NullVoxPopuli: I think the chain is fine as long as it's in its own method and cached.

NullVoxPopuli: that would simplify the look of this method as well, and improve skimability.

bf4: the chain is less efficient than my suggestion and isn't a very revealing method name.  Lemme try to do it:

```ruby
    SerializerForResource = Struct.new(:lookup_serializer, :resource_klass) do
      def lookup
        serializer_class = nil
        potential_serializer_class_names.each do |serializer_class_name|
          serializer_class = serializer_class_name.safe_constantize
          break if serializer_class
        end

        if serializer_class
          serializer_class
        elsif klass.superclass
          lookup_serializer.get_serializer_for(resource_klass.superclass)
        end
      end

      # @example
      #   lookup_serializer = AdminSerializer
      #   resource_klass = Admin::User
      #   SerializerForResource.new(lookup_serializer, resource_klass).lookup
      #   #=> %w[AdminSerializer::UserSerializer Admin::UserSerializer UserSerializer]
      def potential_serializer_class_names
        chain = []
        # Given Admin::User
        resource_class_name = klass.name.demodulize    # User
        resource_namespace = klass.name.deconstantize # Admin

        chain.push("#{name}::#{serializer_class_name}") if lookup_serializer != ActiveModel::Serializer
        chain.push("#{resource_namespace}::#{serializer_class_name}")
        chain.push(serializer_class_name)

        chain
      end
    end
    def self.get_serializer_for(resource_klass)
      serializers_cache.fetch_or_store(resource_klass) do
        SerializerForResource.new(self, resource_klass).lookup
      end
    end
```

beauby: So your points are 1. you don't like the name of the method that computes the lookup chain, 2. you dislike the use of functional operators and 3. you'd rather have this in a separate object.

I don't feel strongly about 1., but I don't find the name you suggest better (you seem to dislike the concept of "lookup chain", any specific reason?). I feel quite strongly about 2. because more ad-hoc code = more bugs, whereas functional expression = less code, clearer behavior. I don't feel strongly about 3. although it seems a bit overkill at this stage.

NullVoxPopuli: For me, it's more a separation of concerns thing. And I believe code should be REALLY fast to just skim over and have an idea about what it's doing -- so I tend to prefer A LOT of method abstraction to help keep lines narrow and potential indentation low.

however, I think the name is fine.

bf4: > you don't like the name of the method that computes the lookup chain

or I don't like the name of the method that build a list of possible serializer class names :)

> 2. you dislike the use of functional operators and

where?  we don't have lazy and the performance of the each with break should be about the same.  I have to admin that I don't understand a lot of your intentions when it comes to functional programming.  I'd love to be schooled

> 3. you'd rather have this in a separate object.

maybe eventually probably.  that's why I wanted to be conservative and keep it all together for now.

> you seem to dislike the concept of "lookup chain", any specific reason?

it doesn't tell me what it's doing except that it is a chain of something called 'lookups'

> more ad-hoc code = more bugs, whereas functional expression = less code, clearer behavior. 

I'm not sure what you're referring to.  I like encapsulating behavior of functions in a class.

> I don't feel strongly about 3. although it seems a bit overkill at this stage.

agreed.  can we come to some agreement plz?  I'm trying.  Can you sell me on it with more details than saying it's more functional?

beauby: For naming, `serializer_lookup_chain` refers to the "lookup chain" for serializers, i.e. the chain of possibilities used for looking up the right serializer (and not a "lookups chain" which would be a chain of lookups, which I'm not even sure what it would mean).

As for functional vs ad-hoc, I'm referring to
```ruby
serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x }
```
vs
```ruby
        serializer_class = nil
        potential_serializer_class_names.each do |serializer_class_name|
          serializer_class = serializer_class_name.safe_constantize
          break if serializer_class
        end
```

NullVoxPopuli:
```ruby
serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x }
```

ran be rewritten as 

```ruby
def class_from_serializer_lookup_chain(klass)
  serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x }
end
```

I think extracting the method is possibly half way between what you have, @beauby and @bf4's struct idea

bf4: > `serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x }`

but that's going to unnecessarily iterate over every element when it doesn't need to.  and it's only shorter because of syntactic sugar and laconic naming

`serializer_class = serializer_lookup_chain_for(klass).find{ |x| s = x.safe_constantize; break s if s}`

beauby: Once we wrop support for 1.9.3, which we all agreed on, we can just have `serializer_lookup_chain_for` return a lazy iterator. Map is not syntactic sugar, it is one of the base constructs of functional programming. Plus, it really describes what we're doing here: iterate over the lookup chain, apply `safe_constantize`, and return the first one for which the result is not nil.

bf4: @beauby seriously, I meant the symbol to proc, not map

beauby: My bad. Then it is, indeed syntactic sugar for
```ruby
serializer_lookup_chain_for(klass).map { |x| x.safe_constantize }.find { |x| x }
```

lib/active_model/serializer/array_serializer.rb
```diff
@@ -11,9 +11,8 @@ def initialize(resources, options = {})
         @root = options[:root]
         @object = resources
         @serializers = resources.map do |resource|
-          serializer_class = options.fetch(:serializer) do
-            ActiveModel::Serializer.serializer_for(resource)
-          end
+          serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)
+          serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
```
bf4: Potential :bug: `serializer_context_class` is a required option.  Would you mind adding a comment to the initializer to that effect? and if it's absent, it'll throw a KeyError.  How do we want to handle that?  

Maybe just say "don't use ArraySerializer directly, Just use SerializableResource" :)

beauby: The previous line(`serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)`) ensures it is not required. Am I misunderstanding?

test/serializers/associations_test.rb
```diff
@@ -125,6 +125,88 @@ def test_associations_custom_keys
...
+        end
+
+        def setup
+          @comment = ResourceNamespace::Comment.new
+          @author = ResourceNamespace::Author.new
+          @description = ResourceNamespace::Description.new
+          @post = ResourceNamespace::Post.new(comments: [@comment],
+                                              author: @author,
+                                              description: @description)
+          @post_serializer = ResourceNamespace::PostSerializer.new(@post)
```
bf4: serializer in global scope for completeness (unless I'm just not seeing it)

lib/active_model/serializer/array_serializer.rb
bf4: nope, I'm :facepunch: oops never mind me

test/serializers/associations_test.rb
beauby: So the previous behavior was not to look for a serializer in the global namespace unless the model itself is in the global namespace, which this PR preserves.

beauby: So, per se, those tests were not breaking before this PR, but nonetheless I think we should have such tests to ensure the behavior.

lib/active_model/serializer.rb
```diff
@@ -108,10 +108,25 @@ def self.digest_caller_file(caller_line)
...
+      resource_namespace = klass.name.deconstantize
+      serializer_class_name = "#{resource_class_name}Serializer"
+
+      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
+      chain.push("#{resource_namespace}::#{serializer_class_name}")
+
+      chain
+    end
+
+    # @api private
```
bf4: giving us the license to change it :) :+1:

test/serializers/associations_test.rb
bf4: do you want to add one? else I'll merge it

docs/general/getting_started.md
```diff
@@ -57,6 +57,37 @@ class CommentSerializer < ActiveModel::Serializer
...
+class PostSerializer < ActiveModel::Serializer
+  has_many :comments
+  class CommentSerializer < ActiveModel::Serializer
+    attributes :body_short
+  end
+end
+```
+
+AMS will use `PostSerializer::CommentSerializer` (thus including only the `:body_short` attribute) when serializing a `Comment` as part of a `Post`, but use `::CommentSerializer` when serializing a `Comment` directly (thus including `:body, :date, :nb_likes`).
+
```
bf4: outside of this PR, but this doesn't include the superclass scenario and it might be interesting to compare `Api::V1::PostSerializer` to `PostSerializer`

beauby:
The build failed on appveyor, not sure why as I can't consult the logs.

beauby:
Actually, I have the answer: 
> Build execution time has reached the maximum allowed time for your plan (40 minutes).

bf4:
Yikes. I'll write a timeout max for ci

B mobile phone

bf4:
> Yikes. I'll write a timeout max for ci

hmm, I don't really know how to do around hooks Minitest without hacking it

bf4:
oh weird, it succeeded `build success`, just never exited, it appears

bf4:
try `git commit -m 'rebuild appveyour' --allow-empty; git push -f` to force it to rebuild?

NullVoxPopuli:
:+1: LGTM

Issue #1025, opened 2015-07-29
Serializing Relationship Without Include


jfelchner:
I've been looking through the code and there doesn't appear to be a way to serialize a `relationship` without also serializing an `include`.  Let me explain:

We have two systems.  In one we have, let's say, a `Product` and in the other we have, for example, an `Alert`.  An `Alert` has a `product_id` on it.  For the purposes of our backend systems, the alert system could care less about the `Product`.  All it needs is the ID.

So when I serialize the `Alert`, I want it to be able to generate the `relationship` but ignore the part where it puts the `include` section in.  I also don't want it to care about the model having a _literal_ association on it.  The `Alert` doesn't `belong_to :product` from an `ActiveRecord` perspective, but when the `Alert` JSON gets down to the client, the client will know how to go and fetch the product if it doesn't already have it.

This is what I was hoping that I could do:

```ruby
class AlertSerializer < ActiveModel::Serializer
  belongs_to :product, include: false,
                       type:    'my_other_app_module_product'
end
```

Obviously, the default could be that if you say `belongs_to :product`, it would assume an attribute of `product_id` and a type of `product`, but overriding the `type` is important because sometimes (as in our case) it will not be able to be inferred from the association name.

Anyway, the above would generate something like:

```json
{
  "data" : {
    "type" : "my_app_module_alert",
    "id" : "my_alert_id",
    "attributes" : {
      /* Attributes */
    },
    "relationships" : {
      "product" : {
        "data" : { "id" : "my_product_id", "type" : "my_other_app_module_product" }
      }
    }
  }
}
```

beauby:
So by default the JsonApi adapter *does not* put the related objects in the `include`, *and* it still serializes the defined *relationship objects*. However, it currently relies on the corresponding object to exist an be available as a method on the parent. You could probably get away with defining a method on your model that would behave as expected by AMS to solve your situation. I am also thinking/working on some way to use the `#{relationship}_id` method when available for to-one relationships.

scttnlsn:
I'm having this issue with the `FlattenJson` adapter.  I'd like to be able to pass in `include: []` and have none of the relationships included in the serializable hash (currently relationships 1 level deep are included by default it seems?).  In other cases I'd like to explicitly include various nested relationships.  Is that possible somehow?

NullVoxPopuli:
@scttnlsn that's a good idea. I'll check on that real quick, I'm not sure if that's currently possible.

[...]

PR #1227, opened 2015-10-02
add test to make sure that it is possible to not have any association…

NullVoxPopuli:
…s serialized

related: https://github.com/rails-api/active_model_serializers/issues/1025#issuecomment-145054908

### DIFF
--- a/test/adapter/json/include_test.rb
+++ b/test/adapter/json/include_test.rb
@@ -1,0 +1,37 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
+      class Json
+        class IncludeTest < Minitest::Test
+          def setup
+            @author = Author.new(id: 1, name: 'Steve K.')
+
+            @post = Post.new(id: 42, title: 'New Post', body: 'Body')
+            @post.author = @author
+            @author.posts = [@post]
+            @author.roles = []
+            @author.bio = {}
+          end
+
+          def test_no_associations
+            resource = SerializableResource.new(
+              @author,
+              adapter: :json,
+              include: [])
+
+            expected = {
+              author: {
+                id: 1,
+                name: 'Steve K.'
+              }
+            }
+
+            assert_equal(expected, resource.serializable_hash)
+          end
+        end
+      end
+    end
+  end
+end


test/action_controller/json/include_test.rb
```diff
@@ -36,6 +36,11 @@ def render_without_include
             render json: @author, adapter: :json
           end
 
+          def render_without_associations
+            setup_data
+            render json: @author, adapter: :json, include: []
```
beauby: There's no need anymore to go through a `render` call to test that kind of thing. Just pass the option to `SerializableResource`.

NullVoxPopuli: do we have a file that tests include output, specifically? or is it kinda scattered?

beauby: It is kinda adapter-specific, so no, not really.

bf4: unless you're testing the controller, unit test is better :)  also, see test/support, there's some serialization helpers in there

bf4: I want to remove most of the controller tests are that really now 'SerializableResource' tests.. would clean things up I think

NullVoxPopuli: :+1:

NullVoxPopuli:
I moved the test. :-)

test/adapter/json/include_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
+      class Json
+        class IncludeTest < Minitest::Test
+          def setup
+            ActionController::Base.cache_store.clear
```
beauby: This is not needed as the cache is automatically cleared before each test.

test/adapter/json/include_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+            @blog = Blog.new(id: 1, name: 'My Blog!!')
+            @post.blog = @blog
+            @author.posts = [@post]
+
+            @first_comment.author = @author
+            @second_comment.author = @author
+            @author.comments = [@first_comment, @second_comment]
+            @author.roles = []
+            @author.bio = {}
```
beauby: Most of this is not needed. Let's try to keep the tests as minimal as possible. For now, only one association is needed for your test.

NullVoxPopuli:
rebased

NullVoxPopuli:
@beauby @bf4 ?

NullVoxPopuli:
I'm pretty sure there are already tests that achieve what this test is doing.

Issue #1206, opened 2015-09-27
[RFC] Clean up/rethink test infrastructure


beauby:
Currently the tests use models and serializers defined in `test/fixtures/poro.rb` and `test/fixtures/active_record.rb` (as well as some dynamically defined/locally scoped serializers inside tests) in combination with actual fixtures defined in the `setup` method of many test classes, which leads to lots of code duplication.
I am opening this RFC to discuss alternatives to this, as I feel writing tests should be easier and cleaner.

NullVoxPopuli:
I've always been a fan of having a support folder under my test directory, and then having files named after groups of what they represent. I guess in our case, it could be: models.rb, serializers.rb, model_support.rb (this one could be the model object itself) etc.

bf4:
ref something I wrote in RSpec on 0.10.0.rc1, https://github.com/rails-api/active_model_serializers/issues/878#issuecomment-96073030 as I will (hopefully) get to in https://github.com/rails-api/active_model_serializers/pull/967

PR #1228, opened 2015-10-02
[WIP] Refactor test infrastructure

beauby:
This PR aims to refactor the test infrastructure so that models and serializers are defined where they are used instead of having a big mess into `fixtures/poro.rb`/`fixtures/active_record.rb`.

Ref: #1206.

### DIFF
--- a/test/adapter/json/belongs_to_test.rb
+++ b/test/adapter/json/belongs_to_test.rb
@@ -18,27 +18,37 @@ module ActiveModel
             @post.blog = @blog
             @anonymous_post.blog = nil
 
-            @serializer = CommentSerializer.new(@comment)
-            @adapter = ActiveModel::Serializer::Adapter::Json.new(@serializer)
             ActionController::Base.cache_store.clear
           end
 
           def test_includes_post
-            assert_equal({ id: 42, title: 'New Post', body: 'Body' }, @adapter.serializable_hash[:comment][:post])
+            resource = SerializableResource.new(@comment, adapter: :json, serializer: CommentSerializer)
+            expected = { id: 42, title: 'New Post', body: 'Body' }
+            actual = resource.serializable_hash[:comment][:post]
+
+            assert_equal expected, actual
           end
 
           def test_include_nil_author
-            serializer = PostSerializer.new(@anonymous_post)
-            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
+            resource = SerializableResource.new(@anonymous_post, adapter: :json, serializer: PostSerializer)
+            expected = {
+              post: {
+                title: 'Hello!!', body: 'Hello, world!!', id: 43,
+                comments: [],
+                blog: { id: 999, name: 'Custom blog' }, author: nil
+              }
+            }
+            actual = resource.serializable_hash
 
-            assert_equal({ post: { title: 'Hello!!', body: 'Hello, world!!', id: 43, comments: [], blog: { id: 999, name: 'Custom blog' }, author: nil } }, adapter.serializable_hash)
+            assert_equal expected, actual
           end
 
           def test_include_nil_author_with_specified_serializer
-            serializer = PostPreviewSerializer.new(@anonymous_post)
-            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
+            resource = SerializableResource.new(@anonymous_post, adapter: :json, serializer: PostPreviewSerializer)
+            expected = { post: { title: 'Hello!!', body: 'Hello, world!!', id: 43, comments: [], author: nil } }
+            actual = resource.serializable_hash
 
-            assert_equal({ post: { title: 'Hello!!', body: 'Hello, world!!', id: 43, comments: [], author: nil } }, adapter.serializable_hash)
+            assert_equal expected, actual
           end
         end
       end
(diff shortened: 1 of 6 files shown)


NullVoxPopuli:
My first task will be decoupling the AR models from the need for a database.

NullVoxPopuli:
Notes:

and fix minor style stuff (such as removing `{}` for hashes inside method calls, like `Post.new(name: …, id: …, …)` instead of `Post.new({ name: …, … })`) - @beauby

test/serializable_resource_test.rb
```diff
@@ -2,8 +2,13 @@
 
 module ActiveModel
   class SerializableResourceTest < Minitest::Test
+    class Profile < ::Model
```
beauby: Note that `::Model` is needed (instead of `Model`) because it would resolve to `ActiveModel::Model`.

NullVoxPopuli: that is a good thing to note! :+1:

bf4: or if you wanna be dumb

```ruby
class Minitest::Test
  class Model
  end
end
```

omg, the Model is only available in tests!

bf4: or even

```ruby
class Model; end
class Minitest::Test
  Model = ::Model
end
```

behaves same as above

beauby: I think a more sane approach would be to rename our `Model` class into `PoroModel`. Thoughts?

NullVoxPopuli: Maybe we shouldn't use the name model, as it's not actually part of AMS, and is part of sister gems?

test/serializers/root_test.rb
```diff
@@ -3,18 +3,23 @@
...
       def test_overwrite_root
-        serializer = VirtualValueSerializer.new(@virtual_value, { root: 'smth' })
+        serializer = VirtualValueSerializer.new(@virtual_value, root: 'smth')
         assert_equal('smth', serializer.json_key)
       end
 
       def test_underscore_in_root
         serializer = VirtualValueSerializer.new(@virtual_value)
-        assert_equal('virtual_value', serializer.json_key)
+        assert_equal('active_model/serializer/root_test/virtual_value', serializer.json_key)
```
bf4: ? why does this change?

test/serializable_resource_test.rb
NullVoxPopuli: Lol, :+1: @beauby

test/serializers/root_test.rb
beauby: Namespaces my friend, namespaces...

beauby: More precisely, `VirtualValue` was defined in the global namespace before, so its `json_key` was not namespaced.

test/serializable_resource_test.rb
bf4: sister gems? there's a gem that uses `Model`? how aweful.  well, how about `SerializableModel` and `SerializableRecord` as the base classes

bf4: > I think a more sane approach would be to rename our Model class into PoroMode

it's all good, just let's not bikeshed.  I'd like to consider distinguishing between 'models' and 'records'

test/serializers/root_test.rb
bf4: right, just without source diving I wasn't sure.  for clarity can be something like `name_space_path = self.class.to_s.underscore; assert "#{name_space_path}/virtual_value"`?

beauby: :+1:

test/adapter/json/belongs_to_test.rb
```diff
@@ -18,27 +18,37 @@ def setup
             @post.blog = @blog
             @anonymous_post.blog = nil
 
-            @serializer = CommentSerializer.new(@comment)
-            @adapter = ActiveModel::Serializer::Adapter::Json.new(@serializer)
             ActionController::Base.cache_store.clear
```
beauby: Not necessary anymore.

test/adapter/json/belongs_to_test.rb
```diff
@@ -18,27 +18,37 @@ def setup
...
           end
 
           def test_includes_post
-            assert_equal({ id: 42, title: 'New Post', body: 'Body' }, @adapter.serializable_hash[:comment][:post])
+            resource = SerializableResource.new(@comment, adapter: :json, serializer: CommentSerializer)
```
joaomdmoura: :+1:

test/serializers/association_macros_test.rb
```diff
@@ -3,8 +3,9 @@
 module ActiveModel
   class Serializer
     class AssociationMacrosTest < Minitest::Test
-      AuthorSummarySerializer = Class.new
-      class AssociationsTestSerializer < Serializer
+      class AuthorSummarySerializer
```
joaomdmoura: Any specific reason fro that?

bf4:
my wishlist: controller tests go in the test/controller namespace (and we can probably have less of them than we do now ;)

joaomdmoura:
Nice, despite of my comment it LGTM, so I'm merging it and we can figure it out in a latter PR together with other improvements. :smile:

beauby:
@joaomdmoura Let's hold on this one for a bit, as the goal is to rewrite all tests in a cleaner way.

test/serializers/association_macros_test.rb
beauby: Nope :D Will revert.

remear:
Should we close this and revisit later?

NullVoxPopuli:
yeah, I expect the rebase on this to be pretty serious

PR #1231, opened 2015-10-05
Remove `root_name` class method from Serializer, as it is used nowhere.

beauby:


lib/active_model/serializer.rb
```diff
@@ -98,10 +98,6 @@ def self.adapter
       ActiveModel::Serializer::Adapter.lookup(config.adapter)
     end
 
-    def self.root_name
-      name.demodulize.underscore.sub(/_serializer$/, '') if name
-    end
-
```
bf4: plz add a changelog entry that this is a breaking change (would be good to see where it was last used?), and I'll merge it

good catch

beauby: Done

beauby:
Introduced [here](https://github.com/rails-api/active_model_serializers/pull/680/files#diff-5624f9cf35be4dc254412d979fd293c4R98).

NullVoxPopuli:
I'm good with this! nice find!

NullVoxPopuli:
@bf4 should the changelog specifically state that this change is breaking?

beauby:
@NullVoxPopuli As it was part of the public API, we have to inform people it is not anymore.

NullVoxPopuli:
yeah, I re-looked at the changelog and noticed I didn't notice earlier that the changelog entry was under `Breaking Changes`

we're good!

lib/active_model/serializer.rb

PR #1232, opened 2015-10-05
Extract attributes filtering from serializer into adapter.

beauby:
Filtering attributes is an adapter concern. Plus, this PR simplifies the `Fieldset` utility class by constraining it to the JSON API use case.

### DIFF
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -57,7 +57,10 @@ module ActiveModel
 
         def resource_object_for(options)
           cache_check(serializer) do
-            serializer.attributes(options)
+            attributes = serializer.attributes
+            attributes.slice!(*options[:fields]) if options[:fields]
+
+            attributes
           end
         end
       end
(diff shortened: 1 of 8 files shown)


lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -57,7 +57,10 @@ def include_meta(json)
 
         def resource_object_for(options)
           cache_check(serializer) do
-            serializer.attributes(options)
+            attributes = serializer.attributes
+            attributes.slice!(*options[:fields]) if options[:fields]
```
NullVoxPopuli: why did this need to be added?

beauby: This is actually the main change of this PR: instead of being handled at serializer level, this is now handled at adapter level which allows for different format per adapter.

NullVoxPopuli: neato

NullVoxPopuli:
lgtm :+1:

PR #1233, opened 2015-10-05
Move `meta`/`meta_key` handling inside adapter.

beauby:
The (toplevel) `meta` handling is clearly an adapter concern.

### DIFF
--- a/lib/active_model/serializer/adapter/base.rb
+++ b/lib/active_model/serializer/adapter/base.rb
@@ -37,11 +37,11 @@ module ActiveModel
         private
 
         def meta
-          serializer.meta if serializer.respond_to?(:meta)
+          instance_options.fetch(:meta, nil)
         end
 
         def meta_key
-          serializer.meta_key || 'meta'.freeze
+          instance_options.fetch(:meta_key, 'meta'.freeze)
         end
 
         def root
--- a/test/array_serializer_test.rb
+++ b/test/array_serializer_test.rb
@@ -41,14 +41,6 @@ module ActiveModel
         refute serializers.first.custom_options.key?(:serializer)
       end
 
-      def test_meta_and_meta_key_attr_readers
-        meta_content = { meta: 'the meta', meta_key: 'the meta key' }
-        @serializer = ArraySerializer.new([@comment, @post], meta_content)
-
-        assert_equal @serializer.meta, 'the meta'
-        assert_equal @serializer.meta_key, 'the meta key'
-      end
-
       def test_root_default
         @serializer = ArraySerializer.new([@comment, @post])
         assert_equal @serializer.root, nil
--- a/test/serializers/meta_test.rb
+++ b/test/serializers/meta_test.rb
@@ -4,7 +4,6 @@ module ActiveModel
   class Serializer
     class MetaTest < Minitest::Test
       def setup
-        ActionController::Base.cache_store.clear
         @blog = Blog.new(id: 1,
                          name: 'AMS Hints',
                          writer: Author.new(id: 2, name: 'Steve'),
@@ -12,8 +11,11 @@ module ActiveModel
       end
 
       def test_meta_is_present_with_root
-        serializer = AlternateBlogSerializer.new(@blog, meta: { total: 10 })
-        adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
+        actual = ActiveModel::SerializableResource.new(
+          @blog,
+          adapter: :json,
+          serializer: AlternateBlogSerializer,
+          meta: { total: 10 }).as_json
         expected = {
           blog: {
             id: 1,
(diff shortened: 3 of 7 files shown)


test/array_serializer_test.rb
```diff
@@ -41,14 +41,6 @@ def test_serializer_option_not_passed_to_each_serializer
...
       end
 
-      def test_meta_and_meta_key_attr_readers
-        meta_content = { meta: 'the meta', meta_key: 'the meta key' }
-        @serializer = ArraySerializer.new([@comment, @post], meta_content)
-
-        assert_equal @serializer.meta, 'the meta'
-        assert_equal @serializer.meta_key, 'the meta key'
-      end
-
```
beauby: Those readers simply do not exist anymore.

test/serializers/meta_test.rb
```diff
@@ -62,13 +71,13 @@ def test_meta_key_is_used_with_json_api
...
+        assert_equal expected, actual
       end
 
       def test_meta_is_not_present_on_arrays_without_root
-        serializer = ArraySerializer.new([@blog], meta: { total: 10 })
-        # Attributes doesn't have support to root
-        adapter = ActiveModel::Serializer::Adapter::Attributes.new(serializer)
+        actual = ActiveModel::SerializableResource.new([@blog], adapter: :attributes,
+                                                                meta: { total: 10 })
+                 .as_json
```
NullVoxPopuli: formatting?




`.as_json`



feels so lonely

beauby: How would you rather have it?

NullVoxPopuli:
```ruby
ActiveModel::SerializableResource.new(
  [@blog],
  adapter: :attributes,
  meta: { total: 10 }
).as_json
```

test/array_serializer_test.rb
NullVoxPopuli: :+1:

test/serializers/meta_test.rb
beauby: I could live with that. Will update.

beauby: Done.

NullVoxPopuli:
can you rebase?

NullVoxPopuli:
thanks!

beauby:
Wait, I'll rebase.

beauby:
Rebased.

NullVoxPopuli:
woo!

lib/active_model/serializer/adapter/base.rb
```diff
@@ -37,11 +37,11 @@ def cache_check(serializer)
         private
 
         def meta
-          serializer.meta if serializer.respond_to?(:meta)
+          instance_options.fetch(:meta, nil)
```
joaomdmoura: :+1:

test/serializers/meta_test.rb
```diff
@@ -12,8 +11,11 @@ def setup
       end
 
       def test_meta_is_present_with_root
-        serializer = AlternateBlogSerializer.new(@blog, meta: { total: 10 })
-        adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
+        actual = ActiveModel::SerializableResource.new(
```
joaomdmoura: :+1:

trek:
Did this break existing examples for supplying meta from with a CollectionSerializer? It doesn't seem like that's possible now. Should those be updated to show to override the adapter for this behavior?

beauby:
@trek Not sure I understand what you meant. Could you sketch some code to show what you have in mind?

trek:
The pagination example [here](https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/add_pagination_links.md) doesn't work after moving `meta` to the Adapter (where it does obviously better fit).

It looks like we didn't also add the documentation for centrally overriding serialization to include meta. The JSON API Adapter, for example, [will include links automatically](https://github.com/rails-api/active_model_serializers/blob/65b8493e5150c76e3c3cbf018729ed202f72930d/lib/active_model/serializer/adapter/json_api.rb#L50-L68) but the only way we could figure out how to get extra keys into the top level object (in this case for pagination) was creating our own adapter and overriding `serializable_hash_for_collection`

```
class JsonApiWithMetaAdapter < ActiveModel::Serializer::Adapter::JsonApi
    hash = super
    object = serializer.object

    if serializer.paginated?
      hash[:meta] = hash[:meta] || {}
      hash[:meta][:pagination] = {
        current_page: object.current_page,
        total_pages: object.total_pages,
        total_entries: object.total_entries
      }
    end

    hash
  end
end
```

There might be a better way, but we couldn't find an interface for it on master. It'd be good to document getting keys into the top level hash.

beauby:
@trek You are right. Note that, in master, you can always provide the `meta` option to your render call, which will be forwarded to the adapter, and included in the response document (`render json: posts, meta: my_pagination_hash`). It is probably worth thinking about the possibility of including a resource/serializer-level `meta` directive for the Json adapter, but the example in the docs was clearly a hack.

PR #1234, opened 2015-10-05
Move handling of `root` option from serializer to adapter.

beauby:
This option is clearly an adapter concern and it is only used in the Json adapter.
+ Custom `root` are not subject to inflection anymore.

Most tests are failing because they are directly testing the root value on the serializers. Will fix.


test/adapter/json_api/belongs_to_test.rb
```diff
@@ -38,7 +38,11 @@ def test_includes_post_id
           end
 
           def test_includes_linked_post
-            @adapter = ActiveModel::Serializer::Adapter::JsonApi.new(@serializer, include: [:post])
+            actual = ActiveModel::SerializableResource.new(
```
bf4: use the `serializable` helper to for less words, if you want

lib/active_model/serializer/array_serializer.rb
```diff
@@ -25,11 +24,6 @@ def initialize(resources, options = {})
         @meta_key = options[:meta_key]
       end
 
-      def json_key
-        key = root || serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
-      end
```
bf4: is this not used anywhere?

beauby: Not anymore: this PR makes it an adapter concern.

bf4:
@beauby Would you clarify what `root` and `json_key` mean in context? I thought it was the root for each resource, which would mean it's per serialized resource, no?  (But now that we have `Serializer::type` maybe that replaces this?)

test/adapter/json_api/belongs_to_test.rb
beauby: Is the helper globally available? I think it should be. If so, I'll update these tests.

beauby:
@bf4: No, they were only referring to the root of the document, that's why I believe it's an adapter concern. For the "roots" of resources, we actually use the association's key.

beauby:
Updated.

test/action_controller/serialization_test.rb
```diff
@@ -171,7 +171,7 @@ def test_render_array_using_custom_root
         with_adapter :json do
           get :render_array_using_custom_root
         end
-        expected = { custom_roots: [{ name: 'Name 1', description: 'Description 1' }] }
+        expected = { custom_root: [{ name: 'Name 1', description: 'Description 1' }] }
```
bf4: this raises a huge red flag for me.  Why is this changing?

beauby: I believe it should not have been there in the first place. Modifying explicitly supplied options is such an antipattern. I could move this change to an other PR though.

lib/active_model/serializer/adapter/json.rb
```diff
@@ -15,6 +15,24 @@ def serializable_hash(options = nil)
...
+                  serializer.object.try(:name)
+                end
+              key.to_s.pluralize
+            else
+              serializer.object.class.model_name.to_s
+            end
+          end
+
+          root.to_s.underscore.to_sym
+        end
```
bf4: vs. 

```diff
   def json_key
-      root || object.class.model_name.to_s.underscore
-    end
```

and

```diff
-      def json_key
-        key = root || serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
-      end
```

?

Shouldn't this just be using the Serializer `type` or am I tired?
I'm not convinced

beauby: Yes. Highly possible that the serializer's `type` is the right thing to use here. Main point of this PR was that an option that is specific to the `Json` adapter should not clutter the `Serializer` and `Adapter::Base` classes with lots of overcomplicated code.

beauby: Note that `type` is nil unless explicitly specified though.

NullVoxPopuli:
is this still applicable?

bf4:
Done, iirc

B mobile phone

remear:
Can this be closed then?

beauby:
Seems not: https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializable_resource.rb#L5
As you can see, `:root` is not an adapter option yet.

bf4:
@beauby I would love if you'd come back to this one. Shouldn't be to hard to finish.

beauby:
Will take care of it over the weekend!

beauby:
Closing in favor of https://github.com/rails-api/active_model_serializers/pull/1782.

PR #1239, opened 2015-10-05
Fix duplicate resources inside included in compound document.

### LINKED ISSUE #1202: Duplicate resource object in compound document
There's a bug here (unless I'm reading something wrong) and I can't find an open issue that describes this. Please forgive me if this is a duplicate.

**Reproduction set-up**

Making a [compound document](http://jsonapi.org/format/#document-compound-documents) using `include`. It contains an `Organization`, which `has_many :people`, and `Person` `belongs_to :organization` (there are many other cases that would reproduce this, I think). I have an organization with ID 1, and some people belonging to that organization. The document would have an array of people in the `included` key.

**The bug**

The JSON API spec says:

> A compound document MUST NOT include more than one resource object for each type and id pair.

The `compound document` consists of a document containing both `data` and `included` top-level keys. The `data` is just the `Organization` with ID 1, which has `relationships` which has `people` which has `data`, which is an array of `Person` objects with just `type` (==`people`) and `id`.

In the `included` top-level key, you have the `Person` entities related to the organization. But also, the `Organization` appears here. So you have two resource objects in the same compound document with the same type (`organization`) and ID (`1`). That's a bug, right?

beauby:
This fixes #1202.

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -81,15 +81,17 @@ module ActiveModel
 
         def serializable_hash_for_collection(options)
           hash = { data: [] }
+          included = []
           serializer.each do |s|
             result = self.class.new(s, instance_options.merge(fieldset: fieldset)).serializable_hash(options)
             hash[:data] << result[:data]
+            next unless result[:included]
 
-            if result[:included]
-              hash[:included] ||= []
-              hash[:included] |= result[:included]
-            end
+            included |= result[:included]
           end
+
+          included.delete_if { |resource| hash[:data].include?(resource) }
+          hash[:included] = included if included.any?
 
           if serializer.paginated?
             hash[:links] ||= {}
@@ -102,9 +104,10 @@ module ActiveModel
         def serializable_hash_for_single_resource
           primary_data = primary_data_for(serializer)
           relationships = relationships_for(serializer)
-          included = included_resources(@include_tree)
+          primary_data[:relationships] = relationships if relationships.any?
           hash = { data: primary_data }
-          hash[:data][:relationships] = relationships if relationships.any?
+
+          included = included_resources(@include_tree, [primary_data])
           hash[:included] = included if included.any?
 
           hash
@@ -171,31 +174,31 @@ module ActiveModel
           end
         end
 
-        def included_resources(include_tree)
+        def included_resources(include_tree, primary_data)
           included = []
 
           serializer.associations(include_tree).each do |association|
-            add_included_resources_for(association.serializer, include_tree[association.key], included)
+            add_included_resources_for(association.serializer, include_tree[association.key], primary_data, included)
           end
 
           included
         end
 
-        def add_included_resources_for(serializer, include_tree, included)
+        def add_included_resources_for(serializer, include_tree, primary_data, included)
           if serializer.respond_to?(:each)
-            serializer.each { |s| add_included_resources_for(s, include_tree, included) }
+            serializer.each { |s| add_included_resources_for(s, include_tree, primary_data, included) }
           else
             return unless serializer && serializer.object
 
-            primary_data = primary_data_for(serializer)
+            resource_object = primary_data_for(serializer)
             relationships = relationships_for(serializer)
-            primary_data[:relationships] = relationships if relationships.any?
+            resource_object[:relationships] = relationships if relationships.any?
 
-            return if included.include?(primary_data)
-            included.push(primary_data)
+            return if included.include?(resource_object) || primary_data.include?(resource_object)
+            included.push(resource_object)
 
             serializer.associations(include_tree).each do |association|
-              add_included_resources_for(association.serializer, include_tree[association.key], included)
+              add_included_resources_for(association.serializer, include_tree[association.key], primary_data, included)
             end
           end
         end
--- a/test/adapter/json_api/linked_test.rb
+++ b/test/adapter/json_api/linked_test.rb
@@ -289,11 +289,12 @@ module ActiveModel
 
           class PostSerializer < ActiveModel::Serializer
             type 'posts'
-            class AuthorSerializer < ActiveModel::Serializer
-              type 'authors'
-              has_many :posts, serializer: PostSerializer
-            end
-            belongs_to :author, serializer: AuthorSerializer
+            belongs_to :author
+          end
+
+          class AuthorSerializer < ActiveModel::Serializer
+            type 'authors'
+            has_many :posts
           end
 
           def setup
@@ -313,7 +314,6 @@ module ActiveModel
 
           def test_no_duplicates
             hash = ActiveModel::SerializableResource.new(@post1, adapter: :json_api,
-                                                                 serializer: PostSerializer,
                                                                  include: '*.*')
                    .serializable_hash
             expected = [
@@ -343,7 +343,6 @@ module ActiveModel
           def test_no_duplicates_collection
             hash = ActiveModel::SerializableResource.new(
               [@post1, @post2], adapter: :json_api,
-                                each_serializer: PostSerializer,
                                 include: '*.*')
                    .serializable_hash
             expected = [
--- a/test/adapter/json_api/linked_test.rb
+++ b/test/adapter/json_api/linked_test.rb
@@ -1,11 +1,16 @@
 require 'test_helper'
+
+NestedPost = Class.new(Model)
+class NestedPostSerializer < ActiveModel::Serializer
+  has_many :nested_posts
+end
+
 module ActiveModel
   class Serializer
     module Adapter
       class JsonApi
         class LinkedTest < Minitest::Test
           def setup
-            ActionController::Base.cache_store.clear
             @author1 = Author.new(id: 1, name: 'Steve K.')
             @author2 = Author.new(id: 2, name: 'Tenderlove')
             @bio1 = Bio.new(id: 1, content: 'AMS Contributor')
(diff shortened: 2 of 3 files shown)


beauby:
@hut8 Could you test this patch?

hut8:
@beauby The test LGTM. It does appear to stop producing the duplicate I saw in the original issue. Thanks!

NullVoxPopuli:
Thanks for verifying @hut8.

@beauby this looks good to me.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -84,13 +84,15 @@ def serializable_hash_for_collection(options)
...
             result = self.class.new(s, instance_options.merge(fieldset: fieldset)).serializable_hash(options)
             hash[:data] << result[:data]
+            next unless result[:included]
 
-            if result[:included]
-              hash[:included] ||= []
-              hash[:included] |= result[:included]
-            end
+            hash[:included] ||= []
+            hash[:included] |= result[:included]
```
NullVoxPopuli: why bitwise `or` here?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -84,13 +84,15 @@ def serializable_hash_for_collection(options)
...
           end
 
+          hash[:included].delete_if { |resource| Array(hash[:data]).include?(resource) } if hash[:included]
+          hash.delete(:included) if hash[:included] && hash[:included].empty?
```
NullVoxPopuli: can these two lines be extracted to a method?

test/adapter/json_api/linked_test.rb
```diff
@@ -277,6 +277,79 @@ def test_nil_link_with_specified_serializer
...
           end
         end
+
+        class NoDuplicatesTest < Minitest::Test
+          Post = Class.new(::Model)
+          Author = Class.new(::Model)
+
+          class PostSerializer < ActiveModel::Serializer
+            type 'posts'
+            class AuthorSerializer < ActiveModel::Serializer
```
NullVoxPopuli: nested serializers :-)

lib/active_model/serializer/adapter/json_api.rb
hut8:
```
[7] pry(main)> lol ||= []
[]
[8] pry(main)> lol |= [1,2,3]
[
    [0] 1,
    [1] 2,
    [2] 3
]
[9] pry(main)> lol
[
    [0] 1,
    [1] 2,
    [2] 3
]
```

It's fancy syntax for a union

NullVoxPopuli: woah. that is fancy. TIL

beauby:
Rebased.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -81,16 +81,18 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+            next unless result[:included]
 
-            if result[:included]
-              hash[:included] ||= []
-              hash[:included] |= result[:included]
-            end
+            included |= result[:included]
           end
 
+          included.delete_if { |resource| hash[:data].include?(resource) }
```
joaomdmoura: I see we are deleting here the duplicated resource, but it seems hacky, it's not addressing the problem itself, just working around it, did you found the main issue causing this?

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -171,31 +174,31 @@ def relationships_for(serializer)
...
 
-            primary_data = primary_data_for(serializer)
+            resource_object = primary_data_for(serializer)
             relationships = relationships_for(serializer)
-            primary_data[:relationships] = relationships if relationships.any?
+            resource_object[:relationships] = relationships if relationships.any?
 
-            return if included.include?(primary_data)
-            included.push(primary_data)
+            return if included.include?(resource_object) || primary_data.include?(resource_object)
```

test/adapter/json_api/linked_test.rb
joaomdmoura: Would be nice to have another scenario using non-nested serializers :smile:

lib/active_model/serializer/adapter/json_api.rb
beauby: It is a structural issue. When the primary resource is a single resource (`serializable_hash_for_single_resource`), we make sure we do not add to `included` any related resource that is the primary resource. However, when the primary resource is a collection (`serializable_hash_for_collection`), the adapter works by computing the hash for every resource independently, and then combining the hashes into a big one. This has the advantage of clear code, but the disadvantage that when serializing `[a, b]` where `a` includes `b`, we have no other choice than manually removing duplicates when aggregating.

joaomdmoura:
@beauby just checked it, made few comments on it for you to check

test/adapter/json_api/linked_test.rb
beauby: We need #1225 for that. The reason is that if we do not use nested serializers here, we have a namespacing issue. With #1225, we can declare the serializers in a non-nested way and it will still use the right one.

beauby:
@joaomdmoura Answered ;)

lib/active_model/serializer/adapter/json_api.rb
joaomdmoura: got it :wink:

joaomdmoura:
LGTM :+1:

test/adapter/json_api/linked_test.rb
```diff
@@ -1,11 +1,16 @@
 require 'test_helper'
+
+NestedPost = Class.new(Model)
+class NestedPostSerializer < ActiveModel::Serializer
```
NullVoxPopuli: I'd say because this model *and* serializer are created here, and specific to this test, they should be namespaced under the test. :-\

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -102,9 +104,10 @@ def serializable_hash_for_collection(options)
         def serializable_hash_for_single_resource
           primary_data = primary_data_for(serializer)
           relationships = relationships_for(serializer)
-          included = included_resources(@include_tree)
+          primary_data[:relationships] = relationships if relationships.any?
```
bf4: simple refactor, nice

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -102,9 +104,10 @@ def serializable_hash_for_collection(options)
...
           hash = { data: primary_data }
-          hash[:data][:relationships] = relationships if relationships.any?
+
+          included = included_resources(@include_tree, [primary_data])
```
bf4: so, `included_resources` now takes in `[primary_data]` as a second argument

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -171,31 +174,31 @@ def relationships_for(serializer)
           end
         end
 
-        def included_resources(include_tree)
+        def included_resources(include_tree, primary_data)
           included = []
 
           serializer.associations(include_tree).each do |association|
-            add_included_resources_for(association.serializer, include_tree[association.key], included)
+            add_included_resources_for(association.serializer, include_tree[association.key], primary_data, included)
```
bf4: and `add_included_resources_for` passes `primary_data` through

lib/active_model/serializer/adapter/json_api.rb
beauby: Yes, this is to ensure, preemptively, that we do not include any of the primary data in the `included` hash when building the hash for a single resource.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -171,31 +174,31 @@ def relationships_for(serializer)
...
           end
 
           included
         end
 
-        def add_included_resources_for(serializer, include_tree, included)
+        def add_included_resources_for(serializer, include_tree, primary_data, included)
           if serializer.respond_to?(:each)
-            serializer.each { |s| add_included_resources_for(s, include_tree, included) }
+            serializer.each { |s| add_included_resources_for(s, include_tree, primary_data, included) }
```
bf4: and primary_data (which is `[primary_data]`) is a new argument to `add_included_resources_for`

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -171,31 +174,31 @@ def relationships_for(serializer)
...
           else
             return unless serializer && serializer.object
 
-            primary_data = primary_data_for(serializer)
+            resource_object = primary_data_for(serializer)
             relationships = relationships_for(serializer)
-            primary_data[:relationships] = relationships if relationships.any?
+            resource_object[:relationships] = relationships if relationships.any?
```
bf4: nice refactor

lib/active_model/serializer/adapter/json_api.rb
bf4: And we add `primary_data.include?(resource_object)` ( which is true when `[primary_data].include(resource_object)` which is approx `[primary_data_for(serializer).update(relationships: relationships_for(serializer))].include(primary_data_for(serializer)update(relationships: relationships_for(serializer))`) i.e. primary data is for same serializer) as condition to return early

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -171,31 +174,31 @@ def relationships_for(serializer)
...
 
-            return if included.include?(primary_data)
-            included.push(primary_data)
+            return if included.include?(resource_object) || primary_data.include?(resource_object)
+            included.push(resource_object)
 
             serializer.associations(include_tree).each do |association|
-              add_included_resources_for(association.serializer, include_tree[association.key], included)
+              add_included_resources_for(association.serializer, include_tree[association.key], primary_data, included)
```
bf4: and pass along our `[primary_data]` again to `add_included_resources_for`

test/adapter/json_api/linked_test.rb
bf4: I think we agree we'll start doing that soon

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -81,16 +81,18 @@ def fragment_cache(cached_hash, non_cached_hash)
...
           serializer.each do |s|
             result = self.class.new(s, instance_options.merge(fieldset: fieldset)).serializable_hash(options)
             hash[:data] << result[:data]
+            next unless result[:included]
 
-            if result[:included]
-              hash[:included] ||= []
-              hash[:included] |= result[:included]
-            end
+            included |= result[:included]
```
bf4: refactor

lib/active_model/serializer/adapter/json_api.rb
bf4: deletes resource from `included` if is a root resource

this included stuff is getting messy. needs a refactor before it sends even more vines through the class

beauby: I believe it is the best we can achieve without refactoring the adapter so that it does not call itself recursively for collections (which I'm in favor of, in a subsequent PR). The scope of this PR was to fix a behavior breaking JSON API spec, without modifying the adapter too much.

bf4:
I'm ok to merge, but we need to admit that this inclusion stuff has been a source of bugs and increasing complexity.  IMHO, it deserves its own object

beauby:
@bf4 We agree. I'm happy to work on a refactor once this is merged.

test/adapter/json_api/linked_test.rb
beauby: Yes. Let's leave them there for now, and take care of it during the Great Test Refactor.

PR #1241, opened 2015-10-05
Docs explaining usage of AMS without AR

NullVoxPopuli:


docs/howto/without_active_record.md
```diff
@@ -0,0 +1,116 @@
+# How to use AMS without ActiveRecord
+
+## Defining Models With ActiveModel
+
+When using active model, there are only a few methods you'll need to implement.
+* attributes
```
beauby: Could you elaborate on this?

docs/howto/without_active_record.md
```diff
@@ -0,0 +1,116 @@
...
+      hash[ivar] = instance_variable_get(ivar)
+    end
+  end
+
+  def serializable_hash
+    attributes
+  end
+
+  def read_attribute_for_serialization(name)
+    attributes[name]
```
beauby: Why not simply `send(name)`?

docs/howto/without_active_record.md
```diff
@@ -0,0 +1,116 @@
...
+```ruby
+require 'active_model'
+
+class BaseModel
+  include ActiveModel::Model
+  include ActiveModel::Serialization
+
+  attr_accessor :id, :updated_at
+
+  def initialize(hash = {})
```
beauby: This is probably outside the scope of the AMS docs.

NullVoxPopuli: initialize is outside the scope?

beauby: What I meant is that putting this in the docs is misleading because it makes people believe AMS needs this constructor to work.

NullVoxPopuli: Understood. :-)
I'll rework this when I get back from dinner

docs/howto/without_active_record.md
```diff
@@ -0,0 +1,100 @@
+# How to use AMS without ActiveRecord
+
+Every model that is to be serialized must implement the following methods
+
+ method name | description
+ ----------- | -----------
+ `attributes` | a hash of all attributes/fields on the model and their corresponding values
```
beauby: This is not required.

bf4: This should reference the `Lint` docs

docs/howto/without_active_record.md
```diff
@@ -0,0 +1,100 @@
+# How to use AMS without ActiveRecord
```
bf4: AMS, lulz

bf4: we could get that name... https://rubygems.org/gems/ams

docs/howto/without_active_record.md
```diff
@@ -0,0 +1,91 @@
+# How to use ActiveModelSerializers without ActiveRecord
+
+Every model that is to be serialized must implement the following methods
```
bf4: this entire page can be boiled down to

```ruby
class MyModel
  include ActiveModel::Model
  include ActiveModel::Serialization
end
```

or otherwise passes Lint

```ruby
class MyModelSerializer < ActiveModel::Serializer
  attribute :potato
  def potato; 'potAHto'; end
end
```

and that's it.

Now, outside of the controller the rules are **exactly** the same as for records.

```ruby
ActiveModel::SerializableResource.new(MyModel.new(potato: "potAHto'), adapter: :json)
```

I'd rather just add to the front page readme that AMS requires the thing-to-be-rendered (resource) to pass the Lint, see documentation there. ActiveRecord::Base records pass the lint, as do objects including ActiveModel::Serialization

NullVoxPopuli: I don't think it hurts to have more general-use examples.
But, you're also saying that `ActiveModel::Model` provides the hash attribute instantiate? I did not know that.

@bf4 you should have write access to this repo/branch - I haven't used the lint before, but if you want to add documentation here (or maybe it would be better elsewhere)? I think that could be beneficial.

NullVoxPopuli:
@beauby @bf4 did we still need this? or was there another PR accomplishing the same/similar thing? (I don't remember)

NullVoxPopuli:
I think https://github.com/rails-api/active_model_serializers/blob/master/docs/ARCHITECTURE.md#what-does-a-serializable-resource-look-like covers the purpose of this PR

bf4:
Also https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/outside_controller_use.md

Thanks for the issue gardening!

PR #1246, opened 2015-10-06
Add support for resource-level JSON API links.

beauby:
The syntax is as follows:
```ruby
  # Block syntax for `meta` and/or dynamic value
  link :self do
    href "//example.com/link_author/#{object.id}"
    meta stuff: 'value'
  end

  # Direct syntax for plain strings without `meta`
  link :other, '//example.com/resource'
end
```

Update: new syntax.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -50,6 +50,9 @@ module ActiveModel
       self._attributes ||= []
       class_attribute :_attributes_keys          # @api private : maps attribute value to explict key name, @see Serializer#attribute
       self._attributes_keys ||= {}
+      class_attribute :_links                    # @api private : links definitions, @see Serializer#link
+      self._links ||= {}
+
       serializer.class_attribute :_cache         # @api private : the cache object
       serializer.class_attribute :_fragmented    # @api private : @see ::fragmented
       serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key
@@ -72,6 +75,7 @@ module ActiveModel
       caller_line = caller.first
       base._attributes = _attributes.dup
       base._attributes_keys = _attributes_keys.dup
+      base._links = _links.dup
       base._cache_digest = digest_caller_file(caller_line)
       super
     end
@@ -81,6 +85,10 @@ module ActiveModel
     #     type 'authors'
     def self.type(type)
       self._type = type
+    end
+
+    def self.link(name, value = nil, &block)
+      _links[name] = block || value
     end
 
     # @example
@@ -247,6 +255,12 @@ module ActiveModel
           hash[name] = send(name)
         end
       end
+    end
+
+    # @api private
+    # Used by JsonApi adapter to build resource links.
+    def links
+      self.class._links
     end
 
     protected
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -102,9 +102,8 @@ module ActiveModel
         end
 
         def serializable_hash_for_single_resource
-          primary_data = primary_data_for(serializer)
-          relationships = relationships_for(serializer)
-          primary_data[:relationships] = relationships if relationships.any?
+          primary_data = resource_object_for(serializer)
+
           hash = { data: primary_data }
 
           included = included_resources(@include_tree, [primary_data])
@@ -142,26 +141,22 @@ module ActiveModel
         end
 
         def resource_object_for(serializer)
-          cache_check(serializer) do
+          resource_object = cache_check(serializer) do
             resource_object = resource_identifier_for(serializer)
 
             requested_fields = fieldset && fieldset.fields_for(resource_object[:type])
             attributes = attributes_for(serializer, requested_fields)
             resource_object[:attributes] = attributes if attributes.any?
-
-            links = links_for(serializer)
-            resource_object[:links] = links if links.any?
-
             resource_object
           end
-        end
 
-        def primary_data_for(serializer)
-          if serializer.respond_to?(:each)
-            serializer.map { |s| resource_object_for(s) }
-          else
-            resource_object_for(serializer)
-          end
+          relationships = relationships_for(serializer)
+          resource_object[:relationships] = relationships if relationships.any?
+
+          links = links_for(serializer)
+          resource_object[:links] = links if links.any?
+
+          resource_object
         end
 
         def relationship_value_for(serializer, options = {})
@@ -201,9 +196,7 @@ module ActiveModel
           else
             return unless serializer && serializer.object
 
-            resource_object = primary_data_for(serializer)
-            relationships = relationships_for(serializer)
-            resource_object[:relationships] = relationships if relationships.any?
+            resource_object = resource_object_for(serializer)
 
             return if included.include?(resource_object) || primary_data.include?(resource_object)
             included.push(resource_object)
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -5,6 +5,7 @@ module ActiveModel
         extend ActiveSupport::Autoload
         autoload :PaginationLinks
         autoload :FragmentCache
+        autoload :Link
 
         # TODO: if we like this abstraction and other API objects to it,
         # then extract to its own file and require it.
@@ -94,7 +95,7 @@ module ActiveModel
 
           if serializer.paginated?
             hash[:links] ||= {}
-            hash[:links].update(links_for(serializer, options))
+            hash[:links].update(pagination_links_for(serializer, options))
           end
 
           hash
@@ -136,12 +137,21 @@ module ActiveModel
           { id: id.to_s, type: type }
         end
 
+        def attributes_for(serializer, fields)
+          serializer.attributes(fields).except(:id)
+        end
+
         def resource_object_for(serializer)
           cache_check(serializer) do
             resource_object = resource_identifier_for(serializer)
+
             requested_fields = fieldset && fieldset.fields_for(resource_object[:type])
-            attributes = serializer.attributes(requested_fields).except(:id)
+            attributes = attributes_for(serializer, requested_fields)
             resource_object[:attributes] = attributes if attributes.any?
+
+            links = links_for(serializer)
+            resource_object[:links] = links if links.any?
+
             resource_object
           end
         end
@@ -204,7 +214,21 @@ module ActiveModel
           end
         end
 
-        def links_for(serializer, options)
+        def links_for(serializer)
+          serializer.links.each_with_object({}) do |(name, value), hash|
+            hash[name] =
+              if value.respond_to?(:call)
+                link = Link.new(serializer)
+                link.instance_eval(&value)
+
+                link.to_hash
+              else
+                value
+              end
+          end
+        end
+
+        def pagination_links_for(serializer, options)
           JsonApi::PaginationLinks.new(serializer.object, options[:context]).serializable_hash(options)
         end
       end
--- a/lib/active_model/serializer/adapter/json_api/link.rb
+++ b/lib/active_model/serializer/adapter/json_api/link.rb
@@ -1,0 +1,34 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        class Link
+          def initialize(serializer)
+            @object = serializer.object
+            @scope = serializer.scope
+          end
+
+          def href(value)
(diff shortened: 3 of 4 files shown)


lib/active_model/serializer.rb
```diff
@@ -50,6 +50,10 @@ def self.type(type)
       self._type = type
     end
 
+    def self.links(&block)
+      self._links_block = block
+    end
```
bf4: My gut is to avoid long-lived blocks, as there may be a lot of these, and the memory would never be released

test/adapter/json_api/links_test.rb
```diff
@@ -0,0 +1,37 @@
...
+
+LinkPost = Class.new(Model)
+class LinkPostSerializer < ActiveModel::Serializer
+  belongs_to :link_author
+end
+
+LinkAuthor = Class.new(Model)
+class LinkAuthorSerializer < ActiveModel::Serializer
+  links do
+    link :self, href: "//link_author/#{object.id}"
```
bf4: Would you confirm this is a valid link for JSON API? I don't believe it is, but I may be mistaken.  (I know there's some discussion about relative vs. absolute. I'm not sure what this form is called. protcol-less?)

test/adapter/json_api/links_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+
+LinkPost = Class.new(Model)
+class LinkPostSerializer < ActiveModel::Serializer
+  belongs_to :link_author
+end
```
bf4: these objects must be defined inside the test to avoid pollution of main

test/adapter/json_api/links_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  class Serializer
+    module Adapter
+      class JsonApi
+        class LinksTest < Minitest::Test
+          def setup
+            @author = LinkAuthor.new(id: 1337)
+          end
+
+          def test_resource_links
+            hash = ActiveModel::SerializableResource.new(@author, adapter: :json_api).serializable_hash
```
bf4: if you'd like, there's a shorthand `serializable` in https://github.com/rails-api/active_model_serializers/blob/d02cd30fe55a3ea85e1d351b6e039620903c1871/test/support/serialization_testing.rb#L34-L36

lib/active_model/serializer.rb
beauby: Alternative? You actually need this block every time to instantiate a serializer.

bf4: only if we design it this way.  simple solution is just pass in `links` to the adapter until we need more

test/adapter/json_api/links_test.rb
beauby: This is a protocol-relative URL. JSON API does not take any explicit stand on the matter.

test/adapter/json_api/links_test.rb
beauby: C.f. discussion with @joaomdmoura, most users define their serializers in the global namespace, so we should test that.

test/adapter/json_api/links_test.rb
beauby: Same answer as in my other PR, I'm all up for this helper, but I do not want to redefine it here. We should move it to `test_helper.rb`.

lib/active_model/serializer.rb
beauby: Sorry, I don't understand your last comment. Could you elaborate?

test/adapter/json_api/links_test.rb
bf4: It is

B mobile phone

beauby: Awesome then!

lib/active_model/serializer.rb
beauby: Note also that there would be at most 1 block per serializer class.

beauby: An other option would be to declare each link as a lambda, but that would make much more stored procs...

bf4: I just don't think we need this code at all until someone needs it.

beauby: I'm not sure I understand. Are you saying we should not support JSON API resource-level links?

test/adapter/json_api/links_test.rb
bf4: How do you read 

- https://github.com/json-api/json-api/issues/840#issuecomment-141631828
- https://github.com/json-api/json-api/pull/834#issuecomment-131192713
- https://github.com/json-api/json-api/pull/834#issuecomment-141631788

>  relative references are simply not allowed.

beauby: They seem not to be sure, plus the spec itself does not say anything. I personally do not mind changing the tests though, as those are arbitrary strings anyways.

bf4: > those are arbitrary strings anyways.

That's exactly why I don't want the links method.  Let's just let `links` from the controller, then, in another PR, build a `ApiObject::Link` and use that. Then, if someone comes up with a good use case for the serializer, we'll add it then

test/adapter/json_api/links_test.rb
bf4: @beauby not the same thing.  They must be defined inside the test or else they will interact and cause crazy bugs.  We could only get away with it when there were all in one file.  Believe me, I've encountered this.  I'm guessing your concern is just the namespacing.

test/adapter/json_api/links_test.rb
beauby: I disagree. We decided to go for full JSON API support, and JSON API states [http://jsonapi.org/format/#document-resource-objects](that resource objects may contain links to the resource, among other). We need a mechanism to handle that.

bf4: @beauby I disagree as well. I'm not sure what to do about it.  I'm saying 'not now' and you're saying 'yes now'.

bf4: also, I think the adapter should build the link href for each resource using a uri template and the request object, which it already has

beauby: It's in the roadmap we defined for 0.10, so I'm not sure what you mean by 'not now'.

As for automatically building links, this would be nice, but wouldn't cover use-cases where you have to give a custom link (not the `self`, but the JSON API spec allows for other links).

bf4: Ok, well, I'll stand back till this PR implements the full spec for resource level link objects since we're obviously miscommunicating about the content of this PR.

steveklabnik: @beauby a protocol relative URI still contains the domain. This discussion's url: https://github.com/rails-api/active_model_serializers/pull/1246#discussion_r41348842

Its 'protocol' relative version: //github.com/rails-api/active_model_serializers/pull/1246#discussion_r41348842

Not //rails-api/active_model_serializers/pull/1246#discussion_r41348842

beauby: @steveklabnik You're right. To be honest, I didn't really put a lof of thought into the choice of the actual URI in the test, as it is just an arbitrary string so far.

steveklabnik: :+1:  it's all good.

lib/active_model/serializer.rb
```diff
@@ -159,6 +166,12 @@ def json_key
       root || object.class.model_name.to_s.underscore
     end
 
+    def links
```
NullVoxPopuli: can you add some documentation to this method since it's publicly available?

lib/active_model/serializer.rb
```diff
@@ -53,6 +56,10 @@ def self.type(type)
       self._type = type
     end
 
+    def self.link(name, &block)
```
NullVoxPopuli: can you add some documentation to this method since it's publicly available?

test/adapter/json_api/links_test.rb
```diff
@@ -15,16 +23,26 @@ def test_toplevel_links
               adapter: :json_api,
               links: {
                 self: {
-                  href: '//posts'
```
NullVoxPopuli: was this an invalid test before? `//posts` isn't a URL :-\

antondomratchev:
@beauby Any updates on resolving the conflicts on this? I think this is a great missing feature! Good job on this.

beauby:
@antondomratchev Thanks! The current issue is not so much about rebasing this, but about choosing the right way to handle links (since they are being declared pre-instantiation, but might rely on some stuff, such as `object` or `scope` that is only available in an instance).

NullVoxPopuli:
I think this way of doing links is pretty good. And it's consistent with all the block stuff you've been pushing.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -142,13 +143,24 @@ def resource_identifier_for(serializer)
           { id: id.to_s, type: type }
         end
 
+        def attributes_for(serializer, fields)
+          attributes = serializer.attributes.except(:id)
+          attributes.slice!(*fields) if fields
```
NullVoxPopuli: slick

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -208,7 +220,21 @@ def add_included_resources_for(serializer, include_tree, primary_data, included)
           end
         end
 
-        def links_for(serializer, options)
+        def links_for(serializer)
+          serializer.links.each_with_object({}) do |(name, value), hash|
+            hash[name] =
+              if value.respond_to?(:call)
```
NullVoxPopuli: why check respond_to? call?

I think `.is_a?(Proc)` reads better, and makes more sense about the intention of what's going on. Even if there are other objects that you want to support that implement `call`, it would be better to explicitly say what those objects are.

cause, what if I set value to an instance of `Phone` a phone can `call` but it's not the same kind of call.

lib/active_model/serializer/adapter/json_api/link.rb
```diff
@@ -0,0 +1,34 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        class Link
```
NullVoxPopuli: should other pieces of the JSON API spec have their own classes as well? (not that it should be part of this PR, but just wondering what the overall plan is)

beauby: Possibly, I don't have a definitive answer.

lib/active_model/serializer/adapter/json_api.rb
beauby: Basic duck-typing for a callable object (mainly `Proc` or `Lambda`). Anyways the value should either be a callable or a `String`.

NullVoxPopuli: So the if should use is_a?

beauby: Not really. Rails uses this kind of duck-typing as well: https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activemodel/lib/active_model/validations/clusivity.rb#L18.

NullVoxPopuli: then can you put a note in here about what sort of objects are expected, because if I have a Phone for my value, it's gonna error......

beauby: Yeah but common sense kinda plays its role here: a `Phone` is not really a link

NullVoxPopuli: fair enough. 

I guess this comment doesn't really matter here. 

I'll bring it up when I see other duck-typing usages

NullVoxPopuli:
any objections to this? I'd like to get it merged.

lib/active_model/serializer.rb
```diff
@@ -249,6 +257,12 @@ def attributes(requested_attrs = nil)
       end
     end
 
+    # @api private
+    # Used by JsonApi adapter to build resource links.
+    def links
+      self.class._links
+    end
```
bf4: adding an instance method is still kind of dangerous, no?

lib/active_model/serializer/adapter/json_api/link.rb
bf4: I think so maybe, but that's what the ApiObjects namespace was for, and also the two RFCs

test/adapter/json_api/links_test.rb
bf4: It's arguably valid, but not recommended by the JSON API team

PR #1247, opened 2015-10-06
Add support for toplevel JSON API links.

beauby:
As an adapter option:
```ruby
render json: posts, links: { self: { href: "http//api.domain.com/posts" } }
```

### DIFF
--- a/lib/active_model/serializable_resource.rb
+++ b/lib/active_model/serializable_resource.rb
@@ -1,7 +1,7 @@
 require 'set'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links])
 
     # Primary interface to composing a resource with a serializer and adapter.
     # @return the serializable_resource, ready for #as_json/#to_json/#serializable_hash.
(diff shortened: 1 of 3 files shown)


lib/active_model/serializable_resource.rb
```diff
@@ -1,7 +1,7 @@
 require 'set'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links])
```
bf4: nice!  shouldn't this integrate with existing code such as pagination links?

NullVoxPopuli:
:+1:

lib/active_model/serializable_resource.rb
beauby: Not really, pagination is automagically added to resources anyways.

joaomdmoura: but would both work well together? What do you think about adding a test covering it? :smile:

joaomdmoura:
Just one small comments, would be nice to have this test IMO. Other then this it LGTM

lib/active_model/serializable_resource.rb
beauby: Yes, the pagination stuff just updates the links.

joaomdmoura:
LGTM then, merging it.

Issue #1098, opened 2015-08-28
Consider having a `SerializableResource` and `DeserializableDocument`


bf4:
Whereas

- Having `SerializableResource.serialize` return an instance of `SerializableResource` is not intuitive.
- It makes more sense for `SerializableResource.new` to return a `serializable_resource`
- A SerializableResource takes a Ruby model and presents it with a serialization interface `serializable_hash`, `as_json`, `to_json` that is usually [consumed by the Renderer in Rails](http://www.benjaminfleischer.com/2015/06/02/understanding-rails-model-serializers/)
- The upcoming deserialization feature takes a JSON document and returns a Ruby model
- The code for `SerializableResource.serialize` and `SerializableResource.deserialize` would get complicated quite quickly as they're really different responsibilities

Therefore

- It may be more intuitive to have a `SerializableResource` that takes a resource and a `DeserializableDocument` that takes a JSON document.  It might even make sense for the `DeserializableDocument` to be a pure function since there may not need to be an intermediate state, as there is with the `SerializableResource`.

Also, we might start a convention of calling subclass of an ActiveModel::Serializer an e.g. PostSerialization 

Relevant PRS

- https://github.com/rails-api/active_model_serializers/pull/1083
- https://github.com/rails-api/active_model_serializers/pull/950
  -  https://github.com/rails-api/active_model_serializers/pull/950#issuecomment-120059189
  - https://github.com/rails-api/active_model_serializers/pull/950#issuecomment-134037938 ) 
  - https://github.com/joaomdmoura/active_model_serializers/commit/5bc262f81bf1b5a3a50bc0668e975def2ecd0c02 ? Joao renamed PostSerializer to PostSerialization 
- https://github.com/rails-api/active_model_serializers/pull/954
  - https://github.com/rails-api/active_model_serializers/pull/954#issuecomment-114022049
  - https://github.com/rails-api/active_model_serializers/pull/954#discussion_r33283723
  - https://github.com/rails-api/active_model_serializers/pull/954#issuecomment-120771297
- https://github.com/rails-api/active_model_serializers/pull/889#discussion_r32258272

Some (lightly amended) discussion of what SerializableResource does (from the [slack team](https://amserializers.slack.com))

```plain
> Hmm why does AM::SerializableResource#serialize return an AM::SerializableResource? Is it not a bit confusing?

bf4 [7:43 PM]
b/c it's `AM::SerializableResource::serialize`, a class method returning an instance :simple_smile:

[...]

PR #1248, opened 2015-10-06
Basic deserialization.

beauby:
# Latest update:
Recap on the current situation:
+ Two methods are exposed (`ActiveModelSerializers::Deserialization.jsonapi_parse` and `ActiveModelSerializers::Deserialization.jsonapi_parse!`)
+ They both take the same parameters:
  - a `document` (either a Hash or an instance of `ActionController::Parameters` representing a JSON API payload)
  - an `options` hash:
    - `only` or `except`: array of white/black-listed fields (recall that in the context of JSON API, "fields" denotes both attributes and relationships)
    - `polymorphic`: array of polymorphic associations (those for which the `#{relationship}_type` must be set for `ActiveRecord` to handle it properly)
    - `keys`: hash of translated keys (for instance when the payload contains an `author` relationship which refers to an `user` association on the model)
+ The "unsafe" method (the one ending with a `!`) throws an `InvalidDocument` exception when the payload does not meet basic criteria for JSON API deserialization.
+ The "safe" method simply returns an empty hash.

Example:
```ruby
def create
  @post = Post.new(jsonapi_create_params)
end

def jsonapi_create_params
  ActiveModelSerializers::Deserialization.jsonapi_parse(params, only: [:title, :content])
end
```

~~There should be a bit more tests, but other than that,~~ I believe this PR is ready. Any thoughts?

# Update:
This PR offers basic JSON API deserialization through `ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(hash, options = {})` which takes a hash ~~or an instance of `ActionController::Parameters`~~ representing a JSON API document (and possibly some options, more on that further down) and returns an `ActiveRecord`-friendly hash, so that you can use it as follows:
```ruby
post_params = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(params.to_h) # or params.to_unsafe_h depending on your rails version
@post = Post.new(post_params)
```

The only available option currently is `fields`, which allows for
 + whitelisting of fields (attributes and relationships) by providing an array of symbols, and 
 + specifying the corresponding attribute name on the model when using the `key` option on the association in the serializer, by providing a Hash at the end of the array.
Example:
```ruby
hash = {
  'data' => {
    'type' => 'photos',
    'id' => 'zorglub',
    'attributes' => {
      'title' => 'Ember Hamster',
      'src' => 'http://example.com/images/productivity.png'
    },
    'relationships' => {
      'author' => {
        'data' => nil
      },
      'photographer' => {
        'data' => { 'type' => 'people', 'id' => '9' }
      },
      'comments' => {
        'data' => [
          { 'type' => 'comments', 'id' => '1' },
          { 'type' => 'comments', 'id' => '2' }
        ]
      }
    }
  }
}

parse(hash, fields: [:id, :title, :src, :comments, author: :user])
```
returns
```ruby
{
  id: 'zorglub',
  src: 'http://example.com/images/productivity.png',
  user_id: nil,
  comment_ids: ['1', '2']
}
```

Note: It currently does not handle polymorphic associations, but could easily be extended to do so.

# Initial post:

This PR offers basic JSON API deserialization through ~~`ActiveModel::Serializer::Adapter::JsonApi.parse(params_or_hash)`~~ which takes a hash or an instance of `ActionController::Parameters` representing a JSON API document and turns it into an `ActiveRecord`-friendly hash, so that you cas use it as follows:
```ruby
post_params = ActiveModel::Serializer::Adapter::JsonApi.parse(params)
@post = Post.new(post_params)
```

~~Note that this PR does not support aliased relationships~~ (that is the name of the relationship in the client-issued JSON API document won't be changed). 
Moreover, ~~it currently does not support any kind of whitelisting directly~~, although you could simply do `post_params.slice!(:id, :name, :description)` to whitelist or `post_params.except!(:is_admin)` to blacklist.
Finally, ~~it currently fails if the JSON API document is ill-formed.~~

We could also make it so that the `parse` method transforms whatever comes its way into a new `ActionController::Parameters` instance, so that people can `require` / `permit` stuff instead of `slice` / `except`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Breaking changes:
 
 Features:
 
+- [#1248](https://github.com/rails-api/active_model_serializers/pull/1248) Experimental: Add support for JSON API deserialization (@beauby)
 - [#1378](https://github.com/rails-api/active_model_serializers/pull/1378) Change association blocks
   to be evaluated in *serializer* scope, rather than *association* scope. (@bf4)
     * Syntax changes from e.g.
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -6,6 +6,7 @@ module ActiveModel
         autoload :PaginationLinks
         autoload :FragmentCache
         autoload :Link
+        autoload :Deserialization
 
         # TODO: if we like this abstraction and other API objects to it,
         # then extract to its own file and require it.
--- a/lib/active_model/serializer/adapter/json_api/deserialization.rb
+++ b/lib/active_model/serializer/adapter/json_api/deserialization.rb
@@ -1,0 +1,207 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        # NOTE(Experimental):
+        # This is an experimental feature. Both the interface and internals could be subject
+        # to changes.
+        module Deserialization
+          InvalidDocument = Class.new(ArgumentError)
+
+          module_function
+
+          # Transform a JSON API document, containing a single data object,
+          # into a hash that is ready for ActiveRecord::Base.new() and such.
+          # Raises InvalidDocument if the payload is not properly formatted.
+          #
+          # @param [Hash|ActionController::Parameters] document
+          # @param [Hash] options
+          #   only: Array of symbols of whitelisted fields.
+          #   except: Array of symbols of blacklisted fields.
+          #   keys: Hash of translated keys (e.g. :author => :user).
+          #   polymorphic: Array of symbols of polymorphic fields.
+          # @return [Hash]
+          #
+          # @example
+          #   document = {
+          #     data: {
+          #       id: 1,
+          #       type: 'post',
+          #       attributes: {
+          #         title: 'Title 1',
+          #         date: '2015-12-20'
+          #       },
+          #       associations: {
+          #         author: {
+          #           data: {
+          #             type: 'user',
+          #             id: 2
+          #           }
+          #         },
+          #         second_author: {
+          #           data: nil
+          #         },
+          #         comments: {
+          #           data: [{
+          #             type: 'comment',
+          #             id: 3
+          #           },{
+          #             type: 'comment',
+          #             id: 4
+          #           }]
+          #         }
+          #       }
+          #     }
+          #   }
+          #
+          #   parse(document) #=>
+          #     # {
+          #     #   title: 'Title 1',
+          #     #   date: '2015-12-20',
+          #     #   author_id: 2,
+          #     #   second_author_id: nil
+          #     #   comment_ids: [3, 4]
+          #     # }
+          #
+          #   parse(document, only: [:title, :date, :author],
+          #                   keys: { date: :published_at },
+          #                   polymorphic: [:author]) #=>
+          #     # {
+          #     #   title: 'Title 1',
+          #     #   published_at: '2015-12-20',
+          #     #   author_id: '2',
+          #     #   author_type: 'people'
+          #     # }
+          #
+          def parse!(document, options = {})
+            parse(document, options) do |invalid_payload, reason|
+              fail InvalidDocument, "Invalid payload (#{reason}): #{invalid_payload}"
+            end
+          end
+
+          # Same as parse!, but returns an empty hash instead of raising InvalidDocument
+          # on invalid payloads.
+          def parse(document, options = {})
+            document = document.dup.permit!.to_h if document.is_a?(ActionController::Parameters)
+
+            validate_payload(document) do |invalid_document, reason|
+              yield invalid_document, reason if block_given?
+              return {}
+            end
+
+            primary_data = document['data']
+            attributes = primary_data['attributes'] || {}
+            attributes['id'] = primary_data['id'] if primary_data['id']
+            relationships = primary_data['relationships'] || {}
+
+            filter_fields(attributes, options)
+            filter_fields(relationships, options)
+
+            hash = {}
+            hash.merge!(parse_attributes(attributes, options))
+            hash.merge!(parse_relationships(relationships, options))
+
+            hash
+          end
+
+          # Checks whether a payload is compliant with the JSON API spec.
+          #
+          # @api private
+          # rubocop:disable Metrics/CyclomaticComplexity
+          def validate_payload(payload)
+            unless payload.is_a?(Hash)
+              yield payload, 'Expected hash'
+              return
+            end
+
+            primary_data = payload['data']
+            unless primary_data.is_a?(Hash)
+              yield payload, { data: 'Expected hash' }
+              return
+            end
+
+            attributes = primary_data['attributes'] || {}
+            unless attributes.is_a?(Hash)
+              yield payload, { data: { attributes: 'Expected hash or nil' } }
+              return
+            end
+
+            relationships = primary_data['relationships'] || {}
+            unless relationships.is_a?(Hash)
+              yield payload, { data: { relationships: 'Expected hash or nil' } }
+              return
+            end
+
+            relationships.each do |(key, value)|
+              unless value.is_a?(Hash) && value.key?('data')
+                yield payload, { data: { relationships: { key => 'Expected hash with :data key' } } }
+              end
+            end
+          end
+          # rubocop:enable Metrics/CyclomaticComplexity
+
+          # @api private
+          def filter_fields(fields, options)
+            if (only = options[:only])
+              fields.slice!(*Array(only).map(&:to_s))
+            elsif (except = options[:except])
+              fields.except!(*Array(except).map(&:to_s))
+            end
+          end
+
+          # @api private
+          def field_key(field, options)
+            (options[:keys] || {}).fetch(field.to_sym, field).to_sym
+          end
+
+          # @api private
+          def parse_attributes(attributes, options)
+            attributes
+              .map { |(k, v)| { field_key(k, options) => v } }
+              .reduce({}, :merge)
+          end
+
+          # Given an association name, and a relationship data attribute, build a hash
+          # mapping the corresponding ActiveRecord attribute to the corresponding value.
+          #
+          # @example
+          #   parse_relationship(:comments, [{ 'id' => '1', 'type' => 'comments' },
+          #                                  { 'id' => '2', 'type' => 'comments' }],
+          #                                 {})
+          #    # => { :comment_ids => ['1', '2'] }
+          #   parse_relationship(:author, { 'id' => '1', 'type' => 'users' }, {})
+          #    # => { :author_id => '1' }
+          #   parse_relationship(:author, nil, {})
+          #    # => { :author_id => nil }
+          # @param [Symbol] assoc_name
+          # @param [Hash] assoc_data
+          # @param [Hash] options
+          # @return [Hash{Symbol, Object}]
+          #
+          # @api private
+          def parse_relationship(assoc_name, assoc_data, options)
+            prefix_key = field_key(assoc_name, options).to_s.singularize
+            hash =
+              if assoc_data.is_a?(Array)
+                { "#{prefix_key}_ids".to_sym => assoc_data.map { |ri| ri['id'] } }
+              else
+                { "#{prefix_key}_id".to_sym => assoc_data ? assoc_data['id'] : nil }
+              end
+
+            polymorphic = (options[:polymorphic] || []).include?(assoc_name.to_sym)
+            hash.merge!("#{prefix_key}_type".to_sym => assoc_data['type']) if polymorphic
+
+            hash
+          end
+
+          # @api private
+          def parse_relationships(relationships, options)
+            relationships
+              .map { |(k, v)| parse_relationship(k, v['data'], options) }
+              .reduce({}, :merge)
+          end
+        end
+      end
+    end
+  end
+end
--- a/test/adapter/json_api/parse_test.rb
+++ b/test/adapter/json_api/parse_test.rb
@@ -1,0 +1,139 @@
+require 'test_helper'
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        module Deserialization
+          class ParseTest < Minitest::Test
+            def setup
+              @hash = {
+                'data' => {
+                  'type' => 'photos',
+                  'id' => 'zorglub',
+                  'attributes' => {
+                    'title' => 'Ember Hamster',
+                    'src' => 'http://example.com/images/productivity.png'
+                  },
+                  'relationships' => {
+                    'author' => {
+                      'data' => nil
+                    },
+                    'photographer' => {
+                      'data' => { 'type' => 'people', 'id' => '9' }
+                    },
+                    'comments' => {
+                      'data' => [
+                        { 'type' => 'comments', 'id' => '1' },
+                        { 'type' => 'comments', 'id' => '2' }
+                      ]
+                    }
+                  }
+                }
+              }
+              @params = ActionController::Parameters.new(@hash)
+              @expected = {
+                id: 'zorglub',
+                title: 'Ember Hamster',
+                src: 'http://example.com/images/productivity.png',
+                author_id: nil,
+                photographer_id: '9',
+                comment_ids: %w(1 2)
+              }
+
+              @illformed_payloads = [nil,
+                                     {},
+                                     {
+                                       'data' => nil
+                                     }, {
+                                       'data' => { 'attributes' => [] }
+                                     }, {
+                                       'data' => { 'relationships' => [] }
+                                     }, {
+                                       'data' => {
+                                         'relationships' => { 'rel' => nil }
+                                       }
+                                     }, {
+                                       'data' => {
+                                         'relationships' => { 'rel' => {} }
+                                       }
+                                     }]
+            end
+
+            def test_hash
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(@hash)
+              assert_equal(@expected, parsed_hash)
+            end
+
+            def test_actioncontroller_parameters
+              assert_equal(false, @params.permitted?)
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(@params)
+              assert_equal(@expected, parsed_hash)
+            end
+
+            def test_illformed_payloads_safe
+              @illformed_payloads.each do |p|
+                parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(p)
+                assert_equal({}, parsed_hash)
+              end
+            end
+
+            def test_illformed_payloads_unsafe
+              @illformed_payloads.each do |p|
+                assert_raises(InvalidDocument) do
+                  ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(p)
+                end
+              end
+            end
+
+            def test_filter_fields_only
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(@hash, only: [:id, :title, :author])
+              expected = {
+                id: 'zorglub',
+                title: 'Ember Hamster',
+                author_id: nil
+              }
+              assert_equal(expected, parsed_hash)
+            end
+
+            def test_filter_fields_except
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(@hash, except: [:id, :title, :author])
+              expected = {
+                src: 'http://example.com/images/productivity.png',
+                photographer_id: '9',
+                comment_ids: %w(1 2)
+              }
+              assert_equal(expected, parsed_hash)
+            end
+
+            def test_keys
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(@hash, keys: { author: :user, title: :post_title })
+              expected = {
+                id: 'zorglub',
+                post_title: 'Ember Hamster',
+                src: 'http://example.com/images/productivity.png',
+                user_id: nil,
+                photographer_id: '9',
+                comment_ids: %w(1 2)
+              }
+              assert_equal(expected, parsed_hash)
+            end
+
+            def test_polymorphic
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(@hash, polymorphic: [:photographer])
+              expected = {
+                id: 'zorglub',
+                title: 'Ember Hamster',
+                src: 'http://example.com/images/productivity.png',
+                author_id: nil,
+                photographer_id: '9',
+                photographer_type: 'people',
+                comment_ids: %w(1 2)
+              }
+              assert_equal(expected, parsed_hash)
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -12,6 +12,7 @@ module ActiveModelSerializers
   extend ActiveSupport::Autoload
   autoload :Model
   autoload :Callbacks
+  autoload :Deserialization
   autoload :Logging
 end
 
--- a/lib/active_model_serializers/deserialization.rb
+++ b/lib/active_model_serializers/deserialization.rb
@@ -1,0 +1,13 @@
+module ActiveModelSerializers
+  module Deserialization
+    module_function
+
+    def jsonapi_parse(*args)
+      ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(*args)
+    end
+
+    def jsonapi_parse!(*args)
+      ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!(*args)
+    end
+  end
+end
--- a/test/action_controller/json_api/deserialization_test.rb
+++ b/test/action_controller/json_api/deserialization_test.rb
@@ -1,0 +1,59 @@
+require 'test_helper'
+
+module ActionController
+  module Serialization
+    class JsonApi
+      class DeserializationTest < ActionController::TestCase
+        class DeserializationTestController < ActionController::Base
+          def render_parsed_payload
+            parsed_hash = ActiveModelSerializers::Deserialization.jsonapi_parse(params)
+            render json: parsed_hash
+          end
+        end
+
+        tests DeserializationTestController
+
+        def test_deserialization
+          hash = {
+            'data' => {
+              'type' => 'photos',
+              'id' => 'zorglub',
+              'attributes' => {
+                'title' => 'Ember Hamster',
+                'src' => 'http://example.com/images/productivity.png'
+              },
+              'relationships' => {
+                'author' => {
+                  'data' => nil
+                },
+                'photographer' => {
+                  'data' => { 'type' => 'people', 'id' => '9' }
+                },
+                'comments' => {
+                  'data' => [
+                    { 'type' => 'comments', 'id' => '1' },
+                    { 'type' => 'comments', 'id' => '2' }
+                  ]
+                }
+              }
+            }
+          }
+
+          post :render_parsed_payload, hash
+
+          response = JSON.parse(@response.body)
+          expected = {
+            'id' => 'zorglub',
+            'title' => 'Ember Hamster',
+            'src' => 'http://example.com/images/productivity.png',
+            'author_id' => nil,
+            'photographer_id' => '9',
+            'comment_ids' => %w(1 2)
+          }
+
+          assert_equal(expected, response)
+        end
+      end
+    end
+  end
+end


lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -41,6 +41,38 @@ def object
...
+        # into an ActiveRecord-ready hash.
+        # NOTE(beauby): Currently this does not handle relationships with modified keys.
+        #
+        # @param [Hash|ActionController::Parameters] document
+        # @return [Hash] ActiveRecord-ready hash
+        #
+        def self.parse(document)
+          hash = {}
+
+          hash[:id] = document['data']['id'] if document['data']['id']
```
NullVoxPopuli: should there be a check for the existence of `'data'`? could be handy to have detailed parse errors

beauby: Sure. C.f. the description of the PR, there is currently no error handling, but it is planned.

NullVoxPopuli: :+1:

bf4: Can we try to stick to the ActiveModel::Serializers::Json API
https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/json.rb#L137
`from_json`?

For completeness I'd also like to compare this to other libraries out there
and what language they use.

I think, as we've discussed, there should be a DeserializableDocument class
that has a from_json method on it, and that perhaps uses a Deserializer
underneath ref:
https://github.com/rails-api/active_model_serializers/issues/1098

beauby: Rails' `from_json` and this method do different stuff. Here we are not building an instance of a model, we are just translating a json payload into a hash that can be passed to an AR model's constructor.

beauby:
Updated with handling of ill-formed documents, whitelisting, and aliased relationships.

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,73 @@
...
+          # Parse a Hash or ActionController::Parameters representing a JSON API document
+          # into an ActiveRecord-ready hash.
+          #
+          # @param [Hash|ActionController::Parameters] document
+          # @param [Hash] options
+          #   fields: Array of symbols and a Hash. Specify whitelisted fields, optionally
+          #      specifying the attribute name on the model.
+          # @return [Hash] ActiveRecord-ready hash
+          #
+          def parse(document, options = {})
```
bf4: [`from_json`?](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/json.rb#L137-L141)

test/adapter/json_api/parse_test.rb
```diff
@@ -0,0 +1,73 @@
...
+                  }
+                }
+              }
+              @expected = {
+                id: 'zorglub',
+                title: 'Ember Hamster',
+                src: 'http://example.com/images/productivity.png',
+                author_id: nil,
+                photographer_id: '9',
+                comment_ids: %w(1 2)
```
bf4: Does Rails use comment_ids on a has_many :comments to infer the relationship via foreign key?

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: The issue with `from_json` is that it should be called on a model. Here, we make no assumption about the model. I guess we could override `ActiveModel::Serializers::JSON` so that it calls the `parse` method and updates the model. But it would be JSON API specific. Thoughts?

test/adapter/json_api/parse_test.rb
beauby: Yes, c.f. http://guides.rubyonrails.org/association_basics.html section 4.3.1.

lib/active_model/serializer/adapter/json_api/deserialization.rb
bf4: I just meant as a consistent interface

beauby: My question still holds though. As a consistent interface of what? to what?

NullVoxPopuli: I think calling from_json on json-api data should generate a model / associations / whatever.

so, imo, from_json could call this method, `parse` and return an instance of a model.

sachse:
Posting something like `{"data":{"attributes":{"email":"john@doe.dev","password":"qwerty"}}}` I get an error: 
```undefined method `each_with_object' for {"email"=>"john@doe.dev", "password"=>"qwerty"}:ActionController::Parameters```

Is this on my end?

**Comment by beauby** [comment]
@sachse Your payload looks ok. Could you give me your ruby version, rails version, platform and a full stacktrace?
Also, if you have time, could you clone this branch and run the tests (`$ rake test`)?

**Comment by sachse** [comment]
@beauby I'm running Ruby 2.2.3 and Rails 5 from master on OS X. The test suite passes without any problems.

Full trace:
```
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/active_model_serializers-a8564dcfe8f0/lib/active_model/serializer/adapter/json_api/deserialization.rb:46:in `parse_attributes'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/active_model_serializers-a8564dcfe8f0/lib/active_model/serializer/adapter/json_api/deserialization.rb:25:in `parse'
app/controllers/v1/sessions_controller.rb:18:in `session_params'
app/controllers/v1/sessions_controller.rb:13:in `create'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal/basic_implicit_render.rb:4:in `send_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/abstract_controller/base.rb:183:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal/rendering.rb:30:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/abstract_controller/callbacks.rb:20:in `block in process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/callbacks.rb:97:in `__run_callbacks__'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/callbacks.rb:743:in `_run_process_action_callbacks'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/callbacks.rb:90:in `run_callbacks'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/abstract_controller/callbacks.rb:19:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal/rescue.rb:29:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal/instrumentation.rb:31:in `block in process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/notifications.rb:164:in `block in instrument'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/notifications.rb:164:in `instrument'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal/instrumentation.rb:29:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal/params_wrapper.rb:248:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activerecord/lib/active_record/railties/controller_runtime.rb:18:in `process_action'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/abstract_controller/base.rb:128:in `process'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal.rb:192:in `dispatch'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_controller/metal.rb:264:in `dispatch'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/routing/route_set.rb:48:in `dispatch'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/routing/route_set.rb:32:in `serve'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/journey/router.rb:42:in `block in serve'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/journey/router.rb:29:in `each'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/journey/router.rb:29:in `serve'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/routing/route_set.rb:717:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionview/lib/action_view/digestor.rb:14:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rack-35599cfc2751/lib/rack/etag.rb:25:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rack-35599cfc2751/lib/rack/conditional_get.rb:38:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rack-35599cfc2751/lib/rack/head.rb:12:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activerecord/lib/active_record/query_cache.rb:36:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb:963:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activerecord/lib/active_record/migration.rb:489:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/callbacks.rb:29:in `block in call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/callbacks.rb:97:in `__run_callbacks__'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/callbacks.rb:743:in `_run_call_callbacks'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/callbacks.rb:90:in `run_callbacks'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/callbacks.rb:27:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/reloader.rb:71:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/remote_ip.rb:79:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb:48:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/show_exceptions.rb:31:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/railties/lib/rails/rack/logger.rb:42:in `call_app'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/railties/lib/rails/rack/logger.rb:24:in `block in call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/tagged_logging.rb:70:in `block in tagged'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/tagged_logging.rb:26:in `tagged'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/tagged_logging.rb:70:in `tagged'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/railties/lib/rails/rack/logger.rb:24:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/request_id.rb:24:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rack-35599cfc2751/lib/rack/runtime.rb:22:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/activesupport/lib/active_support/cache/strategy/local_cache_middleware.rb:28:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/load_interlock.rb:13:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/actionpack/lib/action_dispatch/middleware/static.rb:133:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rack-35599cfc2751/lib/rack/sendfile.rb:111:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rails-09463183867f/railties/lib/rails/engine.rb:522:in `call'
/Users/kristoffer/.gem/ruby/2.2.3/bundler/gems/rack-35599cfc2751/lib/rack/handler/webrick.rb:86:in `service'
/Users/kristoffer/.rubies/ruby-2.2.3/lib/ruby/2.2.0/webrick/httpserver.rb:138:in `service'
/Users/kristoffer/.rubies/ruby-2.2.3/lib/ruby/2.2.0/webrick/httpserver.rb:94:in `run'
/Users/kristoffer/.rubies/ruby-2.2.3/lib/ruby/2.2.0/webrick/server.rb:294:in `block in start_thread'
```

**Comment by peterhicks** [comment]
I'm getting the same issue. I'm running Ruby 2.2.3, Rails 5 (from master), and AMS from your fork as of 10 minutes ago.

With the following code:

`@post_params = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(params)`
`@genre = Genre.new(genre_post_params)`

I get this stack trace:
```
ActionController::RoutingError at /genres
=========================================

> undefined method `each_with_object' for {"name"=>"test"}:ActionController::Parameters

/Users/peter/Sites/rails/actionpack/lib/action_dispatch/routing/route_set.rb, line 35
-------------------------------------------------------------------------------------

``` ruby
   30             controller = controller req
   31             res        = controller.make_response! req
   32             dispatch(controller, params[:action], req, res)
   33           rescue NameError => e
   34             if @raise_on_name_error
>  35               raise ActionController::RoutingError, e.message, e.backtrace
   36             else
   37               return [404, {'X-Cascade' => 'pass'}, []]
   38             end
   39           end
   40   
```

App backtrace
-------------



Full backtrace
--------------

 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/routing/route_set.rb:35:in `rescue in serve'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/routing/route_set.rb:29:in `serve'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/journey/router.rb:42:in `block in serve'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/journey/router.rb:29:in `serve'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/routing/route_set.rb:717:in `call'
 - bullet (4.14.10) lib/bullet/rack.rb:10:in `call'
 - /Users/peter/Sites/rails/actionview/lib/action_view/digestor.rb:14:in `call'
 -  () Users/peter/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-35599cfc2751/lib/rack/etag.rb:25:in `call'
 -  () Users/peter/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-35599cfc2751/lib/rack/conditional_get.rb:38:in `call'
 -  () Users/peter/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-35599cfc2751/lib/rack/head.rb:12:in `call'
 - /Users/peter/Sites/rails/activerecord/lib/active_record/query_cache.rb:36:in `call'
 - /Users/peter/Sites/rails/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb:963:in `call'
 - /Users/peter/Sites/rails/activerecord/lib/active_record/migration.rb:489:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/callbacks.rb:29:in `block in call'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/callbacks.rb:97:in `__run_callbacks__'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/callbacks.rb:743:in `_run_call_callbacks'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/callbacks.rb:90:in `run_callbacks'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/callbacks.rb:27:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/reloader.rb:71:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/remote_ip.rb:79:in `call'
 - better_errors (2.1.1) lib/better_errors/middleware.rb:84:in `protected_app_call'
 - better_errors (2.1.1) lib/better_errors/middleware.rb:79:in `better_errors_call'
 - better_errors (2.1.1) lib/better_errors/middleware.rb:57:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb:48:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/show_exceptions.rb:31:in `call'
 - /Users/peter/Sites/rails/railties/lib/rails/rack/logger.rb:42:in `call_app'
 - /Users/peter/Sites/rails/railties/lib/rails/rack/logger.rb:24:in `block in call'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/tagged_logging.rb:70:in `block in tagged'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/tagged_logging.rb:26:in `tagged'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/tagged_logging.rb:70:in `tagged'
 - /Users/peter/Sites/rails/railties/lib/rails/rack/logger.rb:24:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/request_id.rb:24:in `call'
 -  () Users/peter/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-35599cfc2751/lib/rack/runtime.rb:22:in `call'
 - /Users/peter/Sites/rails/activesupport/lib/active_support/cache/strategy/local_cache_middleware.rb:28:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/load_interlock.rb:13:in `call'
 - /Users/peter/Sites/rails/actionpack/lib/action_dispatch/middleware/static.rb:133:in `call'
 -  () Users/peter/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-35599cfc2751/lib/rack/sendfile.rb:111:in `call'
 - /Users/peter/Sites/rails/railties/lib/rails/engine.rb:522:in `call'
 -  () Users/peter/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/rack-35599cfc2751/lib/rack/handler/webrick.rb:86:in `service'
 - /Users/peter/.rbenv/versions/2.2.3/lib/ruby/2.2.0/webrick/httpserver.rb:138:in `service'
 - /Users/peter/.rbenv/versions/2.2.3/lib/ruby/2.2.0/webrick/httpserver.rb:94:in `run'
 - /Users/peter/.rbenv/versions/2.2.3/lib/ruby/2.2.0/webrick/server.rb:294:in `block in start_thread'
```

**Comment by beauby** [comment]
Thanks @sachse. I believe ActionController::Parameters changed a bit in Rails5. I'll find a workaround tonight.

**Comment by beauby** [comment]
Couldn't resist to see what's going on. Since [this commit](https://github.com/rails/rails/commit/14a3bd520dd4bbf1247fd3e0071b59c02c115ce0) `ActionController::Parameters` do not inherit from `HashWithIndifferentAccess` anymore, hence the issue at hand. Pushing a fix in 2mins.

File: test/adapter/json_api/parse_test.rb [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+            end
+
+            def test_hash
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(@hash)
+              assert_equal(@expected, parsed_hash)
+            end
+
+            def test_parameters
+              parameters = ActionController::Parameters.new(@hash)
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(parameters)
```
jmonteiro: One solution to fixing this would be to send `parameters.to_unsafe_h` instead of only `parameters`. Or to try to call that method before calling `to_h` inside the `#parse` method.

beauby: @jmonteiro The purpose of this test is to ensure the `parse` method can handle `ActionController::Parameters` in order to reduce friction for the user.

beauby: The idea is that deserialization should handle parameters transparently. The preferred way for whitelisting would be through the `parse` method, but in case a user wants to do whitelisting on the  `ActionController::Parameters` instance, we should not override its will, that's why we're using `to_h` instead of `to_unsafe_h` inside `parse`.

beauby: Oh, I may have made a false assumption here. I thought a pristine instance of `ActionController::Parameters` would have all its parameters whitelisted by default.

beauby: Ok the catch here is that when initializing an instance of `ActionController::Parameters` like here, it will automatically whitelist all attributes, hence the test wouldn't catch the problem you're mentioning. However, the `to_unsafe_h` method was only added in rails 4.2, so I'm not sure what the best solution is here.

beauby: For now, I'll remove direct support for `ActionController::Parameters`.

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,72 @@
...
+          module_function
+
+          # Parse a Hash representing a JSON API document into an ActiveRecord-ready hash.
+          #
+          # @param [Hash] document
+          # @param [Hash] options
+          #   fields: Array of symbols and a Hash. Specify whitelisted fields, optionally
+          #     specifying the attribute name on the model.
+          # @return [Hash] ActiveRecord-ready hash
+          #
```
bf4: `@example` please.. needs more docs :)

NullVoxPopuli: :+1:

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,72 @@
...
+            primary_data = document.fetch('data', {})
+            hash[:id] = primary_data['id'] if primary_data['id'] && (fields.nil? || fields[:id])
+
+            hash.merge!(parse_attributes(primary_data['attributes'], fields))
+            hash.merge!(parse_relationships(primary_data['relationships'], fields))
+
+            hash
+          end
+
+          # @api private
```
bf4: this converts options[:fields] into nil or a Hash?  Would `extract_options` work?

lib/active_model/serializer/adapter/json_api/deserialization.rb
bf4: Is this the flow?

1. deserialize_document
  1. deserialize_data
    1. deserialize_resource
      1. deserialize_relationships
  1. deserialize_meta
  1. deserialize_included
  1. deserialize_links

NullVoxPopuli: This should be in a method header comment on the parse method (or something like it)

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: yes and nope, respectively

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: except there is no ii. or iv.

lib/active_model/serializer/adapter/json_api/deserialization.rb
NullVoxPopuli: @beauby looks like you need to add some method header comments :-\

GCorbel:
It works great. It should be merged! Thanks for your work.

beauby:
Rebased. Any feedback on this? I believe it is pretty useful.

bf4:
Any updates since last discussion? Also, ref: https://github.com/rails-api/active_model_serializers/issues/1098

NullVoxPopuli:
my only feedback was for inline documentation, which I know @beauby despises... so...

:-)

NullVoxPopuli:
but other than that, I think it should be merged -- just to get things moving.

:+1:

test/adapter/json_api/parse_test.rb
```diff
@@ -0,0 +1,67 @@
...
+            end
+
+            def test_hash
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(@hash)
+              assert_equal(@expected, parsed_hash)
+            end
+
+            def test_illformed_payload
+              parsed_hash = ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse({})
+              assert_equal({}, parsed_hash)
```
bf4: I don't understand what this is testing.  What would a failure look like?  (also, `malformed` or `invalid`?)

test/adapter/json_api/parse_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  class Serializer
+    module Adapter
+      class JsonApi
+        module Deserialization
+          class ParseTest < Minitest::Test
+            def setup
+              @hash = {
+                'data' => {
+                  'type' => 'photos',
+                  'id' => 'zorglub',
```
bf4: So, this only works on single-resource documents?

test/adapter/json_api/parse_test.rb
```diff
@@ -0,0 +1,67 @@
+require 'test_helper'
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        module Deserialization
+          class ParseTest < Minitest::Test
+            def setup
```
bf4: Needs a 'reversible' test.

1. Given a resource
2. Serialize resource to hash
3. deserialize to resource (attributes?)

test/adapter/json_api/parse_test.rb
beauby: Yes, as per JSON API spec.

bf4: the spec doesn't allow deserializing collections?

beauby: Nope. C.f. [Creating Resources](http://jsonapi.org/format/#crud-creating):
>  The request MUST include a single resource object as primary data.

[Updating Resources](http://jsonapi.org/format/#crud-updating):
> The PATCH request MUST include a single resource object as primary data.

bf4: But that only means the server expects single-resource requests (without the bulk extension)

The client would still need to handle multiple-resource responses

bf4: which I suppose is fine to start out only deserializing requests since that's the most common use-case

beauby: Not sure I understand what you mean. Could you elaborate? Without the bulk extension, the only payloads a server will ever deal with are single-resource ones.

test/adapter/json_api/parse_test.rb
bf4: and an integration test as described in https://github.com/rails-api/active_model_serializers/pull/1248/files#r46101225

test/adapter/json_api/parse_test.rb
bf4: Not sure if you replied before I wrote https://github.com/rails-api/active_model_serializers/pull/1248/files#r46101704

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,72 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        module Deserialization
+          module_function
+
+          # Parse a Hash representing a JSON API document into an ActiveRecord-ready hash.
```
bf4: Maybe `Transform a JSON API document, containing a single data object, from a hash into an ActiveRecord#attributes hash`?

test/adapter/json_api/parse_test.rb
beauby: Oh, right. Yeah, this PR was initially intended to solve request deserialization only. Note that it would be fairly easy to build upon the `parse` method to parse collections.

test/adapter/json_api/parse_test.rb
beauby: This test just makes sure AMS does not go crazy when the hash does not comply to the JSON API spec. I guess ideally there should be a bunch more tests for invalid/missing keys.

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: Looks good, will update.

test/adapter/json_api/parse_test.rb
beauby: Added an integration test. However, after giving it some thoughts, I don't see the relevance of a 'reversible' test. What we are doing here is precisely transforming a JSON API-compliant document into an `ActiveRecord::Attributes`-compliant hash, so I'm not sure what introducing a dependency to the serialization would add.

beauby:
Added an alias from `ActiveModelSerializers::Deserialization.jsonapi_parse` to `ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse`.

This PR now seems ready for merge in my opinion. Thoughts @rails-api/ams?

NullVoxPopuli:
LGTM :-)

lib/active_model/serializer/adapter/json_api/deserialization.rb
bf4:
```ruby
# Given an Array
# For each element in the array, built a hash
# Then a symbol element is is used as both the key and value
# And a Hash element is merged into return value
# And all other values in the array are ignored
# @example
#    parse_fields_options(['a']) # => nil
#    parse_fields_options([:a]) # => { a: :a }
#    parse_fields_options([{'a' => :a]) # => { 'a' => :a }
#    parse_fields_options(nil) # => nil
#    parse_fields_options({}) # => nil
#    parse_fields_options(:a) # => nil
#    parse_fields_options([[:a]]) # => {}
#    parse_fields_options([:a, {a: :b}]) # => { a: :b }
#    parse_fields_options([{a: :b}, :a]) # => { a: :a }
#   etc
#   etc
# @params fields [Array<Hash, Symbol>] object to transform into a hash or nil
# @return [Hash] for array inputs
# @return [Nil] otherwise
```

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,115 @@
...
+              if attr.is_a?(Symbol)
+                hash[attr] = attr
+              elsif attr.is_a?(Hash)
+                hash.merge!(attr)
+              end
+            end
+          end
+
+          # @api private
+          def parse_attributes(attributes, fields)
```
bf4: ?

1. Given fields and truthy attributes
  1. Build a hash from each key/value pair
    1. If fields is truthy
      1. And there is a value for the symbolized key
         1. Add that value as the key for the original hash's value
     1.  And the key can be symbolized
        1. Add that symbolized key as the key for the original hash's value
1. Else `{}`

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,115 @@
...
+            return {} unless attributes
+            attributes.each_with_object({}) do |(key, value), hash|
+              attribute_name = fields ? fields[key.to_sym] : key.to_sym
+              next unless attribute_name
+              hash[attribute_name] = value
+            end
+          end
+
+          # @api private
+          def parse_relationships(relationships, fields)
```
bf4: ?

1. Given fields and truthy relationships
  1. Build a hash from each key/value pair
    1. Derive the `association_name`
      1. If fields is truthy
        1. And there is a value for the symbolized key
         1. Use that value
       1.  And the key can be symbolized
          1. Use the symbolized key
    1. Get the value of `data` from the value
      1. If `data` is an array (of hashes)
        1. Derive the key as `"#{association_name.to_s.singularize.to_sym}_ids".to_sym`
        1. Add the mapped `id` field of each element in `data` as the value to the key
      1. else
        1. derive the key as `#{association_name}_id".to_sym`
        1. if `data` is truthy
          1. set the value of data's `id` field as the value to the key
        1. else
          1. set `nil` as the value to the key

    
1. Else `{}`

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,115 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        module Deserialization
+          module_function
+
+          # Transform a JSON API document, containing a single data object, from a hash
+          # into an ActiveRecord#attributes hash.
```
bf4: What does ` into an ActiveRecord#attributes hash` mean?

bf4:
I like what this does, but worry that it will be hard to maintain since it's kind of hard to follow what it's doing.  Do my comments about what each method does make sense?  Could we make use of `ActiveModel::Serializers::JSON` to just deserialize to that interface so that there's less to do?

frank06:
Hi! what's the status of deserialization?  Any plans of releasing 0.10 soon?

beauby:
@frank06 Deserialization (in the sense of transforming a hash/ActionController::Parameters instance into a hash that is ready for `Model.new()` and such) is ready and working (you can use this branch for now). I'd like to get it merged ASAP, but as you can see there is a tiny bit of documenting work left before we can do that. As far as 0.10 is concerned, we're a bit late on the schedule but it should be ready quite soon.

frank06:
Thanks @beauby for the heads up! People in the Ember community choose AMS and prefer JSON API... so it would be great if I could start recommending a fully working version

beauby:
Update: 
 + two flavors of `parse` now available (either via either `ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse`/``ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse!` or `ActiveModelSerializers::Deserialization.jsonapi_parse`/`ActiveModelSerializers::Deserialization.jsonapi_parse!`), one of which raises `ArugmentError 'Invalid payload'` on invalid payloads, and the other silently fails and returns an empty hash.
 + added tests
 + added yarddocs

beauby:
@bf4: `ActiveModel::Serializers::JSON.from_json` does something totally different, it builds a model from a plain json representation.

beauby:
@bf4: Moreover, I believe what's happening here is quite straightforward, the complexity is inherent to the JSON API format.

bf4:
> Moreover, I believe what's happening here is quite straightforward, the complexity is inherent to the JSON API format

@beauby You make me feel bad when you say it's 'quite straightforward', as if there's something wrong with me that I asked.

I'll review the doc updates later.  I see it still refers to active record attributes ... Doesn't look ar specific to me.

beauby:
@bf4 Sorry, I didn't mean it that way. I meant that clearly there's complexity to it, but this complexity comes from the spec itself. Like "parsing a LALR language is straightforward", in the sense that it's a total headache, but the procedure is known and documented.

beauby:
@bf4 And concerning the `ActiveRecord` reference, it is there because it is the target format. At the end of the day it is just a hash, but I do not know of a better name for "hash mapping attributes to their values and `"#{key.singularize}_ids"` to arrays of ids for to-many relationships, and `"#{key}_id"` to an id for to-one relationships.

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
...
+          #     #   author_id: 2,
+          #     # }
+          #
+          # rubocop:disable Metrics/CyclomaticComplexity
+          def parse!(document, options = {})
+            fields = parse_fields_option(options[:fields])
+
+            document = document.dup.permit!.to_h if document.is_a?(ActionController::Parameters)
+
+            fail ArgumentError, 'Invalid payload' unless payload_valid?(document)
```
bf4: I'd rather not use the exception for flow control.  I'd rather have all the code in `parse` and have it do, maybe


```ruby
if  payload_invalid?(document)
  if block_given?
    yield document
  else
    return {} # I guess this is supposed to be a benign value?
  end
end
```

and then

```ruby
InvalidDocument = Class.new(ArgumentError)
def parse!(document, options = {})
  parse(document, options) do |invalid_document)
    fail InvalidDocument, "Invalid Document: #{reason}" # reason would be good, right?
  end
end
```

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
...
+            parse!(document, options)
+          rescue ArgumentError
+            {}
+          end
+
+          # Checks whether a payload is compliant with the JSON API spec.
+          #
+          # @api private
+          # rubocop:disable Metrics/CyclomaticComplexity
+          def payload_valid?(payload)
```
bf4: Naming concern: this is the only function in the module that doesn't begin with `parse_`.  and it's usage is `unless payment_valid?` rather than `if payment_invalid?`.

Perhaps this method should be `validate_payload` then return an errors object that is `empty?` when valid.

test/adapter/json_api/parse_test.rb
bf4: >  introducing a dependency to the serialization would add

I mean like

1. Given: `{ 'data' : { 'posts' : [ { 'id' : '1' } ] } }`
1. Deserialize to `[ Post.new(id: 1) ]`
1. Serialize to `{ 'data' : { 'posts' : [ { 'id' : '1' } ] } }`

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Breaking changes:
 
 Features:
 
+- [#1248](https://github.com/rails-api/active_model_serializers/pull/1248) Add support for JSON API deserialization (@beauby)
```
bf4: maybe `for transforming a JSON API request document into an attributes hash`?

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
+module ActiveModel
+  class Serializer
+    module Adapter
+      class JsonApi
+        module Deserialization
+          module_function
+
+          # Transform a JSON API document, containing a single data object, from a hash
+          # into a hash that is ready for ActiveRecord::Base.new() and such.
```
bf4: or `ActiveModelSerializers::Model.new`, etc. :).

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: Agreed that the reason is probably a good thing to have around.

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: Kinda true, although the whole `blah_id` / `blah_ids` is not really handled by non-AR models, right?

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: I am personally less comfortable with yielding stuff around for flow-control, but that's probably just me, so it's not really relevant. The way I thought of it: I defined an "unsafe" method (which would throw an exception when encountering an ill-formed document – I intended to make a custom exception with the reason and such, but haven't got there yet), and a "safe" version which would just catch the exception. I don't know whether this is common practice or specifically avoided. I'd have to have a look at how rails does stuff. In the meantime, would you mind elaborating on the reasons why using exceptions for (highly-local) flow-control does not suit you?

beauby: Ping @bf4

bf4: oh, sorry... there's a number of reasons not to use exceptions for flow control, in no particular order

- they're slower than the alternative
- flow control isn't "exceptional"
- flow control isn't a good reason to possibly crash a program
- it's better to use throw/catch or blocks
  - throw/catch for unwinding the stack
  - blocks for allowing user-defined error-handling strategies

which to me I would summarize as "don't do it unless there's a really good reason"

If I had to choose over 'fail' vs. 'throw/catch', without any refactor, I'd pick throw catch, but really, I think the convention that `parse!` raises an exception is fine (if it is a good API for us), as long as the exceptional method doesn't raise an exception!

bf4: I can dig up sources if you're curious.  they're used in Nokogiri::SAX, Sinatra, and in Rails 5 callbacks https://github.com/rails/rails/pull/17227 .

lib/active_model/serializer/adapter/json_api/deserialization.rb
bf4: I leave that up to the non-AR implementor :)

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
...
+          # @api private
+          # rubocop:disable Metrics/CyclomaticComplexity
+          def payload_valid?(payload)
+            return false unless payload.is_a?(Hash)
+            primary_data = payload['data']
+            return false unless primary_data.is_a?(Hash)
+            attributes = primary_data['attributes']
+            return false unless attributes.nil? || attributes.is_a?(Hash)
+            relationships = primary_data['relationships']
+            return false unless relationships.nil? || relationships.is_a?(Hash)
```
jackturnbull: Feel free to correct me here (it's a bit late and I've not fully got my head around the whole project) but I think this logic isn't what we need. On https://github.com/beauby/active_model_serializers/commit/6ba0a1d77ace7da1a86f2afe0a8b8389c3d89ef9 I seem have the two following test cases:

On a hash without relationships:
```ruby
(byebug) hash = {'data'=>{'attributes'=>{'name'=>'test'}, 'type'=>'events'}}                                        
{"data"=>{"attributes"=>{"name"=>"test"}, "type"=>"events"}}

(byebug) ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(hash, fields:[:name])
*** NoMethodError Exception: undefined method `each' for nil:NilClass
```

On a hash with relationships:
```ruby
(byebug) hash = {'data'=>{'attributes'=>{'name'=>'test'}, 'type'=>'events', 'relationships' => {'author'=>{'data'=>nil}}}}
{"data"=>{"attributes"=>{"name"=>"test"}, "type"=>"events", "relationships"=>{"author"=>{"data"=>nil}}}}

(byebug) ActiveModel::Serializer::Adapter::JsonApi::Deserialization.parse(hash, fields:[:name])                     
{:name=>"test"}
```

Looking at the `unless` statements on line 106 and 108, if there is no relationship key in the hash as per the first test case, the `relationships` variable will be `nil` causing the `unless` statement to pass with no return of false.

It feels like the logic may need to be separated:
* If the relationships key does not exist then continue.
* If the relationships key exists, and it is a hash then iterate.
* If the relationships key exists, and it is not a hash then return false.

beauby: That's what I get for not running the tests before pushing... Thanks for spotting it @jackturnbull! Just pushed a fix.

test/adapter/json_api/parse_test.rb
beauby: I kinda disagree. The job of this piece of code is to do a transformation between two specified formats (JSON API on the one side, ActiveRecord#initialize on the other), which is currently being tested. I'm not against an integration test as described in your scenario, though. It just does not seem necessary to me to ensure the validity of this code.

lib/active_model/serializer/adapter/json_api/deserialization.rb
jackturnbull: Thanks again for the quick response and your work on this! I'm looking forward to getting it into production.

lib/active_model/serializer/adapter/json_api/deserialization.rb
bf4: Pong

B mobile phone

ignat-zakrevsky: could I intervene? Why not to use cons of both approachs?   

```ruby
DEFAULT_FALLBACK = ->(error) { fail ArgumentError, 'Invalid payload' }

def parse!(document, options = {}, &fallback)
  fallback ||= DEFAULT_FALLBACK
  
  fields = parse_fields_option(options[:fields])

  document = document.dup.permit!.to_h if document.is_a?(ActionController::Parameters)

  fallback.call(document) unless payload_valid?(document)

  # ...
```

1) It will rise error as soon as possible wich is much more explicit way then simple return empty hash
2) It allows user to provide custom fallback which can be {}

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
...
+            return nil unless fields.is_a?(Array)
+            fields
+              .map { |attr| attr.is_a?(Symbol) ? { attr => attr } : attr }
+              .reduce({}, :merge)
+          end
+
+          # @example
+          #   parse_attributes({ 'a' => 'b', 'c' => 'd'}, nil)
+          #    # => { :a => 'b', :c => 'd' }
+          #   parse_attributes({ 'a' => 'b', 'c' => 'd', 'e' => 'f' }, { :a, :c => :g })
```
ignat-zakrevsky: invalid example 
```ruby
{ :a, :c => :g }
```

lib/active_model/serializer/adapter/json_api/deserialization.rb
ignat-zakrevsky: could I intervene? Why not to use pros of both approaches?   

```ruby
DEFAULT_FALLBACK = ->(error) { fail ArgumentError, 'Invalid payload' }
SAFE_FALLBACK = ->(error) { {} }

def parse!(document, options = {}, &fallback)
  fallback ||= DEFAULT_FALLBACK
  
  fields = parse_fields_option(options[:fields])

  document = document.dup.permit!.to_h if document.is_a?(ActionController::Parameters)

  fallback.call(document) unless payload_valid?(document)

  # ...


def parse(document, options = {})
  parse!(document, options, SAFE_FALLBACK)
end
```

and then call

lib/active_model/serializer/adapter/json_api/deserialization.rb
beauby: You're right, it was meant to be `[ :a, :c => :g ]`. Thanks for spotting this.

lib/active_model/serializer/adapter/json_api/deserialization.rb
jackturnbull: Skipping back to the first post in this thread, and now that I'm starting to experiment with this PR a bit more; the current `Argument Error: Invalid Payload` format is a little too inflexible to return detailed error messages to the client. In terms of technical approach I couldn't comment but I agree with @bf4 on the following line:

```ruby
fail InvalidDocument, "Invalid Document: #{reason}" # reason would be good, right?
```

A custom error class with a reason would allow us to `rescue_from InvalidDocument, with: ...` so the default `param is missing or the value is empty: #{reason_field}` error message can be replicated, at least functionally.

I'd feel less comfortable rescuing an ArgumentError in a base controller since this is a bit of a catch all and without the reason the specific error can't be passed on to the client.

beauby:
Last commit:
+ Error reporting with custom exception and reason
+ Rewrite to allow the following options:
    - only/except
    - keys (hash of translated keys, like `:author => :user`)
    - polymorphic (list of polymorphic fields (mainly relationships), for which the `#{relationship}_type` attribute should be set)

beauby:
Recap on the current situation:
+ Two methods are exposed (`ActiveModelSerializers::Deserialization.jsonapi_parse` and `ActiveModelSerializers::Deserialization.jsonapi_parse!`)
+ They both take the same parameters:
  - a `document` (either a Hash or an instance of `ActionController::Parameters` representing a JSON API payload)
  - an `options` hash:
    - `only` or `except`: array of white/black-listed fields (recall that in the context of JSON API, "fields" denotes both attributes and relationships)
    - `polymorphic`: array of polymorphic associations (those for which the `#{relationship}_type` must be set for `ActiveRecord` to handle it properly)
    - `keys`: hash of translated keys (for instance when the payload contains an `author` relationship which refers to an `user` association on the model)
+ The "unsafe" method (the one ending with a `!`) throws an `InvalidDocument` exception when the payload does not meet basic criteria for JSON API deserialization.
+ The "safe" method simply returns an empty hash.

There should be a bit more tests, but other than that, I believe this PR is ready. Any thoughts?

CHANGELOG.md
bf4: Also, If I'm reading this code right, this isn't integrated anywhere, so maybe it would be to mark it as an experimental feature so we can merge it and iterate faster?

beauby: Yeah that would be nice!

bf4: Well, just point me to however you mark it as 'experimental, don't rely on this' in the code and changelog, and I'm ok with having it merged.  I'd generally like for us to have a way to add experimental features

bf4: (merged after squashing commits, of course)

bf4: probably would be good to create an issue output of this referencing unresolved issues, comments, what's next, etc.

beauby: Added to the changelog and as a comment on the `Deserialization` module.

beauby:
Will merge once green.

beauby:
Ladies and gentlemen: merging!

GCorbel:
Woohoo! Well done @beauby !

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
...
+          #     # {
+          #     #   title: 'Title 1',
+          #     #   date: '2015-12-20',
+          #     #   author_id: 2,
+          #     #   second_author_id: nil
+          #     #   comment_ids: [3, 4]
+          #     # }
+          #
+          #   parse(document, only: [:title, :date, :author],
+          #                   keys: { date: :published_at },
```
joaomdmoura: Nice!

lib/active_model/serializer/adapter/json_api/deserialization.rb
```diff
@@ -0,0 +1,207 @@
...
+          #                   polymorphic: [:author]) #=>
```
joaomdmoura: Not sure I got this polymorphic thing, can you help me to understand the need? :smile:

joaomdmoura: nvm! got it! pretty obvious actually :grin:

test/adapter/json_api/parse_test.rb
joaomdmoura: I agree that this is not the topic here, but would be a great test to have :smile:

maletor:
Should this support transforming the key *back*? Typically an application will store keys in underscore format. AMS should then convert them to dashed for JSON API. But, upon deserialization, they should be converted back to underscores. Seems like that should happen here.

bf4:
@maletor Can you make a new issue? thanks

PR #1251, opened 2015-10-07
Rename ArraySerializer to CollectionSerializer for clarity

bf4:


lib/active_model/serializer/array_serializer.rb
```diff
@@ -1,44 +1,9 @@
...
-      private # rubocop:disable Lint/UselessAccessModifier
-
-      ActiveModelSerializers.silence_warnings do
-        attr_reader :serializers
-      end
+require 'active_model/serializer/collection_serializer'
+class ActiveModel::Serializer
+  class ArraySerializer < CollectionSerializer
+    def initialize(*)
+      warn "Calling deprecated ArraySerializer in #{caller[0]}. Please use CollectionSerializer"
```
NullVoxPopuli: :+1:

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -0,0 +1,44 @@
+module ActiveModel
+  class Serializer
+    class CollectionSerializer
+      NoSerializerError = Class.new(StandardError)
```
NullVoxPopuli: this seems like a weird place for this.
Do we make new error types like this in other places?
should there be a common place where the errors are all kept?

lib/active_model/serializer/configuration.rb
```diff
@@ -7,9 +7,19 @@ module Configuration
       # Configuration options may also be set in
       # Serializers and Adapters
       included do |base|
-        base.config.array_serializer = ActiveModel::Serializer::ArraySerializer
-        base.config.adapter = :attributes
-        base.config.jsonapi_resource_type = :plural
+        config = base.config
+        config.collection_serializer = ActiveModel::Serializer::CollectionSerializer
+
+        def config.array_serializer=(collection_serializer)
```
NullVoxPopuli: why not use `alias`?

beauby:
I have a feeling pagination won't work out of the box with this.

lib/active_model/serializer/collection_serializer.rb
bf4: I'm thinking maybe it should be a throw since it's essentially for flowcontrol

B mobile phone

lib/active_model/serializer/configuration.rb
bf4: Wasn't sure how it would work in context so didn't bother.  What would the receiver be?

B mobile phone

beauby:
@bf4 Can you confirm this does not break handling of paginated collections as it previously was? I'm thinking of: https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json_api.rb#L96

NullVoxPopuli:
I guess if there aren't tests for that, there should be. @beauby, could you
check on that?

beauby:
The problem is that support for pagination currently relies on external gems.

NullVoxPopuli:
Could we add a pagination gem just in the test environment?

beauby:
Better mock it up. Those usually rely on AR.

bf4:
How did you know this was broken?  I didn't see anything.  The gems are for sure used in tests and hence abailable, IIRC 

B mobile phone

beauby:
@bf4 I don't see the gems in the Gemfile, nor have I seen any test about pagination. Could you point me to those tests?

bf4:
I'll add an integration test but where are the failures? I recall it's green

B mobile phone

bf4:
Gemspec

B mobile phone

beauby:
Wait, there have been some code changes since I posted that comment. Last I checked you got rid of the `paginated?` method on the Array/CollectionSerializer.

bf4:
If so not intentional.  It's just a rename

B mobile phone

bf4:
@beauby could you clarify for me what is blocking this?

lib/active_model/serializer/collection_serializer.rb
bf4: In any case, this file is just the array_serializer renamed

test/array_serializer_test.rb
```diff
@@ -1,92 +1,21 @@
...
-
-      def test_json_key_with_root
-        serializer = ArraySerializer.new(@resource, root: 'custom_root')
-        assert_equal serializer.json_key, 'custom_roots'
-      end
-
-      def test_json_key_with_root_and_no_serializers
-        serializer = ArraySerializer.new(build_named_collection, root: 'custom_root')
-        assert_equal serializer.json_key, 'custom_roots'
+    class ArraySerializerTest < CollectionSerializerTest
```
bf4: for backwards compat https://github.com/rails-api/active_model_serializers/pull/1251/files#r41372642 also, is awesome

beauby:
LGTM but needs rebase.

NullVoxPopuli:
:+1:

Issue #1245, opened 2015-10-06
Can not filter attributes dymically?


groe:
In version 0.8.x it was possible to dynamically exclude certain attributes by implementing `include_<name>?` and returning false. In 0.9.x the same could be achieved with the `filter(keys)` method. For 0.10.0.rc3 [this SO answer](http://stackoverflow.com/a/32490558) suggests overwriting the `attributes` method but that requires the attribute to exist anyway (and it is fetched always, even when excluded, which is not ideal from a performance point of view).

What is the recommended way to filter attributes (e.g. based on the current user, or options passed from the view) in 0.10.0.rc3 using the standard attributes adapter?

bf4:
HI @groe would you mind giving an example of what you're trying to do so we can tailor our answer to your use case?

groe:
Hi @bf4, sure! One example would be to exclude attributes the `current_user` does not have access to.

Could be something like this:
```ruby
class PostSerializer < ActiveModel::Serializer
  attributes :id, :name
  attributes :secret_token, :average_score, :if => :admin?

  def average_score
    # time consuming operation, which should only be executed if really needed
    object.scores.joins(...).average("SQRT(score * weight)")
  end

  def admin?
    scope.admin
  end
end
```

Or to exclude things which are not required, depending on the given `params` in the HTTP request:
```ruby
class PostsController < ApplicationController
  def index
    @posts = Post.all
    render json: @posts, extra_columns: params[:extra_columns].split(",")
  end
end

class PostSerializer < ActiveModel::Serializer
  attributes :id
  attributes :name, :if => Proc.new { options[:extra_columns].include?("name") }
  attributes :content, :if => Proc.new { options[:extra_columns].include?("content") }
end
```

And another example would be to depend on the object itself:
```ruby
class PostSerializer < ActiveModel::Serializer
  attributes :id, :name
  attributes :url, :if => :published?

  def published?
    object.published_at.present?
  end
end
```

In my current app the JSON in fact would depend on all of these: `current_user` (which is `scope` in AMS), `params` (which is `options` in AMS), as well as the `object` itself.

The DSL using `:if` is just a suggestion but I think it would be pretty idiomatic.

Let me know if I can help! Would be great to see this feature in 0.10.0 again!

beauby:
I would tend to think those are actually controller concerns and that your situation should be resolved by chosing an appropriate serializer within the controller (although I understand this would amount to more code on your end).

groe:
[...]

Issue #741, opened 2014-11-26
Add serialization_scope to README


rafaelcgo:
Just a small suggestion that might help others...

I was struggling with `undefined method 'authenticate' for nil:NilClass` for my requests using the serializer (Rspec).

Turns out all I needed was to add `serialization_scope :view_context` but I only discovered it when watching the amazing Railcast about the gem: http://railscasts.com/episodes/409-active-model-serializers?view=asciicast

I think you guys should add `serialization_scope` documentation to the readme, to make things easier. And refer to the Railscast as well..

bf4:
@rafaelcgo in case you're interested, I'm documenting this in https://github.com/rails-api/active_model_serializers/pull/1252

Thanks for the view_context idea

rafaelcgo:
I'm glad I could help! :clap:

PR #1252, opened 2015-10-07
Add serialization_scope example

bf4:
ref https://github.com/rails-api/active_model_serializers/issues/1245

- [x] add docs to https://github.com/rails-api/active_model_serializers/blob/master/docs/general/rendering.md#scope
- output bugs
  - https://github.com/rails-api/active_model_serializers/issues/1509

### DIFF
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -156,7 +156,98 @@ PR please :)
 
 #### #scope
 
-PR please :)
+Allows you to include in the serializer access to an external method.
+
+It's intended to provide an authorization context to the serializer, so that
+you may e.g. show an admin all comments on a post, else only published comments.
+
+- `scope` is a method on the serializer instance that comes from `options[:scope]`. It may be nil.
+- `scope_name` is an option passed to the new serializer (`options[:scope_name]`).  The serializer
+  defines a method with that name that calls the `scope`, e.g. `def current_user; scope; end`.
+  Note: it does not define the method if the serializer instance responds to it.
+
+That's a lot of words, so here's some examples:
+
+First, let's assume the serializer is instantiated in the controller, since that's the usual scenario.
+We'll refer to the serialization context as `controller`.
+
+| options | `Serializer#scope` | method definition |
+|-------- | ------------------|--------------------|
+| `scope: current_user, scope_name: :current_user` | `current_user` | `Serializer#current_user` calls `controller.current_user`
+| `scope: view_context, scope_name: :view_context` | `view_context` | `Serializer#view_context` calls `controller.view_context`
+
+We can take advantage of the scope to customize the objects returned based
+on the current user (scope).
+
+For example, we can limit the posts the current user sees to those they created:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  attributes :id, :title, :body
+
+  # scope comments to those created_by the current user
+  has_many :comments do
+    object.comments.where(created_by: current_user)
+  end
+end
+```
+
+Whether you write the method as above or as `object.comments.where(created_by: scope)`
+is a matter of preference (assuming `scope_name` has been set).
+
+##### Controller Authorization Context
+
+In the controller, the scope/scope_name options are equal to
+the [`serialization_scope`method](https://github.com/rails-api/active_model_serializers/blob/d02cd30fe55a3ea85e1d351b6e039620903c1871/lib/action_controller/serialization.rb#L13-L20),
+which is `:current_user`, by default.
+
+Specfically, the `scope_name` is defaulted to `:current_user`, and may be set as
+`serialization_scope :view_context`.  The `scope` is set to `send(scope_name)` when `scope_name` is
+present and the controller responds to `scope_name`.
+
+Thus, in a serializer, the controller provides `current_user` as the
+current authorization scope when you call `render :json`.
+
+**IMPORTANT**: Since the scope is set at render, you may want to customize it so that `current_user` isn't
+called on every request.  This was [also a problem](https://github.com/rails-api/active_model_serializers/pull/1252#issuecomment-159810477)
+in [`0.9`](https://github.com/rails-api/active_model_serializers/tree/0-9-stable#customizing-scope).
+
+We can change the scope from `current_user` to `view_context`.
+
+```diff
+class SomeController < ActionController::Base
++  serialization_scope :view_context
+
+  def current_user
+    User.new(id: 2, name: 'Bob', admin: true)
+  end
+
+  def edit
+    user = User.new(id: 1, name: 'Pete')
+    render json: user, serializer: AdminUserSerializer, adapter: :json_api
+  end
+end
+```
+
+We could then use the controller method `view_context` in our serializer, like so:
+
+```diff
+class AdminUserSerializer < ActiveModel::Serializer
+  attributes :id, :name, :can_edit
+
+  def can_edit?
++    view_context.current_user.admin?
+  end
+end
+```
+
+So that when we render the `#edit` action, we'll get
+
+```json
+{"data":{"id":"1","type":"users","attributes":{"name":"Pete","can_edit":true}}}
+```
+
+Where `can_edit` is `view_context.current_user.admin?` (true).
 
 #### #read_attribute_for_serialization(key)
 
--- a/test/action_controller/serialization_scope_name_test.rb
+++ b/test/action_controller/serialization_scope_name_test.rb
@@ -1,63 +1,221 @@
 require 'test_helper'
-require 'pathname'
 
-class DefaultScopeNameTest < ActionController::TestCase
-  class UserSerializer < ActiveModel::Serializer
+module SerializationScopeTesting
+  class User < ActiveModelSerializers::Model
+    attr_accessor :id, :name, :admin
     def admin?
-      current_user.admin
-    end
-    attributes :admin?
-  end
-
-  class UserTestController < ActionController::Base
-    protect_from_forgery
-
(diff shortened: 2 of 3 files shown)


beauby:
This is important but I find the example confusing.

README.md
```diff
@@ -121,6 +121,46 @@ render json: @posts, each_serializer: PostPreviewSerializer
...
+  serialization_scope :current_admin
+
+  def current_admin
+    User.new(id: 2, name: 'Bob', admin: true)
+  end
+
+  def index
+    render json: User.new(id: 1, name: 'Pete', admin: false), serializer: AdminUserSerializer, adapter: :json_api
+  end
+end
```
bf4: @beauby how could the example be better. I was lazy and just cribbed the test, but maybe you have something in mind?

README.md
```diff
@@ -121,6 +121,46 @@ render json: @posts, each_serializer: PostPreviewSerializer
...
+```
+
+So that when we render the `#index` action, we'll get
+
+```json
+{"data":{"id":"1","type":"users","attributes":{"admin?":true}}}
```
shicholas: should this be ``false``?

bf4: excactly not. the value of `admin?` is the value from the Serializer, which decorates the user.  In the AdminUserSerializer, the value of `admin?` is `current_admin.admin` which is true (though maybe not the best code)

bf4: good question

README.md
beauby: @bf4: Could you fill me in on what exactly we are testing here? It feels like we're making sure we can override the scope?

README.md
shicholas: cool thanks for the explanation.

README.md
bf4: Not testing, documenting usage (no overriding)

B mobile phone

beauby: Haha right, I was a bit confused. What do you think about:
```ruby
class PostSerializer < AMS
  attributes :title, :body, :can_edit?

  def can_edit?
    scope.admin?
  end
end
```

bf4: sure, wanna edit? I'm busy for a while

bf4:
ref: https://github.com/rails-api/active_model_serializers/issues/1281#issuecomment-149618096

bf4:
Updated, but the example needs help from someone who's used the feature

bf4:
@rails-api/ams any thoughts on this?

NullVoxPopuli:
I think the example is good. If there are no other objects, I'll merge it

README.md
```diff
@@ -128,6 +128,59 @@ render json: @posts, each_serializer: PostPreviewSerializer
...
+By default, it is the `:current_user` method on the controller.
+
+Upon instantiation, the serializer defines the instance methods:
+- `scope` (with the value of `options[:scope]`) and
+- a method named the value of `options[:scope_name]`
+  (e.g. `current_user`, if not already present), that calls 'scope'.
+
+In the usual case, the serializer instance method has the same name as the controller method.
+e.g. `serializer.current_user` calls `controller.current_user`. The method's name
+will always be the `serialization_scope` defined in the controller, but may be
```
beauby: This sentence is a bit hard to understand (at least for me). What is "the `serialization_scope`"? The parameter to the (class) method call `serialization_scope :foo`? (i.e. `:foo` in this case) I guess not as you talk about the controller.

README.md
```diff
@@ -128,6 +128,59 @@ render json: @posts, each_serializer: PostPreviewSerializer
...
+end
+```
+
+So that when we render the `#edit` action, we'll get
+
+```json
+{"data":{"id":"1","type":"users","attributes":{"name":"Pete","can_edit":true}}}
+```
+
+Where `can_edit` is `current_admin.admin` which is true.
```
beauby: `can_edit?` and `current_admin.admin?`

README.md
bf4: So, here's my dirty secret: this part of the code is a bit hard to understand.. by putting something in master, it forces a few more questions and I'm hoping maybe someone else can may it clearer :)

NullVoxPopuli: :+1:

bf4:
omg, great examples in https://github.com/rails-api/active_model_serializers/tree/0-9-stable#customizing-scope and see `filter` https://github.com/rails-api/active_model_serializers/tree/0-9-stable#attributes

bf4:
Found another related issue: https://github.com/rails-api/active_model_serializers/issues/741

docs/general/serializers.md
```diff
@@ -136,7 +136,97 @@ PR please :)
...
+
+That's a lot of words, so here's some examples:
+
+First, let's assume the serializer is instantiated in the controller, since that's the usual scenario.
+We'll refer to the serialization context as `controller`.
+
+| options | `Serializer#scope` | method definition |
+|-------- | ------------------|
+| `scope: :current_user, scope_name: :current_user` | `:current_user` | `Serializer#current_user` calls `controller.current_user`
+| `scope: :current_admin, scope_name: :current_admin` | `:current_admin` | `Serializer#current_admin` calls `controller.current_admin`
```
bf4: I think there's bugs hiding here

bf4:
I updated with some more thought and comparison to 0.9.  I like the examples better, and now understand the discussion of 'authorization context'

I think we lost the usage of the `scope` option in the controller, need to confirm with tests.

e.g. `<%= react_component('CommunityRouter', posts: ActiveModel::SerializableResource.new(@posts, scope: current_business).serializable_hash ) %>`

Also, looks like the serializer could break if `scope_name` is passed in but not a `scope`.

buren:
I think in general this documentation is good and a well needed addition :) :+1: @bf4 

However I also noticed some problems trying to pass in `scope` in controllers. I had some other problems there, so will need to confirm that that is actually happening.. Will update when/if I can confirm anything.

bf4:
Feel free to make a pr into my branch

B mobile phone

bf4:
@buren I may have found the bug

docs/general/serializers.md
```diff
@@ -136,7 +136,97 @@ PR please :)
...
+    object.comments.where(created_by: current_user)
+  end
+end
+```
+
+##### Controller Authorization Context
+
+In the controller, the scope/scope_name options are equal to
+the [`serialization_scope`method](https://github.com/rails-api/active_model_serializers/blob/d02cd30fe55a3ea85e1d351b6e039620903c1871/lib/action_controller/serialization.rb#L13-L20),
+which is `:current_user`, by default.
```
bf4: The controller defineds

```ruby
      class_attribute :_serialization_scope
      self._serialization_scope = :current_user

    def serialization_scope
      send(_serialization_scope) if _serialization_scope &&
        respond_to?(_serialization_scope, true)
    end
      def serialization_scope(scope)
        self._serialization_scope = scope
      end
    end

```

then serialization resource calls

```ruby
        serializable_resource.serialization_scope ||= serialization_scope
```

notice that `serialization_scope` is actually the return value of sending the `_serialization_scope` method in the controller (if that method exists). 

Those two setters correspond to the serializar options `scope` and `scope_name` respectively, which may also be passed in as params to render

test/action_controller/serialization_scope_name_test.rb
```diff
@@ -1,63 +1,228 @@
...
+        exception_matcher = 'admin?'
+      else
+        exception_class = NameError
+        exception_matcher = /admin|current_user/
+      end
+      exception = assert_raises(exception_class) do
+        get :render_post_with_no_scope
+      end
+      assert_match exception_matcher, exception.message
+    end
```
bf4: Bug in `Serializer._serializer_instance_methods` is causing the scoped method to sometimes be available to `read_attribute_for_serialization` and sometime not, since the public instance methods are cached once, but they change with scopes.

docs/general/serializers.md
```diff
@@ -136,7 +136,97 @@ PR please :)
...
+
+For example, we can limit the posts the current user sees to those they created:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  attributes :id, :title, :body
+
+  # scope comments to those created_by the current user
+  has_many :comments do
+    object.comments.where(created_by: current_user)
```
bf4: is same as `object.comments.where(created_by: scope)`

docs/general/serializers.md
```diff
@@ -136,7 +136,97 @@ PR please :)
...
+In the controller, the scope/scope_name options are equal to
+the [`serialization_scope`method](https://github.com/rails-api/active_model_serializers/blob/d02cd30fe55a3ea85e1d351b6e039620903c1871/lib/action_controller/serialization.rb#L13-L20),
+which is `:current_user`, by default.
+
+Specfically, the `scope_name` is defaulted to `:current_user`, and may be set as
+`serialization_scope :view_context`.  The `scope` is set to `send(scope_name)` when `scope_name` is
+present and the controller responds to `scope_name`.
+
+Thus, in a serializer, the controller provides `current_user` as the
+current authorization scope when you call `render :json`.
```
bf4: There is currently a bug that prevents you from passing `scope` or `scope_name` via render options.

buren:
@bf4 wow great work! Impressed that you found the "special" Rails 4 behaviour :+1:

bf4:
>  Impressed that you found the "special" Rails 4 behaviour
@buren well, it failed tests.. so..

test/action_controller/serialization_scope_name_test.rb
NullVoxPopuli: is the fix for this bug for another PR?

NullVoxPopuli:
:+1:

Issue #1261, opened 2015-10-09
a "test" attribute on model will cause an exception


ryedog:
If you have an attribute called "test" on your model it will throw an exception 

There is a "test" method from the kernel on Serializer so when serializing in the attributes instance method of ActiveModel::Serializer you'll get

```
 ArgumentError:
       wrong number of arguments (0 for 2..3)
     # /home/vagrant/.rvm/gems/ruby-2.1.1/bundler/gems/active_model_serializers-1f08865a10bb/lib/active_model/serializer.rb:156:in `test'
     # /home/vagrant/.rvm/gems/ruby-2.1.1/bundler/gems/active_model_serializers-1f08865a10bb/lib/active_model/serializer.rb:156:in `block in attributes'
```

The source of the test method is

```
ruby-2.1.1-p76 (#<BaseSerializer:0x0000000afe8e40>):0 > show-source self.test

From: file.c (C Method):
Owner: Kernel
Visibility: private
Number of lines: 137

static VALUE
rb_f_test(int argc, VALUE *argv)
{
    int cmd;

    if (argc == 0) rb_check_arity(argc, 2, 3);
    cmd = NUM2CHR(argv[0]);
    if (cmd == 0) {
```

Digging into the implementation of ActiveModel Serializer it's apparent that any attribute with the same name as a method on the Serializer will have similar issues.

bf4:
yup, sorry about that.  do you not have a `test` method on your resource? I've seen similar issues with `select` which is an IO mixin that's in Kernel for some reason, and even a BasicObject has it

bf4:
would it be helpful if we just defined a `test` and `select` method on the Serializer base object that raises if you try to call it but haven't defined it?

beauby:
That sounds like a terrible idea, as we would keep defining many such methods as more similar issues arise.

beauby:
Plus, it does not fix the situation where you have a `select` attribute on your model, and declare it as `attribute :select` in your serializer (as the serializer will see that `#select` is already defined on the serializer, thus not defining it as `object.select`).

bf4:
Dunno, OP hasn't responded.  As to Ruby stdlib gotchas, those are the only two I can think of.  Anyhow, I just meant AMS could be nice and somehow tell you that you've asked for the `select` attribute but it's hitting Kernel.  I agree that I'm not sure that would necessarily be the better solution.  Hence the question :)

beauby:
[...]

PR #1262, opened 2015-10-09
Add inline syntax to override attributes/associations.

beauby:
This PR brings the following syntax for overriding attributes/associations and declaring virtual attributes/associations in a serializer:
```ruby
  attribute :name do
    "#{object.first_name} #{object.last_name}"
  end

  has_many :comments do
    object.comments.limit(10)
  end
end
```
that is currently equivalent to the usual
```ruby
  attribute :name
  def name
    "#{object.first_name} #{object.last_name}"
  end

  has_many :comments
  def comments
    object.comments.limit(10)
  end
end
```
Note that this change is not only cosmetic, as it is a prerequisite to move away from the old way (defining methods on the serializer, which caused many issues when the attribute name was that of a reserved method name (ref #1261)), and refactor the way attributes are handled in a safer way.

Update: added overriding of associations as well + factoring out of the attributes logic.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -65,11 +65,8 @@ module ActiveModel
       serializer.class_attribute :_cache_digest # @api private : Generated
     end
 
-    # Serializers inherit _attributes, _attributes_keys and _reflections.
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
-      inherit_attributes(base)
-      inherit_associations(base)
       caller_line = caller.first
       base._cache_digest = digest_caller_file(caller_line)
       super
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -2,6 +2,7 @@ require 'thread_safe'
 require 'active_model/serializer/collection_serializer'
 require 'active_model/serializer/array_serializer'
 require 'active_model/serializer/include_tree'
+require 'active_model/serializer/attributes'
 require 'active_model/serializer/associations'
 require 'active_model/serializer/configuration'
 require 'active_model/serializer/fieldset'
@@ -12,7 +13,9 @@ require 'active_model/serializer/lint'
 module ActiveModel
   class Serializer
     include Configuration
+    include Attributes
     include Associations
+
     require 'active_model/serializer/adapter'
 
     # Matches
@@ -46,10 +49,6 @@ module ActiveModel
 
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
-      class_attribute :_attributes               # @api private : names of attribute methods, @see Serializer#attribute
-      self._attributes ||= []
-      class_attribute :_attributes_keys          # @api private : maps attribute value to explict key name, @see Serializer#attribute
-      self._attributes_keys ||= {}
       serializer.class_attribute :_cache         # @api private : the cache object
       serializer.class_attribute :_fragmented    # @api private : @see ::fragmented
       serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key
@@ -66,12 +65,12 @@ module ActiveModel
       serializer.class_attribute :_cache_digest # @api private : Generated
     end
 
-    # Serializers inherit _attributes and _attributes_keys.
+    # Serializers inherit _attributes, _attributes_keys and _reflections.
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
+      inherit_attributes(base)
+      inherit_associations(base)
       caller_line = caller.first
-      base._attributes = _attributes.dup
-      base._attributes_keys = _attributes_keys.dup
       base._cache_digest = digest_caller_file(caller_line)
       super
     end
@@ -81,47 +80,6 @@ module ActiveModel
     #     type 'authors'
     def self.type(type)
       self._type = type
-    end
-
-    # @example
-    #   class AdminAuthorSerializer < ActiveModel::Serializer
-    #     attributes :id, :name, :recent_edits
-    def self.attributes(*attrs)
-      attrs = attrs.first if attrs.first.class == Array
-
-      attrs.each do |attr|
-        attribute(attr)
-      end
-    end
-
-    # @example
-    #   class AdminAuthorSerializer < ActiveModel::Serializer
-    #     attributes :id, :recent_edits
-    #     attribute :name, key: :title
-    #
-    #     def recent_edits
-    #       object.edits.last(5)
-    #     end
-    #
-    # @example
-    #   class AdminAuthorSerializer < ActiveModel::Serializer
-    #     attribute :name do
-    #       "#{object.first_name} #{object.last_name}"
-    #     end
-    def self.attribute(attr, options = {}, &block)
-      key = options.fetch(:key, attr)
-      _attributes_keys[attr] = { key: key } if key != attr
-      _attributes << key unless _attributes.include?(key)
-
-      ActiveModelSerializers.silence_warnings do
-        define_method key do
-          if block_given?
-            instance_eval(&block)
-          else
-            object.read_attribute_for_serialization(attr)
-          end
-        end unless method_defined?(key) || _fragmented.respond_to?(attr)
-      end
     end
 
     # @api private
@@ -244,20 +202,6 @@ module ActiveModel
     # Used by adapter as resource root.
     def json_key
       root || object.class.model_name.to_s.underscore
-    end
-
-    # Return the +attributes+ of +object+ as presented
-    # by the serializer.
-    def attributes
-      attributes = self.class._attributes.dup
-
-      attributes.each_with_object({}) do |name, hash|
-        if self.class._fragmented
-          hash[name] = self.class._fragmented.public_send(name)
-        else
-          hash[name] = send(name)
-        end
-      end
     end
 
     protected
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -101,15 +101,25 @@ module ActiveModel
     #
     #     def recent_edits
     #       object.edits.last(5)
-    #     enr
-    def self.attribute(attr, options = {})
+    #     end
+    #
+    # @example
+    #   class AdminAuthorSerializer < ActiveModel::Serializer
+    #     attribute :name do
+    #       "#{object.first_name} #{object.last_name}"
+    #     end
+    def self.attribute(attr, options = {}, &block)
       key = options.fetch(:key, attr)
       _attributes_keys[attr] = { key: key } if key != attr
       _attributes << key unless _attributes.include?(key)
 
       ActiveModelSerializers.silence_warnings do
         define_method key do
-          object.read_attribute_for_serialization(attr)
+          if block_given?
+            instance_eval(&block)
+          else
+            object.read_attribute_for_serialization(attr)
+          end
         end unless method_defined?(key) || _fragmented.respond_to?(attr)
       end
     end
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -27,7 +27,9 @@ module ActiveModel
       end
 
       module ClassMethods
-        def inherit_associations(base)
+        # Serializers inherit _reflections.
+        def inherited(base)
+          super
           base._reflections = _reflections.dup
         end
 
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -13,9 +13,8 @@ module ActiveModel
       DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
 
       included do |base|
-        class << base
-          attr_accessor :_reflections
-        end
+        base.class_attribute :_reflections
+        base._reflections ||= []
 
         extend ActiveSupport::Autoload
         autoload :Association
@@ -28,8 +27,8 @@ module ActiveModel
       end
 
       module ClassMethods
-        def inherited(base)
-          base._reflections = self._reflections.try(:dup) || []
+        def inherit_associations(base)
+          base._reflections = _reflections.dup
         end
 
         # @param [Symbol] name of the association
@@ -39,8 +38,8 @@ module ActiveModel
         # @example
         #  has_many :comments, serializer: CommentSummarySerializer
         #
-        def has_many(name, options = {})
-          associate HasManyReflection.new(name, options)
+        def has_many(name, options = {}, &block)
+          associate(HasManyReflection.new(name, options), block)
         end
 
         # @param [Symbol] name of the association
@@ -50,8 +49,8 @@ module ActiveModel
         # @example
         #  belongs_to :author, serializer: AuthorSerializer
         #
-        def belongs_to(name, options = {})
-          associate BelongsToReflection.new(name, options)
+        def belongs_to(name, options = {}, &block)
+          associate(BelongsToReflection.new(name, options), block)
         end
 
         # @param [Symbol] name of the association
@@ -61,8 +60,8 @@ module ActiveModel
         # @example
         #  has_one :author, serializer: AuthorSerializer
         #
-        def has_one(name, options = {})
-          associate HasOneReflection.new(name, options)
+        def has_one(name, options = {}, &block)
+          associate(HasOneReflection.new(name, options), block)
         end
 
         private
@@ -73,11 +72,15 @@ module ActiveModel
         #
         # @api private
         #
-        def associate(reflection)
+        def associate(reflection, block)
           self._reflections = _reflections.dup
 
           define_method reflection.name do
-            object.send reflection.name
+            if block_given?
+              instance_eval(&block)
+            else
+              object.send reflection.name
+            end
           end unless method_defined?(reflection.name)
 
           self._reflections << reflection
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -1,0 +1,78 @@
+module ActiveModel
+  class Serializer
+    # Defines the attributes to be serialized.
+    #
+    module Attributes
+      extend ActiveSupport::Concern
+
+      included do |base|
+        base.class_attribute :_attributes      # @api private : names of attribute methods, @see #attribute
+        base._attributes ||= []
+        base.class_attribute :_attributes_keys # @api private : maps attribute value to explict key name, @see #attribute
+        base._attributes_keys ||= {}
+      end
+
+      module ClassMethods
+        def inherit_attributes(base)
+          base._attributes = _attributes.dup
+          base._attributes_keys = _attributes_keys.dup
+        end
+
+        # @example
+        #   class AdminAuthorSerializer < ActiveModel::Serializer
+        #     attributes :id, :name, :recent_edits
+        def attributes(*attrs)
+          attrs = attrs.first if attrs.first.class == Array
+
+          attrs.each do |attr|
+            attribute(attr)
+          end
+        end
+
+        # @example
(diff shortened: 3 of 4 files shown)


bf4:
Nice notion!  But it increases complexity for a solved problem.  If you want to prove the problem isn't solved, and this fixes it, use attributes named 'test' and 'select' that fail before the change and work after.

beauby:
@bf4 You're right, in the current state of things, this does not solve the aforementioned problem. However, in order to solve it, we need to get rid of users defining a virtual/overridden attribute as a method on the serializer, which this PR offers a way to. 
The next PR in this series will extract attributes handling into a mixin and store the attributes' values into a hash instead of defining a method on the serializer.

bf4:
I still don't see how this helps in any way.  Show me! ;)

B mobile phone

beauby:
@bf4 See #1263. Once we remove the `public_send` in `attribute_value`, no more clash with existing methods.

bf4:
Well, I still want to hold on this

B mobile phone

beauby:
@bf4 Sure, let's talk about it when you're back.

test/serializers/attribute_test.rb
```diff
@@ -71,6 +71,21 @@ def id
...
+
+      PostWithVirtualAttribute = Class.new(::Model)
+      class PostWithVirtualAttributeSerializer < ActiveModel::Serializer
+        attribute :name do
+          "#{object.first_name} #{object.last_name}"
+        end
+      end
+
+      def test_virtual_attribute_block
+        post = PostWithVirtualAttribute.new(first_name: 'Lucas', last_name: 'Hosseini')
```
bf4: in this case, you might want to test `post.attributes` since that's the immediate case.. tests less layers

beauby: Agreed

lib/active_model/serializer/attributes.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        def attributes(*attrs)
+          attrs = attrs.first if attrs.first.class == Array
+
+          attrs.each do |attr|
+            attribute(attr)
+          end
+        end
+
+        def attribute(attr, options = {}, &block)
```
beauby: Added the `&block` param.

lib/active_model/serializer/attributes.rb
```diff
@@ -0,0 +1,59 @@
...
+          key = options.fetch(:key, attr)
+          _attributes_keys[attr] = { key: key } if key != attr
+          _attributes << key unless _attributes.include?(key)
+
+          ActiveModelSerializers.silence_warnings do
+            define_method key do
+              if block_given?
+                instance_eval(&block)
+              else
```
beauby: Added the `if block_given?; instance_eval(&block); else` clause.

lib/active_model/serializer/attributes.rb
```diff
@@ -0,0 +1,59 @@
+module ActiveModel
+  class Serializer
+    # Defines the attributes to be serialized.
+    #
+    module Attributes
```
beauby: This module is directly extracted from the serializer, no modifications to the code have been made, except for:
 1. the definition of an `inherit_attributes` method, rather than overriding `inherited` as in the `Associations` module, which made it impossible for the two modules to define the inherited properties
 2. the optional block parameter to the `attribute` method.

lib/active_model/serializer/associations.rb
```diff
@@ -28,8 +27,8 @@ class << base
       end
 
       module ClassMethods
-        def inherited(base)
-          base._reflections = self._reflections.try(:dup) || []
+        def inherit_associations(base)
```
beauby: Overriding the `inherited` method does not work when more than one modules do it.

lib/active_model/serializer.rb
```diff
@@ -57,9 +56,9 @@ def self.digest_caller_file(caller_line)
     end
 
     def self.inherited(base)
+      inherit_attributes(base)
+      inherit_associations(base)
```
bf4: super weird code tells us something isn't right in the design

lib/active_model/serializer.rb
```diff
@@ -168,18 +147,6 @@ def json_key
...
-
-      attributes.each_with_object({}) do |name, hash|
-        if self.class._fragmented
-          hash[name] = self.class._fragmented.public_send(name)
-        else
-          hash[name] = send(name)
-        end
-      end
-    end
-
```
bf4: funny, I see this as **the** fundamental concern of the serializer and probably the only thing I **wouldn't** remove (though maybe I'd extract method object)

lib/active_model/serializer/associations.rb
bf4: sounds like this should be a bug fix, not part of a feature

bf4:
I'm still very :-1: on this. Sorry.

lib/active_model/serializer.rb
beauby: What wasn't right in the design was that the `Associations` module was overriding the `inherited` method of the `Serializer` class, which worked fine as long as only one module was doing that. I'm all up for suggestion as to making this cleaner though.

lib/active_model/serializer.rb
beauby: I personally don't think associations are any less fundamental to serializers than attributes. Plus extracting semantically delimited code to a module does not make it less fundamental in my opinion.

lib/active_model/serializer/associations.rb
beauby: Agreed. I just happened to stumble upon this problem while writing this PR.

beauby:
@bf4 Any specific reason?

bf4:
I don't think it's good for AMS.

beauby:
Would you care to give
> any specific reason?

beauby:
Because as you can imagine, I *do* think it's good for AMS, so it's kind of a Mexican standout.

bf4:
except others are welcome to show up and approve :)

lib/active_model/serializer/associations.rb
NullVoxPopuli: I'm always ok with small bug fixes here and there. Esp if they are stumbled upon, and aren't documented anywhere. Better to catch it early before it becomes an actual problem.

test/serializers/attribute_test.rb
NullVoxPopuli: I'd like to see this too.

lib/active_model/serializer/attributes.rb
```diff
@@ -0,0 +1,59 @@
...
+          end
+        end
+
+        def attribute(attr, options = {}, &block)
+          key = options.fetch(:key, attr)
+          _attributes_keys[attr] = { key: key } if key != attr
+          _attributes << key unless _attributes.include?(key)
+
+          ActiveModelSerializers.silence_warnings do
+            define_method key do
```
NullVoxPopuli: this absolutely needs to be document above the method signature with some examples. 
I am a HUUUGE fan of metaprogramming, but it can be a pain in the butt if people aren't expecting it. So, having some example syntax in a method yardoc would be fabulous.

NullVoxPopuli:
ok, so --

I am a fan of, especially for the json api, how there are numerous things that can be configured via blocks. I think that's very consistent, and doesn't leave our users guessing as to which syntax to use.

I do agree that the way module inheritance has been going has been kinda sketchy, but I believe that is a bigger overall architecture discussion for another time.

@beauby, you have my :+1: 

@bf4, do have any specific concerns about this for AMS?

I do know some static analysis tools will complain if there is too much code at the class level, and prefer things to be done in methods (as they are more easily testable, stubbable, greppable etc)

lib/active_model/serializer/attributes.rb
beauby: Agreed, although this code is already in AMS. I just added the `if block_given?; ...; else`.

NullVoxPopuli: Never hurts to add documentation. Code pre-existing isn't an excuse to *not* add docs. :-\

beauby: Sure, as I said, I agree. I was just making clear that I didn't just throw loads of metaprogramming magic without comments.

NullVoxPopuli: oh :-) :+1:

bf4: 1) we should be conservative in adding new interfaces because then we need to support it
2) if we need this feaure, why? If we don't need it, what are the trade offs of adding it? Not adding it?
3) how does it affect ease of understand usage? Ease of maintenaince? Complexity of either?
4) is it extending existing behavior or modifying it? Could it bd done another way?
5) how do you see this feature evolving?  What are its limitations?  Can it be easily customized? Eg adapter no longer need to be in thd ams namespace or even inherit the base.  So, it's easy to extend (though the boundaries could be clearer and the interfaces more consitent.  Am::s should probably be called a resource and implement the serializers::json interface
(Writing in phone; apologies for anything unclear)
>

beauby:
@bf4:
> 1) we should be conservative in adding new interfaces because then we need to support it

I agree. We should also provide the best possible interfaces to our users.

> 2) if we need this feaure, why? If we don't need it, what are the trade offs of adding it? Not adding it?

This feature has two main benefits:
 + When overriding attributes/associations (or declaring virtual attributes/associations), it forces the user to have the redefinition next to the definition, thus encouraging code clarity.
 + If we want to move away from dynamically declaring methods on the serializer in order to access the object's properties (which causes clashes with existing method names, and hard to understand errors) , we need to provide a different way to override attributes/associations (which this feature does).

> 3) how does it affect ease of understand usage? Ease of maintenaince? Complexity of either?

As stated above, I believe it improves ease of usage. Complexity and ease of maintenance are hardly increased.

> 4) is it extending existing behavior or modifying it? Could it bd done another way?

Currently, it is extending existing behavior. It could be done in an other way.

> 5) how do you see this feature evolving?  What are its limitations?  Can it be easily customized? 

I'm not sure about what you mean by evolving, limitations and customization in this setting.

> Eg adapter no longer need to be in thd ams namespace or even inherit the base.  So, it's easy to extend (though the boundaries could be clearer and the interfaces more consitent. 

Not sure I understand. This PR does not deal with adapters in any way.

> Am::s should probably be called a resource 

Sounds sensible

> and implement the serializers::json interface

The `ActiveModel::Serializers::JSON` interface is just `as_json/from_json`. What do you mean?

lib/active_model/serializer/attributes.rb
beauby: Fixed [the situation](https://github.com/beauby/active_model_serializers/commit/af8fc7e80ed00f67df68ba0f5ad15efcd78d2508). Reverted to overriding `inherited` at module-level, but calling `super` each time in order to decorate the serializers.

joaomdmoura:
Hey everyone, first of all, great work @beauby.
I understand @bf4's point.

We are not solving "defining methods on the serializer" problem **yet**, IMO the main achievement here is aesthetics. Don't take me wrong, it's good aesthetics, and I like it, in favor of it :smile:. 

The concern that I share with @bf4 is,**if** there is a trade-off involved and how big is it. Any piece of code might hunt us down later, even if it seems simple, that's why I think @bf4 made the following question:

> if we need this feature, why?

No. But it's a nice feature, and I agree with @beauby:

> When overriding attributes/associations (or declaring virtual attributes/associations), it forces the user to have the redefinition next to the definition, thus encouraging code clarity.

I'd say to hold overall small/medium features like this one, and focus on the `mandatory` ones first.

About this feature itself, it LGTM, we could merge it now or later after mandatory stuff, but would be nice to know more about this @bf4:

> I don't think it's good for AMS.

Maybe we can talk about it over slack and avoid a huge thread in here :smile:

bf4:
So, I think I would be more :+1: on this DSL if it were optional. i.e. 

Given

```
  attribute :testee, key: :test
end
```
right now the `attribute` macro does one thing and expects another:

1. It defines an instance method `test` on the UserSerializer (if not already defined, hence the bug for e.g. Kernel#test)
2. When the adapter calls `user_serializer.attributes` it calls `:testee` on the serializer or fragment cache

The UserSerializer also makes available the the model it wraps as `object` so that devs can define their own methods for the attributes to refer to.  (serialization_scope is another method that provides an instance method, current_user. We'll be adding a serialization_context method soon, too, as well.)

The problem with this approach is that it requires us to be vigilant not to define instance methods on ActiveModel::Serializer that might also be attributes

So, the general need is to avoid this conflict.

One solution would be for all AMS methods to be prefixed with something like `_` to minimize the chances of collision and maybe have a blacklist for undesired methods.

The type of solution in this PR is reasonable, in that it can solve the problem by removing the need to define methods on the serializer instance at all.  

Every attribute declaration would now define an attribute object that the class keeps track of.  This object may be a method object or a Proc.  So then, when we calls attributes, we call one of these method objects.  We could no longer override or define methods on the instance, but we could define them inline in a Proc/Lambda/Method object etc.

This would be a big breaking change but could have big benefits, depending on how it's implemented.

Now, we can exchange the innards of `Serializer.attribute(attr, options = {})`

```ruby
          _attributes_keys[attr] = { key: key } if key != attr
          _attributes << key unless _attributes.include?(key)
          define_method :test do
              object.read_attribute_for_serialization(:testee)
            end
          end unless method_defined?(:test) || _fragmented.respond_to?(:testee)
```

for

```ruby
          key = options.fetch(:key, attr, &block)
          _attributes[key] = block || proc do 
            object.read_attribute_for_serialization(:testee)
          end
```

which is then used in `Serializer#attributes` similarly

```ruby
      def attributes
        attributes = self.class._attributes.dup

          if self.class._fragmented
            hash[name] = self.class._fragmented.public_send(name)
          else
            hash[name] = send(name)
          end
        end
      end
```

becomes

```
      def attributes
        self.class._attributes.each_with_object({}) do |key, attributes|
          if self.class._fragmented
            attributes[key] = self.class._fragmented.public_send(key)
          else
            attributes[key] = _attributes.call
          end
        end
      end
```

And maybe we can preserve the interface by introducing an `Attribute = Struct.new(:callable) do def call; callable.call; end; end`

Sot that we can define the attribute method as 

```ruby
_attributes[:test] = Attribute.new(block || ->{  object.read_attribute_for_serialization(:testee) })
```

and call it as

```ruby
    self.class._fragmented ? _self.class._fragmented.public_send(key) : self.class._attributes[key].call
  end
  def attributes
    @attributes ||= self.class.attributes.keys.each_with_object({}) do |key, attributes|
      attributes[key] = read_attribute_for_serialization(key)
    end
  end
```

which is super nice because the our `Serializer` implements the [ActiveModel::Serialization](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb) ([JSON](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/json.rb)) interface and is the basically a serialization formatter for the the record. ref: https://github.com/rails-api/active_model_serializers/pull/1301 and https://github.com/rails-api/active_model_serializers/issues/1298

So, a dev that wanted to define a custom method on the Serializer would either

```ruby
attribute :test, -> { object.testee }
# or
attribute :testee, key: :test
# or
_attributes[:test] = Attribute.new(->{ object.read_attribute_for_serialization(:testee) })
```

but that's not quite what is in this PR and doesn't require the concerning :kissing_smiling_eyes: removal of attributes methods to a module since they are now even more fundamental to the class.

:)

(note: written without proofreading, so some details might not mesh quite right, but should still make sense.)

beauby:
That's very close to the following PR I mentioned at the beginning of the conv (except it handles only blocks). I fail to see what you don't feel confortable with here (except for extracting Attributes into its own module, which we could debate about in an other PR).

bf4:
like I said, it's pretty similar, but I don't like passing the block all over the place.  I'd like a more surgical change.

bf4:
Also, what I described solves a problem. It's more than just sugar

beauby:
@bf4 As I stated in the PR description, this is only part 1 of a 2 part PR:
> Note that this change is not only cosmetic, as it is a prerequisite to move away from the old way (defining methods on the serializer, which caused many issues when the attribute name was that of a reserved method name (ref #1261)), and refactor the way attributes are handled in a safer way.

Plus, at the end of the day, this PR is passing a proc around, not a block.

beauby:
So to sum up my vision of things here:
 + I'd like to introduce this inline block syntax to override attributes and associations (and declare virtual ones)
 + Because it promotes cleaner serializers for the users (all the logic for one attribute/association kept in one place of the file)
 + And also so that we can remove this nasty dynamic method generation for each attribute/association, which leads to very surprising errors (c.f. mentioned issues)
 + So that we're left with clean serializer classes, on which all the methods defined are AMS ones, and possibly some user-defined which could be used for
 + conditionally including/excluding attributes/associations (the proposed `include_if` / `exclude_if` options, which would only take symbols (referring to user-defined methods), no lambdas)
 + conditionally chosing the serializer for an association (the proposed extension of the `serializer` options which could take a String, a Constant, or a *Symbol* (referring to a user-defined method).

bf4:
>  passing a proc around, not a block.

I'm not sure I understand the significance of the distinction. I don't follow. It's a callable object. Proc, Lambda, Method are all objects that are all callable.

beauby:
You said 
>  I don't like passing the block all over the place

I said 
> this PR is passing a proc around, not a block.

bf4:
In any case, I think the change in this PR could be much smaller, solve the problem ( no need for #1263 ), and actually improve the design and clarity of the code, as I discussed in https://github.com/rails-api/active_model_serializers/pull/1262#issuecomment-151545353

Still needs work, IMHO. Let's focus this PR on defining/encapsulating attribute conversions at the class level.  Let me know if you want to pair.

bf4:
Where are we on this? want me to submit a PR to your branch?

beauby:
Will update soon.

bf4:
awesomes

beauby:
@bf4 See 1311 which covers half of this PR (the associations part).

lib/active_model/serializer/associations.rb
```diff
@@ -13,9 +13,8 @@ module Associations
       DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
 
       included do |base|
-        class << base
-          attr_accessor :_reflections
-        end
+        base.class_attribute :_reflections
```
bf4: did you intend to allow instance reader/writers? did you intend to make it inheritable?

lib/active_model/serializer.rb
```diff
@@ -83,37 +79,6 @@ def self.type(type)
...
-    # @example
-    #   class AdminAuthorSerializer < ActiveModel::Serializer
-    #     attributes :id, :name, :recent_edits
-    def self.attributes(*attrs)
-      attrs = attrs.first if attrs.first.class == Array
-
-      attrs.each do |attr|
-        attribute(attr)
-      end
-    end
```
bf4: wrote a bunch but github seems to have lost it when I clicked on the tab to get the link to https://github.com/rails-api/active_model_serializers/pull/1262#issuecomment-151545353 re: a latent Attribute class and not removing the attributes api definition from the serializer or changing the behavior of _reflections or _attributes or _attribute_keys 

gonna tweet about it.

tl;dr I'd rather discuss adding a new superclass (the Attributes mxin) separately from adding new behavior (removing the object instance method in favor of defining attribute and association mappings in the attribute / association declarations)

beauby: You'll be delighted to know that I'm doing exactly that as we speak. As said, this is only half the other PR (only the association related stuff).

bf4: :heart: :smile: :heart:

bf4:
@beauby I put together something somewhat minimal based on your work here https://github.com/rails-api/active_model_serializers/pull/1356 (which includes your commits)

bf4:
Closed by https://github.com/rails-api/active_model_serializers/issues/1356

Issue #671, opened 2014-10-02
Versioning ActiveModel::Serializer


jeromeboe:
I'm facing some issues with versioning.

**Controllers**

In `app/controllers/v1/contracts_controller.rb`
```ruby
module V1
    class ContractsController < ApiController

        def index
            @contracts = Contract.all
            render json: @contracts
        end

    end
end
```

In `app/controllers/v2/contracts_controller.rb`
```ruby
module V2
    class ContractsController < ApiController

        def index
            @contracts = Contract.all
            render json: @contracts
        end

    end
end
```

**Serializers**

In `app/serializers/v1/contract_serializer.rb`
```ruby
class ContractSerializer < ActiveModel::Serializer
    attributes :id
end
```

In `app/serializers/v2/contract_serializer.rb`
```ruby
class ContractSerializer < ActiveModel::Serializer
    attributes :id, :name
end
```

Whether I call the route `/v1/contracts` or `/v2/contracts`, the rendered json include the contract name, which means that the serializer in v2 seems to be always called.

FYI, I added `config.autoload_paths += Dir[Rails.root.join('app', 'serializers', '**/')]` in `config/application.rb`

brancusi:
+1

chadwtaylor:
+1

gwagener:
Have you tried adding modules `V1` and `V2` to your serializers so they match the directory structure? I did something similar and I can't remember if I then had to explicitly specify the serializer in the `render` call, but I got it working.

formigarafa:
as @d3chapma said lots of versioning issues have being closed and mostly rejected (if not all). And I believe that there is something missing in all discussions I have seen so far: who defines the resource name?

After all the AMS is responsible to serializer a model for a controller. However, Who owns the namespace? What if the model is also namespaced? Should we search for a serializer accordingly with the controller namespace/name or model namespace/name?

Imagine a admin/v1/media_controller rendering media/photo models.

what is the path resolution for these serializer?

media/photo serializer?
admin/v1/media/serializer?
or even admin/v1/media/photo serializer?

I believe there is one piece missing in this puzzle. (like a view object or a presenter)

controller -> *something* -> serializer -> model
or
controller -> serializer -> * something * -> model

jejacks0n:
[...]

Issue #562, opened 2014-05-14
There needs to be a way to create versioned serializers.


NullVoxPopuli:
I used to use the presenter pattern, which I've found to be very very helpful. But this serializer gem seemed to have all of the functionality I had implemented with presenters, and then additionally relationships.  The reason I was using presenters is because of how we versioned the API. Over time, we've changed the names of attributes and models. 

The presenter file layout was this:

```
app/presenters/api
/v1
  document_presenter.rb
/v2
  document_presenter.rb
```

and in those would be selected based on the the controller route (I know people like to use http headers for api versions, but that's not that hard to do, and this was easier at the time).

The big difference between the api versions is how json collections were rendered.

There are two common formats:

format 1:

```
{
    "resource" : [
          { attributes for this resource }, { attributes for this resource }
     ]
}
```

and 

format 2:

```
[
    "resource": { attributes for this resource },
    "resource": {attributes for this resource }
]
```

format 1 is the more prefered way now a days, which makes sense - it's cleaner, less verbose, and Ember.js uses it.  So that's what we are trying to migrate towards. The ActiveModel Serializers make format 1 really easy.  However, since Serializers don't support API versioning, we are kind of stuck between a rock and a hard place with our older clients' integrations. 

I've been trying to massage the Active Model Serializers for some time now, and came to realize that that I can't override (or don't know how to) the methods that I need to override -- due to the fact that ActiveModel Serializers are all singleton classes. The custom serializers are instances of subclasses of those singletons - which seems reeaaaly weird to me. 

Anywho, this is where I am with trying to extend ActiveModel Serializers to allow different kinds of collection / array json formats.

```
class ActiveModel::Serializer
  class << self

    # Used internally to create a new serializer object based on controller
    # settings and options for a given resource. These settings are typically
    # set during the request lifecycle or by the controller class, and should
    # not be manually defined for this method.
    def build_json(controller, resource, options)
      default_options = controller.send(:default_serializer_options) || {}
      options = default_options.merge(options || {})

      serializer = serializer_for(path: options[:prefixes].first, resource: resource)

      serializer = options.delete(:serializer) || serializer ||
        (resource.respond_to?(:active_model_serializer) &&
         resource.active_model_serializer)

        return serializer unless serializer

      if resource.respond_to?(:to_ary)
        unless serializer <= ActiveModel::ArraySerializer
          raise ArgumentError.new("#{serializer.name} is not an ArraySerializer. " +
[...]

Issue #539, opened 2014-04-02
render json is not finding namespaced model serializer


tboyko:
I have a PORO that includes `ActiveModel::Serialization`. My controller looks like this, for the sake of example (I know that it's nonsensical at the moment):

```
def show
    device = Api::Device.new
    device.name = "Test"

    render json: device
end
```

I also have a serializer in `app/serializers/api/device_serializer.rb` called `Api::DeviceSerializer`.

The above example renders without the influence of the serializer. If I specify `serializer: Api::DeviceSerializer` it renders as expected. What am I missing here?

PR #1265, opened 2015-10-10
Add controller lookup.

beauby:
This PR adds the controller's namespace to the lookup chain, such that the following works:
```ruby
class User < ActiveModel::Base
  ...
end

module Api
  module V1
    class UsersController < ApplicationController
      def index
        render json: User.all
      end
    end
      ...
    end
  end
end
```

### DIFF
--- a/lib/active_model/serializable_resource.rb
+++ b/lib/active_model/serializable_resource.rb
@@ -41,7 +41,7 @@ module ActiveModel
       @serializer ||=
         begin
           @serializer = serializer_opts.delete(:serializer)
-          @serializer ||= ActiveModel::Serializer.serializer_for(resource)
+          @serializer ||= ActiveModel::Serializer.serializer_for(resource, serializer_opts.slice(:_controller_class_name))
 
           if serializer_opts.key?(:each_serializer)
             serializer_opts[:serializer] = serializer_opts.delete(:each_serializer)
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -25,6 +25,7 @@ module ActionController
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
         options[:adapter] = false
       end
+      options[:_controller_class_name] = self.class.name
       serializable_resource = ActiveModel::SerializableResource.new(resource, options)
       if serializable_resource.serializer?
         serializable_resource.serialization_scope ||= serialization_scope
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -92,7 +92,7 @@ module ActiveModel
       elsif resource.respond_to?(:to_ary)
         config.array_serializer
       else
-        options.fetch(:serializer) { get_serializer_for(resource.class) }
+        options.fetch(:serializer) { get_serializer_for(resource.class, options[:_controller_class_name]) }
       end
     end
 
@@ -112,7 +112,7 @@ module ActiveModel
     end
 
     # @api private
-    def self.serializer_lookup_chain_for(klass)
+    def self.serializer_lookup_chain_for(klass, controller_class_name = nil)
       chain = []
 
       resource_class_name = klass.name.demodulize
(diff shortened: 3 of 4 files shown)


NullVoxPopuli:
Nice!! I'm really looking forward to this. 'automatic' api versioning will be a huge help to a lot of people.

lib/active_model/serializable_resource.rb
```diff
@@ -41,7 +41,7 @@ def serializer
       @serializer ||=
         begin
           @serializer = serializer_opts.delete(:serializer)
-          @serializer ||= ActiveModel::Serializer.serializer_for(resource)
+          @serializer ||= ActiveModel::Serializer.serializer_for(resource, serializer_opts.slice(:_controller_class_name))
```
NullVoxPopuli: I think passing the controller name to serializer_for seems weird. Would just passing the namespace make more sense? so that way you don't need to deconstantize that later?

NullVoxPopuli:
related: #886, #671, #562 (@beauby, I actually had a version of this going a long time ago, before that project gave up on the idea of having a versioned API), #539, #499, #381, #144, #108

bf4:
So, is the controller being used as a serializer lookup hint?  If so, I'd rather allow people to add 'hints' (I think this is was Grape calls them).  I think passing the controller class all the way through AMS has pretty low value, since it adds a lot of noise for what could be resolved by specifying the class name.  

I'd rather work on the JSON API unless there's a lot of pressure for something like this.

beauby:
@bf4 Agreed, I made this PR because I knew how to do it without much effort, but it's not among the top priority stuff. 
The one use-case I think this is really helpful is when handling two versions of a same API (say in `Api::V1` and `Api::V2`) where the models do not change, but the serializers do.

bf4:
Yeah, def something to figure out, maybe with a serializer version param or namespace param (which could be derived from controller or specufied)

NullVoxPopuli:
@bf4 could you elaborate?

I really like this approach, as it allows for a mirrored file/folder structure of api versions between the 'serializers' and controllers

bf4:
Yeah, as controllers mirror resource routes, i agree serializers should reflect those paths... 

Let's look at how other libs do it

NullVoxPopuli:
Do you know of other libs that do this?

I think other than how the namespace (re: not the namespace, but the controller class) is passed to the serializable resource, the implementation is pretty good.

NullVoxPopuli:
needs rebase

beauby:
@NullVoxPopuli I'll eventually rebase this if we end up deciding to include it. Currently, it is more meant as a proof of concept than anything else.

NullVoxPopuli:
Well, my vote is for including it :-)

beauby:
I wouldn't mind including it either, but I'd also like for one of @hut8's (I believe) PRs that creates a `context` object, in which we could stick the controller class so that this feature needs less hacking.

NullVoxPopuli:
fair enough :-)

tchak:
I would prefer a way to pass in an api version... I feel like the lookup (`serializer_for`) should take the whole options hash, in order to make possible some advanced use cases.

NullVoxPopuli:
there are many ways to handle api versioning.

Some people like to have it namespaced in their routes, others like to have the version specified in a header. 

We may need to support both.

bf4:
@beauby This PR should be superseded by https://github.com/rails-api/active_model_serializers/pull/1252

beauby:
@bf4 I don't think so, as you don't want to *replace*, say the `current_user` with `current_controller_class` or something. Instead, I think it would be nice to rebuild this PR on top of @trak's PR introducing the `serialization_context` object.

bf4:
@beauby yeah, I pasted the wrong link.. edited comment above to refer to https://github.com/rails-api/active_model_serializers/pull/1289#issuecomment-152555999

bf4:
Should this PR still be open?

elfassy:
Will this get merged? Or is there something else taking its place?

beauby:
@elfassy This PR will probably not get merged, as it was meant as a proof of concept and much has changed in the meantime. I still think the feature would be useful, and would support a PR doing this in a clean way, or maybe issue one myself when I have a bit of free time.

beauby:
Closing in favor of #1436.

PR #1266, opened 2015-10-13
add if/unless options to attribute

LcpMarvel:


lib/active_model/serializer.rb
```diff
@@ -54,10 +57,11 @@ def self.type(type)
     end
 
     def self.attributes(*attrs)
+      options = attrs.last.class == Hash ? attrs.pop : {}
       attrs = attrs.first if attrs.first.class == Array
 
       attrs.each do |attr|
-        attribute(attr)
+        attribute(attr, options)
```
bf4: I don't think we should support any options for `attributes`. That's always just a list.  We use `attribute` for an attribute with options

LcpMarvel: You want to use 
```ruby
attribute :id, if: :admin?
attribute :titile, if: :admin?
```
instead of `attributes :id, :title, if: :admin?` ?

bf4:
Thanks for this. This is a great feature!  And has tests!  

I personally don't think this is a good thing for AMS to support right now. I also don't think the overhead of supporting in on `attributes` is worth it right now.  I'd be more in favor of a less invasive implementation that used a sentinel value to determine if an attributes should be excluded or a whitelist is being used., e.g. something like

```ruby
attributes :banana, :tofu, :burger
def banana
  object.banana || ExcludedAttribute # can't use nil since that's a valid value
  # or 
  object.banana || ExcludedAttribute.new(:banana) # can't use nil since that's a valid value
  # or
  object.banana || :excluded_attribute
  # or 
  condition ? object.banana : :excluded_attribute
end

def tofu
  condition ? object.tofu : WhitelistedAttribute.new(object.tofu) # dunno, just thinking here
end
```

such that the if banana is ExcludedAttribute, attribute keys will be tofu and burger. If tofu is WhitelistedAttribute, then only tofu is a WhitelistedAttribute, then only other whitelisted attribute keys are used, so only tofu will be in the attributes. That way the feature just requires a value-check in the attributes method and the condition logic is easy to use, and maybe could lead to the suggested interface.

Also, I prefer `only`/`except` like Rails uses for this sort of thing

beauby:
Although I am currently not convinced which way is better, it seems lots of people are expecting a feature similar to this one.
@bf4, out of curiosity, how would you see an `only` / `except` system work in this case?

bf4:
@beauby you're right about only/except.  I must have been thinking of something else, like `fields` which is a similar feature, params-driving, and is filtering only

tuvistavie:
Would this work for associations as well? For example:

```
has_one :profile, if: :admin?
```

I think it would be more consistent to support both with the same syntax.

> You want to use
>
> ```
> attribute :id, if: :admin?
> attribute :titile, if: :admin?
> ```
> instead of `attributes :id, :title, if: :admin?` ?

`%i(id title).each { |k| attribute k, if: :admin? }` would do the job without having to make
the implementation more complex than it needs.

lib/active_model/serializer.rb
```diff
@@ -171,8 +186,31 @@ def attributes
       end
     end
 
+    def useless_attributes_keys
```
beauby: `excluded_attribute_keys` might be a better name here.

lib/active_model/serializer.rb
```diff
@@ -171,8 +186,31 @@ def attributes
...
+      self.class
+        ._attributes_conditions
+        .dup
```
beauby: Why the `.dup`?

lib/active_model/serializer.rb
```diff
@@ -171,8 +186,31 @@ def attributes
...
+      end.flatten
+    end
+
     protected
 
     attr_accessor :instance_options
+
+    private
+
+    def judge_attribute(condition)
```
beauby: `eval_attribute_condition` might be a better name here.

lib/active_model/serializer.rb
```diff
@@ -171,8 +186,31 @@ def attributes
...
+      (condition.is_a?(Symbol) && send(condition)) ||
```
beauby: This should be rewritten in a clearer form, something like:
```ruby
switch
when condition.is_a?(Symbol)
  send(condition)
when condition.respond_to?(:call)
  condition.call
end
```

lib/active_model/serializer.rb
```diff
@@ -171,8 +186,31 @@ def attributes
...
+        (condition.respond_to?(:call) && condition.call)
```
beauby: Simply calling `condition.call` here will not work as intended, as the `object` and `current_user` won't be available inside the proc/lambda. You might want to do something like `instance_eval(&condition)`.

lib/active_model/serializer.rb
```diff
@@ -66,6 +70,17 @@ def self.attribute(attr, options = {})
       _attributes_keys[attr] = { key: key } if key != attr
       _attributes << key unless _attributes.include?(key)
 
+      [:if, :unless].each do |switch|
+        next unless options.key?(switch)
+        condition = { switch => options[switch] }
+
+        if _attributes_conditions[condition]
```
beauby: Why not simply store a pair `[Symbol, <Symbol, Proc>]` for each key?

lib/active_model/serializer.rb
beauby: @bf4 I'm not sure I understand why we would not want to support options for `attributes`. Could you elaborate?

beauby:
A bit more work needed, but I'm personally favorable to this implementation. People are clearly expecting such a feature, and this appears to me as the sanest way to go around it.

lib/active_model/serializer.rb
bf4: lI believe you mean

```ruby
case condition
when Symbol then send(condition)
else condition.call
end
```

lib/active_model/serializer.rb
bf4: value to end-user is minimal, cost to library is relatively high

lib/active_model/serializer.rb
beauby: You are right about the `case`, but I believe there should be an `else` clause with a `warn "Invalid condition #{condition} for attribute #{blah}."` in order to avoid weird error messages.

lib/active_model/serializer.rb
beauby: I would usually agree, but since we already support variable number of arguments *and* array of symbols as parameters to `attributes`, I don't believe the added cost is that high.

bf4:
IMHO, this needs to be much simpler. It introduces a lot of complexity (in understanding and maintaining the code, in addition to vectors for bugs) for what could be solved either by polymorphism or by some non-nil value that means 'remove this key'.  A non-nil value that means 'include this key' is more complicated to consider.

beauby:
The non-nil value that means 'exclude this key' puts too much of a burden on the user I think. Could you elaborate on the polymorphism solution? Do you mean subclassing the serializer into various flavors? If so, although I would personally go this way in my projects, several users have expressed their need for conditionally excluding many attributes depending on different conditions, which is a situation in which the polymorphism approach does not scale well (it takes an exponential number of serializers).

bf4:
I don't think any library should do everything

beauby:
@bf4 True, but since this library used to do exactly that, I think we need a better reason than that.

bf4:
Just being philosophical here, but I'm not sure that a feature having been
removed from a library is an argument that it belongs

LcpMarvel:
I had updated the code. :yum:

mrinterweb:
I was using the filter feature to protect attributes in my API. Now that it is gone in 0.10, is there a way to omit certain attributes? Am I supposed to have two serializers one with the attributes and one without? Should I have three serializers to replace this functionality, a base class serializer that the protected and non-protected serializers inherit from? I don't understand why we can't have `filter` or something like it again. Filtering out certain attributes is absolutely essential.

I need a solution to get this type of functionality back.

```ruby
  attributes :id, :name, :email, :settings, :info, token
  def filter(keys)
    unless object == scope
      keys -= [:settings, :email, :token]
    end
    keys
  end
end
```

mrinterweb:
I found a solution that kind of works. My solution does not remove the attribute, but instead nullifies the value. I don't think this is a pretty solution, but it will have to do for now.

```ruby

  [:email, :settings, :token].each do |a|
    define_method(a) do
      object == scope ? object[a] : nil
    end
  end

  attributes :id, :name, :email, :settings, :info, :token
end
```

PR #1260, opened 2015-10-09
Serialization and Cache Documentation

bf4:
-  Document Serializer and FragmentCache

(other work extracted into #1271 #1272)

PR #1271, opened 2015-10-15
Handle no serializer source file to digest.

### LINKED ISSUE #1176: no implicit conversion of nil into String
Got this error on rc3 but not on rc2 and I do not do anything fancy here:

```
  attributes :id
end
```

```
class Api::V2::PostsController < ApplicationController
  def index
    post = Post.all

    render json: post
  end
end
```

Application trace:
```
app/serializers/post_serializer.rb:1:in `<top (required)>'
app/controllers/api/v2/posts_controller.rb:5:in `index'
```

I get the same error on every adapter

bf4:
Handle no serializer source file to digest; output warning. Closes #1176

Outputs warning when no serializer source found.

Extracted from https://github.com/rails-api/active_model_serializers/pull/1260

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -27,6 +27,20 @@ module ActiveModel
       )
     /x
 
+    # Hashes contents of file for +_cache_digest+
+    def self.digest_caller_file(caller_line)
+      serializer_file_path = caller_line[CALLER_FILE]
+      serializer_file_contents = IO.read(serializer_file_path)
+      Digest::MD5.hexdigest(serializer_file_contents)
+    rescue TypeError, Errno::ENOENT
+      warn <<-EOF.strip_heredoc
+       Cannot digest non-existent file: '#{caller_line}'.
+       Please set `::_cache_digest` of the serializer
+       if you'd like to cache it.
+       EOF
+      ''.freeze
+    end
+
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
       class_attribute :_attributes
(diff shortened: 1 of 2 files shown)


lib/active_model/serializer.rb
```diff
@@ -27,6 +27,20 @@ class Serializer
...
+      serializer_file_path = caller_line[CALLER_FILE]
+      serializer_file_contents = IO.read(serializer_file_path)
+      Digest::MD5.hexdigest(serializer_file_contents)
+    rescue TypeError, Errno::ENOENT
+      warn <<-EOF.strip_heredoc
+       Cannot digest non-existent file: '#{caller_line}'.
+       Please set `::_cache_digest` of the serializer
+       if you'd like to cache it.
+       EOF
+      ''.freeze
```
NullVoxPopuli: :+1: very descriptive error message explains the problem, and says what to do about it! yay
:tada:

NullVoxPopuli: but why is the freeze on an empty string?

Or is this something special with how the <<-STRING syntax works?

bf4: No, just a premature optimization :)

B mobile phone

bf4:
@NullVoxPopuli wanna label LGTM?

NullVoxPopuli:
LGTM. Will merge when Travis passes

lib/active_model/serializer.rb
beauby: Have we fixed the case where the error is spit because the file name contains a whitespace?

bf4: Don't think so

B mobile phone

beauby: What does this PR fix actually? Was the problem not all along that we don't handle whitespace in filenames?

bf4: When there's no source file.  More info in description and tests

B mobile phone

beauby: Yeah but what I meant was that the bug actually arose in the case the source file was deemed non-existent because the regex would not allow for whitespace.

bf4: Ok

B mobile phone

PR #1272, opened 2015-10-15
Add PORO serializable base class: ActiveModelSerializers::Model

bf4:
Extracted from https://github.com/rails-api/active_model_serializers/pull/1260

### DIFF
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -1,0 +1,39 @@
+# ActiveModelSerializers::Model is a convenient
+# serializable class to inherit from when making
+# serializable non-activerecord objects.
+module ActiveModelSerializers
+  class Model
+    include ActiveModel::Model
+    include ActiveModel::Serializers::JSON
+
+    attr_reader :attributes
+
+    def initialize(attributes = {})
+      @attributes = attributes
+      super
+    end
+
+    # Defaults to the downcased model name.
+    def id
+      attributes.fetch(:id) { self.class.name.downcase }
+    end
+
+    # Defaults to the downcased model name and updated_at
+    def cache_key
+      attributes.fetch(:cache_key) { "#{self.class.name.downcase}/#{id}-#{updated_at.strftime("%Y%m%d%H%M%S%9N")}" }
+    end
+
+    # Defaults to the time the serializer file was modified.
+    def updated_at
+      attributes.fetch(:updated_at) { File.mtime(__FILE__) }
+    end
+
+    def read_attribute_for_serialization(key)
+      if key == :id || key == 'id'
+        attributes.fetch(key) { id }
+      else
+        attributes[key]
+      end
+    end
+  end
+end
--- a/test/action_controller/adapter_selector_test.rb
+++ b/test/action_controller/adapter_selector_test.rb
@@ -46,7 +46,7 @@ module ActionController
 
       def test_render_skipping_adapter
         get :render_skipping_adapter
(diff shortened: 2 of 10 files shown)


lib/active_model_serializers/model.rb
```diff
@@ -0,0 +1,39 @@
...
+      attributes.fetch(:cache_key) { "#{self.class.name.downcase}/#{id}-#{updated_at.strftime("%Y%m%d%H%M%S%9N")}" }
+    end
+
+    # Defaults to the time the serializer file was modified.
+    def updated_at
+      attributes.fetch(:updated_at) { File.mtime(__FILE__) }
+    end
+
+    def read_attribute_for_serialization(key)
+      if key == :id || key == 'id'
```
NullVoxPopuli: if we make attributes a hash with indifferent access, we wouldn't need this `||` check. 
I just think it's a little weird to only have indifferent access for the id.

test/action_controller/adapter_selector_test.rb
```diff
@@ -46,7 +46,7 @@ def test_render_using_adapter_override
 
       def test_render_skipping_adapter
         get :render_skipping_adapter
-        assert_equal '{"attributes":{"name":"Name 1","description":"Description 1","comments":"Comments 1"}}', response.body
+        assert_equal '{"name":"Name 1","description":"Description 1","comments":"Comments 1"}', response.body
```
NullVoxPopuli: going forward, can the expected value be assigned to an `expected` variable?

bf4: Yeah, would be good 

B mobile phone

lib/active_model_serializers/model.rb
bf4: I was going to explain it but now I'm not sure 

B mobile phone

bf4: For reasons I don't recall there might be an `id` method but not and `:id` attribute but it still hits method_missing...

lib/active_model_serializers/model.rb
```diff
@@ -0,0 +1,39 @@
...
+        attributes.fetch(key) { id }
```
bf4: if the id method is missing, I probably shouldn't be trying to call it here

bf4: But, original code was

```ruby
-    if name == :id || name == 'id'
-      id
```

NullVoxPopuli: :+1:

NullVoxPopuli: Since this is just an example, is there anything wrong with just having attributes[key] ?

bf4: Well, it's used by the test suite like this... I'll double check tonight

B mobile phone

bf4:
I'm going to merge this so people can being using it, even though there's some followup

- [ ] https://github.com/rails-api/active_model_serializers/pull/1272#discussion_r42172925
- [ ] https://github.com/rails-api/active_model_serializers/pull/1272#discussion_r42111785
- [ ] https://github.com/rails-api/active_model_serializers/pull/1272#discussion_r42111871

lib/active_model_serializers/model.rb
paulodiniz: But if id method is missing, shouldn't the resource pass as a object complient resource?

paulodiniz: @NullVoxPopuli I ran some tests, and if only attributes[key] is present, it returns nil when the key is not present on the attributes hash
```
attributes => {"id"=>3, "title"=>"AMS"}
attributes.fetch(key) { id } => 3
attributes[key] => nil
```

cc @bf4

PR #1276, opened 2015-10-16
fix for serializers never including associations on cache hits

kieran:
move @node assignment and the include_associations! call outside the cached response, since including associations produces the side-effect *actually* including the associations.

Previously, subsequent renders of a cached serializer would never include associations.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -361,7 +361,7 @@ module ActiveModel
         _serializable_hash
       end
 
-      include_associations! if _embed
+      include_associations! if @object.present? && _embed
       @node
     end
 
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -353,13 +353,16 @@ module ActiveModel
     # Returns a hash representation of the serializable
     # object without the root.
     def serializable_hash
-      if perform_caching?
+      @node = if perform_caching?
         cache.fetch expand_cache_key([self.class.to_s.underscore, cache_key, 'serializable-hash']) do
           _serializable_hash
         end
       else
         _serializable_hash
       end
+
+      include_associations! if _embed
+      @node
     end
 
     def include_associations!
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Hi there! could you please add a test for this, so we clearly know the problem that is solved? thanks!

chainlink:
+1

kieran:
FYI my spec passes, but there are others which were broken as of the previous release

gf3:
:+1:

beauby:
I am not familiar with the 0.8 codebase so @joaomdmoura I'm assigning you (:

jorgemanrubia:
:+1: to this, thank you so much @kieran 

Current version 0.8 is completely broken when it comes to caching objects with associations. It would be great to have this merged

lib/active_model/serializer.rb
```diff
@@ -476,9 +479,7 @@ def scope
 
     def _serializable_hash
       return nil if @object.nil?
-      @node = attributes
```
bf4: Why the change of node from `attributes` to `_serializable_hash`?

kieran: `@node` needs to be assigned outside of `_serializable_hash`, since associations can differ based on the serializer used, and the `_serializable_hash` method itself is cached inside `serializable_hash`

bf4: Ok, thanks. I'm not as familiar with the 0.8 code flow.

Do you think you can get CI passing?

kieran: @bf4 the specs seem to fail randomly trying to call  `read_attribute_for_serialization` - the failures are not tied to versions of rails or ruby, and can change per run. The issue appears to be in the generated faker data, but I'm not exactly sure what's going on there.

Anyway, we've used this in production without issue since I opened this PR, so it's safe to merge.

bf4: @kieran I'm okay to merge then, since I'm trusting you, here :)  It looks like you have run the specs locally, but you say they fail intermittently for you?

kieran: yeah, they do about 50% of the time. Something to do with the way fake objects are generated for testing.

NullVoxPopuli:
does this still apply? needs rebase if it does

kieran:
rebased, should merge now

Issue #1269, opened 2015-10-14
SerializationContext: request url info, content type, query params, api version, namespace, etc


bf4:
following: https://github.com/rails-api/active_model_serializers/issues/1268#issuecomment-148144139

- [x] rename `options[:context]` in the controller etc. to `options[:serialization_context]` #1289
- [x] test an example of passing in a custom `serialization_context` #1289
- [x] create a `SerializationContext` object that wraps the `request` and exposes `original_url` and `query_parameters` #1289
- [x] consider using the SerializationContext to expose the request `media_type` and `api_version` and `url_helpers`
  - [x] ref api version:  https://github.com/rails-api/active_model_serializers/pull/1265#issuecomment-147267020
  - [x] ref media type: https://github.com/rails-api/active_model_serializers/pull/1082
  - [x] ref url helpers: https://github.com/rails-api/active_model_serializers/pull/1282 https://github.com/rails-api/active_model_serializers/pull/666 https://github.com/rails-api/active_model_serializers/pull/1289#discussion_r42808317

jsdir:
:+1:

vasilakisfil:
Btw why context is always passed down using default Rails render? I am using JSON adapter. Not an expert but I think it's useless to JSON adapter and adds memory for no reason.

beauby:
@vasilakisfil It does not add that much memory though, does it? If I'm not wrong it's just a reference.

vasilakisfil:
@beauby I don't know how much memory it adds but it's a bit huge object. And not used at all (again I am using JSON adapter). I can always remove it by setting context: nil but not sure why it's added by default. Maybe have an option for that just like serializable_scope ?

vasilakisfil:
Just looked up and I really think context shouldn't be included at all :) Why we need it? The only reason I saw was to get the query_parameters. Getting the query_parameters in order to create the pagination links **shouldn't** be AMS responsibility. If you have a paginated object you can retrieve the current/previous/next page/per_page and size numbers and then create the links. Unless I am missing something.

tchak:
I d'like to work on this. Several questions:
- Are we sure about name `request_context`? Maybe `serialization_context` as mentioned in #1282 by symmetry with `serialization_scope`?
- Should we expose url helpers on this object? I find that in rails app, generating urls in serializers is pretty common. It is a "view" layer after all :) If not, what are other options?

tchak:
[...]

PR #1282, opened 2015-10-20
Add supports for :links and :data options to has_many associations for jsonapi adapter

tchak:
This is a rebase from #1028

I hope @lsylvester don't mind :)

I also added more test including the one requested by @joaomdmoura with `links` + `data`. I also added support for passing in a hash of urls or proc in order to be able to provide custom links.

I hope we can move forward from here.


lib/action_controller/serialization.rb
```diff
@@ -27,6 +27,7 @@ def get_serializer(resource, options = {})
       end
       serializable_resource = ActiveModel::SerializableResource.new(resource, options)
       if serializable_resource.serializer?
+        serializable_resource.url_helper = self
```
bf4: Thanks for this

:-1: We will not include the controller. That would be a **huge** bug and complexity attractor

I think there's an issue for this, but I think the `:context` option should be renamed `:request_context` and be an object named `RequestContext` that has the methods `original_url`, `query_params`, and `url_helper`, though  them's just my thoughts.. but the main this is that we will *not* pass around a controller instance

test/fixtures/poro.rb
```diff
@@ -1,6 +1,7 @@
 verbose = $VERBOSE
 $VERBOSE = nil
 class Model
+  extend ActiveModel::Naming
```
bf4: please remove

test/fixtures/poro.rb
```diff
@@ -11,6 +12,10 @@ def initialize(hash = {})
     @attributes = hash
   end
 
+  def to_model
+    self
+  end
```
bf4: please remove

test/fixtures/poro.rb
```diff
@@ -31,9 +36,14 @@ def id
     @attributes[:id] || @attributes['id'] || object_id
   end
 
+  def to_param
+    id.to_s
+  end
```
bf4: please remove

test/fixtures/poro.rb
```diff
@@ -85,6 +95,7 @@ def cache_key
     "#{self.class.name.downcase}/#{self.id}"
   end
 end
+Tag = Class.new(Model)
```
bf4: Please put relevant objects in their respective tests

test/support/rails_app.rb
```diff
@@ -13,6 +13,7 @@ class Foo < Rails::Application
 module TestHelper
   Routes = ActionDispatch::Routing::RouteSet.new
   Routes.draw do
+    get 'tags/:tag_id/posts', to: 'application#index', as: 'tag_posts'
```
bf4: aside: seems to me we're getting close to needing a dummy app

lib/action_controller/serialization.rb
tchak: Sounds good. Would you be ok with this pr only including hash with proc links option? In the mean time it will allow to manually expose links and we can se for automatic link generator in the feature.

test/fixtures/poro.rb
```diff
@@ -241,6 +252,22 @@ def self.root_name
...
+LinkNoDataTagSerializer = Class.new(ActiveModel::Serializer) do
+  has_many :posts, links: true, data: false
+end
+
+LinkUrlTagSerializer = Class.new(ActiveModel::Serializer) do
+  has_many :posts, links: { related: 'http://test.host/post_tags' }, data: false
+end
+
+LinkProcTagSerializer = Class.new(ActiveModel::Serializer) do
+  has_many :posts, links: { related: -> (object, name) { "http://test.host/tags/#{object.id}/#{name}/all" } }, data: false
```
bf4: > related: -> (object, name) { "http://test.host/tags/#{object.id}/#{name}/all" }

I believe what you want here is a uri_template. See https://github.com/hannesg/uri_template 

See e.g. https://developer.github.com/v3/#schema http://tools.ietf.org/html/rfc6570 see e.g. https://api.github.com/ (though not a JSON API, is an example of usage)

bf4:
Thinking aloud: This is a big (important) change. Maybe let's make this smaller and start with the simplest case?

test/fixtures/poro.rb
tchak: Do you want to include `uri_template` library by default with AMS?

tchak:
@bf4 completely agree about starting smaller.

Supporting only

```ruby
has_many :posts, links: { related: url }, data: false
```

and

```ruby
has_many :posts, links: { related: -> () { url } }, data: false
```

sound's like small enough for you?

test/fixtures/poro.rb
bf4: Maybe.. play with it and see if you like it.  It should be pretty simple to implement a simple subset ourselves to start with.

test/fixtures/poro.rb
bf4: Now you can use https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/model.rb !!!

bf4: (Like as of just a moment ago when I merged #1272 )

lib/action_controller/serialization.rb
bf4: yeah, plus there's already a `serialization_scope` options, so maybe `serialization_context` would be good

bf4:
or even just `has_many :posts, links: { related: url }, data: false`, `has_many :posts, links: true` etc.

lib/action_controller/serialization.rb
bf4: ref: https://github.com/rails-api/active_model_serializers/issues/1269

tchak:
@bf4 the problem with `has_many :posts, links: { related: url }, data: false` is that it is rather useless, as usually an association url will contain parent id...

test/fixtures/poro.rb
tchak: I am sure it is the right approach. But we probably should support it on any link, so looks like a separate feature to work on once we have all the basic links working.

beauby:
Yup, @tchak is right. We have basically 3 options here:
 + support uri templates
 + support lambdas/procs
 + define links inside blocks (which is basically the same as the previous option)

tchak:
@beauby my current plan is to simplify this PR to support only proc/lambda. Everything else can be implemented in terms of proc on top of it once we figured out `context` thing in another PR.

beauby:
@tchak uri templates is a very interesting option as well, I'm not sure which is best

tchak:
@beauby I completely agree. I am just thinking making it one step at a time.

bf4:
Link template + link_attributes (like mustache) is easy to understand and switch out impl

Using the template rfc makes building an info response object easier but isn't super high priority

B mobile phone

beauby:
We have to be careful when passing `data: false` not to allow the corresponding associated resource's association to be included via the `include` option in order not to break the full-linkage property.

tchak:
@beauby Right! I will look in to it.

tchak:
@beauby I added a bunch of tests to test different combinations of `data` and `include` options

tchak:
@bf4 @beauby I want to add uri template but I feel like there is already a lot of stuff in this PR. Would prefer to see it merged and then I will work on a next PR with uri template stuff.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -173,16 +173,48 @@ def relationship_value_for(serializer, options = {})
           end
         end
 
+        def relationship_links_for(name, owner, options = {})
+          options[:links].each_with_object({}) do |(key, link), hash|
+            if link.is_a? Proc
+              hash[key] = link.call(owner, name)
```
bf4: should be `respond_to?(:call)`

bf4: here and elsewhere

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -173,16 +173,48 @@ def relationship_value_for(serializer, options = {})
...
+          end
+        end
+
         def relationships_for(serializer)
           serializer.associations.each_with_object({}) do |association, hash|
-            hash[association.key] = { data: relationship_value_for(association.serializer, association.options) }
+            hash[association.key] = {}
+
+            if association.options.fetch(:data, true)
+              hash[association.key][:data] = relationship_data_for(association.serializer, association.options)
```
bf4: oh, interesting, you're re-using the :data key

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -173,16 +173,48 @@ def relationship_value_for(serializer, options = {})
...
+            options[:meta]
+          end
+        end
+
         def relationships_for(serializer)
           serializer.associations.each_with_object({}) do |association, hash|
-            hash[association.key] = { data: relationship_value_for(association.serializer, association.options) }
+            hash[association.key] = {}
+
+            if association.options.fetch(:data, true)
```
bf4: `association.options.fetch(:data, true)` is called a lot, is feature envy, seems association should have an accessor `association.data` `association.data=(data)` or perhaps something better named

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -173,16 +173,48 @@ def relationship_value_for(serializer, options = {})
...
+              hash[association.key][:meta] = relationship_meta_for(association.name, serializer.object, association.options)
+            end
           end
         end
 
         def included_resources(include_tree, primary_data)
           included = []
 
-          serializer.associations(include_tree).each do |association|
+          serializer.associations(include_tree).select { |association| association.options.fetch(:data, true) }.each do |association|
```
bf4: multiple lines plz so it's easier to read

test/adapter/json_api/has_many_links_test.rb
```diff
@@ -0,0 +1,250 @@
+require 'test_helper'
+
+module ActionController
+  module Serialization
+    class JsonApiHasManyLinksTest < ActionController::TestCase
+      LinkTagSerializer = Class.new(ActiveModel::Serializer) do
+        has_many :posts, links: { related: proc { |object, name| "http://test.host/tags/#{object.id}/#{name}" } }, data: false
+      end
```
bf4: plz just `class LinkTagSerializesr < AM::S`. we're moving away from that weird Class.new(AM::S)

bf4:
@tchak Looks really good, :100:!  

Two more things:
- needs some examples of how to use either in code comments or in the `/docs` folder. Explain what `data` and `links`are, how they're used, and that in the absence of the jsonapi adapter, they do nothing.
- the proc usage makes me really nervous.  It seems really inflexible and not very intent-revealing.  Would you mind trying with a (roll your own ok) uri template to see how that feels? I think it'll be a lot more user and dev friendly.

tchak:
@bf4 I was planing on implementing uri template in termes of default proc once this was merged. I think having proc as a backup "power-user" option is good. If you prefer I can add uri template directly to this PR. I am just not sure about all the "variables" I should interpolate with uri template. Obvious one is `parent_id`. I was actually thinking of making it "named", I mean `tag_id` in case of `Tag` serializer with `has_many :posts, links: { related: "http://host/tags/{tag_id}/posts" }`. Not sure if anything else actually needs to be dynamic.

richmolj:
I actually really like the procs since I can imagine scenarios the url should be dynamic. Something like:

```ruby
has_many :posts, proc { |object, name|
  if v2_feature_toggled_on_for_current_user?
    "http://test.host/apt/v2/posts"
  else
    "http://test.host/apt/v1/posts"
  end
}
```

bf4:
@tchak are we still working on this?

tchak:
@bf4 I will have some time to rebase and cleanup this weekend

bf4:
Awesome!

B mobile phone

djsegal:
Is this supposed to resolve: https://github.com/rails-api/active_model_serializers/issues/834?

djsegal:
While using this, I noticed that I didn't have access to `_url` helpers, so I added the following line to the serializer:

+ `include Rails.application.routes.url_helpers`

----------------------
to reduce the repetitiveness of this task,
+ i added a base `class ApplicationSerializer < ActiveModel::Serializer` serializer file
+ and overwrote the serializer template with my own at `lib/templates/rails/serializer/serializer.rb.erb`:

```
<% module_namespacing do -%>
class <%= class_name %>Serializer < ApplicationSerializer
  attributes <%= attributes_names.map(&:inspect).join(", ") %>
<% association_names.each do |attribute| -%>
  has_one :<%= attribute %>
<% end -%>
end
<% end -%>
```

bf4:
@djsegal The url handling is going to be in the [serialization_context](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/serialization_context.rb) [#1313](https://github.com/rails-api/active_model_serializers/pull/1313), but [just hasn't been finished](https://github.com/rails-api/active_model_serializers/pull/1289#discussion_r42808317).  Would you like to [give it a shot](https://github.com/rails-api/active_model_serializers/issues/1269)?

beauby:
ref #1454

remear:
@rails-api/ams Is this still relevant?

PR #1284, opened 2015-10-20
Add optional fields support for associations

vasilakisfil:
I am really fond of JSONAPI but there are people that they still need JSON (mostly) and Attributes adapters :) Plus, we should opt for best migration from 0.8.x and 0.9x adapters.

There is a minor bug in the fields option in master branch. If you have:

```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id],
```

This will apply the fields to all associations too. We don't want that. What we want is:

If there is a fields param, apply it to the main resource. If there are associations, serialize the associations without removing any attributes. For instance:
```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id]
```
Then this should apply the `fields: [:id]` only in the VideoSerializer and not in any association.

If there are associations in the fields array (just like in StrongParams) and these also exist in the main resource, apply the associations fields there. For instance:
```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id, user: [:name]]
```
Then this should apply the `fields: [:id]` only in the VideoSerializer and if there is a user association, apply `fields: [:name]` in the UserSerializer.

I am really looking forward to merge this PR so if you need anything from me just ping me. I will also create a regression test highlighting the bug.

### DIFF
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -35,8 +35,19 @@ module ActiveModel
 
         def resource_relationships(options)
           relationships = {}
+
+          association_fields_hash = {}
+          if options[:fields]
+            association_fields_hash = Hash[
+              *options[:fields].select { |s| s.is_a? Hash }
+            ]
+          end
+
           serializer.associations(@include_tree).each do |association|
-            relationships[association.key] = relationship_value_for(association, options)
+            relationships[association.key] = relationship_value_for(
+              association,
+              options.merge(fields: association_fields_hash[association.name])
+            )
           end
 
           relationships
--- a/test/adapter/json/collection_test.rb
+++ b/test/adapter/json/collection_test.rb
@@ -85,6 +85,72 @@ module ActiveModel
 
             assert_equal(expected, actual)
           end
+
+          def test_fields_option
+            serializer = ArraySerializer.new([@first_post, @second_post])
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
+            actual = adapter.serializable_hash(fields: [:id])
+
+            expected = { posts: [{
+              id: 1,
+              comments: [],
+              author: {
+                id: 1,
+                name: 'Steve K.'
+              },
+              blog: {
+                id: 999,
+                name: 'Custom blog'
+              }
+            }, {
+              id: 2,
+              comments: [],
+              author: {
+                id: 1,
+                name: 'Steve K.'
+              },
+              blog: {
+                id: 999,
+                name: 'Custom blog'
+              }
+            }] }
+
+            assert_equal(expected, actual)
+          end
+
+          def test_fields_with_no_associations_include_option
+            serializer = ArraySerializer.new([@first_post, @second_post])
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer, include: [])
+            actual = adapter.serializable_hash(fields: [:id])
+
+            expected = { posts: [{
+              id: 1
+            }, {
+              id: 2
+            }] }
+
+            assert_equal(expected, actual)
+          end
+
+          def test_fields_with_associations_fields_option
+            serializer = ArraySerializer.new([@first_post, @second_post])
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer, include: [:author])
+            actual = adapter.serializable_hash(fields: [:id, author: [:name]])
+
+            expected = { posts: [{
+              id: 1,
+              author: {
+                name: 'Steve K.'
+              }
+            }, {
+              id: 2,
+              author: {
+                name: 'Steve K.'
+              }
+            }] }
+
+            assert_equal(expected, actual)
+          end
         end
       end
     end


NullVoxPopuli:
@vasilakisfil can you fix the errors mentioned in travis? thanks!

vasilakisfil:
@NullVoxPopuli fixed in travis, waiting for appveyor which takes some time..
Possible linked to https://github.com/rails-api/active_model_serializers/pull/1285 and https://github.com/rails-api/active_model_serializers/pull/1286

beauby:
So if we're going down this way you should make use of the `IncludeTree` class that does exactly what you want, plus it can be built from a string and an extended array/hash syntax.

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -35,8 +35,19 @@ def serializable_hash_for_single_resource(options)
...
+            association_fields_hash = Hash[
+              *options[:fields].select { |s| s.is_a? Hash }
+            ]
+          end
+
           serializer.associations(@include_tree).each do |association|
-            relationships[association.key] = relationship_value_for(association, options)
+            relationships[association.key] = relationship_value_for(
+              association,
+              options.merge(fields: association_fields_hash[association.name])
```
beauby: If you do that, not providing the `fields` option to the adapter will result in no attribute being serialized.

vasilakisfil: not the case since this runs only for associations. The case you said would only happen with the following (given that there is a user association):
```
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id, user: []]
```

Plus, test would cry out if that was true.

vasilakisfil:
@beauby "So if we're going down this way you should make use of the IncludeTree class that does exactly what you want, plus it can be built from a string and an extended array/hash syntax."

I never considered AMS job to parse the query string (+ strings are not the way to go in Ruby but hashes/arrays). Especially when we are talking about JSON adapter. But sure I can fix it.

What do you mean extended array/hash syntax ?

beauby:
@vasilakisfil The query string parsing is here just as a commodity for people dealing with JSON API stuff. The "extended array/hash syntax" is the one described [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/include_tree.rb#L31). Basically it's an array of hashes and symbols (S → Array<Symbol | Hash{Symbol:S}>).

lib/active_model/serializer/adapter/attributes.rb
beauby: Sorry that was an oversight.

vasilakisfil:
@beauby the array/hash syntax is supported in this pull request: `fields: [:id, { user: [:name] }]`. Query string was not supported before this pull request. Do you want me to add support for it?

I am totally **against** it though because unless you are writting an internal API, you will want to filter the query params (query string) beforehand using StrongParams or something else inside the controller so you would need to convert it to an array of hashes/symbols anyway. Extract a helper method that parses this query string in the controller is the way to go in my opinion.

passalini:
+1
ps: nice idea @vasilakisfil

test/adapter/json/collection_test.rb
```diff
@@ -85,6 +85,72 @@ def test_include_option
 
             assert_equal(expected, actual)
           end
+
+          def test_fields_option
+            serializer = ArraySerializer.new([@first_post, @second_post])
+            adapter = ActiveModel::Serializer::Adapter::Json.new(serializer)
+            actual = adapter.serializable_hash(fields: [:id])
```
bf4: Looking good 

1. this needs an integration (controller) test
2. Please use `ActiveModel::SerializableResource.new([@first_post, @second_post], fields: [:id])`

Please don't test passing options into `serializable_hash` since only the renderer would do that

bf4: (that would be a better test usage in the integration test, where the options to `render json: foo, ....` are passed to AMS then to the renderer (unless we mutate them))

@vasilakisfil Are you still interested in this?

vasilakisfil:
@bf4 hmm I am not sure. I have already created my own adapter that is compatible with 0.9.x version and supports association fields among others.

I could work on that but I feel it's kind of deadend. Most attributes/json-adapter related code hasn't been updated for a long time now and most committers in this project doesn't seem to be interested to extend json/attributes adapter features but instead try to keep them as minimal as possible and try to spend more time to jsonapi adapter (which might be a good idea until Rails 5 comes out). That's why I opted for a custom adapter to move faster.

bf4:
custom adapter is good.  I'm not sure what you mean by extending the json/attributes adapters. Do you mean improving or otherwise working on?

williamweckl:
+1

bf4:
> I have already created my own adapter that is compatible with 0.9.x version and supports association fields among others.

@vasilakisfil Is that something you could share?  Would be very helpful.

Any interest in reviving this PR?

NullVoxPopuli:
@vasilakisfil could you rebase this? :-)

NullVoxPopuli:
I'd like to get this merged before taking a look at #1286

vasilakisfil:
Sure I would love to rebase it and give some love to the JSON/Attributes adapters but not now during Easter (1.5 week from now) as I am super busy currently.

I have created 4-5 monkey patched eventually to make it work like GraphQL, didn't create my own adapter.

bf4:
@vasilakisfil 

> I have created 4-5 monkey patched eventually to make it work like GraphQL, didn't create my own adapter.

would love to see them!

vasilakisfil:
Ok if it helps, the following have been tested up until b2cd7bb51ab6f481972b196a9c2915a50aa52826 (the git commit locked in my Gemfile).

There is one more bug (I think it's a bug and not intended) down the line (in ActiveModel::Serializer::Associations#association method), so this pull request doesn't actually solves it. But as I said I think JSON/Attributes adapters have been slightly neglected by the team in favor of JSONAPI adapter and I would love to help out.

```ruby
class ActiveModel::Serializer::Adapter::Attributes
  def relationship_value_for(association, options)
    return association.options[:virtual_value] if association.options[:virtual_value]
    return unless association.serializer && association.serializer.object

    opts = instance_options.merge(
      include: @include_tree[association.key],
      fields: nil
    )
    self.class.new(association.serializer, opts).serializable_hash(
      options.merge(fields: nil)
    )
  end

    options ||= {}
    options[:fields] = instance_options[:fields] if options[:fields].nil?

      serializable_hash_for_collection(options)
    else
      serializable_hash_for_single_resource(options)
    end
  end

    serializer.map { |s|
      self.class.new(s, instance_options).serializable_hash(options)
    }.select{|s| !s.blank?}
  end

  def serializable_hash_for_single_resource(options)
    return {} if serializer.invalid_for_serialization?

    resource = resource_object_for(options)
    relationships = resource_relationships(options)
    resource.merge!(relationships)
  end
end
module ActiveModel
  class Serializer
    module Associations
      def associations(include_tree = DEFAULT_INCLUDE_TREE)
        return unless object

        Enumerator.new do |y|
            next unless include_tree.key?(key)
            y.yield reflection.build_association(
              self, instance_options.merge(fields: nil)
            )
          end
        end
      end
    end
  end
end

class ActiveModel::SerializableResource
  def initialize(resource, options = {})
    @resource = resource
    @adapter_opts, @serializer_opts =
      options.partition { |k, _| ADAPTER_OPTION_KEYS.include? k }.map { |h| Hash[h] }
    @serializer_opts.merge!(fields: options[:fields])
  end
end

class ActiveModel::Serializer
  def attributes
    attributes = self.class._attributes.dup

      next unless instance_options[:fields].nil? || field_allowed?(name)

      if self.class._fragmented
        hash[name] = self.class._fragmented.public_send(name)
      else
        hash[name] = send(name)
      end
    end
  end
  def field_allowed?(name)
   instance_options[:fields] && instance_options[:fields].include?(name)
  end
end
```

vasilakisfil:
The last monkeypatch is irrelevant actually, it allows you to take out an item from a collection just before serializing.

vasilakisfil:
No sorry, that's wrong, I don't even remember my own monkey patches haha. The last one is something else I think but the `invalid_for_serialization?` part does what I said :)

Anyway :) But I would be happy to commit to this project.

remear:
Closing due to inactivity

NullVoxPopuli:
I also believe this functionality already exists

PR #1289, opened 2015-10-21
rename context to serialization_context and add url helpers

tchak:


tchak:
I am not sure about the best way to include url_helpers in the context object. If some people with more knowledge about rails internals have a better idea I am all :ear:  :)

tchak:
I also feel that in the end `serialization_context` should be exposed just like `scope` directly on the serializer instances. I would like to add `serialization_context_name` just like for `scope` and expose it as `context` by default.

lib/action_controller/serialization/context.rb
```diff
@@ -0,0 +1,18 @@
+module ActionController
+  module Serialization
+    class Context
+      include Rails.application.routes.url_helpers if Rails.application.present?
+
+      def self.default_url_options
+        ActionMailer::Base.default_url_options
+      end
```
bf4: I'd rather not include the url helpers and instead stick 'em in a url_helpers method

We could probably set it and default_url_iptions in the railtie.

Also, i don't believe we include action mailer nor do we need it. There should be a similar methid on the controller

bf4:
Awesome work!

lib/action_controller/serialization/context.rb
bf4: Let's also add a context test

tchak: Oh sorry about `ActionMailer` this is me trying things by copy/pasting from SO in a hurry :( Will look in to how do it from the railtie!

beauby:
Pretty cool :+1:

NullVoxPopuli:
LGTM :+1:

NullVoxPopuli:
@tchak I'm good with merging this, unless you think `serialization_context_name` should be included in this PR.. maybe a different PR? All tests pass, so... :-)

tchak:
@NullVoxPopuli let me add a few more test tonight. Yes `serialization_context_name` will go in a separate PR.

NullVoxPopuli:
cool. thanks!

lib/action_controller/serialization/context.rb
```diff
@@ -0,0 +1,13 @@
+module ActionController
+  module Serialization
+    class Context
+      attr_reader :request_url, :query_parameters, :url_helpers
```
bf4: I think I'd rather have `url_helpers` a class method on the Context.  It feels more natural here. Also, no need to include them. you could just

```ruby
class Context
  class << self
    attr_accessor :url_helpers, :default_url_options
  end
end
```

and in the Railtie

```ruby
Context.url_helpers = Rails.application.routes.url_helpers # for now, later would like to encapsulate it to control the boundary e.g. UrlHelpers.new(Rails.application.routes.url_helpers)
Context.default_url_options =  ActionController::Base.default_url_options
```

I'm trying to keep in mind how someone using grape would configure this.  If we keep a good interface, the details of the implementation don't matter as much.  That means more composition and less just mixing things in. Alternatively, there could be a UrlHelpers module that you mix in to the Context when the railtie is called, which allows others to mix in a different helper object

lib/active_model_serializers.rb
```diff
@@ -7,6 +7,8 @@ module ActiveModelSerializers
   mattr_accessor :logger
   self.logger = Rails.logger || Logger.new(IO::NULL)
 
+  mattr_accessor :url_helpers
```
bf4: see https://github.com/rails-api/active_model_serializers/pull/1289/files#r42808317

lib/active_model_serializers.rb
```diff
@@ -7,6 +7,8 @@ module ActiveModelSerializers
   mattr_accessor :logger
   self.logger = Rails.logger || Logger.new(IO::NULL)
```
bf4: This should really be set to Rails.logger in the Railtie, not here. womp womp

lib/action_controller/serialization/context.rb
tchak: hmm as far as I understand `default_url_options` needs to be defined on the same object where `url_helpers` are included in order to be taken in to account. Am I not understanding it correctly?

tchak: also I usually use url helpers in serialisers a lot, so I am thinking that at some point I would like to be able to do in some serializer code something like this:
```
def validation_url
  context.validation_url(object)
end
```

Are we on the same page here? Or do you see the usage differently?

tchak:
I am trying to add some tests to `url_helpers` thing and I feel a bit confused :) I am going to need a better understanding of how rails routes works and compose with everything else... I hope to be able to figure it out in the next few days.

beauby:
In my opinion this PR is worth merging without url_helpers (that we could add later).

NullVoxPopuli:
I agree, @beauby

beauby:
Wait, it wasn't ready though (there were references to url_helpers and stuff). Could you revert the merge until we clear it for merging?

NullVoxPopuli:
that's what I get for not re-reviewing. ha

bf4:
@tchak would you mind re-submitting this PR, referencing it, and referring to the comments in https://github.com/rails-api/active_model_serializers/pull/1289#discussion_r42808317 ? Maybe skip the do it in two parts, without the url helpers, then adding them

beauby:
So in the end, how does one access the `url_helpers` from within an attribute definition in a serializer?
As follows?
```ruby
attribute :url do
  serialization_context.url_helpers.posts_route
end
```

bf4:
I think I left this open for the uri templates

The serialization context url helpers are mixed into Link.  Examples in docs

B mobile phone

PR #1291, opened 2015-10-22
Add Logging

maurogeorge:
Hi guys,

I think today we do not logging nothing right? Based on [this](https://github.com/rails-api/active_model_serializers/pull/1099#discussion_r42184548) I started working on a logging using `ActiveSupport::Notifications` and `ActiveSupport::LogSubscriber`.

The actual log output on this patch is:

```
[AMS] Rendered PostSerializer with ActiveModel::Serializer::Adapter::JsonApi (5.72ms)
```

This makes sense to you guys? If make sense I can continue working on it.


lib/active_model/serializable_resource.rb
```diff
@@ -64,5 +80,14 @@ def serializer?
...
 
     attr_reader :resource, :adapter_opts, :serializer_opts
+
+    def notify_active_support(&block)
+      return unless serializer?
+      event_name = 'render.active_model_serializers'
+      payload = { serializer: serializer, adapter: adapter }
+      ActiveSupport::Notifications.instrument(event_name, payload) do
+        block.call
+      end
```
bf4: did you need to use the `&block` form?  I think you should be able to remove the method argument and replace block.call with `yield` (which is faster and saves object creation

we should also freeze the string here or reduce object allocations on what would be frequently called method, even if the notifications are turned off.

otherwise, :100:

lib/active_model/serializer/logging.rb
```diff
@@ -0,0 +1,32 @@
+module ActiveModel
+  class Serializer
+    module Logging
+      extend ActiveSupport::Concern
+
+      included do
+        cattr_accessor(:logger) do
+          ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
+        end
```
bf4: So, there's a logger already defined on `ActiveModelSerializers.logger` which should be used.  it should default to the null logger and be configured in the railtie, which is extra work, I hope you don't mind

so, you could remove the line from the test helper since it defaults to null

lib/active_model/serializer/logging.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      class LogSubscriber < ActiveSupport::LogSubscriber
+        def render(event)
+          serializer = event.payload[:serializer]
+          adapter = event.payload[:adapter]
+          duration = event.duration.round(2)
+          logger.tagged('AMS') do
+            info do
+              "Rendered #{serializer.name} with #{adapter.class} (#{duration}ms)"
```
bf4: The more work you can do in the `info` block the better (i.e. all the locals here), since it is only evaluated when the info log level. (I think it should be debug, so this should be configurable.. `logger.add(Logger::DEBUG) { "Rendered..." }` would be fine

bf4:
:fireworks: :heart_eyes_cat: yay.. plz see comments re: existing logger

maurogeorge:
@bf4 thanks for the review, I made some changes. I prefer keep the commits for now when it is ok I squash these.

lib/active_model/serializable_resource.rb
maurogeorge: Thanks, I made these changes :+1:

lib/active_model/serializer/logging.rb
maurogeorge: I prefer keep the `ActiveModel::Serializer.logger` since this is the way we define configs today like the adapter. More info about my rationale here https://github.com/maurogeorge/active_model_serializers/commit/d2c2378d55469a17fdef686c845207ae660a7bc2

In fact, what you think about using the config like: `ActiveModel::Serializer.config.logger` the same way we define the adapter we can define a custom log.

lib/active_model/serializer/logging.rb
maurogeorge: I do not understand this line:

> ...so this should be configurable.. logger.add(Logger::DEBUG) { "Rendered..." } would be fine

Could you please clarify?
I follow the same idea of [Active Job](https://github.com/rails/rails/blob/2ad9afe4ff2c12d1a9e4a00d9009d040e636c9b0/activejob/lib/active_job/logging.rb#L56) and [Action View](https://github.com/rails/rails/blob/2ad9afe4ff2c12d1a9e4a00d9009d040e636c9b0/actionview/lib/action_view/log_subscriber.rb#L16) for example the uses the `info` to log.

maurogeorge:
@bf4 I answered each of your comments, they are under the "Show outdated diff" link, since the code changed.

I waiting for your feedback :heart:

lib/action_controller/serialization.rb
```diff
@@ -31,6 +31,7 @@ def get_serializer(resource, options = {})
         serializable_resource.serialization_scope_name = _serialization_scope
         begin
           serializable_resource.adapter
+          serializable_resource
```
beauby: Could you explain why you made that change?

beauby:
I'm generally :+1: with this PR.

docs/general/logging.md
```diff
@@ -0,0 +1,12 @@
+# Logging
+
+If we are using AMS on Rails app the `Rails.logger` will be used.
```
bf4: by default

lib/active_model/serializer/logging.rb
bf4: Sorry. I cannot support this.

docs/general/logging.md
bf4: Also, document that the event is `render.active_model_serializers`

lib/active_model/serializer/railtie.rb
```diff
@@ -2,9 +2,7 @@
 module ActiveModel
   class Railtie < Rails::Railtie
     initializer 'active_model_serializers.logger' do
-      ActiveSupport.on_load(:action_controller) do
-        ActiveModelSerializers.logger = ActionController::Base.logger
-      end
+      ActiveSupport.on_load(:active_model_serializers) { self.logger ||= Rails.logger }
```
bf4: This won't work for two reasons:

1) It's not our lib that needs to be loaded for Rails.logger to be defined, it's Rails.
2) This won't work because there's nothing running this hook

lib/action_controller/serialization.rb
bf4: seconded

lib/active_model/serializer.rb
```diff
@@ -13,6 +14,7 @@ module ActiveModel
   class Serializer
     include Configuration
     include Associations
+    include Logging
```
bf4: :-1: does not belong here

test/logger_test.rb
```diff
@@ -1,18 +0,0 @@
-require 'test_helper'
-
-class ActiveModelSerializers::LoggerTest < Minitest::Test
-  def test_logger_is_set_to_action_controller_logger_when_initializer_runs
-    assert_equal ActiveModelSerializers.logger, ActionController::Base.logger
```
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1291/files#r42951755

bf4:
This PR mixes a desired feature with good code and tests with an architectural and behavior change this maintainer is very :-1: on.  Please separate the two into two PRs so we can get the feature (logging notifications) into master.

bf4:
Also, please fix warnings 

> Calling deprecated ArraySerializer in /home/travis/build/rails-api/active_model_serializers/test/collection_serializer_test.rb:10:in `new'. Please use CollectionSerializer

beauby:
@bf4 I believe the warning fixes do not belong in this PR since it does not touch the file causing the warning. Those should probably have been corrected in the PR that deprecated `ArraySerializer`.

beauby:
Or am I missing something?

bf4:
If they were in there, yeah i must have missed them in a rebase

B mobile phone

lib/action_controller/serialization.rb
maurogeorge: Take a look on this commit message https://github.com/maurogeorge/active_model_serializers/commit/981c2293f4ca46daf1ebcd51468497611a389687

maurogeorge:
Thanks for the review guys, I will made the changes when working on the project again.

maurogeorge:
@bf4 @beauby I made the changes as suggested, waiting for your review.
Thanks :smile:

maurogeorge:
@bf4 @beauby I rollback all changes related with `ActiveModel::Serializer.logger` and keep the actual `ActiveModelSerializers.logger` since this namespace is a discussion on other thread.

When you guys have some time please review.

I think we can remove the labels `Hold` and `Needs Team Discussion` I think this is related with the namespace I am right?

Thanks a lot guys

NullVoxPopuli:
@maurogeorge I think I'm good with this.

i'll give a few hours for other team members to comment

lib/action_controller/serialization.rb
bf4: Didn't clear it up for me

lib/active_model/serializable_resource.rb
```diff
@@ -11,7 +11,23 @@ def initialize(resource, options = {})
...
+      notify_active_support do
+        adapter.as_json(*args)
+      end
+    end
+
+    def to_json(*args)
+      notify_active_support do
+        adapter.to_json(*args)
+      end
+    end
```
bf4: call me crazy, but I think this might be a good use case for an around filter

bf4:
@NullVoxPopuli please don't take off a hold when the raised issues haven't been addressed

lib/active_model/serializer/logging.rb
```diff
@@ -0,0 +1,26 @@
+module ActiveModel
+  class Serializer
+    module Logging
```
bf4: I'm :-1: on adding new namespaces to what I hope to become the legacy namespace, #1310 

It either needs to be in lib/active_model_serializers.rb, in lib/active_model_serializers/logging.rb or have 2 :+1: by other maintainers

lib/active_model/serializer/logging.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        def logger
+          ActiveModelSerializers.logger
+        end
+      end
+    end
+  end
+end
+
+ActiveModel::Serializer::Logging::LogSubscriber.attach_to :active_model_serializers
```
bf4: should be configured in the railtie

test/logging_test.rb
```diff
@@ -0,0 +1,77 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    class LoggingTest < Minitest::Test
+      class TestLogger < ActiveSupport::Logger
+        def initialize
+          @file = StringIO.new
+          super(@file)
+        end
```
bf4: just use your own logger. that's kind of the point, no?

test/logging_test.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      def test_logs_correct_adapter
+        ActiveModel::SerializableResource.new(@post).serializable_hash
+        assert_match(/ActiveModel::Serializer::Adapter::Attributes/, @logger.messages)
+      end
+
+      def test_logs_the_duration
+        ActiveModel::SerializableResource.new(@post).serializable_hash
+        assert_match(/\(\d+\.\d+ms\)/, @logger.messages)
```
bf4: nice tests

lib/active_model/serializer/logging.rb
NullVoxPopuli: @bf4 are you saying everything should be switched to the new namespace before adding anything?

wouldn't that only matter depending on the order of merges? waiting on stuff like that causes needless delays, imo

but maybe I misunderstand? idk

wagenet:
I'm definitely interested in this for use by [Skylight](https://www.skylight.io). The instrumentation part of this looks great.

bf4:
@maurogeorge can we scope this PR to just the notifications?

```diff
+    def notify_active_support
+      event_name = 'render.active_model_serializers'.freeze
+      payload = { serializer: serializer, adapter: adapter }
+      ActiveSupport::Notifications.instrument(event_name, payload) do
+        yield
+      end
```

lib/action_controller/serialization.rb
maurogeorge: The main purpose of this is to the instrumentation be called on the Rails controller, so if I remove, [this test](https://github.com/maurogeorge/active_model_serializers/commit/981c2293f4ca46daf1ebcd51468497611a389687#diff-f290b93cc1e01ae4a2c31174202b50eeR424) will be broken.

As you can see [here](https://github.com/maurogeorge/active_model_serializers/blob/2c4a028bed946d10b2972c975ef9c2501ef43c25/lib/active_model/serializable_resource.rb#L14-L30) the instrumentation is made on the `ActiveModel::SerializableResource` what [I think is the canonical way to use a serializer](https://github.com/maurogeorge/active_model_serializers/tree/2c4a028bed946d10b2972c975ef9c2501ef43c25#using-a-serializer-without-render) and as you can see we simply delegate the method to the adapter. If we call the adapter directly over use the `ActiveModel::SerializableResource` we will need to change all adapters to call instrumentation on they methods `serializable_hash`, `as_json` and `to_json`. 

Keeping the `ActiveModel::SerializableResource` we define this once and all adapter, including new ones, will have instrumentation.

Let me know if you have some doubt.

lib/active_model/serializer/logging.rb
maurogeorge: Why? We still want logging on other environments right?

test/logging_test.rb
maurogeorge: I don't think so. I know that the `ActiveModelSerializers.logger` do his job, so to avoid open the log, parse the log, delete the log and so on. I prefer using a custom logger this way I can have all messages on a easy way. This is the same logic followed by [Active Job](https://github.com/rails/rails/blob/9b663e2679cfdaa2da9bfed0bd4d601309072c22/activejob/test/cases/logging_test.rb#L13-L23)

lib/active_model/serializable_resource.rb
maurogeorge: Hi @bf4 I updated to use this https://github.com/rails-api/active_model_serializers/commit/49c8f69080f9734a0bda5268aa7ede56f753b465. Personally I don't like, but I can live with it :smile:

test/logging_test.rb
maurogeorge: Thanks :blush:

maurogeorge:
@bf4 why? I made the [last change you ask](https://github.com/rails-api/active_model_serializers/pull/1291#discussion_r43975844).

At https://github.com/rails-api/active_model_serializers/pull/1291#discussion_r43976001 you say we need 2 :+1: to go on, I think we have theses on https://github.com/rails-api/active_model_serializers/pull/1291#issuecomment-153873366 and https://github.com/rails-api/active_model_serializers/pull/1291#issuecomment-150979209. If I am not wrong you concern is around the namespace right? I think we can go this way until we solve the namespace issue.

lib/active_model/serializable_resource.rb
```diff
@@ -2,6 +2,15 @@
...
     ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links])
+    extend ActiveModel::Callbacks
+
+    define_model_callbacks :render
+
+    around_render do |_, block, _|
+      notify_active_support do
+        block.call
+      end
+    end
```
bf4: yeah, I think you're right. This doesn't add anything but complexity.  Sorry for having you extra work. I meant it as a thought if it made sense.. and it's more indirection than helpful

lib/active_model/serializer.rb
```diff
@@ -6,6 +6,7 @@
 require 'active_model/serializer/configuration'
 require 'active_model/serializer/fieldset'
 require 'active_model/serializer/lint'
+require 'active_model/serializer/logging'
```
bf4: this should be required in lib/active_model_serializers, I think. It's more of a global concern than local to this file. feel free to disagree

lib/active_model/serializer/logging.rb
bf4: you mean like non-Rails?

bf4: sorry, was referring to notifications above. logging should be configured in rails in the railtie.  outside rails, manually

test/logging_test.rb
bf4: Ok, I think that makes sense. I would have `@file = StringIO.new; logger = ActiveSupport::Logger.new(@file); @file.rewind; @file.read` but this hides that complexity

bf4: re naming, to me

`ActiveJob::Base.logger` corresponds to `ActiveModelSerializers.logger`

and

`ActiveJob::Logging` to `ActiveModelSerializers::Logging`

lib/active_model/serializer/logging.rb
maurogeorge: Thanks for the review, I will update next time I work on the project.

I do not understand what you want here? Do you want that I move `ActiveModel::Serializer::Logging::LogSubscriber.attach_to :active_model_serializers` to the Railtie? Could you please clarify?

Thanks

maurogeorge:
@bf4 after the https://github.com/rails-api/active_model_serializers/pull/1322 what is the idea? I stop working here and continue making revision on the new PR? If the Idea is to stop sending more code to this branch I think we can close this.

Thanks

bf4:
Up for discussion.... I could close that pr and make it into this one

I thought maybe you'd want to do more logging stuff here

Really whatever makes you feel most appreciated.  I wouldn't have couldn't have the 'follow up' pr without this one. 

B mobile phone

maurogeorge:
I will close this and help with the review of the new one.

Thanks for the rebase and keep the reference :heart:

PR #1295, opened 2015-10-23
Allow users to globally opt out of automatic lookup

### LINKED ISSUE #1293: AM::S trying to use non-ActiveModel::Serializer seriaizers

trek:
Adds an option allowing existing applications to opt out of automatic Serialization lookup.

Two use cases where I've already needed this:

  * Existing applications where another serialization system is stomping on the `.*Serializer` suffix and you need to migrate over to AMS piecemeal 
  * Existing applications that have existing public API where they can't change JSON responses and can't change the existing response structure. E.g. if you a `Tweet` model and want to add a `TweetSerializer` for controllers under the `v2` namespace but not change existing responses for `v1` that are created with a bare `render json: a_tweet`

If this looks good, I'll get a test in.

Closes #1293

### DIFF
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -5,6 +5,7 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## General
 
 - `adapter`: The [adapter](adapters.md) to use. Possible values: `:attributes, :json, :json_api`. Default: `:attributes`.
+- `automatic_lookup`: Whether serializer should be automatically looked up or manually provided. Default: `true`
 
 ## JSON API
 
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -20,6 +20,10 @@ module ActionController
     end
 
     def get_serializer(resource, options = {})
+      unless options[:serializer] || options[:each_serializer] || ActiveModel::Serializer.config.automatic_lookup
+        return resource
+      end
+
       if !use_adapter?
         warn 'ActionController::Serialization#use_adapter? has been removed. '\
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -20,6 +20,7 @@ module ActiveModel
 
         config.adapter = :attributes
         config.jsonapi_resource_type = :plural
+        config.automatic_lookup = true
       end
     end
   end


NullVoxPopuli:
@trek for the second example, you could specify a serializer under v1? Maybe I miss the point.

also, for the second point, I think I would be more in favor of configuring teh suffix, rather than opt out of the lookup. The lookup is very powerful. :-)

docs/general/configuration_options.md
```diff
@@ -5,6 +5,7 @@ The following configuration options can be set on `ActiveModel::Serializer.confi
 ## General
 
 - `adapter`: The [adapter](adapters.md) to use. Possible values: `:attributes, :json, :json_api`. Default: `:attributes`.
+- `automatic_lookup`: Whether serializer should be automatically looked up or manually provided. Default: `true`
```
bf4: actually, you can pass `adapter: false` in your render (or `options[:adapter] = false` if you're outside a controller) and it won't look up a serializer or use an adapter.

ref: 

- https://github.com/rails-api/active_model_serializers/blob/f3403c302c46fd470c2c6073d6cc6c364d9c5bd0/lib/active_model/serializable_resource.rb#L54-L58
- https://github.com/rails-api/active_model_serializers/blob/f3403c302c46fd470c2c6073d6cc6c364d9c5bd0/lib/action_controller/serialization.rb#L23-L40

That should work for you, and would still be useful to include in this PR, to call that out :)

It should also be mentioned in https://github.com/rails-api/active_model_serializers/blob/master/docs/ARCHITECTURE.md, I think

bf4: If that's not sufficient, that's the area of the public API that should be affected.. would love to see a code example of what you're trying to do

NullVoxPopuli: :+1:

bf4: So, https://github.com/rails-api/active_model_serializers/pull/1294 was merged, does that address this issue or should we revert it and continue here?

NullVoxPopuli: It's unclear if that PR addresses the issue here. 
My feeling is that it may be a usage issue (the reason for this PR), but we need more information.

trek:
>  for the second example, you could specify a serializer under v1? Maybe I miss the point.

> also, for the first point, I think I would be more in favor of configuring teh suffix, rather than opt out of the lookup. The lookup is very powerful. :-)

v1 in this example (actually several versions in reality) does not use AMS at all nor do we want to through and ensure every every controller specifically opts out. The automatic lookup being _always on_ is very nice for greenfield projects or projects that can go back and update all their calls to `render`, but makes it painful to adopt AMS into an existing project only for the next API version, leaving the response bodies and code that generates them untouched.

NullVoxPopuli:
I wonder if it would be beneficial to override the lookup hierarchy?

like, if you have 4 versions of an api, maybe you want to fallback to previous versions, so you don't need to specify all the serializers fro each version?

docs/general/configuration_options.md
trek: @bf4 (hey!)

#1294 addresses [1 of the 2 problem cases](https://github.com/rails-api/active_model_serializers/pull/1294#issuecomment-152831189). I think the root problem is that automatic lookup is too aggressive of a strategy for larger apps that want to introduce AMS only for new versions of an existing API. It'd be nice – and frankly _safer_ – to opt out globally and opt-in locally in these cases.

Another riff on this PR is to make that an explicit flag on render calls:

```
render json: photos, use_serializer: true
```

But I already personally prefer supplying the specific serializer to the render call. Yes, it's repetitive, but removes the "where the heck is this object turning into _this_ JSON?!?" problem.

bf4: Right now AMS automatically includes the ActionController::Serialization mixin.  What if you could opt out of that and mix it in to controllers as desired. Would that work?  (As opposed to passing in `adapter: false` in most of your controllers)

lib/action_controller/serialization.rb
```diff
@@ -20,6 +20,10 @@ def serialization_scope
     end
 
     def get_serializer(resource, options = {})
+      unless options[:serializer] || options[:each_serializer] || ActiveModel::Serializer.config.automatic_lookup
+        return resource
+      end
+
       if !use_adapter?
```
bf4: I suppose you could also make an argument to bring back `use_adapter?`

bf4: (or add `ActiveModel::Serializer.config.automatic_lookup` as `config.implicit_serializers = false` which would used probably in the serializable resource

docs/general/configuration_options.md
NullVoxPopuli: @bf4 are you think that could be an initialization configuration option? something like:

include_action_controller_serialization = true (by default)

?

bf4:
What do you think of setting `config.implicit_serializers = false` ?

NullVoxPopuli:
I could get behind that, as long as it's documented an well

trek:
Sorry, I don't quite understand the feedback. Are we just bikeshedding the config name or am I missing something?

bf4:
@trek I don't think so.  I'm thinking there are a few good solutions, not exclusive

- Turn off implicit lookup, so that resources are passed through unchanged when no serializer is specified. Something like `ActiveModel::Serializer.config.implicit_serializers = false` (affects: call to [`ActiveModel::Serializer.serializer_for` in SerializableResource](https://github.com/rails-api/active_model_serializers/blob/bca72207ff5a132b6c509ba65f4302a90723b157/lib/active_model/serializable_resource.rb#L48))
- Pass through resource unchanged by setting default adapter to `config.adapter = false`. Using AMS will then require an explicit adapter option to render in any controller. (impact: [`SerializableResource#use_adapter?`](https://github.com/rails-api/active_model_serializers/blob/bca72207ff5a132b6c509ba65f4302a90723b157/lib/active_model/serializable_resource.rb#L58-L66)
- Prevent automatic mixing in of `ActionController::Serialization` into ActionController::Base with some config, maybe `config.action_controller.include = false` (impact: [the railtie](https://github.com/rails-api/active_model_serializers/blob/f3403c302c46fd470c2c6073d6cc6c364d9c5bd0/lib/active_model/serializer/railtie.rb), [2](https://github.com/rails-api/active_model_serializers/blob/f3403c302c46fd470c2c6073d6cc6c364d9c5bd0/lib/active_model_serializers.rb#L52-L59))

I did bring up different names, but implementation-wise, I think the switch should be somewhere besides [`ActionController::Serialization#get_serializer`](https://github.com/rails-api/active_model_serializers/blob/0820d2f4bf5d42628332985a71613eb7d6d5c5ed/lib/action_controller/serialization.rb#L22-L25).

Thoughts?

bf4:
Interestingly, this has come up before: https://github.com/rails-api/active_model_serializers/pull/592 (and https://github.com/rails-api/active_model_serializers/pull/611 )

docs/general/configuration_options.md
bf4: @trek re:

> render json: photos, use_serializer: true

In https://github.com/rails/rails/pull/21496/ I have proposed [removing `use_renderers` since it appears never to have been used or tested](https://github.com/rails/rails/commit/556344639beeed059c1e8c0c27df18c6ef2efdc8).  Since your suggestion here has a similar name, I was wondering if it might be related to the `use_renderers` method in the Rails Render, or just coincidental

trek:
Closing in favor of #1353

PR #1297, opened 2015-10-24
Fix `fields` option to restrict relationships as well.

beauby:
The current implementation breaks the JSON API spec in the way it handles [sparse fieldsets](http://jsonapi.org/format/#fetching-sparse-fieldsets), since the definition of [fields](http://jsonapi.org/format/#document-resource-object-fields) is:
> A resource object's attributes and its relationships are collectively called its "fields".

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -44,13 +44,7 @@ module ActiveModel
         def initialize(serializer, options = {})
           super
           @include_tree = IncludeTree.from_include_args(options[:include])
-
-          fields = options.delete(:fields)
-          if fields
-            @fieldset = ActiveModel::Serializer::Fieldset.new(fields)
-          else
-            @fieldset = options[:fieldset]
-          end
+          @fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(options.delete(:fields))
         end
 
         def serializable_hash(options = nil)
@@ -175,7 +169,7 @@ module ActiveModel
 
         def relationships_for(serializer)
           resource_type = resource_identifier_type_for(serializer)
-          requested_associations = fieldset.try(:fields_for, resource_type) || '*'
+          requested_associations = fieldset.fields_for(resource_type) || '*'
           include_tree = IncludeTree.from_include_args(requested_associations)
           serializer.associations(include_tree).each_with_object({}) do |association, hash|
             hash[association.key] = { data: relationship_value_for(association.serializer, association.options) }
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -174,7 +174,10 @@ module ActiveModel
         end
 
         def relationships_for(serializer)
-          serializer.associations.each_with_object({}) do |association, hash|
+          resource_type = resource_identifier_type_for(serializer)
+          requested_associations = fieldset.try(:fields_for, resource_type) || '*'
+          include_tree = IncludeTree.from_include_args(requested_associations)
+          serializer.associations(include_tree).each_with_object({}) do |association, hash|
             hash[association.key] = { data: relationship_value_for(association.serializer, association.options) }
           end
         end
--- a/lib/active_model/serializer/fieldset.rb
+++ b/lib/active_model/serializer/fieldset.rb
@@ -2,7 +2,7 @@ module ActiveModel
   class Serializer
     class Fieldset
       def initialize(fields)
-        @raw_fields = fields
+        @raw_fields = fields || {}
       end
 
       def fields
@@ -21,7 +21,7 @@ module ActiveModel
 
       def parsed_fields
         if raw_fields.is_a?(Hash)
-          raw_fields.inject({}) { |h, (k, v)| h[k.to_sym] = v.map(&:to_sym); h }
+          raw_fields.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
         else
           {}
         end
--- a/test/adapter/json_api/belongs_to_test.rb
+++ b/test/adapter/json_api/belongs_to_test.rb
@@ -56,7 +56,7 @@ module ActiveModel
           end
 
           def test_limiting_linked_post_fields
-            @adapter = ActiveModel::Serializer::Adapter::JsonApi.new(@serializer, include: [:post], fields: { post: [:title] })
+            @adapter = ActiveModel::Serializer::Adapter::JsonApi.new(@serializer, include: [:post], fields: { post: [:title, :comments, :blog, :author] })
             expected = [{
               id: '42',
               type: 'posts',
--- a/test/adapter/json_api/collection_test.rb
+++ b/test/adapter/json_api/collection_test.rb
@@ -60,7 +60,7 @@ module ActiveModel
           def test_limiting_fields
             actual = ActiveModel::SerializableResource.new(
               [@first_post, @second_post], adapter: :json_api,
-                                           fields: { posts: ['title'] })
+                                           fields: { posts: %w(title comments blog author) })
                      .serializable_hash
             expected = [
               {
--- a/test/adapter/json_api/fields_test.rb
+++ b/test/adapter/json_api/fields_test.rb
@@ -1,0 +1,89 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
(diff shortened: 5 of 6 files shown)


lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -174,7 +174,10 @@ def relationship_value_for(serializer, options = {})
         end
 
         def relationships_for(serializer)
-          serializer.associations.each_with_object({}) do |association, hash|
+          resource_type = resource_identifier_type_for(serializer)
+          requested_associations = fieldset.try(:fields_for, resource_type) || '*'
```
NullVoxPopuli: in what situation does a fieldset not have fields_for?

NullVoxPopuli: I guess, I'd like some docs here explaining why this is happening :-)
I'll be ok with merging once that's done. :-)

beauby: When fieldset is not initialized (which currently happens when the `fields` option is not provided).

beauby: Docs explaining why what is happening?

NullVoxPopuli: the situation you just described with fieldset.
It's important to have docs like this when someone new starts working on the json api adapter.

beauby: Agreed, but I'd rather have it in an `INTERNAL.md` than in the code. Plus, I'd rather have the @fieldset initialized by default. Which I'm going to add to this PR actually.

beauby: Plus I just realized this PR introduces a bug. Fixing it right now.

beauby: Wait, no, it does not actually.

NullVoxPopuli: You'd rather update docs in a separate file than inline? that seems like a ton of work that isn't going to be kept up with. :-\

beauby: I believe code should be clear and self-explanatory. If a new contributor would be confused, then I'd rather change the code to something clearer than add comments. ["Good code is its own best documentation"](https://github.com/bbatsov/ruby-style-guide#no-comments)

NullVoxPopuli: agreed, but until then, it doesn't hurt to spend 10 seconds writing some :-)

NullVoxPopuli: also, that's not an excuse to get out of writing docs.

beauby: Code comments are not docs in my opinion. Maybe we're not talking about the same thing?

NullVoxPopuli: perhaps. 
I'm not meaning to write something that just anyone would look at. but a reminder of what is happening in the code, possibly the scenarios in which certain control flows are followed - so that even the author of the code can have an easier time re-familiarizing with the code 1+ years from when the code was written.
node of that is intended to be in the public documentation that would end up on ruby-doc.org or anything.

lib/active_model/serializer/fieldset.rb
```diff
@@ -21,7 +21,7 @@ def fields_for(type)
 
       def parsed_fields
         if raw_fields.is_a?(Hash)
-          raw_fields.inject({}) { |h, (k, v)| h[k.to_sym] = v.map(&:to_sym); h }
+          raw_fields.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
```
beauby: This (isomorphic) change is not necessary for this PR, but it makes the code slightly cleaner.

lib/active_model/serializer/adapter/json_api.rb
beauby: Then I still disagree. Code that needs explanations is just bad code in my opinion. If you think this bit is obscure, then I'd rather we work to make it less obscure than write a comment to explain it.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -44,13 +44,7 @@ def object
...
           super
           @include_tree = IncludeTree.from_include_args(options[:include])
-
-          fields = options.delete(:fields)
-          if fields
-            @fieldset = ActiveModel::Serializer::Fieldset.new(fields)
-          else
-            @fieldset = options[:fieldset]
-          end
+          @fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(options.delete(:fields))
```
NullVoxPopuli: nice!

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -174,7 +168,10 @@ def relationship_value_for(serializer, options = {})
         end
 
         def relationships_for(serializer)
-          serializer.associations.each_with_object({}) do |association, hash|
+          resource_type = resource_identifier_type_for(serializer)
+          requested_associations = fieldset.fields_for(resource_type) || '*'
```
NullVoxPopuli: I like that the try is gone :-)

bf4:
@beauby @NullVoxPopuli can you help me understand why the order of the fieldset was changed? Also, for a bit I thought the `:fields` option was deleted due to the inlining of the conditional, fwiw.

```diff
-          fields = options.delete(:fields)
-          if fields
-            @fieldset = ActiveModel::Serializer::Fieldset.new(fields)
-          else
-            @fieldset = options[:fieldset]
-          end
+          @fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(options.delete(:fields))
``

Is really the same as

```diff
-          fields = options.delete(:fields)
-          if fields
-            @fieldset = ActiveModel::Serializer::Fieldset.new(fields)
-          else
-            @fieldset = options[:fieldset]
-          end
+         @fieldset = options[:fieldset]
+         if @fieldset.nil?
+           fields = options.delete(:fields))
+           @fieldset = ActiveModel::Serializer::Fieldset.new(fields || {}) # where the fieldset initializer was modified as paraphrased here
```

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -174,7 +168,10 @@ def relationship_value_for(serializer, options = {})
...
+          include_tree = IncludeTree.from_include_args(requested_associations)
+          serializer.associations(include_tree).each_with_object({}) do |association, hash|
```
bf4: In words, what does this diff do?

```ruby
-          serializer.associations.each_with_object({}) do |association, hash|
+          resource_type = resource_identifier_type_for(serializer)
+          requested_associations = fieldset.fields_for(resource_type) || '*'
+          include_tree = IncludeTree.from_include_args(requested_associations)
+          serializer.associations(include_tree).each_with_object({}) do |association, hash|
```

It's supposed to somehow remove associations somehow?

lib/active_model/serializer/adapter/json_api.rb
bf4: Um.. I don't understand the change from the code :)

bf4:
Would like to promote discussion of optimizing code to not make me think: https://github.com/rails-api/active_model_serializers/pull/1297#discussion_r42948318 :smile:

beauby:
@bf4 Any suggestion welcome.

lib/active_model/serializer/adapter/json_api.rb
beauby: It filters out associations not present in the fieldset of the resource (unless no fieldset has been provided for that resource).

bf4: maybe it's just the method and variable names. I can be pretty literal-minded and it looks like whereas before we would iterate over the associations, now we ask the serializer what it's `type` ask the fieldset for its fields, or use `*`, which guess means ALL, the include_tree is something that is passed to associations?

beauby: Exactly. The `IncludeTree` is probably not a perfect name. It is basically a structure that describes a labelled tree in such a way that you can query it using wildcards.

beauby: `serializer.associations(include_tree)` this filters associations

NullVoxPopuli: @beauby I guess this means that an explanation needs to be in code comments. :-)

PR #1299, opened 2015-10-25
add documentation for the type setting on the json api adapters

NullVoxPopuli:


docs/howto/jsonapi/custom_type_value.md
```diff
@@ -0,0 +1,25 @@
+# How to customize the type of a JSON API document
+
+To change the type of a resource, just use the `type` class method in your serializer definition.
+
+Note that by default, type is pluralized.
```
beauby: This is ambiguous. It seems to imply that the `type`, when specified through the `type` method will be subject to inflection.

NullVoxPopuli: when I just used type, if you specify 'post' it comes out as 'post' in the json -- so whatever is passed  to the type method is what gets rendered. So, I do think that it's worth mentioning. maybe it could be worded better, though.

bf4: Docs are misleading. type is the JSON API resource type, not document. document's don't have a type

beauby: @bf4 Not sure I understand. Where does it talk about document's type?

docs/README.md
```diff
@@ -13,8 +13,11 @@ This is the documentation of AMS, it's focused on the **0.10.x version.**
 ## How to
 
 - [How to add root key](howto/add_root_key.md)
-- [How to add pagination links](howto/add_pagination_links.md)
 - [Using AMS Outside Of Controllers](howto/outside_controller_use.md)
+- [How to add pagination links](howto/add_pagination_links.md)
+
+#### Specific to JSON API
+- [Customizing the type](howto/jsonapi/custom_type_value.md)
```
bf4: #### JSON API Resource `type`

##### Implicit

Resource class name is used, e.g. ....

##### Explicitly

In Serializer, declare `type foo`

beauby: :+1:

NullVoxPopuli: well, that's certainly less lines :-)

docs/README.md
```diff
@@ -13,8 +13,18 @@ This is the documentation of AMS, it's focused on the **0.10.x version.**
...
+
+#### JSON API Resource `type`
+
+##### Implicit
+
+Resource class name is used, e.g. the type of a `Post` will be `posts` in the resulting json api resource.
+
+##### Explicitly
+
+In the Serializer, declare `type foo`, and `"type": "foo"` will be present in the corresponding resulting json api resource.
```
bf4: This makes me think we really need a page that describes how  we build (and configure) different parts of a JSON API document.  This would be under Resource. 

And also that we might want to specialize a JSON API Resource vs. JSON Resource (which what I'd like to rename AM::S to), like in https://github.com/cerebris/jsonapi-resources

beauby: I've been thinking about specializing resources (our current serializers) depending on the adapter, but that breaks the possibility to "just switch adapters" (which arguably is not that useful to start with, but still).

docs/howto/jsonapi/custom_type_value.md
bf4: The h1

B mobile phone

beauby: Oh, right.

docs/README.md
NullVoxPopuli: so, any particular action items?

I can close this PR, if maybe @beauby wants to include this information in a broader json api resource document. 

Otherwise, I think it's somewhat beneficial in the short term

bf4: `docs/jsonapi/README.md` maybe

beauby: Consistency: `JSON API` instead of `json api`.

docs/README.md
```diff
@@ -13,8 +13,18 @@ This is the documentation of AMS, it's focused on the **0.10.x version.**
...
 - [How to add root key](howto/add_root_key.md)
-- [How to add pagination links](howto/add_pagination_links.md)
 - [Using AMS Outside Of Controllers](howto/outside_controller_use.md)
+- [How to add pagination links](howto/add_pagination_links.md)
+
+#### JSON API Resource `type`
+
+##### Implicit
+
+Resource class name is used, e.g. the type of a `Post` will be `posts` in the resulting json api resource.
```
beauby: Maybe expand a bit, for instance with the generated `type` for `Namespace::MultiWordResource` or such.

docs/README.md
NullVoxPopuli: done

docs/README.md
NullVoxPopuli: done

NullVoxPopuli:
rebased

docs/README.md
```diff
@@ -13,9 +13,19 @@ This is the documentation of AMS, it's focused on the **0.10.x version.**
...
+
+#### JSON API Resource `type`
+
+##### Implicit
+
+Resource class name is used, e.g. the type of a `Post` will be `posts` in the resulting JSON API resource. For namespaced and multiword classes, AMS will use underscores to seperate words and namespaces. e.g.: `Namespaced::MultiWordClass` becomes `namespaced_multi_word_class`.
+
+##### Explicitly
+
+In the Serializer, declare `type foo`, and `"type": "foo"` will be present in the corresponding resulting JSON API resource.
```
beauby: > declare `type 'foo'`
(quotes).

Except for that, I think we should merge this.

NullVoxPopuli: updated!

beauby: Is it?

docs/README.md
```diff
@@ -13,9 +13,19 @@ This is the documentation of AMS, it's focused on the **0.10.x version.**
...
 - [Using AMS Outside Of Controllers](howto/outside_controller_use.md)
 - [How to use JSON API with Ember](howto/ember-and-json-api.md)
+- [How to add pagination links](howto/add_pagination_links.md)
+
+#### JSON API Resource `type`
+
+##### Implicit
+
+Resource class name is used, e.g. the type of a `Post` will be `posts` in the resulting JSON API resource. For namespaced and multiword classes, AMS will use underscores to seperate words and namespaces. e.g.: `Namespaced::MultiWordClass` becomes `namespaced_multi_word_class`.
+
```
bf4: right now it uses `::` -> `/` but per discussion in jSON Api land I think it should be `__` since `/` is illegal per JSON API

NullVoxPopuli: Is there a test you can point me to that verifies that?

bf4: https://github.com/json-api/json-api/issues/850#issuecomment-144849570

bf4:
@NullVoxPopuli please rebase against master :)

groyoh:
Please take a look at: https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#type

NullVoxPopuli:
@groyoh cool, this PR isn't needed then

Issue #1298, opened 2015-10-24
RFC: Name-spacing of objects under (module) ActiveModelSerializers vs. (core class) ActiveModel::Serializer


maurogeorge:
Hi guys,

Today most of the public API is under `ActiveModel::Serializer` like:

- `ActiveModel::Serializer.config`
- `ActiveModel::Serializer`, which is the base class of our serializer

I think the only 2 public APIs that are under the `ActiveModelSerializers` is `ActiveModelSerializers::Model` and `ActiveModelSerializers.logger`, I think this 2 APIs need to live under `ActiveModel::Serializer` to keep consistency to the user. I created 2 PRs that is related with this https://github.com/rails-api/active_model_serializers/pull/1291 https://github.com/rails-api/active_model_serializers/pull/1296

In this PR https://github.com/rails-api/active_model_serializers/pull/1138 was introduced a `ActiveModel::Adapter::Base`. I think we can do something like that, move from a class to a module and create a `Base` class to follow the Rails idiom.

The main goal here is to keep a good public API so the user will use:

```
ActiveModel::Serializer::Base # When creating a AR Serializer
ActiveModel::Serializer::Model # When creating a PORO Serializer
ActiveModel::Serializer.logger # To define the logger
```

I want to hear from you guys, your thought about it? You are :+1: or :-1:

beauby:
Assigning @bf4 to this one as I believe you made those 2 PRs (`AMS::Model` and the logger). Could you explain the rationale behind exposing those via `ActiveModelSerializers` rather than `ActiveModel::Serializer`?

bf4:
Basically, ActiveModel::Serializer is a terrible name for what it is and I'd like to move away from it.  It's both the gem name, a namespace, and a resource attribute mapper.

Also, using a class as a namespace for unrelated things isn't very Ruby-friendly.

I can write more, but does that make sense?

beauby:
@bf4 I get what you're saying. Maybe it's still worth going into more details here so that we can keep a written trace of the exact reasons why.

bf4:
I'm going to try to write up something tonight that is json api focused that higkights the mismatching of terms

But also, activemodel:: seializers right now attracts data and behavior not related to it being a serializer class since it is also a namespace that sounds like the namenof the gem
B mobile phone

maurogeorge:
Thanks for the comments guys.

[...]

PR #1301, opened 2015-10-26
Mapping JSON API spec / schema to AMS [ci skip]

bf4:
This is something to think about. I'm not 100% sure if it should be
in the docs or in an issue, than a PR. But, I think it might be useful for discussion
As we work on [JSON API](https://github.com/rails-api/active_model_serializers/issues/1235).

[View formatted](https://github.com/rails-api/active_model_serializers/blob/jsonapi_schema/docs/jsonapi/schema.md)

1. Trying to surface which parts of the code map to the JSON API spec
2. Use point 1 to consider design and naming of objects
3. Track completeness of implementation
4. TODO: Use schema/schema.json to validate and document our implementation
  - https://github.com/rails-api/active_model_serializers/pull/1270
  - https://github.com/rails-api/active_model_serializers/issues/1162
  - maybe use prmd? http://www.sitepoint.com/document-your-json-api-schema-with-prmd/
5. TODO: use uri_template in link generation? use json pointer gem? 
  - https://github.com/rails-api/active_model_serializers/pull/1282#discussion_r42528812
  - > I believe what you want here is a uri_template. See https://github.com/hannesg/uri_template See e.g. https://developer.github.com/v3/#schema http://tools.ietf.org/html/rfc6570 see e.g. https://api.github.com/ (though not a JSON API, is an example of usage)

There are for sure formatting and possible correctness issues. I mean it as a point of discussion.

PR #1296, opened 2015-10-24
Move from ActiveModelSerializers::Model to ActiveModel::Serializer::Model

maurogeorge:
Hi guys,

Today the `ActiveModelSerializers::Model` is the only public API under this namespace.

To keep consistency move to `ActiveModel::Serializer::Model`.

Makes sense to you?

Thanks

PR #1259, opened 2015-10-09
Add more info to CONTRIBUTING

bf4:
cc @rails-api/ams 

needs a second set of eyes and some reworking

PR #1310, opened 2015-10-29
RFC: Primary namespace: ActiveModel::Serializer, ActiveModel::Serializers, vs. ActiveModelSerializers

maurogeorge:
[Rendered](https://github.com/maurogeorge/active_model_serializers/blob/patch-12/docs/rfcs/0000-namespace.md).

Following the https://github.com/rails-api/active_model_serializers/issues/1298#issuecomment-151290661 create a RFC about namespace.

I created the RFC based on the [Ember RFCs](https://github.com/emberjs/rfcs).

Thanks for the feedback :heart:

### DIFF
--- a/docs/rfcs/0000-namespace.md
+++ b/docs/rfcs/0000-namespace.md
@@ -1,0 +1,115 @@
+- Start Date: (2015-10-29)
+- RFC PR: (leave this empty)
+- AMS Issue: (leave this empty)
+
+# Summary
+
+Provide a consistent API for the user of the AMS.
+
+# Motivation
+
+The actual public API is defined under `ActiveModelSerializers`,
+`ActiveModel::Serializer` and `ActiveModel`.
+
+At the `ActiveModel::Serializer` we have:
+
+- `ActiveModel::Serializer.config`
+- `ActiveModel::Serializer`
+
+At the `ActiveModelSerializers` we have:
+
+- `ActiveModelSerializers::Model`
+- `ActiveModelSerializers.logger`
+
+At `ActiveModel` we have:
+
+- `ActiveModel::SerializableResource`
+
+The idea here is to provide a single namespace `ActiveModel::Serializers` to the user.
+Following the same idea we have on other gems like
+[Devise](https://github.com/plataformatec/devise/blob/e9c82472ffe7c43a448945f77e034a0e47dde0bb/lib/devise.rb),
+[Refile](https://github.com/refile/refile/blob/6b24c293d044862dafbf1bfa4606672a64903aa2/lib/refile.rb) and
+[Active Job](https://github.com/rails/rails/blob/30bacc26f8f258b39e12f63fe52389a968d9c1ea/activejob/lib/active_job.rb)
+for example.
+
+# Detailed design
+
+## Require statement and main module
+
+We are adding a extension for the Active Model, so
+[following the Rubygens recomendation](http://guides.rubygems.org/name-your-gem/)
+for the gem name we need to change to this.
+
+|Gem name                  | Require statement          | Main class or module     |
+|--------------------------|----------------------------|--------------------------|
+| active_model_serializers | `active_model/serializers` | ActiveModel::Serializers |
+
+The expected gem name, in the gemspec is `active_model-serializers` but we don't
+need to change this, we can change the code without the need of a new gem on Rubygems.
+
+Active Model for example follow the same idea the gem name on gemspec is `activemodel` and to the end user is:
+
+|Gem name                  | Require statement          | Main class or module     |
+|--------------------------|----------------------------|--------------------------|
+| activemodel              | `active_model`             | ActiveModel              |
+
+As you can see we do not require `activemodel`(the gem name in gemspec) insted
+we use `active_model`.
+
+And based on the [bump of `0.10.0.pre` released by Steve Klabnik](https://github.com/rails-api/active_model_serializers/tree/86fc7d7227f3ce538fcb28c1e8c7069ce311f0e1)
+if we take a look on README and gemspec always is used `ActiveModel::Serializers`.
+
+## New classes and modules organization
+
+Since this will be a big change we can do this on baby steps, read small PRs. A
+possible approach is:
+
+- Create the `ActiveModel::Serializers` namespace;
+- Move all content under `ActiveModelSerializers` to be under
+  `ActiveModel::Serializers`, the logger is on this step;
+- Move all content under `ActiveModel::Serializer` to be under
+  `ActiveModel::Serializers`, the adapter is on this steps;
+- Move all content under `ActiveModel` to be under `ActiveModel::Serializers`,
+  the `SerializableResource` is on this step;
+- Now that all the code lives under `ActiveModel::Serializers` we can:
+  - create a better name to the `ActiveModel::Serializers::Serializer`
+    keeping in mind only to keep this in the same namespace
+  - create a better name to the `ActiveModel::Serializers::Serializer::Adapter::JsonApi`
+    probably remove this from the `ActiveModel::Serializers::Serializer`
+    and do something like `ActiveModel::Serializers::Adapter::JsonApi`
+    keeping in mind only to keep this in the same namespace
+  - Change all public API that doesn't make sense, keeping in mind only to keep
+    this in the same namespace
+- Update the README;
+- Update the docs;
+
+The following table represents the current and the desired classes and modules
+at the first moment.
+
+| Current                                               | Desired                                          | Notes              |
+|-------------------------------------------------------|--------------------------------------------------|--------------------|
+|`ActiveModelSerializers` and `ActiveModel::Serializer` | `ActiveModel::Serializers`                       | The main namespace |
+| `ActiveModelSerializers.logger`                       | `ActiveModel::Serializers.logger`                ||
+|`ActiveModelSerializers::Model`                        | `ActiveModel::Serializers::Model`                ||
+|`ActiveModel::SerializableResource`                    | `ActiveModel::Serializers::SerializableResource` ||
+| `ActiveModel::Serializer`                             | `ActiveModel::Serializers::Serializer`           | I know that is probably a bad name, but In a second moment we can rename this to `Resource` [for example following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185)|
+|`ActiveModel::Serializer.config`                       | `ActiveModel::Serializers.config`                ||
+
+# Drawbacks
+
+This will be a breaking change, so all users serializers will be broken.
+All PRs will need to rebase since the architeture will change a lot.
+
+# Alternatives
+
+We can keep the way it is, and keep in mind to not add another namespace as a
+public API.
+
+Or we can start moving the small ones that seems to be the
+`ActiveModelSerializers` and `ActiveModel` and later we can handle the
+`ActiveModel::Serializer`.
+
+# Unresolved questions
+
+What is the better class name to be used to the class that will be inherited at
+the creation of a serializer.
--- a/docs/rfcs/0000-namespace.md
+++ b/docs/rfcs/0000-namespace.md
@@ -1,6 +1,6 @@
 - Start Date: (2015-10-29)
-- RFC PR: (leave this empty)
-- AMS Issue: (leave this empty)
+- RFC PR: https://github.com/rails-api/active_model_serializers/pull/1310
+- ActiveModelSerializers Issue: https://github.com/rails-api/active_model_serializers/issues/1298
 
 # Summary
 
@@ -25,91 +25,82 @@ At `ActiveModel` we have:
 
 - `ActiveModel::SerializableResource`
 
-The idea here is to provide a single namespace `ActiveModel::Serializers` to the user.
+The idea here is to provide a single namespace `ActiveModelSerializers` to the user.
 Following the same idea we have on other gems like
 [Devise](https://github.com/plataformatec/devise/blob/e9c82472ffe7c43a448945f77e034a0e47dde0bb/lib/devise.rb),
 [Refile](https://github.com/refile/refile/blob/6b24c293d044862dafbf1bfa4606672a64903aa2/lib/refile.rb) and
 [Active Job](https://github.com/rails/rails/blob/30bacc26f8f258b39e12f63fe52389a968d9c1ea/activejob/lib/active_job.rb)
 for example.
 
+This way we are clarifing the boundaries of
+[ActiveModelSerializers and Rails](https://github.com/rails-api/active_model_serializers/blob/master/CHANGELOG.md#prehistory)
+and make clear that the `ActiveModel::Serializer` class is no longer the primary
+behavior of the ActiveModelSerializers.
+
 # Detailed design
-
-## Require statement and main module
-
-We are adding a extension for the Active Model, so
-[following the Rubygens recomendation](http://guides.rubygems.org/name-your-gem/)
-for the gem name we need to change to this.
-
-|Gem name                  | Require statement          | Main class or module     |
-|--------------------------|----------------------------|--------------------------|
-| active_model_serializers | `active_model/serializers` | ActiveModel::Serializers |
-
-The expected gem name, in the gemspec is `active_model-serializers` but we don't
-need to change this, we can change the code without the need of a new gem on Rubygems.
-
-Active Model for example follow the same idea the gem name on gemspec is `activemodel` and to the end user is:
-
-|Gem name                  | Require statement          | Main class or module     |
-|--------------------------|----------------------------|--------------------------|
-| activemodel              | `active_model`             | ActiveModel              |
-
-As you can see we do not require `activemodel`(the gem name in gemspec) insted
-we use `active_model`.
-
-And based on the [bump of `0.10.0.pre` released by Steve Klabnik](https://github.com/rails-api/active_model_serializers/tree/86fc7d7227f3ce538fcb28c1e8c7069ce311f0e1)
-if we take a look on README and gemspec always is used `ActiveModel::Serializers`.
 
 ## New classes and modules organization
 
-Since this will be a big change we can do this on baby steps, read small PRs. A
+Since this will be a big change we can do this on baby steps, read small pull requests. A
 possible approach is:
 
-- Create the `ActiveModel::Serializers` namespace;
-- Move all content under `ActiveModelSerializers` to be under
-  `ActiveModel::Serializers`, the logger is on this step;
+- All new code will be in `lib/active_model_serializers/` using
+  the module namespace `ActiveModelSerializers`.
 - Move all content under `ActiveModel::Serializer` to be under
-  `ActiveModel::Serializers`, the adapter is on this steps;
-- Move all content under `ActiveModel` to be under `ActiveModel::Serializers`,
+  `ActiveModelSerializers`, the adapter is on this steps;
+- Move all content under `ActiveModel` to be under `ActiveModelSerializers`,
   the `SerializableResource` is on this step;
-- Now that all the code lives under `ActiveModel::Serializers` we can:
-  - create a better name to the `ActiveModel::Serializers::Serializer`
-    keeping in mind only to keep this in the same namespace
-  - create a better name to the `ActiveModel::Serializers::Serializer::Adapter::JsonApi`
-    probably remove this from the `ActiveModel::Serializers::Serializer`
-    and do something like `ActiveModel::Serializers::Adapter::JsonApi`
-    keeping in mind only to keep this in the same namespace
-  - Change all public API that doesn't make sense, keeping in mind only to keep
-    this in the same namespace
+- Change all public API that doesn't make sense, keeping in mind only to keep
+  this in the same namespace
 - Update the README;
 - Update the docs;
 
 The following table represents the current and the desired classes and modules
 at the first moment.
 
-| Current                                               | Desired                                          | Notes              |
-|-------------------------------------------------------|--------------------------------------------------|--------------------|
-|`ActiveModelSerializers` and `ActiveModel::Serializer` | `ActiveModel::Serializers`                       | The main namespace |
-| `ActiveModelSerializers.logger`                       | `ActiveModel::Serializers.logger`                ||
-|`ActiveModelSerializers::Model`                        | `ActiveModel::Serializers::Model`                ||
-|`ActiveModel::SerializableResource`                    | `ActiveModel::Serializers::SerializableResource` ||
-| `ActiveModel::Serializer`                             | `ActiveModel::Serializers::Serializer`           | I know that is probably a bad name, but In a second moment we can rename this to `Resource` [for example following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185)|
-|`ActiveModel::Serializer.config`                       | `ActiveModel::Serializers.config`                ||
+| Current                                                | Desired                                          | Notes              |
+|--------------------------------------------------------|--------------------------------------------------|--------------------|
+| `ActiveModelSerializers` and `ActiveModel::Serializer` | `ActiveModelSerializers`                         | The main namespace |
+| `ActiveModelSerializers.logger`                        | `ActiveModelSerializers.logger`                  ||
+| `ActiveModelSerializers::Model`                        | `ActiveModelSerializers::Model`                  ||
+| `ActiveModel::SerializableResource`                    | `ActiveModelSerializers::SerializableResource`   ||
+| `ActiveModel::Serializer`                              | `ActiveModelSerializers::Serializer`             | The name can be discussed in a future pull request. For example, we can rename this to `Resource` [following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185) more info about naming in the next section|
+| `ActiveModel::Serializer.config`                       | `ActiveModelSerializers.config`                ||
+
+## Renaming of class and modules
+
+When moving some content to the new namespace we can find some names that does
+not make much sense like `ActiveModelSerializers::Serializer::Adapter::JsonApi`.
+Discussion of renaming existing classes / modules and JsonApi objects will
+happen in separate pull requests, and issues, and in the google doc
+https://docs.google.com/document/d/1rcrJr0sVcazY2Opd_6Kmv1iIwuHbI84s1P_NzFn-05c/edit?usp=sharing
+
+Some of names already have a definition.
+
+- Adapters get their own namespace under ActiveModelSerializers. E.g
+  `ActiveModelSerializers::Adapter`
+- Serializers get their own namespace under ActiveModelSerializers. E.g
+  `ActiveModelSerializers::Serializer`
+
+## Keeping compatibility
+
+All moved classes or modules be aliased to their old name and location with
+deprecation warnings, such as
+[was done for CollectionSerializer](https://github.com/rails-api/active_model_serializers/pull/1251).
 
 # Drawbacks
 
-This will be a breaking change, so all users serializers will be broken.
-All PRs will need to rebase since the architeture will change a lot.
+This will be a breaking change, so all users serializers will be broken after a
+major bump.
+All pull requests will need to rebase since the architeture will change a lot.
 
 # Alternatives
 
 We can keep the way it is, and keep in mind to not add another namespace as a
 public API.
 
-Or we can start moving the small ones that seems to be the
-`ActiveModelSerializers` and `ActiveModel` and later we can handle the
-`ActiveModel::Serializer`.
-
 # Unresolved questions
 
 What is the better class name to be used to the class that will be inherited at
-the creation of a serializer.
+the creation of a serializer. This can be discussed in other RFC or directly via
+pull request.


docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Detailed design
+
+We are adding a extension for the Active Model, so
+[following the Rubygens recomendation](http://guides.rubygems.org/name-your-gem/)
+for the gem name we need to change to this.
+
+|Gem name                  | Require statement          | Main class or module     |
+|--------------------------|----------------------------|--------------------------|
+| ActiveModel::Serializers | `active_model/serializers` | ActiveModel::Serializers |
```
bf4: major :-1: this. will. cause. pain. Plus, we all agree, I think, that serializers has got to go as the main namespace.

beauby: @bf4 When you get a second, could you expand on the kind of pain this will cause?

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,76 @@
...
+- AMS Issue: (leave this empty)
+
+# Summary
+
+Provide a consistent API for the user of the AMS.
+
+# Motivation
+
+The actual public API is defined under `ActiveModelSerializers`,
+`ActiveModel::Serializer` and `ActiveModel`.
```
bf4: reposting https://github.com/rails-api/active_model_serializers/issues/1298#issuecomment-151267763 here

So, I haven't thought it all the way through, and I just wrote it out without any review, but what I'm thinking right now is something like:

| Current | Desired | Notes |
|----|---|------|
| require 'active_model_serializers' | require 'active_model_serializers'
| ActiveModelSerializers | ActiveModelSerializers 
| ActiveModel::SerializableResource | ActiveModel::SerializableResource
| namespace ActiveModel::Serializer | ActiveModelSerializers::Current, ActiveModelSerializers::V08, etc | versioned
| class ActiveModel::Serializer  | ActiveModelSerializers::Current::Resource | is a class that defines the properties and behavior of the data that you present to the user
|  ActiveModelSerializers::Model |  ActiveModelSerializers::Model | e.g. ActiveRecord object but a PORO. Maybe should be ActiveModelSerializers::Record A record is an instance of a model that contains data loaded from a server. Your application can also create new records and save them back to the server.
| ActiveModel::Serializer::Lint | ActiveModelSerializers::Model::Lint
| ActiveModel::Serialization | ActiveModelSerializers::ActionController::Serialization 
| ActiveModel::Serializer::CollectionSerializer | ActiveModelSerializers::Current::ResourceCollection |
| namespace ActiveModel::Serializer::Adapter | ActiveModelSerializers::Adapter |   adapter: knows how to talk to your server. knows how to translate requests from AMS into requests on your server.  object that translates requests from AMS/Rails/Grape etc (such as "find the user with an ID of 123") into a requests to a server.  let you completely change how your API is implemented without impacting your application code.
| class ActiveModel::Serializer::Adapter::JsonApi | ActiveModelSerializers::Adapter::JsonApi, ActiveModelSerializers::Adapter::V08::Json | not versioned
| class ActiveModel::Serializer, ActiveModel::Serializer::Adapter::Attributes | ActiveModelSerializers::Formatter | maps keys and values to desired format, see https://github.com/orbitjs/orbit.js/blob/master/lib/orbit-common/jsonapi-serializer.js
| N/A |  ActiveModelSerializers::Current::Store | store is the central repository of records in your application, use the store to retrieve records, as well to create new ones. The store will automatically cache records for you. 
| N/A | JsonApi::Exceptions 
| ActiveModel::Serializer::Adapter<br>::JsonApi::ApiObject::JsonApi | ActiveModelSerializers::Adapter::JsonApi::ApiObject::{JsonApi,Links,Link,Resource,Error, etc}? vs. JsonApiFormat https://github.com/endpoints/endpoints/blob/master/es5/format-jsonapi/index.js or even schema https://github.com/orbitjs/orbit.js/blob/master/lib/orbit-common/schema.js https://github.com/endpoints/endpoints/blob/master/es5/validate-json-schema/index.js
| ActiveModelSerializers.logger | ActiveModelSerializers.logger 

see 
- http://guides.emberjs.com/v2.1.0/models/
- https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/formatter.rb
- https://andycrum.github.io/ember-data-model-maker/
- https://github.com/cerebris/jsonapi-resources#jsonapiresource
- origins of naming in AMS:
  - https://github.com/rails/rails/commit/d2b78b3594b9cc9870e6a6ebfeb2e56d00e6ddb8#diff-80d5beeced9bdc24ca2b04a201543bdd
  - https://github.com/rails/rails/commit/783db25e0c640c1588732967a87d65c10fddc08e
  - https://github.com/rails/rails/commit/c6bc8e662614be711f45a8d4b231d5f993b024a7
  - https://github.com/rails/rails/commit/fbdf706fffbfb17731a1f459203d242414ef5086#diff-531b52ff1432c7c9f9472188b39d6a43
  - https://github.com/rails/rails/commit/4bfbdc133a37c6046b8854d3659ba75597d1d37e
  - https://github.com/rails/rails/commit/8896b4fdc8a543157cdf4dfc378607ebf6c10ab0
  - https://github.com/rails-api/active_model_serializers/commits/master?page=19
  - https://github.com/rails-api/active_model_serializers/commit/729a823868e8c7ac86c653fcc7100ee511e08cb6#diff-fe7aa2941c19a41ccea6e52940d84016
  - https://github.com/rails-api/active_model_serializers/commit/4a2d9853ba7486acc1747752982aa5650e7fd6e9
  - https://github.com/rails-api/active_model_serializers/commit/f00fe5595ddf741dc26127ed8fe81adad833ead5

docs/rfcs/0000-namespace.md
maurogeorge: @bf4 when you say "serializers" what is the namespace? It is `ActiveModel::Serializers`?

docs/rfcs/0000-namespace.md
maurogeorge: @bf4 I think here you go beyond the namespace issue I am right? You mentioned things like `ActiveModelSerializers::V08` and `ActiveModelSerializers::Current::Store` that I think are new idea of implementations. 

The Idea of RFC is to define a consistent namespace I will try to focus on it.

At the column desired we have two namespaces as public API `ActiveModelSerializers` and `ActiveModel`? In your opinion does not make sense a single namespace?

Thanks for the feedback

docs/rfcs/0000-namespace.md
bf4: re namespacing see https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43512723 @beauby re: pain... it's so obvious to me and all the work we've done together that I'm having a hard time thinking of an answer.  Like using classes as namespaces is buggy, like how not everything is a Serializer, like how Serializers don't serialize and is generally a bad name, like how it's not even the core concept of the library anymore... does that mean anything?  Does the naming comparison in the other comment help?

bf4: In any event, I'm very strongly against changing the name of the library to ActiveModel::Serializer unless someone can make a convincing argument

maurogeorge: @bf4 I know that it is will be a big change, that will hurt. I added some of the pains on the drawbacks sections, could you please expand your idea of the pains.

Even with the pain I think this is a good time to do that, since the AMS does not reach 1.0, if we will follow [Semantic Version](http://semver.org/) it is the right time to break things.

Thanks for the feedback

beauby: @bf4 Sure, I just meant writing it for the record

docs/rfcs/0000-namespace.md
bf4: So, that table is also part of https://github.com/rails-api/active_model_serializers/pull/1301 but I think also helps to explain the namespacing

`ActiveModel` comes from Rails.  Our lib is called `ActiveModelSerializers`  Rails has an `ActiveModel::Serializers::JSON` and `ActiveModel::Serialization` which reflect the history of the two libs, but in any case, I want AMS naming to be less confusing, and its interactions with Rails namespacing makes that harder, in my experience.

This is basically identifying the same naming issue as you and coming to the opposite decision.  The naming, as it stands, makes things more confusing, in my experience.

So, yes, I'm not changing the scope of the RFC, which is `AMS` vs. `AM::S`, but am elaborating on why `AMS` should remain the status quo.

bf4: And I guess I wrote some more at https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185

docs/rfcs/0000-namespace.md
bf4: @maurogeorge 

> Even with the pain I think this is a good time to do that, since the AMS does not reach 1.0, if we will follow Semantic Version it is the right time to break things.

breaking things in pre-1.0 is pretty consistent with semver, unless I misunderstand you

NullVoxPopuli: we can always deprecate in 0.11, for example and remove the aliases to the old namespace in 0.12 (or 1.0 -- not sure what the release schedule looks like)

It's pretty easy to alias classes :-\

maurogeorge: @NullVoxPopuli @bf4 I agree with both of you about the semver, that if we will change the namespace we are in a good place and time to do that.

docs/rfcs/0000-namespace.md
maurogeorge: |Current | Desired | Notes |
|-----------|-------------|---------|
|`ActiveModelSerializers` and `ActiveModel::Serializer` | `ActiveModel::Serializers` | The main namespace |
| `ActiveModelSerializers.logger` | `ActiveModel::Serializers.logger` | |
|`ActiveModelSerializers::Model`| `ActiveModel::Serializers::Model` ||
|`ActiveModel::SerializableResource`| `ActiveModel::Serializers::SerializableResource` | |
| `ActiveModel::Serializer` | `ActiveModel::Serializers::Serializer` | I know that is probably a bad name, but In a second moment we can rename this to `Resource` [for example following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185)|
|`ActiveModel::Serializer.config`| `ActiveModel::Serializers.config` ||

This way we will have a single namespace to keep all the AMS into a single namespace `ActiveModel::Serializers`.

It seens to me that the actual architeture leads to some confusion to the development team as you can see [on this comment](https://github.com/rails-api/active_model_serializers/pull/1313#discussion_r43573370) this is not clear where to put a new implementation, I feel it too when I am working on the test helpers. 

I think if we have a single namespace this will be more clear to the team and to the end users too.

docs/rfcs/0000-namespace.md
maurogeorge: @bf4 about the name I want change to ActiveModel::Serializer**s** the argument is to follow the Rubygems recomendation and with that [keep all the AMS code into a single namespace](https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43813265).

maurogeorge: @bf4 @NullVoxPopuli @beauby thanks a lot for your time and attention at this discussion :heart:

docs/rfcs/0000-namespace.md
NullVoxPopuli: :+1:

beauby:
So what are the pros and cons about using `ActiveModelSerializers` or `ActiveModel::Serializers` (plural) as the base namespace? The latter kinda makes sense to me in the sense that it is inside the `ActiveModel` namespace.

maurogeorge:
@beauby I think the trade-offs are only about preference about the name in this case. I vote on `ActiveModel::Serializers` (plural) since we are adding functionalities to the `ActiveModel` so this way we are following the Rubygems recommendation about naming:

> Use dashes for extensions
>
>If you’re adding functionality to another gem, use a dash. This usually corresponds to a / in the require statement (and therefore your gem’s directory structure) and a :: in the name of your main class or module.

The expected gem name, in the gemspec is `active_model-serializers` but **I don't think we need to change this too**, we can change the other things we can change without the need of a new gem on Rubygems.

|Gem name                  | Require statement          | Main class or module     |
|--------------------------|----------------------------|--------------------------|
| active_model_Serializers | `active_model/serializers` | ActiveModel::Serializers |

All these change are related only with the code, keeping the Rubygems the same way.

Active Model for example follow the same idea the gem name on gemspec is `activemodel` and to the end user is:

|Gem name                  | Require statement          | Main class or module     |
|--------------------------|----------------------------|--------------------------|
| activemodel | `active_model` | ActiveModel |

As you can see we do not require `activemodel`(the gem name in gemspec) insted we use `active_model`. I think we are good to go following the above table and keep the gem name on gemspec the actual name.

NullVoxPopuli:
:+1:

maurogeorge:
I think we are reaching a common sense here, as we can see [at this comment](https://github.com/rails-api/active_model_serializers/pull/1296#issuecomment-153839271).
If I am not wrong:
- @NullVoxPopuli agree with these changes. 
- @beauby your agree and your concern is around `ActiveModelSerializers` or `ActiveModel::Serializers` (plural). I hope I clarified something about it on https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-153867830

If I was wrong please clarify here.

@bf4 after the latest discussions, what is your opinion about this?

bf4:
@maurogeorge wrote:

>  I vote on ActiveModel::Serializers (plural) since we are adding functionalities to the ActiveModel so this way we are following the Rubygems recommendation about naming:

> The expected gem name, in the gemspec is active_model-serializers but I don't think we need to change this too

This logic is backwards.  The gem name is [ActiveModelSerializers](https://github.com/rails-api/active_model_serializers/blob/0200d89da8fdd989ddf6fb6e85c8a9818669c943/active_model_serializers.gemspec#L7) and the file in lib is [active_model_serializers.rb](https://github.com/rails-api/active_model_serializers/blob/0200d89da8fdd989ddf6fb6e85c8a9818669c943/lib/active_model_serializers.rb) containing the namespace ActiveModelSerializers.  The RubyGem convention is for a gem and lib file of those names to put the core library functionality in lib/active_model_serializers (except for some rails engine-related conventions).

The initial name of the class in Rails was [ActiveModel::Serializer](https://github.com/rails/rails/commit/d2b78b3594b9cc9870e6a6ebfeb2e56d00e6ddb8#diff-80d5beeced9bdc24ca2b04a201543bdd) and the gem was originally named by @steveklabnik as [ActiveModelSerializers](https://github.com/rails-api/active_model_serializers/commit/c718915075e6cae4839efc1df448b245f7900ae7), then renamed to [ActiveModel::Serializer](https://github.com/rails-api/active_model_serializers/commit/729a823868e8c7ac86c653fcc7100ee511e08cb6#diff-fe7aa2941c19a41ccea6e52940d84016), then referred to as [ActiveModel::Serializers](https://github.com/rails-api/active_model_serializers/commit/4a2d9853ba7486acc1747752982aa5650e7fd6e9), though the gem name didn't change, and recently [we started thinking about what the name of the gem should be](https://github.com/rails-api/active_model_serializers/pull/1259#discussion_r42073706)

Now, Rails has [ActiveModel::Serializers::JSON](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serializers/json.rb) and [ActiveModel::Serialization](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb) all of which makes me think we're better off taking advantage of the gem already having a different name and prefer to use it.  For example, what would you call ActiveModelSerializers::Model? ActiveModel::Serializer::Model?  (Not to mention that using a class as a namespace causes load-order issues)

cc @kurko @joaomdmoura @bolshakov @bacarini @mateomurphy @guilleiguaran @spastorino @josh

bolshakov:
I like the idea of following rubygems naming convention. But I think it would be better to have namespace not related to the rails internals.

maurogeorge:
@bf4 the logic I followed about the name `ActiveModel::Serializers` (plural) it was the same followed on [Active Model](https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-153867830).

About the name I wrote about it  [on the RFC at the motivation section](https://github.com/maurogeorge/active_model_serializers/blob/patch-12/docs/rfcs/0000-namespace.md#motivation)

Using a class as a namespace be a bad idea, I agree, [this not proposed on this RFC](https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43813265).

Thanks to mention and bring more people to the discussion. Let's wait for more feedback and different vision around this.

bf4:
> 
Right, but your vision doesn't include the whole story or renaming activemodel::serializer (which is no longer a good name for what the class does as i discussed in the notes column)
>

maurogeorge:
@bf4 what you mean with whole story? I try expose my points on the [RFC](https://github.com/maurogeorge/active_model_serializers/blob/patch-12/docs/rfcs/0000-namespace.md) and with the comments, I think the more clarifying are [1](https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43813265) and [2](https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-153867830).

Renaming `ActiveModel::Serializer` it is a real thing we need to discuss, I added this to the [RFC](https://github.com/maurogeorge/active_model_serializers/blob/patch-12/docs/rfcs/0000-namespace.md#unresolved-questions) but as a first step we can simply add this to the namespace `ActiveModel::Serializers::Serializer` and on a second moment rename this to a better name, but at this point all the code will be running under `ActiveModel::Serializers`.

NullVoxPopuli:
Yeah, activemodel::serializer shouldn't be a class, and when it's not a class, I think it solves the problem, and there isn't anything left to discuss aside from personal preference. :-\

I'm *very* in favor of the `ActiveModel::Serializers` *namespace* as it means that AMS is an extension to the rails suite of tools, but also that because no part of the namespace is a class or module that AMS relies on, the namespace `ActiveModel` and `Serializers` shouldn't conflict with any non-rails projects.

NullVoxPopuli:
I think it's also important, if the direction is to get adopted by rails officially -- which I think would be a huge benefit, esp for rails-api apps.

bf4:
@maurogeorge 

> what you mean with whole story? I try expose my points on the RFC and with the comments, I think the more clarifying are 1 and 2.

You didn't discuss the 'adapters'

@NullVoxPopuli 

> Yeah, activemodel::serializer shouldn't be a class, and when it's not a class, I think it solves the problem, and there isn't anything left to discuss aside from personal preference.

The two are intertwined. And if I thought it were a matter of personal preference, I wouldn't bring it up. This is about 1) lessening complexity 2) shrinking the hierarchy 3) future-proofing against namespace collisions between this lib and rails, both of which would build on a second level namespace they own. It would be the only part of rails that doesn't have its own namespace 4) making things more consistent 5) admitting that ActiveModel::Serializer is no longer a good class name for what AMS does, nor is it a good primary namespace 6) deciding to what extent we want to focus on the name 'Serializer' given that the term has narrowed in scope and changed in meaning since it was first used. I included examples of this in the notes for each name https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43512723

beauby:
The only possible issue with `ActiveModel::Serializers` is that it already exists in rails (it contains one class, namely `ActiveModel::Serializers::JSON`). On the other hand, `ActiveModelSerializers` does not look very appealing to me.

maurogeorge:
@bf4 thanks for clarifying the question. About adapter today they live in something like `ActiveModel::Serializer::Adapter::JsonApi` at the first moment to keep the things more easy to change we can keep this on `ActiveModel::Serializers::Serializer::Adapter::JsonApi`(as you can see keep the actual naming change only the base namespace from `ActiveModel::Serializer` to `ActiveModel::Serializers`) yeah a bad name, I know. But for less complexity on this change I think we can do this on some steps:

- Create the `ActiveModel::Serializers` namespace;
- Move all content under `ActiveModelSerializers` to be under `ActiveModel::Serializers`, the logger is on this step;
- Move all content under `ActiveModel::Serializer` to be under `ActiveModel::Serializers`, the adapter is on this steps;
- Move all content under `ActiveModel` to be under `ActiveModel::Serializers`, the `SerializableResource` is on this step ;
- Now that all the code lives under `ActiveModel::Serializers` we can:
  - create a better name to the `ActiveModel::Serializers::Serializer` keeping in mind only to keep this in the same namespace
  - create a better name to the `ActiveModel::Serializers::Serializer::Adapter::JsonApi` probably remove this from the `ActiveModel::Serializers::Serializer` and do something like `ActiveModel::Serializers::Adapter::JsonApi` keeping in mind only to keep this in the same namespace

As you can see the idea still the same even if this a adapter or other class/namespace.

bf4:
@maurogeorge ok, so you still see the Adapter being inside the Serializer namespace? That doesn't fix anything. It just makes the namespace one level deeper for everything **and** ties the library to a namespace that this library was **removed** from

maurogeorge:
@bf4 no. The adapter at the end will be in `ActiveModel::Serializers::Adapter::JsonApi` just to make the transition easy, we can make this on steps, on separate PRs. 

IMO this fixes the [3 different namespaces we use today](https://github.com/maurogeorge/active_model_serializers/blob/patch-12/docs/rfcs/0000-namespace.md#motivation).

bf4:
Ok, @maurogeorge so, do we totally agree on everything in https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43512723 except for whether we use ActiveModelSerializers or ActiveModel::Serializers for the primary namespace and gem name?  For me, how we intend to use it is part of which one to use.  (Besides whether or not we should change the gem name.)

If we change the gem name to `ActiveModel::Serializers`, how would you require/load it? We won't be able to `require active_model/serializers` since that already exists (and would also imply changing the gem name to active_model-serializers).  For me, this is a big reason to avoid re-using the Rails namespace, along with no other parts of Rails being defined under namespaces for other parts.

maurogeorge:
@bf4 I do not totally agree with https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43512723 since in this comment the classes are defined on **2** namespaces `ActiveModelSerializers` and `ActiveModel` (the `ActiveModel::SerializableResource`) and are defined new classes like `ActiveModelSerializers::Current` and `ActiveModelSerializers::V08` and others classes too, about this new classes I don't have a strong option, since I am trying to focus here in the actual classes and to the main namespace.
I think more in this way [1](https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r43813265) and [2](https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-154105259).

Yes we can `require active_model/serializers`, I don't find a file under `lib/active_model/serialzers.rb` on [AMS](https://github.com/rails-api/active_model_serializers/tree/0200d89da8fdd989ddf6fb6e85c8a9818669c943/lib/active_model) or on [Active Model](https://github.com/rails/rails/tree/e262b2d01ccf26deb2345665a748548c7c18a8c1/activemodel/lib/active_model). I am missing something here? And we do not need to rename the gem, we can follow the same idea [Active Model did](https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-153867830).

bf4:
Should we try to move this discussion to a Google Doc? It might help organize it better

maurogeorge:
I don't think change tools here will solve anything.

This PR was created 2 weeks ago, one week ago was called more people to the discussion. I think we can try define what the way we will follow here.

I think actually we have the following opinions about the namespace

##### ActiveModel::Serializers
- Me
- @NullVoxPopuli 
- @beauby 

##### ActiveModelSerializers
- @bf4 
- @bolshakov 

If I was wrong please clarify here.

How can we proceed based on that?

Thanks for the feedback guys

NullVoxPopuli:
I think it makes sense to model after rails' ActiveModel::Serializers
but, I remember a question (don't remember if this was answered): how do you require the gem, if rails includes that namespace?

I don't remember if it was answered or what the answer was, but I think as long as we keep the gem name, we'll be fine.

(I think we could keep the gem name, `active_model_serializers`)

maurogeorge:
@NullVoxPopuli about the require I answered this here https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-154188252 at the second paragraph.

NullVoxPopuli:
looks like we're good then

bf4:
So we're going to have `ActiveModel::Serializers::Serializer` `ActiveModel::Serializers::JSON`, `ActionModel::Serializers::Serializer::Adapter::JsonApi::ApiObjects....`?

bf4:
@beauby I don't like `ActiveModelSerializers` as a namespace, but it is the name of the gem, the file name in the gem path, has less naming baggage than `ActiveModel::Serializers`, esp since we now have `ActiveModel::Serializer` (our serializer doesn't serialize), and isn't building on a non-primary namespace we don't own

sandstrom:
I agree with @bf4 on this. Or use `::AMS` which [seems to be free](https://rubygems.org/gems/AMS).

NullVoxPopuli:
@sandstrom so is [activemodel/serializers](https://rubygems.org/search?utf8=%E2%9C%93&query=activemodel%2Fserializers), [activemodel-serializers](https://rubygems.org/search?utf8=%E2%9C%93&query=activemodel-serializers), [activemodel_serializers](https://rubygems.org/search?utf8=%E2%9C%93&query=activemodel_serializers), and [active_model/serializers](https://rubygems.org/search?utf8=%E2%9C%93&query=active_model%2Fserializers)

the gem names out there don't necessarily correspond to namespaces.

And since namespaces can be in flux so much anyway from project to project, and since `ActiveModel::Serializers` isn't in use by rails, I think this issue has gotten way out of hand. :-\

bf4:
@NullVoxPopuli 

> @sandstrom so is activemodel/serializers, activemodel-serializers, activemodel_serializers, and active_model/serializers

not really. there's the gem name: active_model_serializers and there's what it's been called for it's entire existence ActiveModel::Serializer(s). All four of those examples are idiosyncratic

> the gem names out there don't necessarily correspond to namespaces.

that was true years ago and for some older gems, but isn't really true today.  There's a best practice in naming recommended by the guides that pretty much all new gems use and that all experienced devs use.

> And since namespaces can be in flux so much anyway from project to project, 

Not sure what you might be referring to. It's a rare thing for a gem to change its name.  bcrypt-ruby did it. Can't think of any others off hand.

> and since ActiveModel::Serializers isn't in use by rails

and yet, it is. Not as a gem name, but that code is in there. It's generally considered a bad practice to build on someone else's namespace, unless it's a plugin and that's how they recommend.

> I think this issue has gotten way out of hand. :-\

Or just there's been no consensus nor decree.  I don't think AM::Ss is any better or worse than AMS, but I think it is more dangerous both as a namespace and as a name

maurogeorge:
@bf4 answering your question about class name, you can take a look here https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-154105259.

> I think this issue has gotten way out of hand. :-\

I tend to agree with that. I think we are running in circles, questioning and answering the same questions.

I don't know how the maintainers decide this kind of question, I try with some democracy couting the votes https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-155885745 but I am not sure if is the way you guys are thinking to resolve this.

I don't think we will have 100% consensus but I think we need to decide between `ActiveModelSerializers` or `ActiveModel::Serializers` until there the problem of the 3 namespaces still alive. I like to follow the idea of [Thoughtbot code review](https://github.com/thoughtbot/guides/tree/master/code-review):

> Accept that many programming decisions are opinions. Discuss tradeoffs, which you prefer, and reach a resolution quickly.

[Even I don't 100% like a idea I can live with that if this is the main idea of the group](https://github.com/rails-api/active_model_serializers/pull/1322#discussion_r44408794).

Thanks a lot for your effort on this guys.

NullVoxPopuli:
Would it help if a PR was made or this one was modified to show what the
namespace change would look like?

kurko:
1. lots of duplicated arguments
2. it's better to achieve consensus than votes. Voting only split the community.
3. this is a mess

How about we bring the discussion to IRC and then come back with a more robust solution where everyone sees some shared benefits?

bf4:
I was thinking maybe reviving the google group....

B mobile phone

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,76 @@
...
+if we take a look on README and gemspec always is used `ActiveModel::Serializers`.
+
+To follow this we will need to:
+
+- Create the `ActiveModel::Serializers` namespace;
+- Move all content under `ActiveModelSerializers` to be under `ActiveModel::Serializers`;
+- Move all content under `ActiveModel::Serializer` to be under `ActiveModel::Serializers`;
+- Move all content under `ActiveModel` to be under `ActiveModel::Serializers`;
+- Create a new class to be the inherited at the creation of a serializer,
+  like `ActiveModel::Serializers::Base` to follow the Rails convention;
```
kurko: No please god no. Might as well use https://github.com/garybernhardt/base then

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,76 @@
...
+- Update the docs;
+
+# Drawbacks
+
+This will be a breaking change, so all users serializers will be broken.
+All PRs will need to rebase since the architeture will change a lot.
+
+# Alternatives
+
+We can keep the way it is, and keep in mind to not another namespace as a public
```
kurko: "keep in mind to not another" not add?

docs/rfcs/0000-namespace.md
NullVoxPopuli: but it's namespaced to AM::S?

NullVoxPopuli: it would have the same amount of methods as the current superclass

NullVoxPopuli:
wherever the conversation continues, could someone post the link?

maurogeorge:
@kurko thanks for join the conversation.

What is the advantage of the IRC over the GH?

docs/rfcs/0000-namespace.md
maurogeorge: @kurko I don't like the name either, I keep this on the [unresolved questions](https://github.com/maurogeorge/active_model_serializers/blob/patch-12/docs/rfcs/0000-namespace.md#unresolved-questions).

docs/rfcs/0000-namespace.md
maurogeorge: Good catch thanks.

maurogeorge:
@NullVoxPopuli makes sense to me update the PR, this is the first time that I make a RFC I wasn't updating the PR since GH will hide some comments, but I can do that on the next time I work on the project. I will gather information on the comments and update the PR.

NullVoxPopuli:
as long as you don't squash, I think all the comments will stay put. (some may be hidden)

I just think it may be helpful to see where things *would* go

bf4:
[Here's a public-to-comment google doc we can use for more threaded conversation https://docs.google.com/document/d/1AzQPHCdbJjsPlUQcNqrSD_7tiPLVirXZCmFw0D9XqSM/edit?usp=sharing](https://docs.google.com/document/d/1AzQPHCdbJjsPlUQcNqrSD_7tiPLVirXZCmFw0D9XqSM/edit?usp=sharing)  I'll paste an editable link in the slack since it's more private

maurogeorge:
Hi @NullVoxPopuli I updated the RFC.

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,115 @@
...
+The expected gem name, in the gemspec is `active_model-serializers` but we don't
+need to change this, we can change the code without the need of a new gem on Rubygems.
+
+Active Model for example follow the same idea the gem name on gemspec is `activemodel` and to the end user is:
+
+|Gem name                  | Require statement          | Main class or module     |
+|--------------------------|----------------------------|--------------------------|
+| activemodel              | `active_model`             | ActiveModel              |
+
+As you can see we do not require `activemodel`(the gem name in gemspec) insted
```
NullVoxPopuli: insted should be instead

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,115 @@
...
+| `ActiveModelSerializers.logger`                       | `ActiveModel::Serializers.logger`                ||
+|`ActiveModelSerializers::Model`                        | `ActiveModel::Serializers::Model`                ||
+|`ActiveModel::SerializableResource`                    | `ActiveModel::Serializers::SerializableResource` ||
+| `ActiveModel::Serializer`                             | `ActiveModel::Serializers::Serializer`           | I know that is probably a bad name, but In a second moment we can rename this to `Resource` [for example following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185)|
+|`ActiveModel::Serializer.config`                       | `ActiveModel::Serializers.config`                ||
+
+# Drawbacks
+
+This will be a breaking change, so all users serializers will be broken.
+All PRs will need to rebase since the architeture will change a lot.
```
NullVoxPopuli: My preference would be to add aliases from the old namespaces to the new ones, and just throw some deprecation warnings.

Using this, http://apidock.com/ruby/Module/const_missing
it would be pretty easy to log a deprecation message, and then proceed to the re-assigned namespace.

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,115 @@
...
+
+# Alternatives
+
+We can keep the way it is, and keep in mind to not add another namespace as a
+public API.
+
+Or we can start moving the small ones that seems to be the
```
NullVoxPopuli: I think this would take too long

docs/rfcs/0000-namespace.md
NullVoxPopuli: once the next major version is released, remove the aliases. (I think this is called semantics versioning?)

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,115 @@
...
+At the `ActiveModelSerializers` we have:
+
+- `ActiveModelSerializers::Model`
+- `ActiveModelSerializers.logger`
+
+At `ActiveModel` we have:
+
+- `ActiveModel::SerializableResource`
+
+The idea here is to provide a single namespace `ActiveModel::Serializers` to the user.
```
bf4: @maurogeorge thanks for all your work on this

We're going to need to either rewrite this doc or create a new one.  I'm going to enumerate below the result what's been happening in the gem and what we've been discussed in issues and in slack.

Scope of RFC: existing and future class and module names in gem used in namespacing. Does not include renaming of existing classes.

- We're not going to change the gem require lib. It remains `active_model_serializers`
- We're not going to change the namespace in the required file. It remains `ActiveModelSerializers`.
- We're going to wean ourselves off of `ActiveModel::Serializer` as the primary namespace.
- We're going to begin writing new code in `lib/active_model_serializers/` using the module namespace `ActiveModelSerializers`.
- We're going to move functionality, as we work in it, to live under  `lib/active_model_serializers/` and use the module namespace `ActiveModelSerializers`.
- For clarity, will we refer to the gem as `ActiveModelSerializers`, one word.
- Up for discussion: 
  - referring to it as `AMS`. Proposal: in discussion and code review yes, not in code or code comments or documentation, unless defined as an alias.
  - should `ActiveModel::SerializableResource` be moved under the `ActiveModelSerializers` namespace? Proposal: yes
  - should `ActiveModel::Serializer` be moved under the `ActiveModelSerializers` namespace? Proposal: yes
  - should `ActionController::Serialization` be moved under the `ActiveModelSerializers` namespace? Proposal: yes
  - should Adapters get their own namespace under `ActiveModelSerializers`? Proposal: yes
  - should Serializers get their own namespace under `ActiveModelSerializers`? Proposal: yes
  - should all `ActiveModel`-namespaced files in the gem be moved under `ActiveModelSerializers`? Proposal: yes
  - should moved classes or module will initially be aliased to their old name and location with deprecation warnings, such [as was done for `CollectionSerializer`](https://github.com/rails-api/active_model_serializers/pull/1251). Proposal: yes


Goal: 
  - clarify boundaries of gem and Rails
  - clarify that the `Serializer` class is no longer the primary behavior of the gem.

cc @beauby @NullVoxPopuli @joaomdmoura @kurko @steveklabnik

maurogeorge:
@bf4 thanks for the feedback.

So the consensus of the team is to use `ActiveModelSerializers`? Or this is still under discussion? If this is decided I can update the docs the next time I work on the project.

Thanks

maurogeorge:
Hi @bf4 just a ping, when you have some time please answer my last comment https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-160086361

Thanks

maurogeorge:
Hi @bf4 ping again :grin:, when you have some time please answer my comment https://github.com/rails-api/active_model_serializers/pull/1310#issuecomment-160086361

Thanks

docs/rfcs/0000-namespace.md
bf4: @maurogeorge yeah, please update the PR so we review in-line. Sorry for the delay in getting back to you.

maurogeorge: @bf4 thanks for the feedback. For now I will be working on the assert_response_schema later I can back to this.

Thanks

bf4:
oh, I thought I answered in my comment https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r45947587

maurogeorge:
@bf4 you are right, I answered you too https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r47278143, it is how the Github show the thread :smile:

beauby:
@maurogeorge I believe the current move is to use `ActiveModelSerializers`, yes. @rails-api/ams?

bf4:
Y'eup. I'll repost https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r45947587 here

-----


Goal of RFC:

  - clarify boundaries of gem and Rails
  - clarify that the `Serializer` class is no longer the primary behavior of the gem.

Scope of RFC: 

- existing and future class and module names in gem, with respect to namespacing. Does not include renaming of existing classes.

Output of RFC:

- We're not going to change the gem require lib. It remains `lib/active_model_serializers.rb`
- We're not going to change the namespace in the required file. It remains `ActiveModelSerializers`.
- We're going to wean ourselves off of `ActiveModel::Serializer` as the primary namespace.
- We're going to begin writing new code in `lib/active_model_serializers/` using the module namespace `ActiveModelSerializers`.
- We're going to move functionality, as we work in it, to live under  `lib/active_model_serializers/` and use the module namespace `ActiveModelSerializers`.
- For clarity we will refer to the gem as `ActiveModelSerializers`, one word. (For some history on this, see https://github.com/rails-api/active_model_serializers/pull/1388 )
- Discussion of renaming existing classes / modules and JsonApi objects will happen in separate PRs, and issues, and in the [google doc https://docs.google.com/document/d/1rcrJr0sVcazY2Opd_6Kmv1iIwuHbI84s1P_NzFn-05c/edit?usp=sharing](https://docs.google.com/document/d/1rcrJr0sVcazY2Opd_6Kmv1iIwuHbI84s1P_NzFn-05c/edit?usp=sharing), e.g.
  - [ ] should moved classes or modules be aliased to their old name and location with deprecation warnings, such [as was done for `CollectionSerializer`](https://github.com/rails-api/active_model_serializers/pull/1251). Proposal: yes
  - [ ] should Ruby representations of JSON API object be [defined in their own namespace](https://github.com/rails-api/active_model_serializers/pull/1004#discussion_r46783510, and not embedded in the JSON API adapter? Proposal: yes
- Up for discussion: 
  - [ ] referring to it as `AMS`. Proposal: in discussion and code review yes, not in code or code comments or documentation, unless defined as an alias.
  - [ ] per [proposal by Adrian Mungolo @xymbol](https://groups.google.com/d/msg/rails-api-core/pGSWLiWUsCw/-BdKF4pAPwAJ), *change gem name to `AMS`*. Proposal: table
  - [ ] should `ActiveModel::SerializableResource` be moved under the `ActiveModelSerializers` namespace? Proposal: yes
  - [ ] should `ActiveModel::Serializer` be moved under the `ActiveModelSerializers` namespace? Proposal: yes
  - [ ] should `ActionController::Serialization` be moved under the `ActiveModelSerializers` namespace? Proposal: yes
  - [ ] should Adapters get their own namespace under `ActiveModelSerializers`? Proposal: yes
  - [ ] should Serializers get their own namespace under `ActiveModelSerializers`? Proposal: yes
  - [ ] should all `ActiveModel`-namespaced files in the gem be moved under `ActiveModelSerializers`? Proposal: yes


refs:

- https://github.com/rails-api/active_model_serializers/blob/master/docs/ARCHITECTURE.md
- https://github.com/rails-api/active_model_serializers/blob/master/docs/jsonapi/schema.md#json-api-document-schema
- [Discussion doc](google doc https://docs.google.com/document/d/1rcrJr0sVcazY2Opd_6Kmv1iIwuHbI84s1P_NzFn-05c/edit?usp=sharing](https://docs.google.com/document/d/1rcrJr0sVcazY2Opd_6Kmv1iIwuHbI84s1P_NzFn-05c/edit?usp=sharing)

bf4:
@maurogeorge I just noticed you [wrote a lot of the ActiveModel Basics guide in Rails](https://github.com/rails/rails/pull/15494) (4.2) :100: .  I was reading through it and saw the [`ActiveModel::Serializers`](https://github.com/rails/rails/blob/v5.0.0.beta1/guides/source/active_model_basics.md#activemodelserializers) section and thought "That's weird. I thought this wasn't going to happen until 5.1".  Then I realized it was talking about the `ActiveModel::Serializers::JSON` that is in Rails. ay yai yai

maurogeorge:
@bf4 thanks :heart: 

In fact, [when I wrote about that](https://github.com/rails/rails/pull/15494/files#diff-398eb7039ae9934ebc7ebc2222c7ac02R295) we have the [`ActiveModel::Serializers::JSON`](https://github.com/rails/rails/pull/15494/files#diff-398eb7039ae9934ebc7ebc2222c7ac02R301) and [`ActiveModel::Serializers::Xml`](https://github.com/rails/rails/pull/15494/files#diff-398eb7039ae9934ebc7ebc2222c7ac02R357) inside the `ActiveModel::Serializers`.

About the content I will update when I work on the project again, probably after the `assert_` helpers

maurogeorge:
@bf4 @beauby @NullVoxPopuli guys, sorry for the delay. I updated the RFC following the https://github.com/rails-api/active_model_serializers/pull/1310#discussion_r45947587

This is ok for you?

Thanks

bf4:
Reading on my phone, looks good.  Did you see the prehistory I added to the changelog?  Ams was a reverted rails pr that was supposed to incubate outside of rails, but never got pulled back in, hence tight namespace coupling

maurogeorge:
@bf4 cool, it is something that need to be made before the merge?

Thanks

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,105 @@
+- Start Date: (2015-10-29)
+- RFC PR: (leave this empty)
```
bf4: this should be filled int now, right?

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+- AMS Issue: (leave this empty)
```
bf4: s//ActiveModelSerializers

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+The idea here is to provide a single namespace `ActiveModelSerializers` to the user.
+Following the same idea we have on other gems like
+[Devise](https://github.com/plataformatec/devise/blob/e9c82472ffe7c43a448945f77e034a0e47dde0bb/lib/devise.rb),
+[Refile](https://github.com/refile/refile/blob/6b24c293d044862dafbf1bfa4606672a64903aa2/lib/refile.rb) and
+[Active Job](https://github.com/rails/rails/blob/30bacc26f8f258b39e12f63fe52389a968d9c1ea/activejob/lib/active_job.rb)
+for example.
+
+This way we are clarifing the boundaries of ActiveModelSerializers and Rails and
+make clear that the `ActiveModel::Serializer` class is no longer the primary
+behavior of the ActiveModelSerializers.
```
bf4: Maybe note that reasons for naming confusion is in its origin in Rails https://github.com/rails-api/active_model_serializers/blob/master/CHANGELOG.md#prehistory

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+The following table represents the current and the desired classes and modules
+at the first moment.
+
+| Current                                                | Desired                                          | Notes              |
+|--------------------------------------------------------|--------------------------------------------------|--------------------|
+| `ActiveModelSerializers` and `ActiveModel::Serializer` | `ActiveModelSerializers`                         | The main namespace |
+| `ActiveModelSerializers.logger`                        | `ActiveModelSerializers.logger`                  ||
+| `ActiveModelSerializers::Model`                        | `ActiveModelSerializers::Model`                  ||
+| `ActiveModel::SerializableResource`                    | `ActiveModelSerializers::SerializableResource`   ||
+| `ActiveModel::Serializer`                              | `ActiveModelSerializers::Serializer`             | The name can be discussed in a second moment we can rename this to `Resource` [for example following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185) more info about naming in the next section|
```
bf4: > The name can be discussed in a second moment

pull request?

maurogeorge:
@bf4 Thanks for the review.
I updated the docs, following your last comments, I think we are good to merge.

docs/rfcs/0000-namespace.md
```diff
@@ -0,0 +1,106 @@
...
+The following table represents the current and the desired classes and modules
+at the first moment.
+
+| Current                                                | Desired                                          | Notes              |
+|--------------------------------------------------------|--------------------------------------------------|--------------------|
+| `ActiveModelSerializers` and `ActiveModel::Serializer` | `ActiveModelSerializers`                         | The main namespace |
+| `ActiveModelSerializers.logger`                        | `ActiveModelSerializers.logger`                  ||
+| `ActiveModelSerializers::Model`                        | `ActiveModelSerializers::Model`                  ||
+| `ActiveModel::SerializableResource`                    | `ActiveModelSerializers::SerializableResource`   ||
+| `ActiveModel::Serializer`                              | `ActiveModelSerializers::Serializer`             | The name can be discussed in a pull requst, we can rename this to `Resource` [for example following this idea](https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185) more info about naming in the next section|
```
bf4: `in a future pull request. For example, we can rename this to 'Resource'`

maurogeorge: @bf4 updated, please take a look when you have some time.

bf4:
@rails-api/ams LGTM. ok to merge?

bf4:
Merged.  Enough time passed without objection

maurogeorge:
Thanks for the merge :heart:

PR #1311, opened 2015-10-30
Block syntax for associations.

beauby:
This PR is half of #1262, introducing the block syntax for overriding associations (and removing the legacy method defining syntax).

Currently 2 tests fail, I believe those tests are ill (they seem to expect an association to be treated as an attribute).

### DIFF
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -14,7 +14,7 @@ module ActiveModel
 
       included do |base|
         base.class_attribute :_reflections
-        self._reflections ||= []
+        base._reflections ||= []
 
         extend ActiveSupport::Autoload
         autoload :Association
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -13,9 +13,8 @@ module ActiveModel
       DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
 
       included do |base|
-        class << base
-          attr_accessor :_reflections
-        end
+        base.class_attribute :_reflections
+        self._reflections ||= []
 
         extend ActiveSupport::Autoload
         autoload :Association
@@ -29,58 +28,44 @@ module ActiveModel
 
       module ClassMethods
         def inherited(base)
-          base._reflections = self._reflections.try(:dup) || []
+          super
+          base._reflections = _reflections.dup
         end
 
         # @param [Symbol] name of the association
         # @param [Hash<Symbol => any>] options for the reflection
+        # @param [Block] optional block to customize value of the reflection
         # @return [void]
         #
         # @example
         #  has_many :comments, serializer: CommentSummarySerializer
         #
-        def has_many(name, options = {})
-          associate HasManyReflection.new(name, options)
+        def has_many(name, options = {}, &block)
+          _reflections << HasManyReflection.new(name, options, block)
         end
 
         # @param [Symbol] name of the association
         # @param [Hash<Symbol => any>] options for the reflection
+        # @param [Block] optional block to customize value of the reflection
         # @return [void]
         #
         # @example
         #  belongs_to :author, serializer: AuthorSerializer
         #
-        def belongs_to(name, options = {})
-          associate BelongsToReflection.new(name, options)
+        def belongs_to(name, options = {}, &block)
+          _reflections << BelongsToReflection.new(name, options, block)
         end
 
         # @param [Symbol] name of the association
         # @param [Hash<Symbol => any>] options for the reflection
+        # @param [Block] optional block to customize value of the reflection
         # @return [void]
         #
         # @example
         #  has_one :author, serializer: AuthorSerializer
         #
-        def has_one(name, options = {})
-          associate HasOneReflection.new(name, options)
-        end
-
-        private
-
-        # Add reflection and define {name} accessor.
-        # @param [ActiveModel::Serializer::Reflection] reflection
-        # @return [void]
-        #
-        # @api private
-        #
-        def associate(reflection)
-          self._reflections = _reflections.dup
-
-          define_method reflection.name do
-            object.send reflection.name
-          end unless method_defined?(reflection.name)
-
-          self._reflections << reflection
+        def has_one(name, options = {}, &block)
+          _reflections << HasOneReflection.new(name, options, block)
         end
       end
 
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -17,7 +17,7 @@ module ActiveModel
     #
     # So you can inspect reflections in your Adapters.
     #
-    Reflection = Struct.new(:name, :options) do
+    Reflection = Struct.new(:name, :options, :proc) do
       # Build association. This method is used internally to
       # build serializer's association by its reflection.
       #
@@ -40,7 +40,11 @@ module ActiveModel
       # @api private
       #
       def build_association(subject, parent_serializer_options)
-        association_value = subject.send(name)
+        association_value = if proc.nil?
+                              subject.object.send(name)
+                            else
+                              subject.instance_eval(&proc)
+                            end
         reflection_options = options.dup
         serializer_class = subject.class.serializer_for(association_value, reflection_options)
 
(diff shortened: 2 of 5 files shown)


lib/active_model/serializer/associations.rb
```diff
@@ -29,7 +28,8 @@ class << base
 
       module ClassMethods
         def inherited(base)
-          base._reflections = self._reflections.try(:dup) || []
```
NullVoxPopuli: why was this a `try` before?

lib/active_model/serializer/reflection.rb
```diff
@@ -54,7 +61,7 @@ def build_association(subject, parent_serializer_options)
             reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
           end
         elsif !association_value.nil? && !association_value.instance_of?(Object)
-          reflection_options[:virtual_value] = association_value
+          reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
```
NullVoxPopuli: why is there a try here?

test/fixtures/poro.rb
```diff
@@ -212,13 +208,12 @@ def self.root_name
 VirtualValueSerializer = Class.new(ActiveModel::Serializer) do
   attributes :id
 
-  has_many :reviews, virtual_value: [{ id: 1 }, { id: 2 }]
-  has_one :maker, virtual_value: { id: 1 }
-
-  def reviews
+  has_many :reviews do
```
NullVoxPopuli: much cleaner than `virtual_value` :-)

lib/active_model/serializer/associations.rb
beauby: Because the `self._reflections` was not initialized in the constructor.

lib/active_model/serializer/reflection.rb
beauby: For consistency with the way we handle collections for which we can't find serializers.

test/fixtures/poro.rb
beauby: Yeah, plus I'm not sure `virtual_value` was officially part of the public API.

lib/active_model/serializer/reflection.rb
NullVoxPopuli: how do you mean? what scenarios are those?

beauby: When doing 
```ruby
has_many :blah do 
  [ { id: 3, type: 'blah' }, {...} ]
end
```
or
```ruby
belongs_to :blah do
  { id: blah, type: 'blah' }
end
```

lib/active_model/serializer/associations.rb
NullVoxPopuli: so any class that inherits `Associations` must have a `_reflections` accessor? 
can you document that, please? 
I know it's outside the scope of a refactor, but since we have next to none of this type of documentation, I think it will be good to do it as we go.

lib/active_model/serializer/reflection.rb
NullVoxPopuli: can you write some comments about why try is being used and a quick blurb about when association_value might not have `as_json`

thanks ;-)

test/fixtures/poro.rb
NullVoxPopuli: even better!

lib/active_model/serializer/associations.rb
beauby: No, including `Associations` *does* declare a `_reflections` accessor.

beauby: c.f. https://github.com/rails-api/active_model_serializers/pull/1311/files#diff-b2b037951247cfa436c1f2201f5c12dfL16

NullVoxPopuli: I see below on line 32, that `base._reflections` is set to a dup of `_reflections` there does the duped `_reflections` come from?

beauby: That's upon inheritance, so it is duped from the parent class's

NullVoxPopuli: oh ok, so you're setting `base._reflections` just to an empty array?

beauby: depending if the parent class has some reflections already or not

NullVoxPopuli: I'm confused. Sorry. Can you summarize?

beauby: Module `Associations` defines (on the class that includes it) a class attribute `_reflections` that is initialized to `[]` and is inherited.

lib/active_model/serializer/reflection.rb
beauby: `association_value` might be user-supplied and it might be anything

lib/active_model/serializer/associations.rb
beauby: To be clear, those changes are mainly (if not only) cosmetic, in order to be consistent with how we deal with inheritance of class attributes all around, they're not necessary for this PR.

lib/active_model/serializer/reflection.rb
```diff
@@ -17,7 +17,21 @@ class Serializer
     #
     # So you can inspect reflections in your Adapters.
```
NullVoxPopuli: what is a reflection?

beauby: https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/reflection.rb

NullVoxPopuli: oh. ha. didn't see there was more. thanks

beauby:
Should be green.

lib/active_model/serializer/associations.rb
```diff
@@ -13,9 +13,8 @@ module Associations
       DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
 
       included do |base|
-        class << base
-          attr_accessor :_reflections
-        end
+        base.class_attribute :_reflections
```
bf4: I don't think we want this to be instance readable/writable? Do we want these to be inheritable?

bf4: re: inhertiable, yes, oops 

```diff
-          base._reflections = self._reflections.try(:dup) || []
+          super
+          base._reflections = _reflections.dup
```

lib/active_model/serializer/associations.rb
```diff
@@ -29,58 +28,44 @@ class << base
...
         # @param [Symbol] name of the association
         # @param [Hash<Symbol => any>] options for the reflection
+        # @param [Block] optional block to customize value of the reflection
         # @return [void]
         #
         # @example
         #  has_many :comments, serializer: CommentSummarySerializer
         #
-        def has_many(name, options = {})
-          associate HasManyReflection.new(name, options)
```
bf4: can we keep the `associate` method?  I don't feel strongly about it, but I like it...

lib/active_model/serializer/reflection.rb
```diff
@@ -17,7 +17,7 @@ class Serializer
...
     #
-    Reflection = Struct.new(:name, :options) do
+    Reflection = Struct.new(:name, :options, :proc) do
```
bf4: I prefer `callable` to refer to a `callable` when, like here, it can be a block, proc, or lambda

lib/active_model/serializer/reflection.rb
```diff
@@ -40,7 +40,11 @@ class Serializer
       # @api private
       #
       def build_association(subject, parent_serializer_options)
-        association_value = subject.send(name)
+        association_value = if proc.nil?
+                              subject.object.send(name)
+                            else
+                              subject.instance_eval(&proc)
```
bf4: Do we need a design that requires us to instance_eval?

lib/active_model/serializer/reflection.rb
bf4: I don't think we need the `try(:as_json)` change for this pr, or in general

test/fixtures/poro.rb
```diff
@@ -62,12 +62,10 @@ module Spam; end
   attributes :id, :title, :body
 
   has_many :comments
-  belongs_to :blog
-  belongs_to :author
-
-  def blog
+  belongs_to :blog do
     Blog.new(id: 999, name: 'Custom blog')
```
bf4: What does this mean?

bf4: n/m it just looks weird to create an artificial association, but that's pre-existing

bf4:
Closed by https://github.com/rails-api/active_model_serializers/issues/1356

PR #1313, opened 2015-10-31
rename context to serialization_context

tchak:
A resubmit of my earlier PR. Removed all references to `url_helpers` for now. See https://github.com/rails-api/active_model_serializers/pull/1289#discussion_r42808317 for reference

cc @bf4

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -1,4 +1,5 @@
 require 'active_support/core_ext/class/attribute'
+require 'active_model_serializers/serialization_context'
 
 module ActionController
   module Serialization
@@ -46,7 +47,7 @@ module ActionController
 
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
-        options.fetch(:context) { options[:context] = request }
+        options.fetch(:serialization_context) { options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request) }
         serializable_resource = get_serializer(resource, options)
         super(serializable_resource, options)
       end
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -222,7 +222,7 @@ module ActiveModel
         end
 
         def pagination_links_for(serializer, options)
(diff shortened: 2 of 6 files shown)


NullVoxPopuli:
LGTM :+1:

lib/action_controller/serialization/context.rb
```diff
@@ -0,0 +1,12 @@
+module ActionController
```
beauby: I'm not sure the `Context` class belongs in `ActionController`, as we might want to use it for non-AC related stuff as well. Some mixin/concern definitely belongs here, but I think the base `Context` class should be part of the `ActiveModel(::)Serializer(s)` namespace. Thoughts?

NullVoxPopuli: I agree - action controller isn't ever going to really make use of the context, where as it is AMS that utilizes context.

NullVoxPopuli: The action controller only creates the context, and that's it

bf4: It's not in action_controller, it's namespaced within actioncontroller::serialization which is fine by me

bf4: Alternative: `lib/active_model_serializers/serialization_context.rb` 

```ruby
module ActiveModelSerializers
  class SerializationContext
    #etc
  end
end
```

tchak: I am fine with whatever is getting consensus :) Looks like I should rename to `active_model_serializers/serialization_context`?

bf4: speaking for myself, I would mere that in the next 10 minutes

beauby: Yeah, let's put it in `active_model_serializers/serialization_context` and I'm OK to merge.

NullVoxPopuli: :+1:

bf4: @tchak amend commit, force push, and we'll merge

bf4: @tchak want one of us to do it?

tchak:
@bf4 fixed and rebased. Sorry for the delay!

NullVoxPopuli:
Looks good to me. We are still in disarray about namespacing, but that's outside the scope of this.

good work!

Issue #1307, opened 2015-10-29
Performance optimization with fields option: AMS should compute only the requested fields


vasilakisfil:
I am using JSON adapter but I am pretty sure it's the same on every adapter: When you request specific field from a serializer, it shouldn't compute everything in advance, only the fields you requested. For instance when you request only for :id: it should be much faster than computing all fields because some of them might be VERY expensive.

I found that the issue is here: https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L241-L251

Personally I solved that by monkey patching AMS: passing the fields in the serializer and checking the requested fields before computing them in that method.

beauby:
You're very right. There is work being done in that direction, but let's use this issue as a tracker for what I refer to as "lazy computation of attributes and associations" (the attributes case is pretty straightforward, but for associations it requires a tad more work).

beauby:
Fixed on master.

PR #1315, opened 2015-10-31
Compute only requested attributes.

beauby:
Ref #1307.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -238,10 +238,9 @@ module ActiveModel
 
     # Return the +attributes+ of +object+ as presented
     # by the serializer.
-    def attributes
-      attributes = self.class._attributes.dup
-
-      attributes.each_with_object({}) do |name, hash|
+    def attributes(requested_attrs = nil)
+      self.class._attributes.each_with_object({}) do |name, hash|
+        next unless requested_attrs.nil? || requested_attrs.include?(name)
         if self.class._fragmented
           hash[name] = self.class._fragmented.public_send(name)
         else
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -57,10 +57,7 @@ module ActiveModel
 
         def resource_object_for(options)
           cache_check(serializer) do
-            attributes = serializer.attributes
-            attributes.slice!(*options[:fields]) if options[:fields]
-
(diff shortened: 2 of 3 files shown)


lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -57,10 +57,7 @@ def include_meta(json)
 
         def resource_object_for(options)
           cache_check(serializer) do
-            attributes = serializer.attributes
-            attributes.slice!(*options[:fields]) if options[:fields]
-
-            attributes
+            serializer.attributes(options[:fields])
```
NullVoxPopuli: niiice

NullVoxPopuli:
looks good :-)

lib/active_model/serializer.rb
```diff
@@ -238,10 +238,9 @@ def json_key
 
     # Return the +attributes+ of +object+ as presented
     # by the serializer.
-    def attributes
-      attributes = self.class._attributes.dup
-
-      attributes.each_with_object({}) do |name, hash|
+    def attributes(requested_attrs = nil)
+      self.class._attributes.each_with_object({}) do |name, hash|
+        next unless requested_attrs.nil? || requested_attrs.include?(name)
```
bf4: why should `request_attrs` default to nil? Shouldn't it be an array?

and why not just assume they're giving valid attribute names? or compare the two lists?

```ruby
def attributes(attribute_names = self.class._attributes)
  attribute_names.each_with_object({}) do |name, hash|
```

Also, this might be a good spot to pass in `options[:fields]` as `{only: options[:fields]}` so that we can also support `except`. Shrug.

lib/active_model/serializer.rb
```diff
@@ -238,10 +238,9 @@ def json_key
 
     # Return the +attributes+ of +object+ as presented
     # by the serializer.
-    def attributes
```
bf4: nice rolling in field request from elsewhere

```diff
-            attributes = serializer.attributes
-            attributes.slice!(*options[:fields]) if options[:fields]
-
-            attributes
+            serializer.attributes(options[:fields])
```

to here


```diff
-    def attributes
+    def attributes(requested_attrs = nil)
```

lib/active_model/serializer.rb
beauby: The `fields` option is an array of attributes *and* associations, that's why.

Issue #340, opened 2013-06-19
In Rails JSON.dump does not work with ActiveModel::Serializer instance, while MultiJson does.


mlangenberg:
The fix for this issue might be found in Rails or JSON, but I am reporting it since I ran into it while working with AMS. 

Bug reproduced with following gem versions.
- ruby-1.9.3p392
- rails-3.2.13
- active_model_serializers-0.8.1
- multi_json-1.7.7
- json-1.8.0

``` ruby
MultiJson.dump(
  ActiveModel::Serializer.new(Object.new)
) #=> "{\"serializer\":{}}" 
```

``` ruby
JSON.dump(
  ActiveModel::Serializer.new(Object.new)
) #=> NoMethodError: undefined method `fetch' for #<JSON::Ext::Generator::State:0x007fd07a5f9928>
  # from /active_model_serializers-0.8.1/lib/active_model/serializer.rb:340:in `as_json'
  # from /activesupport-3.2.13/lib/active_support/json/encoding.rb:47:in `block in encode'
  # from /activesupport-3.2.13/lib/active_support/json/encoding.rb:77:in `check_for_circular_references'
  # from /activesupport-3.2.13/lib/active_support/json/encoding.rb:46:in `encode'
  # from /activesupport-3.2.13/lib/active_support/json/encoding.rb:31:in `encode'
  # from /activesupport-3.2.13/lib/active_support/core_ext/object/to_json.rb:16:in `to_json'
  # from /active_model_serializers-0.8.1/lib/active_model/serializer.rb:333:in `to_json'
  # from /json-1.8.0/lib/json/common.rb:223:in `generate'
  # from /json-1.8.0/lib/json/common.rb:223:in `generate'
  # from /json-1.8.0/lib/json/common.rb:394:in `dump'
  # from (irb):9
  # from /railties-3.2.13/lib/rails/commands/console.rb:47:in `start'
  # from /railties-3.2.13/lib/rails/commands/console.rb:8:in `start'
  # from /railties-3.2.13/lib/rails/commands.rb:41:in `<top (required)>'
  # from script/rails:6:in `require'
  # from script/rails:6:in `<main>'
```

You will run into this when using Sidekiq-2.12 in combination with AMS. Since Sidekiq-2.12 it uses JSON instead of multi_json for serializing arguments passed to a Sidekiq worker.

> Remove usage of multi_json as json is now robust on all platforms.
> https://github.com/mperham/sidekiq/blob/master/Changes.md#2120

[...]

PR #1318, opened 2015-11-03
Fix `#as_json` glitch caused by JSON and Rails

mhuggins:
This addresses issue #340 for the 0.8.x series.  It looks like the existing fix for that issue was not included in that series of active_model_serializers.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -338,7 +338,7 @@ module ActiveModel
     # object including the root.
     def as_json(options={})
       options ||= {}
-      if root = options.fetch(:root, @options.fetch(:root, root_name))
+      if root = options.to_hash.fetch(:root, @options.fetch(:root, root_name))
         @options[:hash] = hash = {}
         @options[:unique_values] = {}
 
--- a/test/serialization_test.rb
+++ b/test/serialization_test.rb
@@ -242,7 +242,7 @@ class RenderJsonTest < ActionController::TestCase
 
   def test_render_json_with_callback
     get :render_json_hello_world_with_callback
-    assert_equal 'alert({"hello":"world"})', @response.body
+    assert_equal '/**/alert({"hello":"world"})', @response.body
     # For JSONP, Rails 3 uses application/json, but Rails 4 uses text/javascript
     assert_match %r(application/json|text/javascript), @response.content_type.to_s
   end
--- a/test/serializer_test.rb
+++ b/test/serializer_test.rb
@@ -1363,6 +1363,7 @@ class SerializerTest < ActiveModel::TestCase
   def test_inheritance_does_not_used_cached_attributes
     parent = Class.new(ActiveModel::Serializer) do
       attributes :title
+      alias :read_attribute_for_serialization :send
     end
 
     child = Class.new(parent) do
@@ -1371,6 +1372,7 @@ class SerializerTest < ActiveModel::TestCase
 
     data_class = Class.new do
       attr_accessor :title, :body
+      alias :read_attribute_for_serialization :send
     end
 
     item = data_class.new
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,7 @@
 language: ruby
 rvm:
-  - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
-  - ree
-  - jruby-18mode
   - jruby-19mode
   - rbx-2
 gemfile:
@@ -14,15 +10,3 @@ gemfile:
 matrix:
   allow_failures:
     - gemfile: Gemfile.edge
-  exclude:
-    # Edge Rails is only compatible with 1.9.3
-    - gemfile: Gemfile.edge
-      rvm: 1.8.7
-    - gemfile: Gemfile.edge
-      rvm: 1.9.2
-    - gemfile: Gemfile.edge
-      rvm: ree
-    - gemfile: Gemfile.edge
-      rvm: jruby-18mode
-    - gemfile: Gemfile.edge
-      rvm: rbx-18mode
--- a/Rakefile
+++ b/Rakefile
@@ -16,3 +16,6 @@ task :bench do
 end
 
 task :default => :test
+
+desc 'CI test task'
+task :ci => :default
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,29 @@
+version: '{build}'
+
+skip_tags: true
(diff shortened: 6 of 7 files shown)


lib/active_model/serializer.rb
```diff
@@ -338,7 +338,7 @@ def to_json(*args)
     # object including the root.
     def as_json(options={})
       options ||= {}
-      if root = options.fetch(:root, @options.fetch(:root, root_name))
+      if root = options.to_hash.fetch(:root, @options.fetch(:root, root_name))
```
bf4: failing test?

mhuggins: I'll check it out tomorrow.

mhuggins: Interesting, these tests are already failing on 0.8.3 prior to this change.

mhuggins: For reference: https://travis-ci.org/rails-api/active_model_serializers/builds/43605004

mhuggins: It looks like the 0-8-stable branch has been failing since 2013, looks like this PR broke it: https://github.com/rails-api/active_model_serializers/pull/477

mhuggins: @bf4 I fixed some of your gem's tests, so this is now in a better state than the current 0-8-stable branch.  It's not passing on some older versions of ruby that you guys are testing on travis because:

> mime-types requires Ruby version >= 1.9.2

[as seen here](https://travis-ci.org/rails-api/active_model_serializers/builds/89587378).

bf4: @mhuggins s/my/our gem.  Thanks for looking into this.  Sorry for being slow here...

mhuggins: No problem, hopefully this is helpful and looks good. :)

beauby:
Somebody familiar with the 0.8 codebase to review this? cc @joaomdmoura

mhuggins:
Just thought I'd follow up to see if there's anything else I can offer here.

beauby:
Considering ruby 1.8.7 has been EOLed a few years ago, and considering changes necessitating ruby >= 1.9.3 have been committed to the 0.8 branch, I believe we can safely remove support for ruby < 1.9.3, and merge this commit.
What I suggest is to add a commit that changes the `.travis.yml` to this PR, so that it also removes old rubies tests, and then keep on rocking.

beauby:
Also, thanks a lot @mhuggins for taking the time to make this PR happen, and for following up!

mhuggins:
Cool, happy to do that.  I've got some work critical stuff this week, but I'll plan on updating this week.  I appreciate you getting back to me! :)

mhuggins:
@beauby I've updated the `.travis.yml` file, and tests are now passing.  In addition to removing 1.8.7 and 1.9.2, I had to remove ree for the same reason since it's built upon the 1.8 series.

Let me know if there's anything else that needs to be addressed.  Thanks! :smile:

test/serialization_test.rb
```diff
@@ -242,7 +242,7 @@ def test_render_json_with_status
 
   def test_render_json_with_callback
     get :render_json_hello_world_with_callback
-    assert_equal 'alert({"hello":"world"})', @response.body
+    assert_equal '/**/alert({"hello":"world"})', @response.body
```
bf4: oh weird, shouldn't it have been this the whole time?

test/serializer_test.rb
```diff
@@ -1363,6 +1363,7 @@ def name
   def test_inheritance_does_not_used_cached_attributes
     parent = Class.new(ActiveModel::Serializer) do
       attributes :title
+      alias :read_attribute_for_serialization :send
```
bf4: This is needed? If so, it probably shouldn't just be in the test, but generally in the serializer, no?

Also, see https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/serialization.rb#L141

test/serialization_test.rb
mhuggins: Yep, it is currently failing on the 0.8 branch.  This fixes it.

bf4:
Seems like you'll need to add an [`appveyor.yml`](https://github.com/rails-api/active_model_serializers/blob/f562449bd439d1295b91b41b01d5b780c793100a/appveyor.yml) to get Travis to go green

test/serializer_test.rb
mhuggins: Good callout.  Happy to change it if this is the desired fix.

mhuggins:
I've never used appveyor.  The 0.8 branch was failing historically, so nothing new broke here.  Any insight into what I can do to get that working?

bf4:
@mhuggins I think the big problems with CI on the 0-8-branch were related to incompatibilities with Rails < 4, and the lack of a Rails version constraint in CI.

I've addressed those in https://github.com/rails-api/active_model_serializers/pull/1376 basically copying what we have in master.  Hopefully, that should let you focus on the bug your fixing.

I included all your CI-specific commits in #1376

mhuggins:
Cool, I think it's safe to close this PR in favor of that one then. Thanks!!

bf4:
Oh, but I left out  2dbc4ddbaeb00388068843a9aecf00356b11abf7

Do you think it's no longer needed?

bf4:
Do you think it's worth working on the ruby < 1.9.3, rails < 4.0 test support?

bf4:
In the referenced bug you wrote

> I'm still seeing this error on Rails 3.2.22 with ActiveModel Serializers 0.8.3. Am I missing something?

    ```ruby
    serializer = Service::Notifications::NewsroomSerializer.new(newsroom)
    s = serializer.to_json
    # => "{\"name\":\"Huggie\",\"subdomain\":\"huggie\"}" 

    JSON.dump(s)
    # => NoMethodError: undefined method `fetch' for #<JSON::Ext::Generator::State:0x007faf44dc3048>

    MultiJson.dump(s)
    # => "{\"name\":\"Huggie\",\"subdomain\":\"huggie\"}" 
    ```

Which I haven't confirmed if it's still a problem

mhuggins:
Ahh, okay, I misunderstood.  I'll reopen a new PR with just that change then, if that works for you. :)

bf4:
No need, you can just force push to this branch

mhuggins:
:+1: Will do shortly

mhuggins:
Sorry for the delay, I'm going to make sure to address this during the week.

PR #1320, opened 2015-11-06
Add test and bugfix to include an array of string

barodeur:
Hello,

I ran into an issue yesterday as I was trying to include multiple attributes of a nested object.

The following section of the README made me think that I would have the result writing `['comments.author', 'comments.attachment']` or `'comments.author,comments.attachment'` but I didn't.

```ruby
# IN README

render @posts, include: ['author', 'comments', 'comments.author']
# or
render @posts, include: 'author,comments,comments.author'
```

The problem came from 

```ruby
include_args_to_hash(['comments.author', `comments.attachment`])
# => { :comments => { :attachment => {} } }

include_args_to_hash('comments.author,comments.attachment')
# => { :comments => { :author => {}, :attachment => {} } }
```

So I think I found a bug, and here is the fix.

### DIFF
--- a/test/include_tree/include_args_to_hash_test.rb
+++ b/test/include_tree/include_args_to_hash_test.rb
@@ -44,6 +44,19 @@ module ActiveModel
 
             assert_equal(expected, actual)
           end
+
+          def test_array_of_string
+            expected = {
+              comments: { author: {}, attachment: {} }
+            }
+            input = [
+              'comments.author',
+              'comments.attachment'
+            ]
+            actual = Parsing.include_args_to_hash(input)
+
+            assert_equal(expected, actual)
+          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
looks good to me! thanks for this!

test/include_tree/include_args_to_hash_test.rb
```diff
@@ -44,6 +44,19 @@ def test_include_args_to_hash_from_array_of_hashes
...
+              comments: { author: {}, attachment: {} }
+            }
+            input = [
+              'comments.author',
+              'comments.attachment'
+            ]
+            actual = Parsing.include_args_to_hash(input)
+
+            assert_equal(expected, actual)
+          end
```
bf4: Thanks for the test!

beauby:
Cool :+1:

PR #1322, opened 2015-11-08
Instrumenting rendering of resources

bf4:
Building on the very excellent PR by @maurogeorge (all squashed)

https://github.com/rails-api/active_model_serializers/pull/1291


lib/active_model/serializer/railtie.rb
```diff
@@ -1,9 +1,11 @@
...
 require 'rails/railtie'
+
 module ActiveModel
   class Railtie < Rails::Railtie
     initializer 'active_model_serializers.logger' do
-      ActiveSupport.on_load(:action_controller) do
-        ActiveModelSerializers.logger = ActionController::Base.logger
+      ActiveSupport.on_load(:active_model_serializers) do
+        self.logger = ActionController::Base.logger
       end
```
bf4: bonus code, out of scope... but shouldn't be that controversial.. pattern copied from active job

lib/active_model_serializers.rb
```diff
@@ -50,6 +51,7 @@ def silence_warnings
 
 require 'action_controller/serialization'
 ActiveSupport.on_load(:action_controller) do
+  ActiveSupport.run_load_hooks(:active_model_serializers, ActiveModelSerializers)
```
bf4: - [ ] TODO: document

lib/action_controller/serialization.rb
```diff
@@ -30,7 +30,7 @@ def get_serializer(resource, options = {})
         serializable_resource.serialization_scope ||= serialization_scope
         serializable_resource.serialization_scope_name = _serialization_scope
         begin
-          serializable_resource.adapter
+          serializable_resource.tap(&:adapter)
```
beauby: Could you explain the rationale here?
`serializable_resource.tap(&:adapter)` returns `serializable_resource` but calls the `serializable_resource.adapter` method in the meantime. Is this to ensure the serializers get built and the `CollectionSerializer::NoSerializerError` gets raised if necessary?

beauby:
Looks cool :+1:

lib/action_controller/serialization.rb
bf4: Yeah, we need to do a better job with the class's api... Should have less branching in controlker

Needs to return self to be ibstrumented at this time

B mobile phone

maurogeorge: It is possible to write a test around it?

lib/active_model/serializer/railtie.rb
maurogeorge: Pretty cool :+1:

lib/active_model_serializers.rb
```diff
@@ -1,14 +1,16 @@
-require 'logger'
 require 'active_model'
 require 'active_support'
+require 'active_support/tagged_logging'
+require 'active_support/logger'
```
maurogeorge: I think we can remove these 2 requires, since we already require `active_support`

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,112 @@
+##
+# ActiveModelSerializers::Logging
+#
+#   https://github.com/rails/rails/blob/280654ef88/activejob/lib/active_job/logging.rb
+#
+module ActiveModelSerializers::Logging
```
maurogeorge: Any special reason to use `ActiveModelSerializers::Logging` over a nested module? I think the more verbose is prefered http://blog.elpassion.com/ruby-gotchas/. Makes sense?

lib/active_model_serializers/callbacks.rb
```diff
@@ -0,0 +1,55 @@
...
+
+module ActiveModelSerializers
+  # = ActiveModelSerializers Callbacks
+  #
+  # ActiveModelSerializers provides hooks during the life cycle of serialization and
+  # allow you to trigger logic. Available callbacks are:
+  #
+  # * <tt>around_render</tt>
+  #
+  module Callbacks
```
maurogeorge: This callbacks is a public API? I mean I can use this on my `PostSerializer` for example? What is a use case for this? If this is the idea I think we need to add some documentation to it.

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,112 @@
...
+  #
+  #       include ActiveModelSerializers::Logging
+  #       notify :instance_method, :render
+  #
+  #       class << self
+  #         extend ActiveModelSerializers::Logging::NotificationMacro
+  #         notify :klass_method, :render
+  #       end
+  #     end
+  module NotificationMacro
```
maurogeorge: Make sense this leave in a `ActiveModelSerializers::Notification` namespace over the `Logging`?

lib/active_model_serializers.rb
```diff
@@ -1,14 +1,16 @@
...
 require 'action_controller/railtie'
 module ActiveModelSerializers
-  mattr_accessor :logger
-  self.logger = Rails.logger || Logger.new(IO::NULL)
+  mattr_accessor(:logger) { ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT)) }
 
   extend ActiveSupport::Autoload
   autoload :Model
+  autoload :Callbacks
+  autoload :Logging
```
maurogeorge: :100:

lib/active_model_serializers/callbacks.rb
bf4: You could... Docs just in the source for now... Copied out of activejob

B mobile phone

lib/active_model_serializers/logging.rb
bf4: I thought about it... Wasn't sure where it belonged... Do you think one class is more obvious than the other? Or too manyvor few files?

B mobile phone

lib/active_model_serializers/logging.rb
bf4: I was on the fence about this

It requires the ams module to be loaded first.  That's a feature, not a gotcha, in thid case

But i'm not sure how useful it us here

Copied pattern from gem;:deprecate, which is really the only reason i considered it

B mobile phone

lib/action_controller/serialization.rb
bf4: > It is possible to write a test around it?

if you mean the exception behavior, yes, there are tests.

lib/active_model_serializers.rb
bf4: So, for this I followed the pattern in active job

It's always nice to be explicit

You're right that it just so happens `require 'active_support'` requires logging and autoloads TaggedLogging, but there's no guarantee https://github.com/rails/rails/blob/4-0-stable/activesupport/lib/active_support.rb#L27

lib/active_model_serializers.rb
bf4: :grinning:

lib/active_model_serializers.rb
bf4: the only thing that worries me here is that it was removed from AMS at one point

lib/active_model_serializers/callbacks.rb
bf4: how about a test that uses `render` rather than the `notify` 'macro'?

lib/action_controller/serialization.rb
maurogeorge: What I mean is, if I change the `serializable_resource.tap(&:adapter)` to `serializable_resource` any specs will be broken? I wrote a [spec to test the instrumentation on the controller](https://github.com/bf4/active_model_serializers/commit/eaa9c6b084eb665f194570ed3a95094979ae7270#diff-f290b93cc1e01ae4a2c31174202b50eeR424) I think we need a spec to check the behavior we are doing here.

lib/active_model_serializers.rb
maurogeorge: I think if we have tests covering this we are good to go, to remove.

lib/active_model_serializers/logging.rb
maurogeorge: I think the rubygems use this way in more contexts, I think the AMS codebase use the nested module in most cases. Do you think that make sense keep this exception or change to the nested?

lib/action_controller/serialization.rb
bf4: If specs don't break, it's a regression in the specs, since it was added in response to behavior https://github.com/rails-api/active_model_serializers/pull/962#issuecomment-114204858

> There was a crash when rendering arrays that don't consist of "serializables" classes.

lib/active_model_serializers/logging.rb
bf4: honestly, your comments are sufficient for me to change it

lib/active_model_serializers/callbacks.rb
maurogeorge: I ask about the public API because this callbacks is more code that we need to maintain, If the idea is to provide a public API I think we need to add to the docs, if this is to be used only internally I think we can make this more simple without the callback.

lib/active_model_serializers/logging.rb
maurogeorge: I think about the class to keep only the notification concern and the Loggin doing only logging.

lib/active_model_serializers/callbacks.rb
bf4: Well it's public and documented.. just not much functionality.. yet..

lib/active_model_serializers/logging.rb
bf4: Yeah, but since it's really just a way to make sure as_json -> serializable_hash -> to_json doesn't notify three times... and I'm not sure it deserves it's own file right now...

lib/action_controller/serialization.rb
bf4: Spec that breaks is without calling the adapter is `ActionController::Serialization::ImplicitSerializerTest#test_render_json_array_object_without_serializer` with `ActiveModel::Serializer::CollectionSerializer::NoSerializerError:   No serializer found for resource: {:error=>"Result is Invalid"}``

bf4:
@maurogeorge updated with two incremental commits responding to your feedback

lib/active_model_serializers/callbacks.rb
maurogeorge: Since it just not much functionality. I vote to keep the things simple and remove the callback. But apparently all maintainers are ok with that, I can live with that.

maurogeorge:
@bf4 I am ok with that :+1:

docs/general/instrumentation.md
```diff
@@ -0,0 +1,19 @@
+# Instrumentation
+
+ActiveModelSerializers uses the ActiveSupport::Notification API, which
```
NullVoxPopuli: should there be an example here of how to tie this in to the Notification API?

lib/active_model_serializers.rb
NullVoxPopuli: what does it do?

lib/active_model_serializers/callbacks.rb
NullVoxPopuli: I'm ok with it. As long as it's documented and there are automated tests for it :-)

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,118 @@
...
+          tag_logger do
+            notify_render do
+              block.call(args)
+            end
+          end
+        end
+      end
+    end
+
+    # Macros that can be used to customize the logging of class or instance methods,
```
NullVoxPopuli: this is some solid documentation :-)

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,118 @@
...
+            run_callbacks callback_name do
+              send old, *args, &block
+            end
+          end
+        end
+      end
+    end
+
+    def notify_render(*)
+      event_name = 'render.active_model_serializers'.freeze
```
NullVoxPopuli: should this key be in a constant somewhere? I've seen it elsewhere in this PR as well.

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+    end
+
+    def notify_render_payload
+      { serializer: serializer, adapter: adapter }
+    end
+
+    private
+
+    def tag_logger(*tags)
```
NullVoxPopuli: I know these are private, but for the sanity of future maintainers, can a little bit of documentation go with the following methods  / references to tagged?

like, I want to know why we are calling tagged here on the logger, and passing a blog to it.

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,118 @@
...
+      if ActiveModelSerializers.logger.respond_to?(:tagged)
+        tags.unshift 'AMS'.freeze unless logger_tagged_by_active_model_serializers?
```
NullVoxPopuli: should `'AMS'.freeze` be a constant somewhere? I see it all over the place

lib/active_model_serializers/logging.rb
```diff
@@ -0,0 +1,118 @@
...
+      else
+        yield
+      end
+    end
+
+    def logger_tagged_by_active_model_serializers?
+      ActiveModelSerializers.logger.formatter.current_tags.include?('AMS'.freeze)
+    end
+
+    class LogSubscriber < ActiveSupport::LogSubscriber
```
NullVoxPopuli: how do we use this class?
could docs be added?

lib/active_model_serializers/logging.rb
bf4: I think it's more flexible this way.  Straight out of active job

B mobile phone

lib/active_model_serializers/logging.rb
bf4: Dunno.  I might have but it's not a pattern in rails

B mobile phone

lib/active_model_serializers/logging.rb
bf4: I'd consider it... Or even rename it to activemodelserializers... This is pretty simple and in line w rails and activejob

B mobile phone

lib/active_model_serializers/logging.rb
bf4: Copied pretty much :)

B mobile phone

lib/active_model_serializers/logging.rb
bf4: Link to activesupport logsubscriber should be sufficient 

B mobile phone

lib/active_model_serializers/callbacks.rb
bf4: I thought it made the loggng file easier to use and understand . File and organization based on active job b so seems fine to me

B mobile phone

lib/active_model_serializers/logging.rb
NullVoxPopuli: that's ok, we don't need to not use the same patterns rails is not using

lib/active_model_serializers/logging.rb
NullVoxPopuli: more flexible as opposed to what? 
This is what I think should be noted inline

lib/active_model_serializers/logging.rb
NullVoxPopuli: aye, but DRYness, etc, reducing typos, etc, etc

bf4: that's not really what DRY means... https://speakerdeck.com/bf4/david-chelimsky-rubyconf-2010-maintaining-balance-while-reducing-duplication?slide=7 

I didn't change the name 'cause that's what was in the original tests

lib/active_model_serializers/logging.rb
bf4: so the actual logging doesn't care if it's a tagged logger or not.  This allows us to treat it as a tagged logger if it is

bf4:
Appveyour failure is transitory

wagenet:
Is this waiting on anything? Would love to get to work on Skylight integration once this is official :)

bf4:
@wagenet not sure. Have you tried using it?  @NullVoxPopuli ok?

maurogeorge:
@bf4 I am ok with that :shipit: 
Please just keep my single commit https://github.com/bf4/active_model_serializers/commit/51424963da3ce009a2dc22fe9e976f87f3fbabc2 :heart:

bf4:
Merged

maurogeorge:
Thanks :tada: :confetti_ball:

wagenet:
Thanks! Haven't had a chance to use it yet, but hope to soon :)

wagenet:
Started to play around with this some. Looks good to me so far!

PR #1323, opened 2015-11-09
Use Minitest::Reporters to capture warnings && show error output

bf4:


.travis.yml
```diff
@@ -27,6 +26,8 @@ env:
 
 matrix:
   include:
+  - rvm: 2.1
+    env: CAPTURE_STDERR=true
```
bf4: only capture errors on 2.1

Issue #1205, opened 2015-09-27
Recurse associations by default


brauliobo:
Given these two serializers
```ruby
class TeamsPlugin::ContextSerializer < ActiveModel::Serializer
  has_many :teams, serializer: TeamsPlugin::TeamSerializer
end
class TeamsPlugin::TeamSerializer < ActiveModel::Serializer
  has_many :members, serializer: TeamsPlugin::MemberSerializer
end
```
If I call
```ruby
ActiveModel::SerializableResource.new(@article, serializer: TeamsPlugin::ContextSerializer)
```
It will serialize `teams`, but not `members`. For `members` I would have to specify `include: {teams: :members}` option for it to work.

But this seems very strange to me. Intuitively, by default it should go recursively.

beauby:
You are right that the current default behavior of AMS with the Json/Attributes adapters is to embed only the *first level* associations. I am personally in favor of making the default embedding policy `**`, i.e. embed all levels, although we would need to add a check for infinite recursion in case the data graph has cycles. Thoughts @bf4, @NullVoxPopuli, @joaomdmoura?
As a workaround, you can specify `include: '**'` when serializing.

brauliobo:
@beauby this is workaround is quite handy :)

bf4:
I'd rather make deep nesting an intentional oractice, rather than easy, and so therefore like that more than two levels requires explicitness

B mobile phone

beauby:
@bf4 I don't feel strongly either way, but it should be made clear in the docs so that people don't get surprised by this behavior (intuitively, at least I, would expect all relationships to be serialized for the Json/Attributes adapters).

brauliobo:
@bf4 @beauby to make sense that the top association (`teams`) is going to get serialized, an option `include: true` on `has_many :teams` is necessary. otherwise, this option could be true recursively, by default.

NullVoxPopuli:
I'm against having ** by default, as it could lead to an infinite loop of including, and redundant data. :-\

brauliobo:
@NullVoxPopuli so what do you think about associations explicity saying they need to be included?

beauby:
@NullVoxPopuli I agree this would happen, although it would just be the consequence of bad design (Json adapter + cyclic data graph).

NullVoxPopuli:
[...]

PR #1327, opened 2015-11-10
change attibutes adapter to show associations at "second" level by as…

passalini:
…sociation's option.

issue: #1205

### DIFF
--- a/lib/active_model/serializer/adapter/attributes.rb
+++ b/lib/active_model/serializer/adapter/attributes.rb
@@ -46,7 +46,11 @@ module ActiveModel
           return association.options[:virtual_value] if association.options[:virtual_value]
           return unless association.serializer && association.serializer.object
 
-          opts = instance_options.merge(include: @include_tree[association.key])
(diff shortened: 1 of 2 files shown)


bf4:
@passalini Thanks for this. Can you explain the need it fills? I'm not sure why the default should be two levels.  If a Blog has Posts and Posts have Authors and Comments and I want a basic representation of the Blog, I don't I'd want anything more than the blog's attributes and its associated posts, which may have some information about their authors and comments.

passalini:
@bf4 now I see that show associations at second level isn't good, but when I did this pr I was an a task to implement an api's response that shows associations and its attributes at second (sometimes at third oO) level and I think that I thought that it will be helpfull for someone else....

I thought that it is better than...

```ruby
class PostTestSerializer < Serializer
  attributes :id, :title
  
  has_many :comments
  belongs_to :author

  def comments
    ActiveModel::SerializableResource.new(object.comments).serializable_hash
  end
end
```

remear:
Is there anything more to discuss here, or can we close this?

Issue #1339, opened 2015-11-22
Resource level meta key


rafael:
Does active model serializer supports jsonapi resource meta key at the resource object level? I would like to do something like this:

```
// ...
{
  "type": "articles",
  "id": "1",
  "meta": { "my_metadata": true" },
  "attributes": {
    "title": "Rails is Omakase"
  },
  "relationships": {
    "author": {
      "links": {
        "self": "/articles/1/relationships/author",
        "related": "/articles/1/author"
      },
      "data": { "type": "people", "id": "9" }
    }
  }
}
// ...
```

I've been trying to see how to achieve this, but only been able to achieve it at the top level, not at the resource object level.

tylerwillingham:
It looks like that doesn't conform to the jsonapi spec. The spec points to the meta key being specific to the document/response.

You can implement a custom key through a serializer to get the data that you want but if conforming to the spec is important to you I would choose a different name for your key.

rafael:
@tylerwillingham are you sure that doesn't conform the spec? I was reading this part and thought it should be valid: 

```
Resource Objects
"Resource objects" appear in a JSON API document to represent resources.
A resource object MUST contain at least the following top-level members:
	•	id
	•	type
Exception: The id member is not required when the resource object originates at the client and represents a new resource to be created on the server.
In addition, a resource object MAY contain any of these top-level members:
	•	attributes: an attributes object representing some of the resource's data.
	•	relationships: a relationships object describing relationships between the resource and other JSON API resources.
	•	links: a links object containing links related to the resource.
	•	meta: a meta object containing non-standard meta-information about a resource that can not be represented as an attribute or relationship.
```

Is this not the right interpretation? Reading the docs looks to me that is allowed both at the `top level` and  the `resource object` level.

beauby:
The spec, indeed allows it. It is not yet possible in AMS, but there's a PR in the making for this.

rafael:
@beauby oh that's cool. Could add a reference to the PR?

tylerwillingham:
My mistake. I should have provided a link to the section I was looking at it in the docs. It didn't give any mention of that :)

beauby:
@rafael Just issued one. If you have time to write a few tests, it would certainly help getting it merged quickly. Otherwise I'll write some when I get a chance.

rafael:
@beauby. Awesome. Will do, I can get some tests written by EOD tomorrow. Closing this ticket in favor of #1340.

[...]

Issue #1235, opened 2015-10-05
Full JSON-API support


joaomdmoura:
### **tl;dr**
We need to cover what we are missing, mostly the links (all-levels)

---
Right now we are covering most of JSON-API conventions but there are some tiny parts missing and we need sync this up.
We are really missing all-levels links support, right now I think there might be just one missing, but we also need to double check overall convention to make sure we are good to go.

tommyblue:
Hi @joaomdmoura, is it possible with 0.10.0.rc3 to use `links`? I tried with:

```
render json: @categories, links: { "self": "http://example.com/api/posts" }, each_serializer: Api::V1::CategorySerializer
```

but I only have the "data" key in the JSON object :/

beauby:
@tommyblue Could you try #1247?

tommyblue:
it's ok!
I'd like to find a good way to automatically add at least the `self` link.
Something like this: `links: { self: { href: request.url } }`
Probably it could be automatically added if set using a config variable

bf4:
The adapter has access to the request object (as context)

bf4:
@joaomdmoura I'd like to break up the errors rfc into 1) the mvp error objects  2) all error objects 3) decide if we require an errors key or an ActiveModel::Errors or something that responds to errors 4) to what extent error codes and status codes live in AMS

joaomdmoura:
@bf4 :+1: thx for update the issue. 
What exactly do you mean by break up? different issues / PRs? check boxes?
Let's keep the focus on MVP

tchak:
It seems relationship-level links and meta are missing from the checklist

tchak:
Relationship-level pagination links are also not supported right now I belive

beauby:
@tchak Very true, I edited the checklist. Thanks!

tchak:
@beauby I found a pull request you can link to #1028

bf4:
@rails-api/ams per http://jsonapi.org/schema pasted below we can break our goals up into

- [ ] success
  - [ ] data: `"$ref": "#/definitions/data"`
  - [ ] included: array of unique items of type `"$ref": "#/definitions/resource"`
  - [ ] meta: `"$ref": "#/definitions/meta"`
  - [ ] links:
    - [ ] link: `"$ref": "#/definitions/links"`
    - [ ] pagination: ` "$ref": "#/definitions/pagination"`
[...]

PR #1340, opened 2015-11-22
Add support for resource-level meta.

beauby:
Untested, but should work. Would love for somebody to provide a few tests, as I am rather busy myself at the moment.
Ref: #1339, #1235.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -23,6 +23,23 @@ module ActiveModel
     include Type
     require 'active_model/serializer/adapter'
 
+    with_options instance_writer: false, instance_reader: false do |serializer|
+      serializer.class_attribute :_meta # @api private : meta definition, @see Serializer#meta
+    end
+
+    # Register a meta attribute for the corresponding resource.
+    #
+    # @param [Hash] hash Optional hash
+    # @param [Block] block Optional block
+    def self.meta(hash = nil, &block)
+      self._meta =
+        if !block.nil?
+          block
+        else
+          hash
+        end
+    end
+
     # @param resource [ActiveRecord::Base, ActiveModelSerializers::Model]
     # @return [ActiveModel::Serializer]
     #   Preferentially returns
@@ -125,6 +142,14 @@ module ActiveModel
         self.class._fragmented.read_attribute_for_serialization(attr)
       else
         object.read_attribute_for_serialization(attr)
+      end
+    end
+
+    def meta
+      if self.class._meta.respond_to?(:call)
+        instance_eval(&self.class._meta)
+      else
+        self.class._meta
       end
     end
 
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -9,6 +9,7 @@ require 'active_model/serializer/configuration'
 require 'active_model/serializer/fieldset'
 require 'active_model/serializer/lint'
 require 'active_model/serializer/links'
+require 'active_model/serializer/meta'
 require 'active_model/serializer/type'
 
 # ActiveModel::Serializer is an abstract class that is
@@ -20,25 +21,9 @@ module ActiveModel
     include Attributes
     include Caching
     include Links
+    include Meta
     include Type
     require 'active_model/serializer/adapter'
-
-    with_options instance_writer: false, instance_reader: false do |serializer|
-      serializer.class_attribute :_meta # @api private : meta definition, @see Serializer#meta
-    end
-
-    # Register a meta attribute for the corresponding resource.
-    #
-    # @param [Hash] hash Optional hash
-    # @param [Block] block Optional block
-    def self.meta(hash = nil, &block)
-      self._meta =
-        if !block.nil?
-          block
-        else
-          hash
-        end
-    end
 
     # @param resource [ActiveRecord::Base, ActiveModelSerializers::Model]
     # @return [ActiveModel::Serializer]
@@ -142,14 +127,6 @@ module ActiveModel
         self.class._fragmented.read_attribute_for_serialization(attr)
       else
         object.read_attribute_for_serialization(attr)
-      end
-    end
-
-    def meta
-      if self.class._meta.respond_to?(:call)
-        instance_eval(&self.class._meta)
-      else
-        self.class._meta
       end
     end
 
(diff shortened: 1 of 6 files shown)


kurko:
@gaganawhad following your offer in

> I would be happy to try submitting a PR For Resource-level meta. Does that work? Any reason not to ? Let me know! Thanks. #1235 

Do you think you could give a hand here? All help is welcome :)

beauby:
Thanks @kurko! @gaganawhad in case you'd like to give a hand, stuff is happening [on my branch](https://github.com/beauby/active_model_serializers/pull/1).

beauby:
Waiting for travis, but otherwise looks good to me. @bf4 @NullVoxPopuli, thoughts?

test/adapter/json_api/resource_meta_test.rb
```diff
@@ -0,0 +1,63 @@
...
+          def setup
+            @post = Post.new(id: 1337, comments: [], author: nil)
+          end
+
+          def test_meta_hash_object_resource
+            hash = ActiveModel::SerializableResource.new(
+              @post,
+              serializer: MetaHashPostSerializer,
+              adapter: :json_api
+            ).serializable_hash
```
bf4: fwiw, there's a test support helper `serializable` but it wouldn't make things much shorter.  Do you think we should test Ruby hashes made from `as_json` instead of serializable_hash in the future?

bf4:
Just thinking out loud:

Given that `Meta` is JSON API object that can appear in a number of places (https://github.com/rails-api/active_model_serializers/blob/master/docs/jsonapi/schema.md#json-api-document-schema), perhaps we can re-use a JSON API meta object?  I'm not quite sure how it would look.

Where meta is used:

- topLevel.meta
- topLevel.jsonapi.meta
- resource.meta
- resource.links; link.linkObject.meta
- resource.relationships.meta
- resource.relationships.relationshipToOne.linkage.meta
- resource.relationships.relationshipToMany.linkages; linkage.meta
- errors; error.meta

I'm a little iffy with defining the meta on the serializer, but I'm not sure where it would otherwise go.  [JSONAPI-Resources](https://github.com/cerebris/jsonapi-resources/blob/7b3cdffc5740647ac5a854bc3d3707895220b212/README.md#resource-meta) puts the meta in the `Resource`, so, that's good enough for me.  Should we consider defining it at the instance level? `_serializer_meta`?

beauby:
@bf4 The way I understand the spec, resource-level `meta` is a legitimate part of a resource's representation (unlike the toplevel meta which contains data related to the document itself, and therefore belongs in the adapter), which in our case is specified by the serializers, so I think it does belong in the serializer.

lib/active_model/serializer.rb
```diff
@@ -233,6 +248,14 @@ def links
       self.class._links
     end
 
+    def meta
```
bf4: Should we prefer  `_meta` even though there shouldn't be attribute method conflicts anymore?

test/serializers/meta_test.rb
```diff
@@ -125,6 +127,20 @@ def test_meta_is_present_on_arrays_with_root
...
+        assert_equal(blog_meta_serializer.meta, stuff: 'value')
+      end
+
+      def test_meta_is_set_with_block
+        MetaBlogSerializer.meta do
+          { articles_count: object.articles.count }
+        end
+        blog_meta_serializer = MetaBlogSerializer.new(@blog)
+        assert_equal(blog_meta_serializer.meta, articles_count: @blog.articles.count)
+      end
```
bf4: I'd like to validate that meta creates valid JSON API objects.  Do you think this PR is a good spot for that?

bf4:
@beauby would you mind adding the `Serializer::meta` method to https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md ? and maybe reference this PR and its interface in https://github.com/rails-api/active_model_serializers/blob/fce856fc83/docs/jsonapi/schema.md#json-api-document-schema and https://github.com/rails-api/active_model_serializers/blob/fce856fc83/docs/jsonapi/schema.md#definitions

lib/active_model/serializer.rb
```diff
@@ -50,6 +50,8 @@ def self.digest_caller_file(caller_line)
       serializer.class_attribute :_type, instance_reader: true
       serializer.class_attribute :_links         # @api private : links definitions, @see Serializer#link
       self._links ||= {}
+      class_attribute :_meta                     # @api private : meta definition, @see Serializer#meta
```
beauby: This should probably be `serializaer.class_attribute :_meta`.

beauby:
Resource-level meta tests are expected to fail after last commit, since the handling of the option itself (`instance_eval`uating the block) is now left to the adapter.

kurko:
@beauby what does that mean? Expected to fail, which means this PR is WIP?

beauby:
@kurko I meant that those 2 failing tests were testing implementation-specific stuff that does not exist anymore. I'll update them (or get rid of them if they don't make sense at all anymore) tonight if I get a chance, or tomorrow. Thanks for the ping!

bf4:
Merged. Added followup issues and addressed more urgent ones

rafael:
Hey guys - question that I'm not sure from the jsonapi spec. Should the meta be rendered when you include a resource in a request? Let's say you have something like this:

```
GET /users/:id

=> 

    {
      "id": "e877c0cb-83a9-4058-a033-aa67d3683feb",
      "type": "users",
      "attributes": {
        "username": "test"
      },
      "meta": {
        "test_meta": "test
      }
    }
```

Should the following request include the meta information for the user?

```
GET /products/:id?include=users

{
  "data": {
    "id": "ce3402d7-23c8-4c27-bc0f-08b47a5edc39",
    "type": "products",
    "attributes": {
      "title": "asdfsaf"
    },
    "relationships": {
      "user": {
        "data": {
          "id": "6dbf0bb8-b9f5-4658-ab5d-f16663b1f3c2",
          "type": "users"
        }
      }
    }
  },
  "included": [
    {
      "id": "6dbf0bb8-b9f5-4658-ab5d-f16663b1f3c2",
      "type": "users",
      "attributes": {
        "username": "test"
      },
      "meta": {
        "test_meta": "test
      }
    }
  ]
}


```

groyoh:
@rafael Since the `meta` is used for custom information, I guess that's "up to you". Using AMS, I don't think there is a way of removing it or providing a different `meta` when the resource is in `included` though.

Since you asked about the spec itself, you would probably get a more appropriate answer at discuss.jsonapi.org. :wink:

PR #1336, opened 2015-11-19
Grape formatter feature requested in #1258 - Rebased and Repushed (#1273)

johnhamelink:
I've rebased, squashed and repushed #1273 so we can get this merged!

 - adds handling for when the returned resource is not serializable via ams
 - fix for when resource is an Array
 - Moves grape include to grape namespace. Changes Enumerable to Array because a plain hash is enumerable.
 - Add integration test
 - Refine scope of Grape version dependency
 - Assert that the response is equal to a manually defined JSON string
 - Add single module to include in Grape projects
 - Create a Serializable Resource to test rails-api from Grape
 - Update docs
 - Fix discrepency between ActiveRecord 4.0 - 4.1 and 4.2
 - Remove different behaviour for Arrays
 - Clean up copy in documentation

**Edit:** Just want to explicitly point out that credit for code goes to Julian Paas (https://github.com/jpaas)

## Todos:

- [x] Add instructions to docs
- [x] Use `self.call` instead of a singleton
- [x] Remove `.gitignore`
- [x] Add Unit tests
- [x] Add single module for including the formatter & helper in their Grape project
- [x] Assert JSON response & not the ability for the JSON parser to parse it (see [here](https://github.com/rails-api/active_model_serializers/pull/1336/files#r45619253))
- [x] Fix tests for Rails < 4.1
- [x] Squash & Rebase

PR #1341, opened 2015-11-24
Fix bundler caching in travis & Appveyor

johnhamelink:
I noticed while working on #1336 that Grape is not caching Gems correctly: 

screen shot 2015-11-24 at 00 39 31

This PR is an attempt to fix this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: ruby
 
 sudo: false
 
-cache:
-  bundler: true
-
 rvm:
   - 1.9.3
   - 2.0.0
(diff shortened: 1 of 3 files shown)


johnhamelink:
Caching is now working. Future builds should happen much faster!

screen shot 2015-11-24 at 01 03 33

.travis.yml
```diff
@@ -12,8 +9,10 @@ rvm:
   - ruby-head
   - rbx-2
 
-install:
-  - bundle install --retry=3
+install: bundle install --path=vendor/bundle --retry=3 --jobs=3
```
bf4: weird. I thought travis just did this

johnhamelink: The reason it wasn't caching was because the bundle install command was explicitly defined, so it overrode whatever default travis defines. I fixed this by setting a dir to install into, then explicitly telling travis to cache that dir :smile:

PR #1346, opened 2015-11-25
Modify cache key to prevent errors and incorrect data returned. Fixes #1344

### LINKED ISSUE #1344: Issues with #object_cache_key method
In my use case, I have 2 separate issues that I've come across with the current implementation of the `object_cache_key` method.

The first is when I use a PORO object that does not have an `updated_at` attribute, I've defined a cache key, but `object_cache_key` does not look for that until after it builds `object_time_safe` and therefore throws an error.

I overrode `object_cache_key` in my app to look to see if `cache_key` is defined up front and just use that is it's there.

My other issue is, there are times I use different adapters for the same object in different parts of the app.  Since different adapters use the same cache_key, the second adapter to get used returns the wrong format.  

I overrode `object_cache_key` to include the adapter in the cache key to prevent this.

Here is my override.  

```
      # overridden so it uses the object's cache key if it's defined instead of only calling it if @klass._cache_key is not present
      # and to include the adapter type in the cache key.
      def object_cache_key
        @klass._cache_options = Hash(@klass._cache_options) # while we are overriding things, we might as well convert nil options to a Hash to prevent a ReadThis error.
        if @cached_serializer.object.respond_to? :cache_key
          cache_key = "#{self.class.to_s.demodulize.underscore}/#{@cached_serializer.object.cache_key}"
          cache_key.gsub!(cache_key.split('/').second, @klass._cache_key.to_s) if @klass._cache_key
          return cache_key
        end
        object_time_safe = @cached_serializer.object.updated_at
        object_time_safe = object_time_safe.strftime('%Y%m%d%H%M%S%9N') if object_time_safe.respond_to?(:strftime)
        "#{self.class.to_s.demodulize.underscore}/#{@klass._cache_key}/#{@cached_serializer.object.id}-#{object_time_safe}"
      end
```

kevintyll:
When caching, return the object's cache_key up front if it's defined.  This will prevent objects PORO objects that don't have updated_at defined, from throwing an error.  Not as big a deal now that PORO objects can inherit ActiveModelSerializers::Model, but still necessary if it's not inherited for whatever reason.

Add the Adapter type to the cache key.  This prevents incorrect results when the same object is serialized with different adapters.

fixes #1344


lib/active_model/serializer.rb
```diff
@@ -55,7 +55,7 @@ def self.digest_caller_file(caller_line)
 
       serializer.class_attribute :_cache         # @api private : the cache object
       serializer.class_attribute :_fragmented    # @api private : @see ::fragmented
-      serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key
+      serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key.  Ignored if the serialized object defines #cache_key.
```
bf4: `s/serialized/serializable` ?

test/fixtures/poro.rb
```diff
@@ -16,8 +16,8 @@ def method_missing(meth, *args)
     end
   end
 
-  def cache_key_with_digest
-    "#{cache_key}/#{FILE_DIGEST}"
+  def digest
+    FILE_DIGEST
   end
```
bf4: Could you add a comment summarizing this change?

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -27,17 +29,21 @@ def fragment_cached?
           @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
         end
 
-        def cache_key
+        def cache_key(adapter_instance)
           parts = []
           parts << object_cache_key
+          parts << adapter_instance.class.to_s.demodulize.underscore
```
bf4: So, the adapter name is now part of the cache_key. Aside, seems the adapter should define its own name rather than rely on https://github.com/rails-api/active_model_serializers/blob/8aaaa44ba3d5b328fbbf92d59b9f3ecf1b802c8f/lib/active_model/serializer/adapter.rb#L48-L49 making it up...

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -27,17 +29,21 @@ def fragment_cached?
...
         end
 
         def object_cache_key
+          return @cached_serializer.object.cache_key if @cached_serializer.object.respond_to? :cache_key
+
+          fail(UndefinedCacheKey, "#{@cached_serializer.object} must define #cache_key, or the cache_key option must be passed into cache on #{@cached_serializer}") if @klass._cache_key.blank?
           object_time_safe = @cached_serializer.object.updated_at
           object_time_safe = object_time_safe.strftime('%Y%m%d%H%M%S%9N') if object_time_safe.respond_to?(:strftime)
-          (@klass._cache_key) ? "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{object_time_safe}" : @cached_serializer.object.cache_key
+          "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{object_time_safe}"
```
bf4: So, the diff is essentially

```diff
- if @klass._cache_key
+ if !@cached_serializer.object.respond_to?(:cache_key)
+    if @klass._cache_key.blank?
+     fail UndefinedCacheKey, "'#{@cached_serializer.object}' must define #cache_key, " <<
+       "or the cache_key option must be passed into cache on '#{@cached_serializer}'."
+    end
  object_time_safe = @cached_serializer.object.updated_at
  object_time_safe = object_time_safe.strftime('%Y%m%d%H%M%S%9N') if object_time_safe.respond_to?(:strftime)
  "#{@klass._cache_key}/#{@cached_serializer.object.id}-#{object_time_safe}" 
else
  @cached_serializer.object.cache_key
end
```

bf4: So that it 

1. prefers the `@cached_serializer.object.cache_key` over `@klass._cache_key`
1. raises an exception if there's no cache key on the cached object nor the serializer

Where previously it would

1. Prefer `@klass._cache_key`
1. When `@klass._cache_key` was not defined, use `@cached_serializer.object.cache_key` even it weren't defined

test/fixtures/poro.rb
```diff
@@ -38,9 +38,16 @@ class ProfilePreviewSerializer < ActiveModel::Serializer
...
+Author   = Class.new(Model) do
+  # To confirm cache_key is set using updated_at and cache_key option passed to cache
+  undef_method :cache_key
+end
 Bio      = Class.new(Model)
 Blog     = Class.new(Model)
+Article  = Class.new(Model) do
+  # To confirm error is raised when cache_key is not set and cache_key option not passed to cache
+  undef_method :cache_key
+end
```
bf4: maybe we can subclass this in the test as

```ruby
class UncachedArticle < Article
  undef_method :cache_key
end
```

so the fixture here doesn't need to change? etc.

Also, the ArticleSerializer below can be defined in the test, rather than adding a new one here

bf4:
@kevintyll hi :)

kevintyll:
Hi, sorry I haven't responded.  I haven't had a chance to work on your suggestions yet.  I'm hoping to be able to get to them after the first of the year.

bf4:
soon... :)

bf4:
:sunrise:

kevintyll:
Hoping to get to it Wednesday.

lib/active_model/serializer.rb
kevintyll: Don't know what you're asking here.

bf4: in the diff, to change `Ignored if the serialized object defines #cache_key.` to `Ignored if the serializable object defines #cache_key.`

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -27,17 +29,21 @@ def fragment_cached?
           @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
         end
 
-        def cache_key
+        def cache_key(adapter_instance)
           parts = []
           parts << object_cache_key
+          parts << adapter_instance.class.to_s.demodulize.underscore
           parts << @klass._cache_digest unless @klass._cache_options && @klass._cache_options[:skip_digest]
           parts.join('/')
```
bf4: urgh, we need to document this option

test/serializers/cache_test.rb
```diff
@@ -64,15 +66,21 @@ def test_cache_key_definition
         assert_equal(nil, @comment_serializer.class._cache_key)
       end
 
-      def test_cache_key_interpolation_with_updated_at
+      def test_error_is_raised_if_cache_key_is_not_defined_on_object_or_passed_as_cache_option
+        assert_raises ActiveModel::Serializer::Adapter::CachedSerializer::UndefinedCacheKey do
```
bf4: lulz to the long constant

test/serializers/cache_test.rb
```diff
@@ -64,15 +66,21 @@ def test_cache_key_definition
...
+          render_object_with_cache(@article)
+        end
+      end
+
+      def test_cache_key_interpolation_with_updated_at_when_cache_key_is_not_defined_on_object
         render_object_with_cache(@author)
-        assert_equal(nil, ActionController::Base.cache_store.fetch(@author.cache_key))
-        assert_equal(@author_serializer.attributes.to_json, ActionController::Base.cache_store.fetch("#{@author_serializer.class._cache_key}/#{@author_serializer.object.id}-#{@author_serializer.object.updated_at.strftime("%Y%m%d%H%M%S%9N")}").to_json)
+        key = cache_key_with_adapter("#{@author_serializer.class._cache_key}/#{@author_serializer.object.id}-#{@author_serializer.object.updated_at.strftime("%Y%m%d%H%M%S%9N")}")
+        assert_equal(@author_serializer.attributes.to_json, ActionController::Base.cache_store.fetch(key).to_json)
```
bf4: - [ ] TODO should use `ActiveModelSerializers.config.cache_store` instead, but maybe in another PR

lib/active_model/serializer/adapter/cached_serializer.rb
bf4: Can we move the `cache_key` method into the serializer? Would simplify things I think

[Serializer::Caching](https://github.com/rails-api/active_model_serializers/blob/7d4f0c5c8af93c32bd69ed36b2433de42243f93c/lib/active_model/serializer/caching.rb)

```ruby
module ActiveModel
  class Serializer
    module Caching
      extend ActiveSupport::Concern

      # Use object's cache_key if available, else derive a key from the object
      # Pass the `key` option to the `cache` declaration or override this method to customize the cache key
      def cache_key
        if object.respond_to?(:cache_key)
          object.cache_key
        else
          object_time_safe = object.updated_at
          object_time_safe = object_time_safe.strftime('%Y%m%d%H%M%S%9N') if object_time_safe.respond_to?(:strftime)
          "#{self.class._cache_key}/#{object.id}-#{object_time_safe}"
        end
      end
    end
  end
end
```

bf4: And 

```ruby
module ActiveModel
  class Serializer
    module Adapter
      class CachedSerializer
        def cache_key
          parts = []
          parts << @cached_serializer.cache_key
          parts << @klass._cache_digest unless @klass._cache_options && @klass._cache_options[:skip_digest]
          parts.join('/')
        end
      end
    end
  end
end
```

and the serializer really should handle 


```ruby
#          parts << @klass._cache_digest unless @klass._cache_options && @klass._cache_options[:skip_digest]
def self._skip_digest?
  _cache_options && _cache_options[:skip_digest]
end

# so we can
parts << @klass._cache_digest unless @klass._skip_digest?
```

test/fixtures/poro.rb
bf4: looks like it should be a test expectation helper as it's not really a model concern, as I review it.  I don't think the new `digest` is used at all, right?

lib/active_model/serializer/adapter/cached_serializer.rb
bf4: I'm not sure if we need an exception for the case where the object has `cache_key` defined but it is nil or empty string?

kevintyll:
Let me know about the change I made to the register method with the adapter#name.  I'll continue tomorrow.

bf4:
The inclusion of adapter type is a good idea but the impl change is too big for this pr, needs to be discussed elsewhere... I'm ok surfacing the adapter name as a class method, but interface of registering shouldn't change here (reviewing on phone, pardon if I missed something.)

bf4:
I think serializer#cache_key is the more fundamental change https://github.com/rails-api/active_model_serializers/pull/1346#discussion_r48897427

test/fixtures/poro.rb
kevintyll: Only a single test is using this.  The method in unnecessary, the test can just use the constant directly.

lib/active_model/serializer/adapter/cached_serializer.rb
bf4: @maurogeorge @kevintyll thoughts? --^

test/fixtures/poro.rb
kevintyll: Why wouldn't you prefer for all the objects to be defined in a single location given the existence of this file.  It seems to me that if you want the test using these objects to create the objects they will test, then all tests should do the same and eliminate the need for this file.  Is that the direction you're going?

bf4: @kevintyll yup.  I'd love to get rid of most of it, but some is useful, and maybe we can just create a dummy rails app for testing like in https://github.com/rails-api/active_model_serializers/pull/1412 

I think that would be better in the long run, e.g. how rails plugins create a `dummy` app

kevintyll: Great, I'll move my new objects.  I'm a fan of each test being responsible for it's own objects, but I'm a bigger fan of consistency.  Wanted to double check.

lib/active_model/serializer/adapter/cached_serializer.rb
kevintyll: I have it in the initializer now with tests passing.  Let me know if you'd prefer not to throw the error.

```

        def initialize(serializer)
          @cached_serializer = serializer
          @klass             = @cached_serializer.class
          if cached? && !@cached_serializer.object.respond_to?(:cache_key) && @klass._cache_key.blank?
            fail(UndefinedCacheKey, "#{@cached_serializer.object} must define #cache_key, or the cache_key option must be passed into cache on #{@cached_serializer}")
          end
        end
```

kevintyll:
Okay, I think I got it all.  Not sure what happened to the diff.  Looks like it's picking things up from the merge of master into my fork.

bf4:
Probably rebase and force push helps

B mobile phone

kevintyll:
Rebasing didn't do anything.  Had some issues merging upstream into my fork.  I probably screwed up the commit history.  I confident the code is correctly merged though.

bf4:
Continued in #1642

Issue #923, opened 2015-05-21
Cache not working


rainerborene:
Not sure about how the `cache` method works but seems not to be working.

```ruby
class CommentSerializer < ActiveModel::Serializer
  cache only: [:content]
  attributes :id, :content

  def content
     # nokogiri parsing stuff.
  end
end
```

Even with the `cache` enabled I can see some objects being allocated by the `nokogiri` gem.

joaomdmoura:
The cache feature uses the Rails cache, so ensure that it's enable for the the environment that you are testing :smile:

rainerborene:
@joaomdmoura `perform_caching` option is already enabled.

rainerborene:
I suggest you to debug it using the `rack-mini-profiler` gem to measure the number of objects allocated during a request.

MrWEST:
+1

eliduke:
I recently upgraded my AMS gem from `0.8.3` to `0.10.0.rc1` in hopes of getting some better caching functionality. I am trying to cache some App Label and the serializer there looks like this:

```
class AppLabelSerializer < ActiveModel::Serializer
  cache key: 'app_label', expires_in: 3.hours
  attributes :id, :key, :label, :label_plural

  def key
    object.app_label_dictionary.key
  end

end
```

But I'm not seeing any caching in my local server. Previously, in the older version of AMS, I used this caching syntax:

```
class AppLabelSerializer < ActiveModel::Serializer
  cached
  delegate :cache_key, to: :object

  attributes :id, :key, :label, :label_plural
end
```

And there was a **very clear** indication in the server output showing when the db was being queried vs a cache_read. Since switching to the release candidate and using the new caching syntax I'm not seeing any caching. Is there something else that I am missing? Some specific config?

groyoh:
@rainerborene do you have a non-cached attribute that calls `content`,e.g.
```ruby
attribute :another
def another
  content
  # and other stuffs
end
```
@joaomdmoura non-cached attributes (overriden) that call cached attributes won't use the cached values because [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L205) we woul call the attribute on the original serializer and not the cached one.

rainerborene:
@groyoh I don't think so.

```ruby
class BillSerializer < ActiveModel::Serializer
  include BillsHelper

  attributes :id, :content
  cache only: [:content]

  def content
    clean(object.content)
  end
end
```

[...]

PR #1352, opened 2015-11-30
Fix generators (@dgynn); load Railtie only with Rails, ensures caching configured

bf4:
## Fixing the generators (@dgynn)

Only load generators when needed
- use hook_for to hook in the serializer and remove load_generators
- move generators so they can be found by rails
- move to_prepare block to railtie config

This commit improves the way the generators are loaded and how
they extend the resource generator.

* The initializer block has been changed to a `generator` block which is only executed when generators are needed.
* The call to `app.load_generators` has been removed. There is no need to load *all* generators.
* The `resource_override.rb` has been changed to use `hook_for` to extend the resource generator.
* The directory for the generators has been moved to match the way Rails looks to load generators.

With `hook_for` it would now be possible for a user to pass `--no-serializer` to skip that option.
The `--serialize` option also now shows up in the generator help with `rails g resource --help`.

These changes follow the way the Draper gem extends the `controller` generator.

## Putting Rails Code in the Railties (@bf4)

akak let people use AMS [without requiring 'rails' or 'action{pack,view}'](https://github.com/rails-api/active_model_serializers/issues/1389)

Only call railtie when Rails is defined; assume controller loaded

Isolated Testing

- Rake test inspired by https://github.com/rails/rails/blob/v5.0.0.beta1/activejob/Rakefile
- Isolated unit inspired by
  - https://github.com/rails/rails/blob/v5.0.0.beta1/railties/test/isolation/abstract_unit.rb
  - https://github.com/rails/rails/blob/v5.0.0.beta1/activemodel/test/cases/railtie_test.rb

Misc

- Turns out `mattr_accessor(:logger) {
  ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT)) }`
  was always nil until the Railtie was loaded, since mattr_accessor
  block defaults don't really work on modules, but on the classes that
  include them.
- Should resolve https://github.com/rails-api/active_model_serializers/issues/1389 where Railties was being too-eagerly loaded.
- Commented on *importance of Rails being required first for caching to
  work*. Ref https://github.com/rails-api/active_model_serializers/issues/923#issuecomment-167067238

  ```diff
+    # To be useful, this hook must run after Rails has initialized,
 +    # BUT before any serializers are loaded.
 +    # Otherwise, the call to 'cache' won't find `cache_store` or `perform_caching`
 +    # defined, and serializer's `_cache_store` will be nil.
 +    # IF the load order cannot be changed, then in each serializer that that defines a `cache`,
 +    # manually specify e.g. `PostSerializer._cache_store = Rails.cache` any time
 +    # before the serializer is used.  (Even though `ActiveModel::Serializer._cache_store` is
 +    # inheritable, we don't want to set it on `ActiveModel::Serializer` directly unless
 +    # we want *every* serializer to be considered cacheable, regardless of specifying
 +    # `cache # some options` in a serializer or not.
 +    initializer 'active_model_serializers.caching' => :after_initialize do
 +      ActiveModelSerializers.config.cache_store     = ActionController::Base.cache_store
 +      ActiveModelSerializers.config.perform_caching = Rails.configuration.action_controller.perform_caching
 +    end
  ```
- In isolated tests, `active_support/core_ext/object/with_options` is required.


lib/active_model_serializers/railtie.rb
```diff
@@ -1,7 +1,17 @@
...
   class Railtie < Rails::Railtie
+    initializer 'active_model_serializers.action_controller' do
+      ActiveSupport.on_load(:action_controller) do
+        ActiveSupport.run_load_hooks(:active_model_serializers, ActiveModelSerializers)
+        include ::ActionController::Serialization
+        ActionDispatch::Reloader.to_prepare do
+          ActiveModel::Serializer.serializers_cache.clear
+        end
+      end
+    end
```
bf4: I wonder if maybe this shouldn't be in an initializer

lib/active_model_serializers/railtie.rb
```diff
@@ -15,7 +25,7 @@ class Railtie < Rails::Railtie
       end
     end
 
-    initializer 'generators' do |app|
+    initializer 'active_model_serializers.generators' do |app|
```
bf4: I don't think this name is important, but it is old

lib/active_model_serializers/railtie.rb
bf4: It does, however, make it easier to conditionally disable including the module into action controller. e.g. https://github.com/rails-api/active_model_serializers/pull/592/files

bf4:
cc @rails-api/ams

lib/active_model_serializers/railtie.rb
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1374

lib/active_model_serializers/railtie.rb
dgynn: Having the `include ::ActionController::Serialization` in an initializer is good for being able to conditionally disable it based on whatever gets decided from #1295. And having the `on_load(:action_controller)` is good for lazy-loading in case `ActionController::Base` is never loaded. Those both seem good.

The `to_prepare` block does not need to be in the initializer. It could be outside the initializers on the Railtie itself with just...
```ruby
  config.to_prepare do
    ActiveModel::Serializer.serializers_cache.clear
  end
```

Using an initializer, and a load hook (`:action_controller`), and then running another load hook (`:active_model_serializers`) to set the logger seems like overly complicated. Depending on load ordering, someone could have an initializer that explicitly sets `ActiveModelSerializers.logger` but then gets overwritten. 
Is there any other reason to have the `:active_model_serializers` lazy-load hook? Could that logic just get folded in where the `run_load_hooks` call is? Would you ever expect someone else to hook into that? And if so, would it only be after :action_controller has been loaded?

I'm still poking around a bit to see if I've got the order right on some of these.

bf4: @dgynn awesome! Thanks for looking at this! Feel free to submit a PR into it if you'd like.

Also, see https://github.com/rails-api/active_model_serializers/issues/1353 which is similar to https://github.com/rails-api/active_model_serializers/pull/1295

bf4: @dgynn 

> Using an initializer, and a load hook (:action_controller), and then running another load hook (:active_model_serializers) to set the logger seems like overly complicated. Depending on load ordering, someone could have an initializer that explicitly sets ActiveModelSerializers.logger but then gets overwritten. 
> Is there any other reason to have the :active_model_serializers lazy-load hook? 

Since the logger isn't necessarily  derived from the controller, I'm just being paranoid that someone might configure the logger before the controller was loaded.  This makes sure the user has the final say.. but I could me misthinking that

bf4:
ref: @dgynn has made a PR into the merge source https://github.com/bf4/active_model_serializers/pull/1

bf4:
@dgynn merged in https://github.com/bf4/active_model_serializers/pull/1

Rakefile
```diff
@@ -44,7 +44,20 @@ Rake::TestTask.new do |t|
...
+namespace :test do
+  namespace :isolated do
+    desc 'Run isolated tests for Railtie'
+    task :railtie do
+      dir = File.dirname(__FILE__)
+      file = "#{dir}/test/active_model_serializers/railtie_test_isolated.rb"
+      sh(Gem.ruby, '-w', "-I#{dir}/lib", "-I#{dir}/test", file) or fail 'Failures' # rubocop:disable Style/AndOr
+    end
+  end
+end
```
bf4: see commit message, but basically this is adapted from ActiveJob Rakefile

.travis.yml
```diff
@@ -33,8 +33,8 @@ matrix:
   include:
   - rvm: 2.2
     env: CAPTURE_STDERR=true
-  - rvm: jruby-9000
-    env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false -Xcli.debug=true --debug'
+  - rvm: jruby-9.0.4.0
+    env: JRUBY_OPTS='-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug'
```
beauby: Does this change belong here?

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Fixes:
 - [#1239](https://github.com/rails-api/active_model_serializers/pull/1239) Fix duplicates in JSON API compound documents (@beauby)
 - [#1214](https://github.com/rails-api/active_model_serializers/pull/1214) retrieve the key from the reflection options when building associations (@NullVoxPopuli, @hut8)
 - [#1358](https://github.com/rails-api/active_model_serializers/pull/1358) Handle serializer file paths with spaces (@rwstauner, @bf4)
+- [#1352](https://github.com/rails-api/active_model_serializers/pull/1352) Isolate Rails-specifc code in Railties (@dgynn, @bf4)
```
beauby: Should new changelogs be added on top or at the bottom? (Real question, I have no answer)

beauby:
I'm not familiar enough with railties. Would you mind summarizing what this PR does?

.travis.yml
bf4: No, from the jruby pr.  Had a failute

B mobile phone

CHANGELOG.md
bf4: I've always added at the bottom but on reflection, i think all thr earlier and generated are age ascending, newest on top. I should change that

B mobile phone

bf4:
@beauby Updated the description from the commit messages https://github.com/rails-api/active_model_serializers/pull/1352#issuecomment-167889219. Thanks for the nudge

Rakefile
```diff
@@ -44,7 +44,32 @@ Rake::TestTask.new do |t|
...
+        command = "-w -I#{dir}/lib -I#{dir}/test #{file}"
+        full_command = %("#{Gem.ruby}" #{command})
+        system(full_command) or fail 'Failures' # rubocop:disable Style/AndOr
+      end
+    end
+  end
+end
+
+if ENV['RAILS_VERSION'].to_s > '4.0' && RUBY_ENGINE == 'ruby'
+  task default: [:isolated, :test, :rubocop]
```
bf4: isolated tests as set up here break on 4.0 and on JRuby

test/active_model_serializers/railtie_test_isolated.rb
```diff
@@ -0,0 +1,69 @@
+# Execute this test in isolation
```
bf4: note that file is named `_isolated.rb` so that it can be picked up and tested separately

test/active_model_serializers/railtie_test_isolated.rb
```diff
@@ -0,0 +1,69 @@
...
+require 'support/isolated_unit'
+
+class RailtieTest < ActiveSupport::TestCase
+  include ActiveSupport::Testing::Isolation
+
+  class WithRails < RailtieTest
```
bf4: Using subclassing to define different contexts

test/active_model_serializers/railtie_test_isolated.rb
```diff
@@ -0,0 +1,69 @@
...
+    setup do
+      require 'rails'
```
bf4: Compare to `WithoutRails`

```diff
- require 'rails'
+ require 'active_model_serializers'
+ make_basic_app
```

test/active_model_serializers/railtie_test_isolated.rb
```diff
@@ -0,0 +1,69 @@
...
+        "ActionController::Serialization should be included in ActionController::Base, but isn't"
+    end
+
+    test 'sets the ActiveModelSerializers.logger to Rails.logger' do
+      refute_nil Rails.logger
+      refute_nil ActiveModelSerializers.logger
+      assert_equal Rails.logger, ActiveModelSerializers.logger
+    end
+
+    test 'it is configured for caching' do
```
bf4: :boom:

bf4:
@dgynn @beauby @maurogeorge Look ok?  The PR description and individual commits describe it pretty well, I think

dgynn:
@bf4 I have some concerns about the Railtie in the last version. I see that you've been working to get the order right so caching will be set up correctly. And you removed the `on_load(:action_controller)` to get that working.

That works (I assume) but what that means is that the call to `ActionController::Base.send` will cause ActionController to get loaded during initialization. It looks like the ActionController Railtie is *specifically* trying to delay loading `Base` during initialization.
https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/railtie.rb

See how all of their initializers only set up the config and then set hooks for  `on_load(:action_controller)` to complete the initialization. That hook gets called once `ActionController::Base` is referenced.

I think the way around this is that AMS needs to not use the *configured* logger and cache store from ActionController::Base but should instead use the same configuration *options* as ActionController.
 
That would mean:
* Put back the on_load hook that does the `include ::ActionController::Serialization`
* Change the initializer that sets up caching to use `Rails.configuration.action_controller.cache_store || Rails.cache` rather than `ActionController::Base.cache_store`
* Change the initializer that sets up logging to do a similar thing with the configured logger.

Or... maybe AMS just does need to cause ActionController to fully load during initialization. I don't think that is the end of the world. But if it can be avoided it should.

Also, as it stands right now, the load hook for `on_load(:active_model_serializers)` is not doing anything. The first initializer fires the hook before the second initializer has even registered for it. So that could be removed.

dgynn:
I've pushed up a PR to your branch that shows the changes I'd suggest. I did not update all the tests or change the comment about cache setup. Feel free to edit or disregard. :)

BTW, you might be interested in a tool I'm working on that inspects Rails configuration info.
https://github.com/dgynn/tuttle
In particular, it can list out all of the initializers configured for an app and show their order and dependencies. Here is a demo of that with my suggested changes installed. Look at the initializers tab...
https://tuttle-demo.herokuapp.com/tuttle/rails

Working on that is how I came across the generators issue initially.

lib/active_model_serializers.rb
```diff
@@ -1,32 +1,22 @@
...
   autoload :Callbacks
   autoload :Deserialization
   autoload :Logging
-end
 
-require 'active_model/serializer'
-require 'active_model/serializable_resource'
-require 'active_model/serializer/version'
+  class << self; attr_accessor :logger; end
+  self.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
```
maurogeorge: I think the `mattr_accessor(:logger) { ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT)) }` is more legible than this one.

lib/active_model_serializers.rb
```diff
@@ -1,32 +1,22 @@
...
-  include ::ActionController::Serialization
-  ActionDispatch::Reloader.to_prepare do
-    ActiveModel::Serializer.serializers_cache.clear
+  def self.config
+    ActiveModel::Serializer.config
   end
-end
 
-require 'active_model/serializer/railtie'
+  require 'active_model/serializer/version'
```
maurogeorge: What you think about put all the `require`'s on the top of the file?

lib/active_model_serializers.rb
dgynn: I believe mattr_accessor does not support blocks in ActiveSupport 4.0.

lib/active_model_serializers.rb
bf4: https://github.com/rails-api/active_model_serializers/commit/fc04b1400a623b3fefc95bc1f5e5672ccbf61a8a but I don't remember exactly what the failures were

bf4:
@maurogeorge @dgynn updated

lib/active_model_serializers.rb
maurogeorge: :+1:

maurogeorge:
LGTM. I only recommend take a look at the @dgynn comments https://github.com/rails-api/active_model_serializers/pull/1352#issuecomment-171528265 https://github.com/rails-api/active_model_serializers/pull/1352#issuecomment-171554933 that worked on this part of the project :+1:

bf4:
@maurogeorge I merged in his code... but I'm tired.. besides the value in re-reading the comments, was there more work to do that I missed?

maurogeorge:
@bf4 no, to me looks good. @dgynn let us know if you have more concerns about this.

dgynn:
This looks good. I do think the comment in the Railtie about caching can now be removed but that should be verified. 
Also, the 2 initializer blocks for caching and logging could be combined. There really isn't much benefit to having them separate. ActionController itself used to have 2 separate initializers for logging and caching and they combined them here... https://github.com/rails/rails/commit/fbc9d0f44f090a9873834f4966760c3e80682559

bf4:
Can I give you commit to my fork to finish it? I'm on vacation

Re comment there are tests for it

B mobile phone

dgynn:
I've updated the Railtie to combine the initializers and update the comment. I believe this initializer ordering should resolve the caching issue from #923 and also not trigger too-early loading of ActionController::Base. I'd say this is ready to be merged.

bf4:
@dgynn looks good. pr description still good summary?

lib/active_model_serializers.rb
bf4: See pr description. It didn't work

lib/active_model_serializers/railtie.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails/railtie'
+require 'action_controller'
+require 'action_controller/railtie'
```
bf4: This still necessary?

lib/active_model_serializers/railtie.rb
```diff
@@ -0,0 +1,36 @@
...
+require 'action_controller/serialization'
+
+module ActiveModelSerializers
+  class Railtie < Rails::Railtie
+    config.to_prepare do
+      ActiveModel::Serializer.serializers_cache.clear
+    end
```
bf4: We should comment why we do this at some point

lib/active_model_serializers/railtie.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+
+    # This hook is run after the action_controller railtie has set the configuration
+    # based on the *environment* configuration and before any config/initializers are run
+    # and also before eager_loading (if enabled).
+    initializer 'active_model_serializers.set_configs', :after => 'action_controller.set_configs' do
+      ActiveModelSerializers.logger = Rails.configuration.action_controller.logger
+      ActiveModelSerializers.config.cache_store     = Rails.configuration.action_controller.cache_store
+      ActiveModelSerializers.config.perform_caching = Rails.configuration.action_controller.perform_caching
```
bf4: How would someone set different defaults? A regular initializer? I think that's why I had the ams load hook.

lib/active_model_serializers/railtie.rb
dgynn: The AMS railtie *does* require that the action_controller/railtie has been loaded so that `Rails.configuration.action_controller` exists. But there isn't any normal that wouldn't have already been required. It should have been loaded by rails/all or even active_record/railtie (see the comment here https://github.com/rails/rails/blob/master/activerecord/lib/active_record/railtie.rb#L5).

lib/active_model_serializers/railtie.rb
dgynn: Different defaults could get set in an a `config/environments/*.rb` file using `config.cache_store` or `config.action_controller.*`.

In a  regular initializer, a user could set `ActiveModelSerializers.config.cache_store` directly. Basically this hook is setting the default AMS config to match ActionController's default config (including `environments/*.rb` configuration) before regular initializers run.

dgynn: Related to this... It seems like it would be better to *not* copy the `ActiveModelSerializers.config.cache_store` to _cache on the serializers themselves. That creates a situation where the individual serializer's cache store is dependent on the order of configuration files, serializers getting loaded, and lazy-loading of action_controller. That is outside the scope of this PR but something that could be considered.

bf4: Yup

B mobile phone

bf4: I think myvcomment about loaf order is still true given tests pass

B mobile phone

dgynn: Right. The PR comment is still true. It is now just very unlikely that a Serializer could get loaded before ActiveModelSerializer has at least the default cache configuration set up. If someone reconfigures caching in a custom initializer or somehow loads a Serializer before the Railties do basic configuration then they would need to manually set the `_cache` on the Serializer.

lib/active_model_serializers.rb
maurogeorge: :+1:

lib/active_model_serializers/railtie.rb
bf4: Agreed.  Maybe merge this and submit a follow up?  If we're quick we can get it in rc4

I'm on phone so don't have full context of latest comments. I think it's good to go with some points ti follow up on?  Any ideas for a good label name for that?

B mobile phone

bf4:
Ok, anything than needs to be resolved before merging this?  I'll add an issue with label 'pr followup' for what we've discussed

dgynn:
@bf4 This looks good to merge to me.

bf4:
merged. Let the heavens be glad; let the earth rejoice

PR #1354, opened 2015-11-30
Enable use of an `object` key in serilaized forms

noahsilas:
Because all serializers already implement an `object` method (part of the API expected when customizing a serializer), attempting to add a key to the serialized object with the name `object` would not output the expected value, and would instead dump a serialized version of the object under serialization.

This prevents a commonly used identifier from appearing in serialized output. As one real world example, the Stripe API includes an `object` key in each serialized object describing its type.

I introduced a change in the storage for the serializer configuration, and added methods to access it in the same way as before. Open question: I stored the key/attribute mapping in a Hash<Symbol => Symbol>; would the project be better served mapping to a nested hash to support additional configuration options in the future without changing the storage again?

This may have introduced a breaking change; before, methods were defined on the serializer based on the name of the key; now they will be defined based on the name of the underlying attribute. This behavior doesn't seem to be explicitly mentioned in docs or tested anywhere, though, so it may be a non-issue.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -46,10 +46,8 @@ module ActiveModel
 
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
-      class_attribute :_attributes               # @api private : names of attribute methods, @see Serializer#attribute
-      self._attributes ||= []
-      class_attribute :_attributes_keys          # @api private : maps attribute value to explict key name, @see Serializer#attribute
-      self._attributes_keys ||= {}
+      class_attribute :_attributes_map           # @api private : maps attribute key names to names to names of implementing methods, @see Serializer#attribute
+      self._attributes_map ||= {}
       class_attribute :_links                    # @api private : links definitions, @see Serializer#link
       self._links ||= {}
 
@@ -69,12 +67,11 @@ module ActiveModel
       serializer.class_attribute :_cache_digest # @api private : Generated
     end
 
-    # Serializers inherit _attributes and _attributes_keys.
+    # Serializers inherit _attributes_map
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
       caller_line = caller.first
-      base._attributes = _attributes.dup
-      base._attributes_keys = _attributes_keys.dup
+      base._attributes_map = _attributes_map.dup
       base._links = _links.dup
       base._cache_digest = digest_caller_file(caller_line)
       super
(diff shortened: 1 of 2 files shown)


bf4:
@noahsilas Thanks for this. I believe this will be resolved by a PR such as @beauby 's https://github.com/rails-api/active_model_serializers/pull/1262

noahsilas:
@bf4 I tried checking that out locally and adding my test case, but it still fails:
```
 FAIL["test_object_attribute_override", ActiveModel::Serializer::AttributeTest, 0.16997674852609634]
 test_object_attribute_override#ActiveModel::Serializer::AttributeTest (0.17s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -{:blog=>{:object=>"AMS Hints"}}
        +{:blog=>{:object=>#<Blog:0xXXXXXX @attributes={:id=>1, :name=>"AMS Hints", :type=>"stuff"}>}}
        /home/noah/src/active_model_serializers/test/serializers/attribute_test.rb:96:in `test_object_attribute_override'
```

I'll start looking into #1262 and see how I can resolve my commit against it.

bf4:
@noahsilas I'll have a PR soon, just some finishing touches

test/serializers/attribute_test.rb
```diff
@@ -43,6 +43,15 @@ def test_id_attribute_override
...
 
+      def test_object_attribute_override
+        serializer = Class.new(ActiveModel::Serializer) do
+          attribute :name, key: :object
+        end
+
+        adapter = ActiveModel::Serializer::Adapter::Json.new(serializer.new(@blog))
+        assert_equal({ blog: { object: 'AMS Hints' } }, adapter.serializable_hash)
+      end
+
```
bf4: nice

lib/active_model/serializer.rb
```diff
@@ -112,17 +109,30 @@ def self.attributes(*attrs)
...
 
     # @api private
+    # An accessor for the old _attributes internal API
+    def self._attributes
+      _attributes_map.keys
+    end
+
+    # @api private
+    # An accessor for the old _attributes_keys internal API
+    def self._attributes_keys
```
NullVoxPopuli: can you provide an example in the method comment about what this method does (before and after).

This is to help future developers debug more painlessly

NullVoxPopuli:
Just requires a little more inline documentation, but otherwise LGTM! :-)

noahsilas:
@NullVoxPopuli: This ended up getting merged via #1356 as 7bde7bf75. I'm going to close this PR, but I'll come back later with another to improve those docs.

PR #1263, opened 2015-10-09
Refactor attributes

beauby:

Issue #1325, opened 2015-11-10
Avoid DB hits for JSONAPI relationships


richmolj:
I've been searching through the issues, and this may be relevant to bringing `filter` to 0.10.x, but I couldn't find anything on this specifically.

If I use JSONAPI adapter, and put `has_many :comments` in my `PostSerializer`, there will be a DB hit to fill out all the ids for the comment `relationships`. Even when no `include` is specified.

I thought of avoiding this by something like

```ruby
def comments
  object.comments.loaded ? object.comments : []
end
```

Unfortunately the key `comments` will still appear in the response. This means if my client (Ember) has already loaded the post with its comments, then hits this separate endpoint, it will think all the comments for the post have been deleted (versus just not included in the response).

Currently I've been creating a base serializer for every model, then subclassing and adding relationships for each specific endpoint. This not only gets really tedious, but every time I add a relationship I am also specifying `includes`, which makes me think the default should be to not load relationships unless `include` is specified. Finally, it makes it quite hard to support an API like `/posts/1?include=comments` since I need to conditionally select a serializer that actually includes the requested relationships.

Is the correct solution here to wait for `filter`, do the subclassing-style, or something else? Thanks!

beauby:
You could use the `fields` option to filter out your `comments` association. The db should not get hit.

richmolj:
This would work, and the upcoming `filter` would make it even better. My question is, shouldn't this be a default? This would make it much easier to build a `?include=comments`-style API that would not hit the database when the relationship is not included.

richmolj:
Also worth noting if you're using the JSONAPI `links` feature for your relationships (https://github.com/rails-api/active_model_serializers/issues/834, https://github.com/rails-api/active_model_serializers/pull/1282) then putting the `relationships` in the response is not needed and may be invalid. Maybe that should be the logic - if the relationship has `links` then don't include `relationships` in the response?

beauby:
@richmolj the `include` parameter of the JSON API spec does only specify which *associated resources* should be returned in the `included` part of the document, not which relationship `field` should appear in the primary resource (for this, as mentioned, the `fields` option is what you are looking for). So it is expected behavior that the db gets hit, and it is expected behavior that the key appears under `relationships`, even when the relationship is not `included`.

richmolj:
[...]

PR #1356, opened 2015-11-30
Add inline syntax for attributes and associations

bf4:
### Goals:

1. [x] Allow defining attributes so that they don't conflict with existing methods
1. [x] Allow customizing associations e.g. `has_many :comments do last(5) end`
  - [ ] compare to using [`virtual_value`](https://github.com/rails-api/active_model_serializers/pull/1356#discussion_r47146466) such as  `has_many :reviews, virtual_value: [{ id: 1 }, { id: 2 }]`
  and how association virtual values differ from attribute values.
1. [x] Remove dynamically defined methods on the serializer

Based on work in #1262, #1354

Ref: #1263, #1261, #1311

This PR is intended as a **starting point** for further work improving how
serializers map their resources attributes and associations

### Introduced, in support of this:

- [x] `Serializer#_attribute_mappings`
- [x] Extracted Serializer attribute methods to an Attributes module. I'd consider a class but this is a simpler first step and doesn't block changing it to an Attributes class than has many Attribute objects in the future

### [Possible regression](https://github.com/rails-api/active_model_serializers/pull/1356#discussion_r46381686): 

removed referring to object attributes in the serializer without calling them directly on object. (I didn't even know this was possible. We could add a respond_to_missing / method_missing for this)

```diff
   def slug
-    "#{name}-#{id}"
+    "#{object.name}-#{object.id}"
```
- [ ] Discuss

### Associations as 'includes'

It seems like we might benefit from thinking of associations, at least in the JSON API adapter, as includes, and only including them when requested (which is also similar to earlier naming and usage in the library, which I like).  This PR made me think of it, in part, because the difference between a reflection, an association, and associations, I think should be more clear (i.e. just an include).

- [ ] Discuss

### Blocks in serializer attribute definitions

I like this.  They remove the need for defining methods on the serializer, and make these attributes seem more like transformations.  I don't mind that they require a call to object, but would be okay eval'ing them in the context of the object, instead of the serializer.

- [x] `attribute :name do 'AMS' end`

### Blocks in serializer association definitions

Since there may be more options to pass in here, and to distinguish the context of an association block, which is on the association, from the attribute block, which is on the serializer, I'm tossing around the idea of having a `scope` option the receives a proc or lambda.  e.g. 

- [x]  `has_many :comments, ->{ last(5) }` (NOT `->{ comments.last(5) }`
  - [ ] consider `has_many :comments, scope: ->{ last(5) }`

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -46,10 +46,8 @@ module ActiveModel
 
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
-      class_attribute :serialized_attributes, instance_writer: false # @api public: maps attribute name to 'Attribute' function
-      self.serialized_attributes ||= {}
-      class_attribute :_attributes_keys          # @api private : maps attribute value to explict key name, @see Serializer#attribute
-      self._attributes_keys ||= {}
+      class_attribute :_attributes_map           # @api private : maps attribute key names to names to names of implementing methods, @see Serializer#attribute
+      self._attributes_map ||= {}
       class_attribute :_links                    # @api private : links definitions, @see Serializer#link
       self._links ||= {}
 
@@ -73,8 +71,7 @@ module ActiveModel
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
       caller_line = caller.first
-      base.serialized_attributes = serialized_attributes.dup
-      base._attributes_keys = _attributes_keys.dup
+      base._attributes_map = _attributes_map.dup
       base._links = _links.dup
       base._cache_digest = digest_caller_file(caller_line)
       super
@@ -89,10 +86,6 @@ module ActiveModel
 
     def self.link(name, value = nil, &block)
       _links[name] = block || value
-    end
-
-    def self._attributes
-      serialized_attributes.keys
     end
 
     # @example
@@ -121,19 +114,33 @@ module ActiveModel
     #     end
     def self.attribute(attr, options = {}, &block)
       key = options.fetch(:key, attr)
-      _attributes_keys[attr] = { key: key } if key != attr
+      reader = if block
+                 ->(instance) { instance.instance_eval(&block) }
+               else
+                 ->(instance) { instance.send(attr) }
+               end
 
-      if block_given?
-        serialized_attributes[key] = ->(instance) { instance.instance_eval(&block) }
-      else
-        serialized_attributes[key] = ->(instance) { instance.object.read_attribute_for_serialization(attr) }
-      end
+      _attributes_map[key] = { attr: attr, reader: reader }
 
       ActiveModelSerializers.silence_warnings do
-        define_method key do
-          serialized_attributes[key].call(self)
-        end unless method_defined?(key) || _fragmented.respond_to?(attr)
+        define_method attr do
+          object.read_attribute_for_serialization(attr)
+        end unless method_defined?(attr) || _fragmented.respond_to?(attr)
       end
+    end
+
+    # @api private
+    # An accessor for the old _attributes internal API
+    def self._attributes
+      _attributes_map.keys
+    end
+
+    # @api private
+    # An accessor for the old _attributes_keys internal API
+    def self._attributes_keys
+      _attributes_map
+        .select { |key, details| key != details[:attr] }
+        .each_with_object({}) { |(key, details), acc| acc[details[:attr]] = { key: key } }
     end
 
     # @api private
@@ -261,14 +268,14 @@ module ActiveModel
     # Return the +attributes+ of +object+ as presented
     # by the serializer.
     def attributes(requested_attrs = nil)
-      self.class._attributes.each_with_object({}) do |name, hash|
-        next unless requested_attrs.nil? || requested_attrs.include?(name)
-        hash[name] = read_attribute_for_serialization(name)
+      self.class._attributes_map.each_with_object({}) do |(key, details), hash|
+        next unless requested_attrs.nil? || requested_attrs.include?(key)
+        if self.class._fragmented
+          hash[key] = self.class._fragmented.public_send(details[:attr])
+        else
+          hash[key] = details[:reader].call(self)
+        end
       end
-    end
-
-    def read_attribute_for_serialization(key)
-      self.class._fragmented ? self.class._fragmented.public_send(key) : send(key)
     end
 
     # @api private
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -69,7 +69,7 @@ module ActiveModel
       serializer.class_attribute :_cache_digest # @api private : Generated
     end
 
-    # Serializers inherit serialized_attributes and _attributes_keys.
+    # Serializers inherit serialized_attributes, _attributes_keys, and _reflections.
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
       caller_line = caller.first
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -46,8 +46,8 @@ module ActiveModel
 
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
-      class_attribute :_attributes               # @api private : names of attribute methods, @see Serializer#attribute
-      self._attributes ||= []
+      class_attribute :serialized_attributes, instance_writer: false # @api public: maps attribute name to 'Attribute' function
+      self.serialized_attributes ||= {}
       class_attribute :_attributes_keys          # @api private : maps attribute value to explict key name, @see Serializer#attribute
       self._attributes_keys ||= {}
       class_attribute :_links                    # @api private : links definitions, @see Serializer#link
@@ -69,11 +69,11 @@ module ActiveModel
       serializer.class_attribute :_cache_digest # @api private : Generated
     end
 
-    # Serializers inherit _attributes and _attributes_keys.
+    # Serializers inherit serialized_attributes and _attributes_keys.
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
       caller_line = caller.first
-      base._attributes = _attributes.dup
+      base.serialized_attributes = serialized_attributes.dup
       base._attributes_keys = _attributes_keys.dup
       base._links = _links.dup
       base._cache_digest = digest_caller_file(caller_line)
@@ -91,6 +91,10 @@ module ActiveModel
       _links[name] = block || value
     end
 
+    def self._attributes
+      serialized_attributes.keys
+    end
+
     # @example
     #   class AdminAuthorSerializer < ActiveModel::Serializer
     #     attributes :id, :name, :recent_edits
@@ -102,6 +106,7 @@ module ActiveModel
       end
     end
 
+    # TODO: remove the dynamic method definition
     # @example
     #   class AdminAuthorSerializer < ActiveModel::Serializer
     #     attributes :id, :recent_edits
@@ -109,15 +114,17 @@ module ActiveModel
     #
     #     def recent_edits
     #       object.edits.last(5)
-    #     enr
+    #     end
     def self.attribute(attr, options = {})
       key = options.fetch(:key, attr)
       _attributes_keys[attr] = { key: key } if key != attr
       _attributes << key unless _attributes.include?(key)
 
+      serialized_attributes[key] = ->(object) { object.read_attribute_for_serialization(attr) }
+
       ActiveModelSerializers.silence_warnings do
         define_method key do
-          object.read_attribute_for_serialization(attr)
+          serialized_attributes[key].call(object)
         end unless method_defined?(key) || _fragmented.respond_to?(attr)
       end
     end
@@ -249,12 +256,12 @@ module ActiveModel
     def attributes(requested_attrs = nil)
       self.class._attributes.each_with_object({}) do |name, hash|
         next unless requested_attrs.nil? || requested_attrs.include?(name)
-        if self.class._fragmented
-          hash[name] = self.class._fragmented.public_send(name)
-        else
-          hash[name] = send(name)
-        end
+        hash[name] = read_attribute_for_serialization(name)
       end
+    end
+
+    def read_attribute_for_serialization(key)
+      self.class._fragmented ? self.class._fragmented.public_send(key) : send(key)
     end
 
     # @api private
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -112,19 +112,26 @@ module ActiveModel
     #     attributes :id, :recent_edits
     #     attribute :name, key: :title
     #
+    #     attribute :full_name do
+    #       "#{object.first_name} #{object.last_name}"
+    #     end
+    #
     #     def recent_edits
     #       object.edits.last(5)
     #     end
-    def self.attribute(attr, options = {})
+    def self.attribute(attr, options = {}, &block)
       key = options.fetch(:key, attr)
       _attributes_keys[attr] = { key: key } if key != attr
-      _attributes << key unless _attributes.include?(key)
 
-      serialized_attributes[key] = ->(object) { object.read_attribute_for_serialization(attr) }
+      if block_given?
+        serialized_attributes[key] = ->(instance) { instance.instance_eval(&block) }
+      else
+        serialized_attributes[key] = ->(instance) { instance.object.read_attribute_for_serialization(attr) }
+      end
 
       ActiveModelSerializers.silence_warnings do
         define_method key do
-          serialized_attributes[key].call(object)
+          serialized_attributes[key].call(self)
         end unless method_defined?(key) || _fragmented.respond_to?(attr)
       end
     end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -184,6 +184,15 @@ module ActiveModel
       end
     end
 
+    def self._serializer_instance_method_defined?(name)
+      _serializer_instance_methods.include?(name)
+    end
+
+    def self._serializer_instance_methods
+      @_serializer_instance_methods ||= (public_instance_methods - Object.public_instance_methods).to_set
+    end
+    private_class_method :_serializer_instance_methods
+
     attr_accessor :object, :root, :scope
 
     # `scope_name` is set as :current_user by default in the controller.
@@ -209,7 +218,7 @@ module ActiveModel
     end
 
     def read_attribute_for_serialization(attr)
-      if _serializer_method_defined?(attr)
+      if self.class._serializer_instance_method_defined?(attr)
         send(attr)
       elsif self.class._fragmented
         self.class._fragmented.read_attribute_for_serialization(attr)
@@ -227,15 +236,5 @@ module ActiveModel
     protected
 
     attr_accessor :instance_options
-
-    private
-
-    def _serializer_instance_methods
-      @_serializer_instance_methods ||= (public_methods - Object.public_instance_methods).to_set
-    end
-
-    def _serializer_method_defined?(name)
-      _serializer_instance_methods.include?(name)
-    end
   end
 end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -14,6 +14,9 @@ module ActiveModel
     include Configuration
     include Associations
     require 'active_model/serializer/adapter'
+    Attribute = Struct.new(:name, :reader) do
+      delegate :call, to: :reader
+    end
 
     # Matches
     #  "c:/git/emberjs/ember-crm-backend/app/serializers/lead_serializer.rb:1:in `<top (required)>'"
@@ -46,8 +49,8 @@ module ActiveModel
 
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
-      class_attribute :_attributes_map           # @api private : maps attribute key names to names to names of implementing methods, @see Serializer#attribute
-      self._attributes_map ||= {}
+      class_attribute :_attribute_mappings # @api private : maps attribute key names to names to names of implementing methods, @see Serializer#attribute
+      self._attribute_mappings ||= {}
       class_attribute :_links                    # @api private : links definitions, @see Serializer#link
       self._links ||= {}
 
@@ -71,7 +74,7 @@ module ActiveModel
     # Generates a unique digest for each serializer at load.
     def self.inherited(base)
       caller_line = caller.first
-      base._attributes_map = _attributes_map.dup
+      base._attribute_mappings = _attribute_mappings.dup
       base._links = _links.dup
       base._cache_digest = digest_caller_file(caller_line)
       super
@@ -120,7 +123,7 @@ module ActiveModel
                  ->(instance) { instance.send(attr) }
                end
 
-      _attributes_map[key] = { attr: attr, reader: reader }
+      _attribute_mappings[key] = Attribute.new(attr, reader)
 
       ActiveModelSerializers.silence_warnings do
         define_method attr do
@@ -130,17 +133,22 @@ module ActiveModel
     end
 
     # @api private
-    # An accessor for the old _attributes internal API
+    # names of attribute methods
+    # @see Serializer::attribute
     def self._attributes
-      _attributes_map.keys
+      _attribute_mappings.keys
     end
 
     # @api private
-    # An accessor for the old _attributes_keys internal API
+    # maps attribute value to explict key name
+    # @see Serializer::attribute
+    # @see Adapter::FragmentCache#fragment_serializer
     def self._attributes_keys
-      _attributes_map
-        .select { |key, details| key != details[:attr] }
-        .each_with_object({}) { |(key, details), acc| acc[details[:attr]] = { key: key } }
+      _attribute_mappings
+        .each_with_object({}) do |(key, attribute_mapping), hash|
+          next if key == attribute_mapping.name
+          hash[attribute_mapping.name] = { key: key }
+        end
     end
 
     # @api private
@@ -268,12 +276,12 @@ module ActiveModel
     # Return the +attributes+ of +object+ as presented
     # by the serializer.
     def attributes(requested_attrs = nil)
-      self.class._attributes_map.each_with_object({}) do |(key, details), hash|
+      self.class._attribute_mappings.each_with_object({}) do |(key, attribute_mapping), hash|
         next unless requested_attrs.nil? || requested_attrs.include?(key)
         if self.class._fragmented
-          hash[key] = self.class._fragmented.public_send(details[:attr])
+          hash[key] = self.class._fragmented.public_send(attribute_mapping.name)
         else
-          hash[key] = details[:reader].call(self)
+          hash[key] = attribute_mapping.call(self)
         end
       end
     end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -119,17 +119,13 @@ module ActiveModel
       key = options.fetch(:key, attr)
       reader = if block
                  ->(instance) { instance.instance_eval(&block) }
+               elsif _fragmented
+                 ->(instance) { instance.class._fragmented.read_attribute_for_serialization(attr) }
                else
-                 ->(instance) { instance.send(attr) }
+                 ->(instance) { instance.read_attribute_for_serialization(attr) }
                end
 
       _attribute_mappings[key] = Attribute.new(attr, reader)
-
-      ActiveModelSerializers.silence_warnings do
-        define_method attr do
-          object.read_attribute_for_serialization(attr)
-        end unless method_defined?(attr) || _fragmented.respond_to?(attr)
-      end
     end
 
     # @api private
@@ -278,11 +274,15 @@ module ActiveModel
     def attributes(requested_attrs = nil)
       self.class._attribute_mappings.each_with_object({}) do |(key, attribute_mapping), hash|
         next unless requested_attrs.nil? || requested_attrs.include?(key)
-        if self.class._fragmented
-          hash[key] = self.class._fragmented.public_send(attribute_mapping.name)
-        else
-          hash[key] = attribute_mapping.call(self)
-        end
+        hash[key] = attribute_mapping.call(self)
+      end
+    end
+
+    def read_attribute_for_serialization(attr)
+      if _serializer_method_defined?(attr)
+        send(attr)
+      else
+        object.read_attribute_for_serialization(attr)
       end
     end
 
@@ -295,5 +295,15 @@ module ActiveModel
     protected
 
     attr_accessor :instance_options
+
+    private
+
+    def _serializer_instance_methods
+      @_serializer_instance_methods ||= (public_methods - Object.public_instance_methods).to_set
+    end
+
+    def _serializer_method_defined?(name)
+      _serializer_instance_methods.include?(name)
+    end
   end
 end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -3,6 +3,7 @@ require 'active_model/serializer/collection_serializer'
 require 'active_model/serializer/array_serializer'
 require 'active_model/serializer/include_tree'
 require 'active_model/serializer/associations'
+require 'active_model/serializer/attributes'
 require 'active_model/serializer/configuration'
 require 'active_model/serializer/fieldset'
 require 'active_model/serializer/lint'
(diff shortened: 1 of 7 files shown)


lib/active_model/serializer/associations.rb
```diff
@@ -73,12 +76,19 @@ def has_one(name, options = {})
...
-            object.send reflection.name
-          end unless method_defined?(reflection.name)
+          reflection_name = reflection.name
+          if block
+            serialized_associations[reflection_name] = ->(instance) { instance.instance_eval(&block) }
+          else
+            serialized_associations[reflection_name] = ->(instance) { instance.object.send(reflection_name) }
+          end
+
+          define_method reflection_name do
```
beauby: Is there still a need/reason to define such methods on the serializer?

bf4: No, except that it breaks some tests, and I didn't want to go and fix 'em :)

bf4: So, we can do it in two steps

beauby: Yeah that's what was holding me back, because even after fixing the tests, there are caching issues (I believe caching is broken tbh).

lib/active_model/serializer.rb
```diff
@@ -159,64 +175,66 @@ def self.cache(options = {})
...
 
-        if serializer_class
-          serializer_class
-        elsif klass.superclass
-          get_serializer_for(klass.superclass)
-        end
+      if block_given?
+        serialized_attributes[key] = Attribute.new(->(instance) { instance.instance_eval(&block) })
+      else
+        serialized_attributes[key] = Attribute.new(->(instance) { instance.object.read_attribute_for_serialization(attr) })
```
beauby: Why encapsulate the lambda in an `Attribute` object? (Especially considering we're not doing the same for associations)

lib/active_model/serializer.rb
```diff
@@ -159,64 +175,66 @@ def self.cache(options = {})
...
+      end
+
+      ActiveModelSerializers.silence_warnings do
+        define_method key do
+          serialized_attributes[key].call(self)
+        end unless method_defined?(key) || _fragmented.respond_to?(attr)
```
beauby: This line, although not introduced in this commit, scares the sh*t out of me. It looks inconsistent (`key` vs `attr`) and removing it does not break any test.

lib/active_model/serializer/associations.rb
bf4: oh, and re having an Attributes file.. I'm pro moving more stuff in there (though I think it should be a class), but even the current Struct isn't really necessary now.

lib/active_model/serializer.rb
```diff
@@ -159,64 +175,66 @@ def self.cache(options = {})
...
+    end
 
-      chain
+    # @example
+    #   class AdminAuthorSerializer < ActiveModel::Serializer
+    #     attributes :id, :name, :recent_edits
+    def self.attributes(*attrs)
+      attrs = attrs.first if attrs.first.class == Array
+
+      attrs.each do |attr|
```
beauby: Wouldn't something like `attrs.each(&:attribute)` work here?

lib/active_model/serializer.rb
```diff
@@ -69,56 +43,98 @@ def self.digest_caller_file(caller_line)
...
       base._cache_digest = digest_caller_file(caller_line)
       super
     end
 
-    # @example
-    #   class AdminAuthorSerializer < ActiveModel::Serializer
-    #     type 'authors'
-    def self.type(type)
-      self._type = type
+    # Matches
```
beauby: As we're already moving things around in this PR, shouldn't we use the opportunity to extract all caching-related methods/constants into its own module? It would make this file less confusing IMHO.

lib/active_model/serializer.rb
bf4: no reason right now. I was thinking I'd do more with it, but I didn't.  So, can be removed.

lib/active_model/serializer.rb
bf4: :+1:

lib/active_model/serializer.rb
beauby: Yeah, I think we talked about it and I agree that at some point it would be nice, but at the moment I think it just adds complexity.

lib/active_model/serializer.rb
bf4: yeah, that's related to the logic in `#attributes` and `#read_attribute_for_serialization`.  All it means is that if we use the fragment cache, it should be able to cache this attribute, otherwise, we need to be able to get its value.... not terribly complex, just crazy looking as it's spread about 3 places

beauby: Yeah but in which possible situation would the method `attr` be defined on the cached serializer, when `attr != key`?

lib/active_model/serializer.rb
bf4: I thought I did, but maybe that was adapters.  I'm thinking I might back out the re-arranging to improve the focus in this PR

The rearranging is about moving the generic class functions away from the attribute/association mapping stuff but is a little distracting in the diff

lib/active_model/serializer.rb
```diff
@@ -14,40 +14,14 @@ class Serializer
...
+      delegate :call, to: :callable
     end
 
     with_options instance_writer: false, instance_reader: false do |serializer|
       class_attribute :_type, instance_reader: true
-      class_attribute :_attributes               # @api private : names of attribute methods, @see Serializer#attribute
-      self._attributes ||= []
+      class_attribute :serialized_attributes, instance_writer: false # @api public: maps attribute name to 'Attribute' instance
+      self.serialized_attributes ||= {}
       class_attribute :_attributes_keys          # @api private : maps attribute value to explict key name, @see Serializer#attribute
```
bf4: I left attributes_keys alone since it's a little funky

lib/active_model/serializer/associations.rb
beauby: The idea behind `serializer/attributes.rb` was mainly to separate various serializer concerns into their own modules/files. The goal was that `serializer.rb` would contain as little stuff as possible and that the logic for handling attributes-related stuff would be in one file (class and instance methods).

lib/active_model/serializer.rb
beauby: Yeah, it is used for caching (although I think it could/should be avoided, but let's worry about that in an other PR).

lib/active_model/serializer.rb
beauby: Yeah, that's why I initially extracted the attributes stuff into `serializer/attributes.rb` (cleaner diff + cleaner files).

lib/active_model/serializer/associations.rb
bf4: Yeah, I started on it, but it wasn't worth the extra effort in this PR

bf4:
rebased and force pushed. I don't think I broke anything...

noahsilas:
It's not clear to me that this PR satisfies goal 1 (Allow defining attributes so that they don't conflict with existing methods). It looks like trying to add a key whose name is already a method name on the serializer will add an entry to serialized_attributes, but that it still delegates to the existing method. I tried to get around this in #1354 by avoiding the usage of the key name as the method name, but it seems like there is room for improvement over that here in evaluating the proc directly instead of (or in addition to?) adding methods to the serializer.

test/serializers/attribute_test.rb
```diff
@@ -71,6 +71,21 @@ def id
 
         assert_equal('custom', hash[:blog][:id])
       end
+
+      PostWithVirtualAttribute = Class.new(::Model)
+      class PostWithVirtualAttributeSerializer < ActiveModel::Serializer
+        attribute :name do
+          "#{object.first_name} #{object.last_name}"
+        end
+      end
```
bf4: @noahsilas I do want to remove the dynamic methods on the serializer, just not in this PR.

However, I do think it's a good idea for there to be a test against attribute names/keys that conflict with methods on the serializer.  I think the `read_attribute_for_serialization` should get around that

bf4: I mean, right now it is `self.class._fragmented ? self.class._fragmented.public_send(key) : send(key)` but it could me `self.class._fragmented ? self.class._fragmented.public_send(key) : object.read_attribute_for_serialization(key)`

noahsilas: I think that the problem here is that the key may not be the name of the attribute that we want to read, as in:

```
class MyThingSerializer < ActiveModel::Serializer
  attribute :name, key: :object
end
```

This should render something like `{ "object": "my name" }`, but to get that, we actually need to ~~invoke the `name` method on the object~~ read the `name` attribute for serialization.

noahsilas:
If you want, I can rebuild #1354 on top of this change and submit a new PR, separating the goal of block definition from the goal of avoiding namespace conflicts.

bf4:
You can re-use this pr

Take it away (please)!

B mobile phone

noahsilas:
I'm not sure that I can push onto this PR, but rebasing #1354 onto this branch ends up looking like this: https://github.com/brigade/active_model_serializers/commits/attribute_objects (specifically, see brigade/active_model_serializers@ca76be1a46 ).

There is still some weird interplay with defining methods on the serializer so that the old style method overrides still work, which is a little disappointing, but a major break in the API, so I didn't feel comfortable removing it.

bf4:
@noahsilas re: 'you can re-use this pr', my mistake. I was replying from email and thought it was the one you authored.

I was thinking that a way to get rid of the dynamic methods would be to pretend they're there. Maybe with a respond_to_missing, or in the 'reader'

bf4:
@noahsilas I added your commit from that branch and made one major and a few minor changes

1. removed dynamically define methods (yay!)
  - added the `_fragment` attributes to the mapping. This handles the `_fragmented.respond_to?(attr)` the prevented the dynamic method from being defined
  - added `Serializer#read_attribute_for_serialization` (part of the ActiveModel::Serialization API) to handle first getting the attribute from a serializer method, then from the object.
    - Checking if the attribute is defined on the serializer handles the dynamic method check for `method_defined?(attr)`, except now it just checks if there's a public instance method by that name not defined in Object
  - **possible regression**: removed referring to object attributes in the serializer without calling them directly on object.  (I didn't even know this was possible. We could add a respond_to_missing / method_missing for this)
  - 
1. changed `_attributes_map` to `_attribute_mappings` so that the collection element could be referred to as each of an `_attribute_mapping`.
  - Changed the mapping from a hash to an object
  ```diff
  -      _attributes_map[key] = { attr: attr, reader: reader }
  +      _attribute_mappings[key] = Attribute.new(attr, reader)
  ```
  - Defined the object as a callable that knows the attribute `name` it represents
  ```diff
  +    Attribute = Struct.new(:name, :reader) do
  +      delegate :call, to: :reader
  +    end
  ```

lib/active_model/serializer.rb
```diff
@@ -14,6 +14,9 @@ class Serializer
     include Configuration
     include Associations
     require 'active_model/serializer/adapter'
+    Attribute = Struct.new(:name, :reader) do
+      delegate :call, to: :reader
+    end
```
bf4: See https://github.com/rails-api/active_model_serializers/pull/1356#issuecomment-161193512 

I could go either way on this.

lib/active_model/serializer.rb
```diff
@@ -69,12 +70,11 @@ def self.digest_caller_file(caller_line)
       serializer.class_attribute :_cache_digest # @api private : Generated
     end
 
-    # Serializers inherit _attributes and _attributes_keys.
+    # Serializers inherit serialized_attributes, _attributes_keys, and _reflections.
```
bf4: typo: `s/serialized_attributes/_attributes_mappings`

typo: need to remove `_attributes_keys`

lib/active_model/serializer.rb
```diff
@@ -102,24 +102,49 @@ def self.attributes(*attrs)
       end
     end
 
+    # TODO: remove the dynamic method definition
```
bf4: typo: remove TODO

lib/active_model/serializer.rb
```diff
@@ -102,24 +102,49 @@ def self.attributes(*attrs)
...
       key = options.fetch(:key, attr)
-      _attributes_keys[attr] = { key: key } if key != attr
-      _attributes << key unless _attributes.include?(key)
+      reader = if block
+                 ->(instance) { instance.instance_eval(&block) }
+               elsif _fragmented
+                 ->(instance) { instance.class._fragmented.read_attribute_for_serialization(attr) }
+               else
+                 ->(instance) { instance.read_attribute_for_serialization(attr) }
+               end
```
bf4: I'm not a fan of this style of if/else/end assignment. I think I'd rather a case statement. @noahsilas ?

There's a possible order-of-evaluation failure here if _fragmented isn't set before the attribute is defined. We'd have to move the logic into either `#attributes` or `#read_attribute_for_serialization` to make it class function order independent, but I think that just muddies up the code a bit

lib/active_model/serializer.rb
```diff
@@ -102,24 +102,49 @@ def self.attributes(*attrs)
...
 
-      ActiveModelSerializers.silence_warnings do
-        define_method key do
-          object.read_attribute_for_serialization(attr)
-        end unless method_defined?(key) || _fragmented.respond_to?(attr)
-      end
+      _attribute_mappings[key] = Attribute.new(attr, reader)
```
bf4: Even though it's called `attr` here, I called it `name` in `Attribute` since `attribute.name` sounds better than `attribute.attr`

lib/active_model/serializer.rb
```diff
@@ -102,24 +102,49 @@ def self.attributes(*attrs)
...
+    end
+
+    # @api private
+    # names of attribute methods
+    # @see Serializer::attribute
+    def self._attributes
```
bf4: I sorta think this and the other `_attribute*` methods should be public, and that the underscore just means that the Serializer owns it.. it could be useful for testing, but keeping it unchanged till we need it public (disclaimer: I've used it for testing)

lib/active_model/serializer.rb
```diff
@@ -102,24 +102,49 @@ def self.attributes(*attrs)
...
+      _attribute_mappings.keys
+    end
+
+    # @api private
+    # maps attribute value to explict key name
+    # @see Serializer::attribute
+    # @see Adapter::FragmentCache#fragment_serializer
```
bf4: This is still used, but maybe it doesn't need to be.  We'd have to look at the FragmentCache

lib/active_model/serializer.rb
```diff
@@ -247,13 +272,17 @@ def json_key
...
     # by the serializer.
     def attributes(requested_attrs = nil)
-      self.class._attributes.each_with_object({}) do |name, hash|
-        next unless requested_attrs.nil? || requested_attrs.include?(name)
-        if self.class._fragmented
-          hash[name] = self.class._fragmented.public_send(name)
-        else
-          hash[name] = send(name)
-        end
+      self.class._attribute_mappings.each_with_object({}) do |(key, attribute_mapping), hash|
```
bf4: This should probably be memoized. There's no reason I can think of that it would change, right?

lib/active_model/serializer.rb
```diff
@@ -247,13 +272,17 @@ def json_key
...
+        next unless requested_attrs.nil? || requested_attrs.include?(key)
+        hash[key] = attribute_mapping.call(self)
+      end
+    end
+
+    def read_attribute_for_serialization(attr)
+      if _serializer_method_defined?(attr)
```
bf4: see https://github.com/rails-api/active_model_serializers/pull/1356#issuecomment-161193512

lib/active_model/serializer/associations.rb
```diff
@@ -13,9 +13,10 @@ module Associations
       DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
 
       included do |base|
-        class << base
-          attr_accessor :_reflections
-        end
+        base.class_attribute :serialized_associations, instance_writer: false # @api public: maps association name to 'Reflection' instance
```
bf4: should probably follow the same convention in the Serializer and be `_association_mappings`

lib/active_model/serializer/associations.rb
```diff
@@ -73,12 +76,19 @@ def has_one(name, options = {})
...
+          reflection_name = reflection.name
+          if block
+            serialized_associations[reflection_name] = ->(instance) { instance.instance_eval(&block) }
+          else
+            serialized_associations[reflection_name] = ->(instance) { instance.object.send(reflection_name) }
+          end
+
+          define_method reflection_name do
+            serialized_associations[reflection_name].call(self)
+          end unless method_defined?(reflection_name)
```
bf4: TODO: [ ] remove dynamic method definition from associations.  (I sorta wish this weren't a module mixed into Serializer, but oh well)

test/fixtures/poro.rb
```diff
@@ -116,7 +116,7 @@ def custom_options
   attributes :id, :name, :description, :slug
 
   def slug
-    "#{name}-#{id}"
+    "#{object.name}-#{object.id}"
```
bf4: Possible regression, see https://github.com/rails-api/active_model_serializers/pull/1356#issuecomment-161193512

test/serializers/associations_test.rb
```diff
@@ -126,6 +126,35 @@ def test_associations_custom_keys
         assert expected_association_keys.include? :site
       end
 
+      class InlineAssociationTestPostSerializer < ActiveModel::Serializer
+        has_many :comments
+        has_many :last_comments do
+          object.comments.last(1)
```
bf4: This is a great example @beauby made of why you'd want to define your own association value

However, the way that this works, it's really the same as `attribute :last_comments do end`, right?  Unless this affects deserialization, right?

test/serializers/attribute_test.rb
```diff
@@ -43,6 +43,15 @@ def test_id_attribute_override
         assert_equal({ blog: { id: 'AMS Hints' } }, adapter.serializable_hash)
       end
 
+      def test_object_attribute_override
+        serializer = Class.new(ActiveModel::Serializer) do
+          attribute :name, key: :object
+        end
```
bf4: Great test @noahsilas wrote

lib/active_model/serializer.rb
noahsilas: Yeah, this starts getting rather ugly with if/elsif/else branches, a case seems like it could be a nice cleanup.

On the other hand, it does seem like pushing this switch based on the `_fragmented` status into `#read_attribute_for_serialization` would help keep that API clean, in that consumers who wanted to get an attribute wouldn't need to know anything about the fragment cache state.

groyoh:
@bf4 I see that there are already some work going on for relationship links (#1282) and I will like to give some feedback on this. I also started working on an implementation for this where the links could be defined as follows:

```ruby
has_many :comments do
  link :self do
    href "http://www.example.com/#{object.id}"
    meta some_meta: :a_meta_value
  end
end
```

This follows the design initiated by #1246 and could be extended like so:

```ruby
has_many :comments do
  link :self do
    href "http://www.example.com/#{object.id}"
    meta some_meta: :a_meta_value
  end
  meta count: object.post_comments.count
  data object.post_comments
end
```

or something similar. 

I would really like to have your opinion on this since I find it more flexible and extensible (even though it's more verbose) and might somehow contradicts some changes in this PR.

beauby:
@groyoh :+1: on the relationship-links. I believe this is the way to go.

beauby:
Only drawback is when you simply want to override the relationship like
```ruby
has_many :comments do
  object.comments.last(10)
end
```
which would become
```ruby
has_many :comments do
  data object.comments.last(10)
end
```
but it seems acceptable, especially considering that it solves the "links only relationships" case quite nicely:
```ruby
has_many :comments do
  link :self do ... end
  data :SpecialConstantMeaningNoData
end
```

groyoh:
@beauby I think there are way more advantages than drawbacks here and adding 4 letters shouldn't be too much to ask :grin: .
This solution also allows us to specify only cetain keys (e.g. render `links` but not `data` or display `data` conditionnaly):

```ruby
has_many :comments do
  link :self do ... end
end
# or
has_many :comments do
  link :self do ... end
  data :SpecialConstantMeaningNoData if something
end
```

lib/active_model/serializer/associations.rb
```diff
@@ -12,9 +12,10 @@ module Associations
 
       DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
 
-      included do |base|
-        class << base
-          attr_accessor :_reflections
+      included do
+        with_options instance_writer: false, instance_reader: true do |serializer|
```
bf4: I don't know if there's a difference between the block variables `base` and `serializer` in what we want here.

lib/active_model/serializer.rb
```diff
@@ -45,14 +47,9 @@ def self.digest_caller_file(caller_line)
     end
 
     with_options instance_writer: false, instance_reader: false do |serializer|
-      class_attribute :_type, instance_reader: true
-      class_attribute :_attributes               # @api private : names of attribute methods, @see Serializer#attribute
```
bf4: not sure why these weren't `serializer.class_attribute`

lib/active_model/serializer.rb
```diff
@@ -244,16 +208,13 @@ def json_key
...
-    # by the serializer.
-    def attributes(requested_attrs = nil)
-      self.class._attributes.each_with_object({}) do |name, hash|
-        next unless requested_attrs.nil? || requested_attrs.include?(name)
-        if self.class._fragmented
-          hash[name] = self.class._fragmented.public_send(name)
-        else
-          hash[name] = send(name)
-        end
+    def read_attribute_for_serialization(attr)
```
bf4: I think this reads pretty well and doesn't need documentation

lib/active_model/serializer/reflection.rb
```diff
@@ -17,7 +17,28 @@ class Serializer
...
+      def initialize(*)
+        super
+        @reader = self.class.build_reader(name, block)
+      end
+
+      def value(instance)
+        call(instance)
+      end
+
+      def self.build_reader(name, block)
```
bf4: like AttributeBlock and AttributeReader, but needs to be in the Reflection rather than Association due to how the code is currently written, so I just build the reader in the initialize rather than use polymorphism

bf4:
@groyoh @beauby Updated removing all dynamic methods on attributes and associations with full backwards compatibility

Anything you think is needed in this PR to support future work?

noahsilas:
@groyoh @beauby The relationship links work seems like it is primarily applicable to the JsonApi adapter; is there a way to satisfy this desire that doesn't introduce overhead for all other adapters? (I may be misunderstanding, but having to specify the relationship via a `data` method seems confusing for the AttributesAdapter, for instance).

bf4:
@noahsilas 

> The relationship links work seems like it is primarily applicable to the JsonApi adapter; 

True, but it's not in this PR :)

lib/active_model/serializer.rb
```diff
@@ -266,5 +227,15 @@ def links
     protected
 
     attr_accessor :instance_options
+
+    private
+
+    def _serializer_instance_methods
```
groyoh: Wouldn't that be better as a class method performance wise?

bf4: I think you're right.  I was working on the instance-level and got tunnel-vision

lib/active_model/serializer/reflection.rb
```diff
@@ -17,7 +17,28 @@ class Serializer
...
+        if block
+          ->(instance) { instance.instance_eval(&block) }
```
bf4: I'm looking at 0.8 and how JSON API considers associations 'includes' and am thinking that association block should be applied to the association, rather than the serializer.  Otherwise, there's really nothing that ties the block to an association.

So, usage from current impl would be

```diff
has_many :comments do
+  comments.limit(5)
-  limit(5)
end
```

which is also more inline with how the `scope` method works in Rails.  I think this makes the api a lot more clear.  If you want a custom attribute, use the attribute block. If you want a scoped association, I think it the instance_eval should be on the association.

See 

- [0.8](https://github.com/rails-api/active_model_serializers/blob/0-8-stable/lib/active_model/serializer.rb#L115-L142)

```ruby
      def associate(klass, attrs) #:nodoc:
        options = attrs.extract_options!
        self._associations = _associations.dup

        attrs.each do |attr|
          unless method_defined?(attr)
            define_method attr do
              object.send attr
            end
          end

          define_include_method attr

          self._associations[attr] = klass.refine(attr, options)
        end
      end

      def define_include_method(name)
        method = "include_#{name}?".to_sym

        INCLUDE_METHODS[name] = method

        unless method_defined?(method)
          define_method method do
            true
          end
        end
      end
    def include_associations!
      _associations.each_key do |name|
        include!(name) if include?(name)
      end
    end

    def include?(name)
      return false if @options.key?(:only) && !Array(@options[:only]).include?(name)
      return false if @options.key?(:except) && Array(@options[:except]).include?(name)
      send INCLUDE_METHODS[name]
    end

    def include!(name, options={})
```

- [0.9](https://github.com/rails-api/active_model_serializers/blob/0-9-stable/lib/active_model/serializer.rb#L26-L27)

```ruby
      EMBED_IN_ROOT_OPTIONS = [
        :include,
      ]
    def associations(options={})
      associations = self.class._associations
      included_associations = filter(associations.keys)
      associations.each_with_object({}) do |(name, association), hash|
        if included_associations.include? name
        # ...
  end
    def filter(keys)
      if @only
        keys & @only
      elsif @except
        keys - @except
      else
        keys
      end
    end
```

lib/active_model/serializer/reflection.rb
```diff
@@ -7,8 +7,16 @@ class Serializer
...
     #     has_many :comments
+    #     has_many :comments, key: :last_comments do
+    #       last(1)
+    #     end
     #  end
     #
+    #  Notice that the association block is evaluated in the context of the association.
+    #  Specifically, the association 'comments' is evaluated two different ways:
+    #  1) as 'comments' and named 'comments'.
+    #  2) as 'comments.last(1)' and named 'last_comments'.
```
bf4: See comment https://github.com/rails-api/active_model_serializers/pull/1356/files#r46723777 about how we should think of association blocks (and how we might want to also consider them as only used when included under JSON API) and see https://github.com/rails-api/active_model_serializers/issues/1325#issuecomment-162016685

groyoh:
@noahsilas I'm pretty sure we can work around this issue. For example, making sure that `meta`, and `link` methods return a certain object (e.g. a symbol as in @beauby's example) and serialize the relationship depending on the block return value:
* Use `object.comments`
```ruby
has_many :comments
```
* Use the block return value
```ruby
has_many :comments do
  object.special_comments
end
```
* With link but use `object.comments`
```ruby
has_many :comments do
  link :self do
    #something
  end # => :SpecialSymbolToPreventInlineAssociation
end
```
* With link and use block return value
```ruby
has_many :comments do
  link :related do
    #something
  end
  object.association # or data object.association
end
```

The first two would work pretty well with any adapter while the last one would only fit JSONAPI and everybody is happy :grin: .

Of course, this probably require more thoughts and maybe have some flaws too.

lib/active_model/serializer/associations.rb
beauby: Unclear to me as well.

lib/active_model/serializer.rb
beauby: Is it actually different?

lib/active_model/serializer/reflection.rb
```diff
@@ -17,7 +25,28 @@ class Serializer
     #
     # So you can inspect reflections in your Adapters.
     #
-    Reflection = Struct.new(:name, :options) do
+    Reflection = Struct.new(:name, :options, :block) do
+      delegate :call, to: :reader
```
beauby: Why delegating it rather than calling `@reader.call` in `value`?

lib/active_model/serializer.rb
bf4: I suppose it would be easy enough to find out, but I didn't...

lib/active_model/serializer/associations.rb
```diff
@@ -76,10 +78,6 @@ def has_one(name, options = {})
         def associate(reflection)
           self._reflections = _reflections.dup
```
beauby: I believe this `dup` is useless. There is no test breaking when I remove it, and I can't think of any case where it would be necessary.

lib/active_model/serializer/reflection.rb
bf4: I just think it reads better. It might be a performance issue.

I think there's definitely room for refactoring and revising this, but I don't think it should be a blocker.  I don't really like how the reflection differs from the attribute, but figure we can kick that can down the road a bit more into its own PR.

lib/active_model/serializer/attributes.rb
```diff
@@ -0,0 +1,109 @@
...
+          super(name)
+          @reader = ->(instance) { instance.read_attribute_for_serialization(name) }
+        end
+      end
+      class AttributeBlock < Attribute
+        def initialize(name, block)
+          super(name)
+          @reader = ->(instance) { instance.instance_eval(&block) }
+        end
+      end
```
beauby: I see what you did here but I have a feeling the whole `Attribute` / `AttributeReader` / `AttributeBlock` tryptic is a bit overcomplicated compared to what we actually want to achieve here: storing a proc built from either a block or an attribute name.

lib/active_model/serializer/associations.rb
bf4: Yeah, It doesn't seem to make sense, but it's pre-existing, and there's enough changes in here already

beauby: Agreed.

lib/active_model/serializer/attributes.rb
bf4: I like the polymorphism. Do you mean as opposed to just a conditional statement that returns the appropriate proc?

beauby: I don't have anything against polymorphism in general, but I don't really find it necessary here. This whole part of code could be replaced by a single method, and I am not sure the polymorphism really adds anything here.

bf4: Would you mind showing me what you have in mind? I think this is 90% a style thing, at this point, and I just found this more readable and extendable.  Having the `Attribute` object let me keep track of the name, which I needed to use, which is why I introduced it (as opposed to just wanting it there for a future need)

beauby: Sure. I meant defining a ("private") class method as follows:
```ruby
def _attribute_mapping(name, block)
  if block
    ->(instance) { instance.instance_eval(&block) }
  else
    ->(instance) { instance.read_attribute_for_serialization(name) }
  end
end
```
and the `attribute` method would just be
```ruby
+        def attribute(attr, options = {}, &block)
+          key = options.fetch(:key, attr)
+          _attribute_mappings[key] = _attribute_mapping(attr, block)
+        end
```

bf4: That would be fine by me, except I'd prefer not to add another class method, and without a reference to `name`, I think we'd have to go back to the old way of defining `_attributes_keys`, no?  All of which drive me to find polymorphic classes easier to reason about and use.

```diff
def attribute(name, options = {}, &block)
-      _attributes_keys[attr] = { key: key } if key != attr
end

+        # @api private
+        # maps attribute value to explict key name
+        # @see Serializer::attribute
+        # @see Adapter::FragmentCache#fragment_serializer
+        def _attributes_keys
+          _attribute_mappings
+            .each_with_object({}) do |(key, attribute_mapping), hash|
+              next if key == attribute_mapping.name
+              hash[attribute_mapping.name] = { key: key }
+            end
```

bf4: Do you think it would make sense for you to make a PR into this branch so we can review it that way?  Sounds like we're both ok with this, except a few not-too-major things

lib/active_model/serializer/reflection.rb
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1356/files#r46783854

I'm also not sure about 'value' as a name here... or even `instance` in place of `serializer_instance` or whatever.. just trying to get a feel for which names work better

bf4:
@rails-api/ams Anything blocking merging this? I'll merge in 24 hours unless I hear otherwise, since it seems pretty well-discussed, and we can always make changes later

NullVoxPopuli:
@bf4 I think it'll be a good addition :-)

lib/active_model/serializer/attributes.rb
beauby: Will add a PR to this branch, we'll see how it works.

bf4: Ok, but if you it takes too long, we'll just make that a PR against master, gith?

test/serializers/attribute_test.rb
bf4: I should add a case here using the `virtual_attribute` option.  e.g. the poro in our fixtures file
```ruby
class VirtualValueSerializer < ActiveModel::Serializer
  attributes :id

  has_many :reviews, virtual_value: [{ id: 1 }, { id: 2 }]
end
```

lib/active_model/serializer/attributes.rb
bf4: ref: https://github.com/rails-api/active_model_serializers/blob/d3837f9466f53d0e518611c3e0987ac95da8721f/docs/general/serializers.md

PR #1357, opened 2015-11-30
fix description of USAGE

y-yagi:
The usage It is written as to generate a test, but actually it looks like not generated.


bf4:
Awesome!

y-yagi:
Thanks!

PR #1358, opened 2015-12-01
Match file paths with spaces in caller regexp

rwstauner:
When matching only `\S+` I get warnings like this on my ci server:

```
Cannot digest non-existent file: '/tmp/space char20151201-28026-1sux619/some_ruby.rb20151201-28026-g481xg:1:in `<top (required)>''.
Please set `::_cache_digest` of the serializer
if you'd like to cache it.
```

Matching the trailing space after `:\d+:in` wasn't strictly necessary, but seemed like it would help it to be more accurate.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -23,7 +23,7 @@ module ActiveModel
     #  c/git/emberjs/ember-crm-backend/app/serializers/lead_serializer.rb
     CALLER_FILE = /
       \A       # start of string
-      \S+      # one or more non-spaces
+      .+       # file path (one or more characters)
       (?=      # stop previous match when
         :\d+     # a colon is followed by one or more digits
         :in      # followed by a colon followed by in
--- a/test/serializers/cache_test.rb
+++ b/test/serializers/cache_test.rb
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'tmpdir'
 require 'tempfile'
 module ActiveModel
   class Serializer
@@ -160,9 +161,23 @@ module ActiveModel
         assert_equal caller_line[ActiveModel::Serializer::CALLER_FILE], path
       end
 
+      def test_serializer_file_path_with_space
+        path = '/Users/git/ember js/ember-crm-backend/app/serializers/lead_serializer.rb'
+        caller_line = "#{path}:1:in `<top (required)>'"
+        assert_equal caller_line[ActiveModel::Serializer::CALLER_FILE], path
+      end
+
+      def test_serializer_file_path_with_submatch
+        # The submatch in the path ensures we're using a correctly greedy regexp.
+        path = '/Users/git/ember js/ember:123:in x/app/serializers/lead_serializer.rb'
+        caller_line = "#{path}:1:in `<top (required)>'"
+        assert_equal caller_line[ActiveModel::Serializer::CALLER_FILE], path
+      end
+
       def test_digest_caller_file
         contents = "puts 'AMS rocks'!"
-        file = Tempfile.new('some_ruby.rb')
+        dir = Dir.mktmpdir('space char')
+        file = Tempfile.new('some_ruby.rb', dir)
         file.write(contents)
         path = file.path
         caller_line = "#{path}:1:in `<top (required)>'"
@@ -170,6 +185,7 @@ module ActiveModel
         assert_equal ActiveModel::Serializer.digest_caller_file(caller_line), Digest::MD5.hexdigest(contents)
       ensure
         file.unlink
+        FileUtils.remove_entry dir
       end
 
       def test_warn_on_serializer_not_defined_in_file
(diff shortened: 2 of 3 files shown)


bf4:
duplicate : https://github.com/rails-api/active_model_serializers/pull/1355

rwstauner:
rats.  i searched for the "cache" warning but i guess i didn't search for "spaces".

bf4:
you solved it somewhat differently. would you like to merge in changes from the other (mine), or have me add some changes?  I don't have failing tests for the difference between the two regexen

bf4:
(As a maintainer, I'd rather someone else get credit...)

beauby:
Great, I was hoping someone would fix the spaces situation. Thanks :+1:

rwstauner:
After fiddling with some extra tests I think your pattern is more accurate.

My original thinking was that the string after the file path (an error message) could include a substring match.
But in this case you aren't matching errors, you're matching an explicit caller line, so I assume the ending will only be something like `` `<top (required)>' `` (or `` `<main>' ``),
in which case the greedy regexp is more correct.

An example that passes for non-greedy and fails for greedy:

```ruby
def test_serializer_file_path_with_path_in_string
  # $ ruby =(echo $'def foo(s); raise s; end;\nfoo("file:123:in string")')
  # /tmp/zshr2gfAE:1:in `foo': file:123:in string (RuntimeError)
  #         from /tmp/zshr2gfAE:2:in `<main>'
  path = '/Users/git/ember js/ember-crm-backend/app/serializers/lead_serializer.rb'
  caller_line = "#{path}:1:in `foo': file:123:in string (RuntimeError)"
  assert_equal path, caller_line[ActiveModel::Serializer::CALLER_FILE]
end
```

but as I said, I don't think that's appropriate in this case.

So I adjusted mine, added your test, and added an additional (more convoluted) test to confirm the difference between the two.  Use whichever pull req makes sense.

test/serializers/cache_test.rb
```diff
@@ -160,9 +161,23 @@ def test_serializer_file_path_on_windows
...
 
+      def test_serializer_file_path_with_space
+        path = '/Users/git/ember js/ember-crm-backend/app/serializers/lead_serializer.rb'
+        caller_line = "#{path}:1:in `<top (required)>'"
+        assert_equal caller_line[ActiveModel::Serializer::CALLER_FILE], path
+      end
+
+      def test_serializer_file_path_with_submatch
+        # The submatch in the path ensures we're using a correctly greedy regexp.
+        path = '/Users/git/ember js/ember:123:in x/app/serializers/lead_serializer.rb'
```
bf4: If someone has that in their file path I pity them :smiley_cat:

test/serializers/cache_test.rb
```diff
@@ -160,9 +161,23 @@ def test_serializer_file_path_on_windows
...
+        caller_line = "#{path}:1:in `<top (required)>'"
+        assert_equal caller_line[ActiveModel::Serializer::CALLER_FILE], path
+      end
+
       def test_digest_caller_file
         contents = "puts 'AMS rocks'!"
-        file = Tempfile.new('some_ruby.rb')
+        dir = Dir.mktmpdir('space char')
+        file = Tempfile.new('some_ruby.rb', dir)
```
bf4: good thinking for the integration test here

lib/active_model/serializer.rb
```diff
@@ -23,10 +23,11 @@ class Serializer
     #  c/git/emberjs/ember-crm-backend/app/serializers/lead_serializer.rb
     CALLER_FILE = /
       \A       # start of string
-      \S+      # one or more non-spaces
+      .+       # file path (one or more characters)
       (?=      # stop previous match when
         :\d+     # a colon is followed by one or more digits
         :in      # followed by a colon followed by in
+        \x20     # followed by a space
```
bf4: Did you notice if the extra space in the lookahead made a difference?

bf4:
@rwstauner looks good. Could you add a changelog entry as a Fix?

I have a few other comments but nothing big.

I keep thinking there should be a way for the subclass to calculate its own digest outside of the inherited hook, which would be simpler, no?  I think the method would just need to be dynamically defined `define_singleton_method(:cache_digest) { MD5::Digest.hexdigest(__FILE__) }` or something like that, right?

lib/active_model/serializer.rb
rwstauner: Quite right.  Now that the regexp is greedy that isn't needed.

test/serializers/cache_test.rb
rwstauner: Yes, quite.

In my case the file path was generated by another process with apparently little sanitation on the naming ;-)

rwstauner:
Sounds like a good idea, but we'd have to find another way to get the file from where the class is inheriting (the caller) since the value of `__FILE__` is based on where the token is present.
And `class_eval "..."` isn't any better (`__FILE__` evaluates to `(eval)`).

bf4:
Closes https://github.com/rails-api/active_model_serializers/pull/1355

rwstauner:
Thanks!

PR #1370, opened 2015-12-10
Simplify attributes handling.

beauby:
ref: https://github.com/rails-api/active_model_serializers/pull/1356#discussion_r47145259

Rationale: I believe the level of abstraction introduced here adds more complexity than required.

### DIFF
--- a/lib/active_model/serializer/attribute.rb
+++ b/lib/active_model/serializer/attribute.rb
@@ -1,0 +1,13 @@
+module ActiveModel
+  class Serializer
+    Attribute = Struct.new(:name, :key, :block) do
+      def value(serializer)
+        if block
+          serializer.instance_eval(&block)
+        else
+          serializer.read_attribute_for_serialization(name)
+        end
+      end
+    end
+  end
+end
--- a/lib/active_model/serializer/attribute.rb
+++ b/lib/active_model/serializer/attribute.rb
@@ -1,6 +1,6 @@
 module ActiveModel
   class Serializer
-    Attribute = Struct.new(:name, :key, :block) do
+    Attribute = Struct.new(:name, :block) do
       def value(serializer)
         if block
           serializer.instance_eval(&block)
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -5,28 +5,19 @@ module ActiveModel
 
       included do
         with_options instance_writer: false, instance_reader: false do |serializer|
-          serializer.class_attribute :_attribute_procs # @api private : maps attribute key names to names to names of implementing methods, @see #attribute
-          self._attribute_procs ||= {}
-          serializer.class_attribute :_attribute_keys # @api private : maps attribute names to keys, @see #attribute
-          self._attribute_keys ||= {}
+          serializer.class_attribute :_attributes_data # @api private
+          self._attributes_data ||= {}
         end
+
+        autoload :Attribute
 
         # Return the +attributes+ of +object+ as presented
         # by the serializer.
         def attributes(requested_attrs = nil, reload = false)
           @attributes = nil if reload
-          @attributes ||= self.class._attribute_keys.each_with_object({}) do |(name, key), hash|
-            next unless requested_attrs.nil? || requested_attrs.include?(key)
-            hash[key] = _attribute_value(name)
-          end
-        end
-
-        # @api private
-        def _attribute_value(name)
-          if self.class._attribute_procs[name]
-            instance_eval(&self.class._attribute_procs[name])
-          else
-            read_attribute_for_serialization(name)
+          @attributes ||= self.class._attributes_data.values.each_with_object({}) do |attr, hash|
+            next unless requested_attrs.nil? || requested_attrs.include?(attr.key)
+            hash[attr.key] = attr.value(self)
           end
         end
       end
@@ -34,8 +25,7 @@ module ActiveModel
       module ClassMethods
         def inherited(base)
           super
-          base._attribute_procs = _attribute_procs.dup
-          base._attribute_keys = _attribute_keys.dup
+          base._attributes_data = _attributes_data.dup
         end
 
         # @example
@@ -62,15 +52,15 @@ module ActiveModel
         #       object.edits.last(5)
         #     end
         def attribute(attr, options = {}, &block)
-          _attribute_keys[attr] = options.fetch(:key, attr)
-          _attribute_procs[attr] = block
+          key = options.fetch(:key, attr)
+          _attributes_data[attr] = Attribute.new(attr, key, block)
         end
 
         # @api private
         # keys of attributes
         # @see Serializer::attribute
         def _attributes
-          _attribute_keys.values
+          _attributes_data.values.map(&:key)
         end
 
         # @api private
@@ -78,10 +68,10 @@ module ActiveModel
         # @see Serializer::attribute
         # @see Adapter::FragmentCache#fragment_serializer
         def _attributes_keys
-          _attribute_keys
-            .each_with_object({}) do |(name, key), hash|
-              next if key == name
-              hash[name] = { key: key }
+          _attributes_data.values
+            .each_with_object({}) do |attr, hash|
+              next if attr.key == attr.name
+              hash[attr.name] = { key: attr.key }
             end
         end
       end
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -5,8 +5,8 @@ module ActiveModel
 
       included do
         with_options instance_writer: false, instance_reader: false do |serializer|
-          serializer.class_attribute :_attribute_mappings # @api private : maps attribute key names to names to names of implementing methods, @see #attribute
-          self._attribute_mappings ||= {}
+          serializer.class_attribute :_attribute_procs # @api private : maps attribute key names to names to names of implementing methods, @see #attribute
+          self._attribute_procs ||= {}
           serializer.class_attribute :_attribute_keys # @api private : maps attribute names to keys, @see #attribute
           self._attribute_keys ||= {}
         end
@@ -17,7 +17,16 @@ module ActiveModel
           @attributes = nil if reload
           @attributes ||= self.class._attribute_keys.each_with_object({}) do |(name, key), hash|
             next unless requested_attrs.nil? || requested_attrs.include?(key)
-            hash[key] = self.class._attribute_mappings[name].call(self)
+            hash[key] = _attribute_value(name)
+          end
+        end
+
+        # @api private
+        def _attribute_value(name)
+          if self.class._attribute_procs[name]
+            instance_eval(&self.class._attribute_procs[name])
+          else
+            read_attribute_for_serialization(name)
           end
         end
       end
@@ -25,7 +34,7 @@ module ActiveModel
       module ClassMethods
         def inherited(base)
           super
-          base._attribute_mappings = _attribute_mappings.dup
+          base._attribute_procs = _attribute_procs.dup
           base._attribute_keys = _attribute_keys.dup
         end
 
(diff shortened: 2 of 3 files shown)


bf4:
On first read looks great!

BTW, that's why I set the things to api private :)

beauby:
@bf4 yup, sorry I didn't make it in time for #1356. I'm not sure this PR deserves a changelog though.

bf4:
Up to you

B mobile phone

beauby:
Added a commit to avoid storing a lambda calling a proc. How are we feeling on this PR?

lib/active_model/serializer/attributes.rb
```diff
@@ -1,55 +1,32 @@
...
+            hash[key] = _attribute_value(name)
+          end
+        end
+
+        # @api private
+        def _attribute_value(name)
+          if self.class._attribute_procs[name]
+            instance_eval(&self.class._attribute_procs[name])
+          else
+            read_attribute_for_serialization(name)
```
bf4: - so, I kind of like the polymorphism better.
- but at the end of the day, this is less code and the switch is entirely contained here, which is good
- on the other hand, now there's a data clump of ` _attribute_procs` and ` _attribute_keys` passed around
- and ` _attribute_keys` sounds a lot like `_attributes_keys`, which seems to be a bug attractor to me
- but `_attribute_value(name)` is a nicer interface than `attribute_mapping.call(self)` and is more consistent with the reflection value

So, how about

```ruby
class Attribute
  attr_reader :name, :block
  def initialize(attr, block = nil)
    @name = attr
    @block = block
  end
  def value(instance)
    if @block
      instance.instance_eval(&block)
    else
      instance.read_attribute_for_serialization(name) 
    end
end
```

and we've removed storing the unnecessary lambdas, introduced a smaller change to the serializer interface, and isolated the attribute vs. block switch with a consistent interface to Reflection?

```ruby
         def attribute(attr, options = {}, &block)
           _attribute_mappings[key] = Attribute.new(attr, block)
```

bf4: @beauby your thoughts?

bf4:
failures in CI probably fixed by #1384

beauby:
@bf4 How about now?

lib/active_model/serializer/attribute.rb
```diff
@@ -0,0 +1,13 @@
+module ActiveModel
+  class Serializer
+    Attribute = Struct.new(:name, :key, :block) do
+      def value(serializer)
+        if block
+          serializer.instance_eval(&block)
+        else
+          serializer.read_attribute_for_serialization(name)
+        end
+      end
```
bf4: I feel better about this l, since it appears to be a pattern we're using all over. However, I still don't have a good sense if we'll want to keep this style, and would like to mark the object as part of the internal api.

beauby: Agreed. Every reference to an `Attribute` object is currently marked as `@api private`.

lib/active_model/serializer/attributes.rb
```diff
@@ -1,55 +1,23 @@
...
 
       included do
         with_options instance_writer: false, instance_reader: false do |serializer|
-          serializer.class_attribute :_attribute_mappings # @api private : maps attribute key names to names to names of implementing methods, @see #attribute
-          self._attribute_mappings ||= {}
+          serializer.class_attribute :_attributes_data # @api private
+          self._attributes_data ||= {}
         end
 
+        autoload :Attribute
```
bf4: I'd prefer the `autoload` below the `extend ActiveSupport::Concern`.  I believe that's how we're doing it elsewhere.  Any particular reason it's here?

bf4: or should it be auto-loaded at all?

beauby: I don't feel strongly about this but I followed what was in [Associations](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/associations.rb#L22). However, I didn't put the `extend ActiveSupport::Autoload`, which usually goes above the `autoload :things`. What's the use of `extend ActiveSupport::Autoload`?

bf4: It's a different autoload from the one in ruby.  I forget what it adds.  Autoload is now thread safe in our supported rubies 

B mobile phone

lib/active_model/serializer/attributes.rb
```diff
@@ -1,55 +1,23 @@
...
-      end
-
       extend ActiveSupport::Concern
 
       included do
         with_options instance_writer: false, instance_reader: false do |serializer|
-          serializer.class_attribute :_attribute_mappings # @api private : maps attribute key names to names to names of implementing methods, @see #attribute
-          self._attribute_mappings ||= {}
+          serializer.class_attribute :_attributes_data # @api private
+          self._attributes_data ||= {}
```
bf4: @beauby For anyone looking back at this, would mind jotting down the advantage of the name change?

I know that part of it is that 
- in both cases the `Attribute` can get the serialization value given a block or record's attribute name, and then
- whether or not the `Attribute` should *also* know the serialization key used for its value/

```diff
-          serializer.class_attribute :_attribute_mappings # @api private : maps attribute key names to names to names of implementing methods, @see #attribute
+          serializer.class_attribute :_attributes_data # @api private
```

```diff
       def attributes(requested_attrs = nil, reload = false)
            @attributes = nil if reload
 -          @attributes ||= self.class._attribute_mappings.each_with_object({}) do |(key, attribute_mapping), hash|
 +          @attributes ||= self.class._attributes_data.values.each_with_object({}) do |attr, hash|
```

```diff
         def attribute(attr, options = {}, &block)
 -          _attribute_mappings[key] = Attribute.build(attr, block)
 +          _attributes_data[attr] = Attribute.new(attr, key, block)
```
```diff
-        # names of attribute methods
+        # keys of attributes
          # @see Serializer::attribute
          def _attributes
-          _attribute_mappings.keys
+          _attributes_data.values.map(&:key)
```

```diff
      def _attributes_keys
 -          _attribute_mappings
+          _attributes_data.values
 -            .each_with_object({}) do |(key, attribute_mapping), hash|
+            .each_with_object({}) do |attr, hash|
 -              next if key == attribute_mapping.name
+              next if attr.key == attr.name
-              hash[attribute_mapping.name] = { key: key }
+              hash[attr.name] = { key: attr.key }
```

```diff
+    Attribute = Struct.new(:name, :key, :block) do
-                                        (name, block)
```

beauby: To be honest, I'd rather have it called `_attributes`, but it was already used... So `_attributes_data` instead of `_attribute_mappings` is simply because the `Attribute` object stores all data about an attribute definition, rather than a mapping.

bf4: I wrote above:

> - in both cases the `Attribute` can get the serialization value given a block or record's attribute name, and then
> - whether or not the `Attribute` should *also* know the serialization key used for its value/

So, the difference in the data structure is that given a serializer containing

```ruby
attribute :object_attribute
attribute :aliased_attribute, :attribute_alias
attribute :dynamic_attribute { 'much_dynamic' }
```

The difference in _mappings vs. _data key/value pairs would be

| `_attribute_mappings` | `_attributes_data` |
|------------------------------ |-------------------------|
| `object_attribute: { name: serializer.read_attribute_for_serialization(:object_attribute) }` | `object_attribute: { key: :object_attribute, name: serializer.read_attribute_for_serialization(:object_attribute) }`
| `attribute_alias: { name: :aliased_attribute, value: serializer.read_attribute_for_serialization(:aliased_attribute) }` | `aliased_attribute: { key: :attribute_alias, name: :aliased_attribute, value: serializer.read_attribute_for_serialization(:aliased_attribute) }`
| `dynamic_attribute: { name: :dynamic_attribute, value: serializer.instance_eval(&block) }` | `dynamic_attribute: { key: :dynamic_attribute, name: :dynamic_attribute, value: serializer.instance_eval(&block) }` |


both of which resolve to:

```ruby
{
  object_attribute: object.send(:object_attribute),
  attribute_alias: object.send(:aliased_attribute),
  dynamic_attribute: 'dynamic_attribute'
}
```

by applying _attributes to either the _mappings or _data where they are:

| ---- | for `_attribute_mappings` | for `_attributes_data` |
|------------------------------ |-------------------------|----------------------|
| `_attributes` | `_attribute_mappings.keys #=> [:object_attribute, :attribute_alias, :dynamic_attribute]` |  `_attributes_data.values.map(&:key) #=> [:object_attribute, :aliased_attribute, :dynamic_attribute]`

which means that the definition of `_attributes` has changed when an attribute specifies `:key`; it would no longer be the serialization key, but the object attribute name.

I think this is why in every case I see, the `_attributes_data` ignores the key in each pair and calls `_attributes_data.values`, which is also why I called it a mapping. (Which I think I got from a comment on the original PR, IIRC).

That makes sense? That's the heart of what I was really asking about.

beauby: Oh I see. Thing is, `_attributes_data` could really be just an array, but access indexed by attribute name makes it easier to avoid/catch duplicate attribute definitions (although it shouldn't happen).

bf4: So you're using a hash to ensure unique attributes via its lookup keys, but otherwise only using the values?

Interesting. I believe that's what Set does internally, or functionally similar.

What do you see as the downside of just declaring the key in the serializer's data structure and let the attribute just be the value (ie previous vs current)?

B mobile phone

beauby: Aaand I won't, because you sold me on removing the redundant `key` from the `Attribute` structure. Which leaves, in this PR, the renaming from `attribute_mappings` to `attributes_data`, and the actual modification of your `Attribute` class. I don't feel strongly about the former, although I do not like the name `mapping` when really the structure holds information abous the attribute (in a philosophical sense). The latter though, I believe, adds clarity (much shorter) and simplicity (avoids storing lambdas calling blocks). Thoughts?

bf4: I still think it maps the serialization's attribute key to its value (or something that can get its value). But, I won't hold the PR for that. Too bad no one else is commenting :disappointed: 

Code looks good.  Just:

- Mind adding a changelog
- [you're right about the location of the autoload](https://github.com/rails-api/active_model_serializers/pull/1370/files#r48460097). I somehow saw the ActiveSupport::Concern and read it as ActiveSupport::Autoload.  In any case, we should use `extend ActiveSupport::Autoload`, is what I meant.  It does have a slightly terser (convention-driven) syntax available, but also works with eager_loading, etc.  And since it *doesn't* work with 'require', I'd rather just be consistent. http://api.rubyonrails.org/classes/ActiveSupport/Autoload.html

beauby: Agreed on both, will update later today.

beauby: Those two remarks were taken care of today. I believe the only thing left is naming? I'd like for other people to chip in as well, but it does not seem like it's going to happen...

beauby: I'm willing to revert the naming back to `attribute_mappings` as `attributes_data` does not fully satisfy me either. Apart from that, are we good to merge?

lib/active_model/serializer/attributes.rb
```diff
@@ -1,55 +1,24 @@
...
         end
 
+        extend ActiveSupport::Autoload
+        autoload :Attribute
```
bf4: - [ ] I don't know that referring to this as a mixin is a useful changelog entry, but I'm going to merge it.. just to end the back and forth on it ;)

> Simplify attributes handling via a mixin

lib/active_model/serializer/attributes.rb
bf4: - [ ] I don't want to nitpick.  I would support it, but don't want it to delay us.

PR #1371, opened 2015-12-10
[DOCS] Refactor, update, create documentation

bf4:
I did a lot of re-organizing, included deduplication of content.

I also wrote some new docs.

There are likely bugs:

- spelling
- doesn't make sense
- broken link
- missing link
- out of place

## Please help review!

- Click `view` on any of the files in the the diff to see how they render
- Or just start from the rendered [readme](https://github.com/bf4/active_model_serializers/blob/documentation_updates/README.md) or the [guides](https://github.com/bf4/active_model_serializers/blob/documentation_updates/docs/README.md)

Note to self, https://github.com/rails-api/active_model_serializers/blob/0-8-stable/DESIGN.textile is neat. It's not in the commit history since 0.10 starts from it's own root node.

Issue #827, opened 2015-03-06
Adding fetch_multi to Cache Strategy


joaomdmoura:
As pointed out by @thibaudgg on my last PR #810 (Adding Fragment Cache to AMS), [fetch_multi](http://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-fetch_multi) would be a great optimisation to the new cache implementation (#693).

One feature of the new implementation is the individual cache strategy that enables your application to re-use the cache in different responses.

Ex. Use an object cached from a ```index``` method response on a ```show``` method response. The opposite is also true.

When retrieving this cached objects AMS will also do this individually, and that's when ```fetch_multi``` comes in. It improve performance retrieving multiple cache keys for a collection in a single go.

Despite of a simple improvement it isn't easy, this will change how the ```Adapter``` fetches cache and integrate with ```FragmentCache```   

I'm planning to work on this, but would like to hear some thoughts :)

thibaudgg:
Great news, thanks @joaomdmoura!
I'm not familiar with the implementation, but IMHO adding `fetch_multi` seems really mandatory for production. 
It would works and reuse the object cached for associations as well right?

joaomdmoura:
@thibaudgg I don't think it's mandatory (we are already conducting some production tests) but yeah, it it's definitively a great optimisation! Really useful! 
In this case it wouldn't be necessary in cached associations. The associations doesn't generate an individual cache for the associated object. It's cached along the original requested object.

thibaudgg:
@joaomdmoura you're right "mandatory" is too strong, but definitely great to have.
My point with associations is to have each object cached separately, so for example if you load & cache a `post` with `comments` and you want to load just a `comment` after it should already be cached.
I guess that @kurko's [comment](https://github.com/rails-api/active_model_serializers/pull/810#issuecomment-77917040) goes in that direction.

joaomdmoura:
Yeah @thibaudgg we discussed it yesterday and we'll move forward with this strategy! :smile: I'm not sure if ```fetch_multi``` will be used to associations yet, but they will be cache separately

thibaudgg:
Great, keep up the good work!

bf4:
The related thoughtbot article is pretty good, even though for 0.8 https://robots.thoughtbot.com/fast-json-apis-in-rails-with-key-based-caches-and

philipgiuliani:
[...]

PR #1372, opened 2015-12-11
support read_multi

LcpMarvel:


lib/active_model/serializer.rb
```diff
@@ -116,7 +116,7 @@ def self.fragmented(serializer)
     # @todo require less code comments. See
     # https://github.com/rails-api/active_model_serializers/pull/1249#issuecomment-146567837
     def self.cache(options = {})
-      self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
+      self._cache = ActionController::Base.cache_store if ActiveModelSerializers.config.perform_caching
```
bf4: Any change to AMS.config.cache_store should be where it is set

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -24,6 +24,13 @@ def fragment_cache(cached_hash, non_cached_hash)
         private
 
         def serializable_hash_for_collection(options)
+          unless options[:batch_cache].present?
+            keys = CachedSerializer.object_cache_keys(serializer, @include_tree)
+            values = ActionController::Base.cache_store.read_multi(*keys)
```
bf4: Here and elsewhere, use the configured cache_store

bf4:
@LcpMarvel Thanks for this :100: 

I'll review.  It will need some tests and documentation. Can you add that?

ref: https://github.com/rails-api/active_model_serializers/issues/827

lib/active_model/serializer.rb
LcpMarvel: I found that the `ActiveModelSerializers.config.cache_store` is always nil.

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -39,6 +39,33 @@ def object_cache_key
...
+                association.serializer.each do |sub_serializer|
+                  cache_keys << object_cache_key(sub_serializer)
+                end
+              else
+                cache_keys << object_cache_key(association.serializer)
+              end
+            end
+          end
+
+          cache_keys.reject(&:nil?)
```
bf4: vs. compact?

LcpMarvel: ok.

lib/active_model/serializer.rb
LcpMarvel: Should i do extra config for `ActiveModelSerializers.config.cache_store`? @bf4

lib/active_model/serializer/adapter/cached_serializer.rb
LcpMarvel:
```ruby
require 'benchmark'

n = 100000

Benchmark.bm do |x|
  x.report do
    n.times do
      Array.new(100) { nil }.reject(&:nil?)
    end
  end

  x.report do
    n.times do
      Array.new(100) { nil }.compact
    end
  end
end
```

```
   0.970000   0.020000   0.990000 (  0.984381)
   0.630000   0.010000   0.640000 (  0.633962)
```

lib/active_model/serializer.rb
bf4: What did yoi have in mind?

B mobile phone

LcpMarvel:
```ruby
    initializer 'active_model_serializers.caching' do
        binding.pry

        ActiveModelSerializers.config.cache_store     = ActionController::Base.cache_store
        ActiveModelSerializers.config.perform_caching = Rails.configuration.action_controller.perform_caching
      end
    end
```

I found that it does not work for Rails 4.2.5.
The `ActionController::Base.cache_store` is nil, but the`Rails.configuration.action_controller.perform_caching` is `true`.

bf4:
I wonder why.  That's a strange default

B mobile phone

LcpMarvel:
I think we should use `ActiveModelSerializers.config.cache_store     = ActiveSupport::Cache.lookup_store(Rails.configuration.cache_store)` to fix this problem.

bf4:
@LcpMarvel Thanks for looking into this.  I'm learning more about this part of the Rails stack. 

It looks to me like we are safest getting the 'Rails' cache store via `ActiveSupport::Cache.lookup_store(ActionController::Base.cache_store || Rails.cache)` since that won't blow up in the absence of a Rails app (i.e. `Rails.configuration.cache_store` won't raise undefined method 'config' on nil).  Otherwise, we can append the condition `Rails.application && Rails.configuration.cache_store` (i.e. Rails.application.config.cache_store).  Then, the default for nil cache_store would be the Memory cache.

I also noticed in the ActionController::Cache code that there's a `caching_enabled?` check that combines perform_caching with the presence of config.cache_store. I like that idea. In that case, we probably wouldn't want to default to the memory store when no store is set.

Any reason you're using `read_multi` over `fetch_multi`? `fetch_multi` calls `read_multi`.

LcpMarvel:
Yes, here we have no necessary for calling block because of the `cache_check` method.

LcpMarvel:
:+1:

LcpMarvel:
I wonder know how to set cache key by `current_user.id` because i have to handle the response by user.

My solution is 
```ruby
def attributes(requested_attrs = nil)
  attrs = super

  self.class.cache key: "#{scope.id}/#{scope.updated_at.to_i}"
  # business logic

  attrs
end
```

Can you give some suggestions?

bf4:
@LcpMarvel 

> I wonder know how to set cache key by current_user.id because i have to handle the response by user.

see https://github.com/rails-api/active_model_serializers/pull/1371/files#diff-683b15b4d7c82f436aaeada1733d9eecR97 https://github.com/bf4/active_model_serializers/blob/1b2324bd021b7e9adc5be11c123dd90286f3541b/docs/general/serializers.md#cache_key

```ruby
# Uses a custom non-time-based cache key
def cache_key
  "#{self.class.name.downcase}/#{self.id}"
end
```

Do you use `scope`? If so, I could use some help documenting it https://github.com/rails-api/active_model_serializers/pull/1252

LcpMarvel:
Yes, I think we can use any variables from the `instance_options`.

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -24,6 +24,15 @@ def fragment_cache(cached_hash, non_cached_hash)
         private
 
         def serializable_hash_for_collection(options)
+          if options[:batch_cache].blank? && ActiveModelSerializers.config.cache_store.present?
```
bf4: batch_cache is a purely internal serialization option?

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -24,6 +24,15 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+            keys = CachedSerializer.object_cache_keys(serializer, @include_tree)
+            if keys.present?
+              values = ActiveModelSerializers.config.cache_store.read_multi(*keys)
+
+              options.merge!(batch_cache: values)
+            end
+          end
+
```
bf4: I think this both blocks of code added in this diff should be their own method. Maybe something like (pardon the naming... could be better)

```ruby
  options.merge!(batch_cache(options))
  serializer.map { |s| Attributes.new(s, instance_options).serializable_hash(options) }
end
def batch_cache(options)
  return {} unless options[:batch_cache].blank? && ActiveModelSerializers.config.cache_store.present?
  if keys = CachedSerializer.object_cache_keys(serializer, @include_tree)
    { batch_cache: ActiveModelSerializers.config.cache_store.read_multi(*keys) }
  else
    {}
  end
end
def batch_cached_serializer(options)
  return nil unless options[:batch_cache].present?
  cache_key = CachedSerializer.new(serializer).cache_key
  options[:batch_cache][cache_key]
end
  if batch_cached_serializer_attributes = batch_cached_serializer(options)
    batch_cached_serializer_attributes
  else
    cache_check(serializer) do
      serializer.attributes(options[:fields])
    end
end
```
  
  
  
lib/active_model/serializer/adapter/attributes.rb
LcpMarvel: I think it's free to users. haha.

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -24,9 +24,27 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+          return {} if options[:batch_cache].present? || ActiveModelSerializers.config.cache_store.blank?
+
+          keys = CachedSerializer.object_cache_keys(serializer, @include_tree)
+
+          return {} unless keys.present?
+
+          { batch_cache: ActiveModelSerializers.config.cache_store.read_multi(*keys) }
+        end
+
+        def batch_cached_serializer(options)
```
bf4: I know I wrote this, but I was just guessing at names without knowing the return type.  Would you mind adding a comment what it returns and maybe change the name of the method?

e.g. using YARDoc

```ruby
# @return [Hash] cached attributes
def batch_cached_serializer(options)
```

bf4: Also, you can now test these method individually

lib/active_model/serializable_resource.rb
```diff
@@ -1,7 +1,7 @@
 require 'set'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :batch_cache])
```
bf4: I'm pretty sure this is wrong, because the options passed to the adapter are passed by the renderer, but maybe I'm forgetting something.  I believe the below is how it would be used.

```ruby
ActiveModel::SerializableResource.new(@post, adapter: :json_api).as_json(batch_cache: { disable: true})
```

bf4: BTW, where does the `disable: true` come from?  It's just truthy?

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -24,9 +24,27 @@ def fragment_cache(cached_hash, non_cached_hash)
...
+
           serializer.map { |s| Attributes.new(s, instance_options).serializable_hash(options) }
         end
 
+        def batch_cache(options)
+          return {} if options[:batch_cache].present? || ActiveModelSerializers.config.cache_store.blank?
+
+          keys = CachedSerializer.object_cache_keys(serializer, @include_tree)
+
+          return {} unless keys.present?
```
bf4: could be `return {} if keys.blank?` (but could really be `if keys.nil?`, right?)

LcpMarvel:
@bf4  I have refactored

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -39,6 +39,38 @@ def object_cache_key
...
+                association.serializer.each do |sub_serializer|
+                  cache_keys << object_cache_key(sub_serializer)
+                end
+              else
+                cache_keys << object_cache_key(association.serializer)
+              end
+            end
+          end
+
+          cache_keys.compact.uniq
```
LcpMarvel: @bf4 I removed duplicate keys for less IO

LcpMarvel:
What else do I need to modify?

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -24,9 +24,31 @@ def fragment_cache(cached_hash, non_cached_hash)
...
         def serializable_hash_for_collection(options)
+          options.merge!(batch_cache(options))
+
           serializer.map { |s| Attributes.new(s, instance_options).serializable_hash(options) }
         end
 
+        # Read cache from cache_store, and set those into options[:batch_cache]
+        # @return [Hash] a batch of cache
+        def batch_cache(options)
+          return {} if options[:batch_cache].present? || ActiveModelSerializers.config.cache_store.blank?
```
bf4: simpler to follow if the logic is something likeI think

```ruby
  batch_cache(options)
  serializer.map { |s| Attributes.new(s, instance_options).serializable_hash(options) }
end

def batch_cache(options)
  options[:batch_cache] ||=
    begin
      if ActiveModelSerializers.config.cache_store.nil?
        {}
      else
        keys = CachedSerializer.object_cache_keys(serializer, @include_tree)
        if keys.any?
          ActiveModelSerializers.config.cache_store.read_multi(*keys)
        else
          {}
        end
      end
    end
end
```

then

```diff
         def resource_object_for(options)
 -          cache_check(serializer) do
+          cached_serializer = CachedSerializer.new(serializer)
+          batch_cache(options).fetch(cached_serializer.cache_key) do
+            cached_serializer.cache_check(self) do
                serializer.attributes(options[:fields])
              end
            end
          end
```

bf4:
@LcpMarvel Thanks for this. I don't recall why we didn't merge it. I added a changelog entry in https://github.com/rails-api/active_model_serializers/commit/b45f7b4ffea36b44fdf7be075295ad9db6bea1b0

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -56,8 +86,12 @@ def include_meta(json)
...
 
         def resource_object_for(options)
-          cache_check(serializer) do
-            serializer.attributes(options[:fields])
+          cached_serializer = CachedSerializer.new(serializer)
+
+          cached_attributes(cached_serializer) do
+            cached_serializer.cache_check(self) do
+              serializer.attributes(options[:fields])
+            end
```
bf4: I don't recall why we removed the `cache_check` method here

lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -56,8 +86,12 @@ def include_meta(json)
         end
 
         def resource_object_for(options)
-          cache_check(serializer) do
-            serializer.attributes(options[:fields])
+          cached_serializer = CachedSerializer.new(serializer)
+
+          cached_attributes(cached_serializer) do
+            cached_serializer.cache_check(self) do
```
LcpMarvel: @bf4  we still use `cache_check` method if the cache is not be hit.

bf4: Yeah, it the switch from `cache_check(serializer) do` to `cached_serializer = CachedSerializer.new(serializer); cached_serializer.cache_check(self) do` that isn't immediately obvious and I'm writing rather than digging into it :)

Issue #1335, opened 2015-11-19
ams does not generate an initializer


JiHunJeon:
i tried install version which 0.9.3 and 0.10.0.rc3

but didn't generate ac initailzer

my rails version is 4.2.4 
gem version is 2.4.5

bf4:
It's not supposed to. Was anything not working? Did you run any commands?

bf4:
see https://github.com/rails-api/active_model_serializers/pull/1374

bf4:
Closing as OP has not responded. Please either comment on what you ended up doing or re-open if this is still a problem.  Ref: https://github.com/rails-api/active_model_serializers/blob/master/CONTRIBUTING.md#after

PR #1374, opened 2015-12-13
Only load generators when needed

dgynn:
This PR improves the way the generators are loaded and how they extend the `resource` generator.
* The initializer block has been changed to a `generator` block which is only executed when generators are needed.
* The call to `app.load_generators` has been removed. There is no need to load *all* generators.
* The `resource_override.rb` has been changed to use `hook_for` to extend the resource generator.
* The directory for the generators has been moved to match the way Rails looks to load generators.

With `hook_for` it would now be possible for a user to pass `--no-serializer` to skip that option.
The `--serialize` option also now shows up in the generator help with `rails g resource --help`.

These changes follow the way the Draper gem extends the `controller` generator.


bf4:
@dgynn thanks for this. Do you know if it these changes may solve the problem in https://github.com/rails-api/active_model_serializers/issues/1335

Generator code is one I find hard to lookup.  Did you find a good source, or just go source diving?  If you know the rails plugins infrastructure well, I'd also appreciated your thoughts on  https://github.com/rails-api/active_model_serializers/issues/1352

Could you rebase this and add some more information to the generator section?  I'd be ok with delaying that until https://github.com/rails-api/active_model_serializers/pull/1371/ is merged since it moves the docs around quite a bit.  (Your review there is also appreciated :)

dgynn:
I've mostly been source diving to look into generators, initializers, and lazy hooks. It seems like a lot of people have copied code from similar examples. And possibly Rails 2 required doing extra stuff that is now unnecessary. I'll add some comments on #1352 but it does seem headed in the right direction.

I can add some docs and rebase after #1371 or see about doing a PR to your `documentation` branch.

For #1335, I think there was just confusion expecting that a config/initializer sample would have been generated. That doesn't seem necessary but I suppose one could be added if there are going to be more configuration options.

bf4:
> see about doing a PR to your documentation branch.

Sounds great!

Anything you can share about your learnings, even in a gist or blog post or PR to rails guides would be much appreciated by me :)

dgynn:
I've submitted a rebased PR to your bf4/railties branch since there would be a conflict there. The documentation seems OK. And since the generator docs are in the Getting Started section it may confuse people to tell them how to skip the generator.

I've found another link that talks about this exact issue (extending the `resource` generator) and it even points back to this project. :) Here is what Jbuilder did to extend the `scaffolding` generator... https://github.com/rails/rails/pull/9647

PR #1383, opened 2015-12-20
Simplify reflections handling.

beauby:


bf4:
@beauby Yeah, that totally makes more sense. :100: 

BTW, I still don't really like `value` as the name, but haven't thought of anything better, and at least there's some history around the usage.

Once #1384 or something like it is merged, I'll merge this.

PR #1384, opened 2015-12-20
Fix db state leaking across tests

bf4:
See e.g. https://travis-ci.org/rails-api/active_model_serializers/builds/97975719

db records created specs weren't being cleared, so that
other specs calling `all` had unexpected records in them.

TODO: include a mixin to automatically clear records.  Maybe

```ruby
module TransactionalTests
   # Minitest.run_one_method isn't present in minitest 4
   if $minitest_version > 4 # rubocop:disable Style/GlobalVars
     def run_one_method(*)
       super
     ensure
       ActiveRecord::Base.descendants.each do |record_class|
         record_class.delete_all
       end
     end
  end
end
module SomeTest < Minitest::Test
  extend TransactionalTests
end
```

Perhaps consider replacing all usages of `Minitest::Test` with `ActiveSupport::Test`
for better compatibility.

### DIFF
--- a/test/grape_test.rb
+++ b/test/grape_test.rb
@@ -4,6 +4,24 @@ require 'grape/active_model_serializers'
 
 class ActiveModelSerializers::GrapeTest < Minitest::Test
   include Rack::Test::Methods
+  module Models
+    def self.model1
+      ARModels::Post.new(id: 1, title: 'Dummy Title', body: 'Lorem Ipsum')
+    end
+
+    def self.model2
+      ARModels::Post.new(id: 2, title: 'Second Dummy Title', body: 'Second Lorem Ipsum')
+    end
+
+    def self.all
+      @all ||=
+      begin
+        model1.save!
+        model2.save!
+        ARModels::Post.all
+      end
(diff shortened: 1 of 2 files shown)


bf4:
The lack of passing builds on master is distracting to active PRs.  So, I'm going to merge this.

beauby:
Raising my voice a bit too late about this, but this way of handling stuff does not seem right at all. We should find a better solution.

bf4:
:100:

bf4:
@beauby Yeah, I would have waited longer, but it did fix the immediate issue, and allowed builds to pass, which was blocking good PRs.  I'm not how we should handle 'this PR needs a followup'. Maybe a label? Maybe an issue that lists issues? Maybe a new issue?

beauby:
Not sure either. A label is probably good.

bf4:
Between that and forgetting to use milestones.. I could use a page that says 'start looking for priority work here and here's some stuff not to forget about', or something like that... rails-bot does some cool stuff.. also auto-marks as stale

PR #1385, opened 2015-12-20
Favor ActiveSupport::TestCase over Minitest::Test

bf4:
- Better minitest 4/5 support
- Better DSL
- Already available with no changes
- Consistent interface

### DIFF
--- a/test/active_model_serializers/logging_test.rb
+++ b/test/active_model_serializers/logging_test.rb
@@ -2,7 +2,7 @@ require 'test_helper'
 
 module ActiveModel
   class Serializer
-    class LoggingTest < Minitest::Test
+    class LoggingTest < ActiveSupport::TestCase
       class TestLogger < ActiveSupport::Logger
         def initialize
           @file = StringIO.new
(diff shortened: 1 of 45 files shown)


beauby:
What would be the drawbacks?

bf4:
@beauby I have no idea what the drawback would be.  I do know that at one point AS::TC was removed in favor of raw minitest.  Perhaps speed?

beauby:
After investigating a bit, it appears `ActiveSupport::TestCase` is just an extension of `Minitest::Test`, so this PR should not be an issue. I'll merge it tomorrow unless somebody raises their voice.

spastorino:
It's ok to merge this, when I was building AMS 0.10 I used Minitest::Test just because I didn't need AS::TC and wanted to depend as minimum as possible on Rails stuff.

beauby:
Great, thanks for the heads up @spastorino. As 0.10 is already depending on quite a bit of rails stuff, I don't believe this PR will hurt much, so I'm merging.

PR #1386, opened 2015-12-21
Remove capture_warnings testing. Is fickle and Rubocop linting is good enough

bf4:
This is just something that's been bothering me, and without it, we get extra warnings from grape, so ...


test/capture_warnings.rb
```diff
@@ -17,17 +17,19 @@ def initialize(fail_on_warnings = true)
     @output = STDOUT
   end
 
-  def execute!
+  def execute!(minitest_run)
     $VERBOSE = true
     $stderr.reopen(stderr_file.path)
-
-    Minitest.after_run do
+    at_exit do
```
bf4: since we run minitest in `END` now, we can rely on `at_exit` running *after* minitest ends

test/test_helper.rb
```diff
@@ -20,27 +20,35 @@
...
   # Minitest 4
   # https://github.com/seattlerb/minitest/blob/644a52fd0/lib/minitest/autorun.rb
   # https://github.com/seattlerb/minitest/blob/644a52fd0/lib/minitest/unit.rb#L768-L787
   # Ensure backward compatibility with Minitest 4
   Minitest = MiniTest unless defined?(Minitest)
   Minitest::Test = MiniTest::Unit::TestCase
-  def Minitest.after_run(&block)
-    MiniTest::Unit.after_tests(&block)
-  end
+  minitest_run = ->(argv) { MiniTest::Unit.new.run(argv) }
```
bf4: basically copied from minitest autorun in versions 4 and 5. the tl;dr is that autorun is just sugar for `at_exit { MiniTest::Unit.new.run(ARGV) }` in 4 and `at_exit { Minitest.run(ARGV) }` in 5.  Copied because we're kind of writing our own 'autorun' here, by controlling when minitest starts in the `MiniTestHack` class below

test/test_helper.rb
```diff
@@ -20,27 +20,35 @@
...
+  $minitest_version = 5
+  # Minitest 5
+  # https://github.com/seattlerb/minitest/blob/e21fdda9d/lib/minitest/autorun.rb
+  # https://github.com/seattlerb/minitest/blob/e21fdda9d/lib/minitest.rb#L45-L59
+  require 'minitest/spec'
+  require 'minitest/mock'
+  minitest_run = ->(argv) { Minitest.run(argv) }
 end
+require 'minitest/reporters'
+Minitest::Reporters.use!
```
bf4: I think @zenspider said he doesn't like Minitest::Reporters, but I don't recall why.  It makes controlling the output(s) a lot easier :).  

(@zenspider this basically let's me guarantee I can run code after minitest finishes by moving the autorun from `at_exit` to `END`, which resolves before `at_exit` resolves.  This is sort of what I was talking about at RubyConf, that I couldn't run code around minitest.  This is Just Ruby, right? :)

test/test_helper.rb
```diff
@@ -71,6 +79,36 @@ def Minitest.after_run(&block)
...
+        @@after_tests.reverse_each(&:call)
+      end
+
+      exit exit_code || false
+    end
+
+    exit_code = minitest_run.call(ARGV)
+  end
+end
+END { MiniTestHack.autorun(minitest_run) }
```
bf4: behold, the `END` block

test/test_helper.rb
```diff
@@ -71,6 +79,36 @@ def Minitest.after_run(&block)
...
+    # don't run if there was a non-exit exception
+    return if $! and not ($!.kind_of? SystemExit and $!.success?)
+
+    # the order here is important. The at_exit handler must be
+    # installed before anyone else gets a chance to install their
+    # own, that way we can be assured that our exit will be last
+    # to run (at_exit stacks).
+    exit_code = nil
+
+    trap('EXIT') do
```
bf4: We can't use `END` inside a method, but trapping EXIT is just fine, especially since we don't need to run with any other signal

Issue #616, opened 2014-08-28
ActionController subscriptions overwritten.


tristandunn:
In 0.9.0 `ActionController::SerializationAssertions` defines `setup_subscriptions` and `teardown_subscriptions` methods, but doesn't call super. Which means the `ActionController::TemplateAssertions` subscriptions are never created, which is breaking the `assert_template` assertion.

I'm not sure if calling `super` is the best fix. I'm up for creating a patch once I know the preferred solution.

PR #1390, opened 2015-12-22
Bring back assert_serializer for controller testing

bf4:
(This PR includes and extends the work done by @maurogeorge in https://github.com/rails-api/active_model_serializers/pull/1099)

The `assert_serializer` test helper was added in 0.9.0.apha1 [1],
and was not included in 0.10.

This patch brings back the `assert_serializer` test helper from the last
revision [2] that has the helper. The original helper was used as base.

[1]: https://github.com/rails-api/active_model_serializers/pull/596
[2]: https://github.com/rails-api/active_model_serializers/tree/610aeb2e9297fa31b8d561f0be9a4597f0258f8c

- Create the AssertSerializer

- Use the Test namespace

- Make the tests pass on the Rails master
  - Rails 5 does not include `assert_template` but we need this on the tests of
the helper.
  - This add the `rails-controller-testing` to keep support on `assert_template`.

- Only load test helpers in the test environment


bf4:
@maurogeorge I made a bunch of changes in small commits, mostly explained in the commit message.  I look forward to hearing your thoughts.

test/active_model_serializers/test/serializer_test.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      def test_raises_argument_error_when_asserting_with_invalid_object
+        get :render_using_serializer
+        e = assert_raise ArgumentError do
+          assert_serializer Hash
+        end
+        assert_match 'assert_serializer only accepts a String, Symbol, Regexp, ActiveModel::Serializer, or nil', e.message
+      end
+
+      def test_does_not_overwrite_notification_subscriptions
```
bf4: The intent of this test has changed a bit, and it should probably be renamed.  I'm not really sure why asserting a template was rendered would relate to 'not overwriting notification subscriptions'. The assert template test does use the [`!render_template.action_view'`](https://github.com/rails/rails-controller-testing/blob/f756b33c138c593eabe37f6085f8bac477b99bfe/lib/rails/controller/testing/template_assertions.rb) event and I wonder if it was just that the original implementation used the same setup/teardown names as the rails assertion. (   setup :setup_subscriptions          teardown :teardown_subscriptions )

bf4: Compare 

- https://github.com/rails-api/active_model_serializers/pull/596/ which uses the same setup/teardown methods as the rails assertion to 
- https://github.com/rails-api/active_model_serializers/pull/620 which fixed https://github.com/rails-api/active_model_serializers/issues/616, that action controller subscriptions were being overwritten.  

So, the `render_a_template` tests are really failing tests prior to #620 that showed that #620 works. I'm not sure they're useful as regression tests.

docs/README.md
```diff
@@ -22,6 +22,7 @@ This is the documentation of ActiveModelSerializers, it's focused on the **0.10.
 - [How to add root key](howto/add_root_key.md)
 - [How to add pagination links](howto/add_pagination_links.md)
 - [Using ActiveModelSerializers Outside Of Controllers](howto/outside_controller_use.md)
+- [Testing ActiveModelSerializers](howto/test.md)
```
joaomdmoura: :heart_eyes:

docs/howto/test.md
```diff
@@ -0,0 +1,18 @@
...
+
+```ruby
+class PostsControllerTest < ActionController::TestCase
+  test "should render post serializer" do
+    get :index
+    assert_serializer "PostSerializer"
+  end
+end
+```
+
```
trek: Outside the scope of this PR, but something link https://github.com/rails-api/active_model_serializers/issues/1011#issuecomment-127608121 should probably be included here as well for unit testing serializers themselves

bf4: @trek Yeah, I intend to add that if someone else doesn't... maybe it would be good to make an issue? I've been meaning to write it up as code in https://github.com/rails-api/active_model_serializers/pull/967 but it hasn't 'happened' yet.

In the meantime, I guess people can post links here :)

lib/active_model_serializers/test/serializer.rb
```diff
@@ -0,0 +1,126 @@
...
+      #
+      def assert_serializer(expectation, message = nil)
+        @assert_serializer.expectation = expectation
+        @assert_serializer.message = message
+        @assert_serializer.response = response
+        assert(@assert_serializer.matches?, @assert_serializer.message)
+      end
+
+      class AssertSerializer
+        EVENT_NAME = ::ActiveModelSerializers::Logging::RENDER_EVENT
```
joaomdmoura: is there a need for this to be a constant?

bf4: Just prevents typos https://github.com/bf4/active_model_serializers/commit/bcc2c97ba421c905ba5011c670e729994f161bd3

joaomdmoura: Yup, totally agree with `RENDER_EVENT`, was talking about `EVENT_NAME`, AFAIK it's only used in https://github.com/rails-api/active_model_serializers/pull/1390/files#diff-e30fc2f0317dd31337115e3d092b1d5aR109

bf4: Oh, mean just stick `::ActiveModelSerializers::Logging::RENDER_EVENT` in `event_name`? I don't have a strong opinion.  I'll make that change

joaomdmoura: Yup :grin: tkx

joaomdmoura:
LGTM :+1:

lib/active_model_serializers/test/serializer.rb
bf4: Fixed

maurogeorge:
@bf4 great job, I added some comments.

bf4:
lulz, I forgot I did this [in my Rails serializers PR](https://github.com/rails/rails/pull/21496/files#diff-d843ada9fd68ee95c5f6ed8eadfe0ef0R86)

```diff
+    def assert_serializer!(serializer_name)
 +      return true if has_serializer?(serializer_name)
 +      raise ActionController::MissingSerializer, "There is no '#{serializer_name}' serializer.\n" <<
 +            "Known serializers are #{_serializers.keys}"
 +    end
```

lib/active_model_serializers/test/serializer.rb
```diff
@@ -0,0 +1,125 @@
+require 'set'
+module ActiveModelSerializers
+  module Test
+    module Serializer
+      extend ActiveSupport::Concern
+
+      included do
+        setup :setup_serialization_subscriptions
+        teardown :teardown_serialization_subscriptions
```
maurogeorge: We really need this `teardown` ? If we need, I think we need to write a broken test to it, what behavior will be broken if we do not have this `teardown`.

lib/active_model_serializers/test/serializer.rb
```diff
@@ -0,0 +1,125 @@
...
+      #
+      def assert_serializer(expectation, message = nil)
+        @assert_serializer.expectation = expectation
+        @assert_serializer.message = message
+        @assert_serializer.response = response
+        assert(@assert_serializer.matches?, @assert_serializer.message)
+      end
+
+      class AssertSerializer
+        attr_reader :serializers, :message, :_subscribers
```
maurogeorge: I think this `_subscribers` can be a protected `attr_acessor`

maurogeorge:
Hi @bf4 I bring back some comments, I guess the last time I made these comments on the commit, probably they are lost after some forced push.

Are only 2 small comments and :shipit:

lib/active_model_serializers/test/serializer.rb
bf4: I don't think it needs a test. 

 It just unsubscribes the subscriber created in the test...

 See more in the commit https://github.com/rails-api/active_model_serializers/commit/d9eefdba7111909bedfc85766fd1d05fe191cdc3


I can try to show what may break without it, but given that the teardown is copied from rails, and I understand the behavior with and without...

B mobile phone

lib/active_model_serializers/test/serializer.rb
bf4: Ok

B mobile phone

lib/active_model_serializers/test/serializer.rb
bf4: Ref original comment https://github.com/rails-api/active_model_serializers/commit/11fda591b04f673dafcbcb1a51af273ae8f64305#commitcomment-15145827

maurogeorge: Ok, I am good with that.

bf4:
Should be ready to go now

lib/active_model_serializers/test/serializer.rb
```diff
@@ -0,0 +1,129 @@
...
+        assert(@assert_serializer.matches?, @assert_serializer.message)
+      end
+
+      class AssertSerializer
+        attr_reader :serializers, :message
+        attr_accessor :response, :expectation
+
+        def initialize
+          @serializers = Set.new
+          self._subscribers = []
```
maurogeorge: I think we can use `@subscribers` over `self._subscribers`, I don't think we need to use `_` to declare visibility.

test/fixtures/template.html.erb
```diff
@@ -0,0 +1 @@
+<p>Hello.</p>
```
maurogeorge: Since this https://github.com/bf4/active_model_serializers/commit/6508637ee9bb0e6459e2bdfc6499771cdd29fc61 test was removed I think we can remove this file.

maurogeorge:
@bf4 sorry for that, I added 2 small comments. For me after that :shipit:

lib/active_model_serializers/test/serializer.rb
beauby: Agreed!

bf4: I never expected this to get so much discussion!  I just cargo-culted it from Rails and...

bf4:
Updated

test/active_model_serializers/test/serializer_test.rb
```diff
@@ -0,0 +1,63 @@
...
+      include ActiveModelSerializers::Test::Serializer
+
+      class MyController < ActionController::Base
+        def render_using_serializer
+          render json: Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+        end
+
+        # For Rails4.0
+        def render_some_text
+          Rails.version > '4.1' ? render(plain: 'ok') : render(text: 'ok')
```
bf4: womp womp  (Really for Rails 5, but it works on 4.1+)

lib/active_model_serializers/test/serializer.rb
beauby: I think the point is not that much about the underscore, but more about ivar vs accessors.

maurogeorge:
@bf4 LGTM :shipit:

maurogeorge:
Thanks for the merge :heart:

dbwieland18:
Can someone provide the steps I need to take in order to make the ```assert_serializer``` helper available in an rspec test?

remear:
@dbwieland18 #1470

PR #1394, opened 2015-12-23
Remove ActiveModelSerializers.silence_warnings dead code

maurogeorge:
The ActiveModelSerializers.silence_warnings was used to avoid warnings on the
Ruby interpreter when define a private attr_acessor. This method is not used in
any part of the code and the recommend way to handle this case is to use
protected instead the silence_warnings [1].

This patch remove the method from the project, this way we avoid people using
this by mistake.

I hope you guys enjoy :smile: 

[1]: https://bugs.ruby-lang.org/issues/10967

### DIFF
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -13,39 +13,6 @@ module ActiveModelSerializers
   autoload :Model
   autoload :Callbacks
   autoload :Logging
-
-  module_function
-
-  # @note
-  #   ```ruby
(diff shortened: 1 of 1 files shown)


bf4:
@maurogeorge FWIW, besides the protected attr_readers, it was also used in the dynamically defined 'attribute' methods on the serializer /shrug

maurogeorge:
@bf4 thanks for the info and for the merge :heart:

PR #1360, opened 2015-12-01
Update travis build matrix to include Ruby 2.2.3 for Rails5

KaraAJC:
contributing to issue #1303 
Utilized the following resources for the changes: 
- https://docs.travis-ci.com/user/languages/ruby/
- https://docs.travis-ci.com/user/customizing-the-build/
- comments and changes from #1324

PR #1270, opened 2015-10-14
Proposal: Create assert_response_schema test helper

maurogeorge:
It is a common pattern to use JSON Schema to validate a API response [1], [2]
and [3].

This patch creates the `assert_response_schema` test helper that helps people do
this kind of validation easily on the controller tests.

This makes sense to you guys? If make sense I can continue with CHANGELOG, Docs and so on.

[1]: https://robots.thoughtbot.com/validating-json-schemas-with-an-rspec-matcher
[2]: https://github.com/sharethrough/json-schema-rspec
[3]: https://github.com/rails-api/active_model_serializers/issues/1011#issuecomment-127608121

PR #1395, opened 2015-12-23
Drop Ruby 1.9.3 on AppVeyor

maurogeorge:
Hi guys,

I think we can remove this from AppVeyor since we are dropping 1.9.3 support #1360.

This is causing the build from my other PR #1270 be broken :disappointed:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,8 +4,6 @@ skip_tags: true
 
 environment:
   matrix:
-    - ruby_version: "193"
-    - ruby_version: "193-x64"
     - ruby_version: "200"
     - ruby_version: "200-x64"
     - ruby_version: "21"


bf4:
Omg, [rubinius](https://travis-ci.org/rails-api/active_model_serializers/jobs/98592942), I banish you to the allow failures. I think that's three today! And it happens all the time! cc (with :heart: to @brixen )

brixen:
@bf4 that issue has been fixed. FWIW, you can thank rubygems for loading a massive Regexp of "acceptable" licenses (yes, they think they can specify what is acceptable) when installing a gem, when it should only be relevant to building a gem (if ever, arguably). But hey, why use `#each` when you can have the Regexp engine do iteration for you.

bf4:
@brixen I love you, man, but sometimes... 

> "acceptable"

That list, SPDX, is just standard abbreviations of various licenses.  It's not supposed to be checking what's acceptable, just against a standard list.  (Full-disclosure: I was involved in getting [SPDX](https://github.com/rubygems/rubygems/pull/917) into the RubyGems.org spec and [collaborated on that PR](https://github.com/rubygems/rubygems/pull/1249#issuecomment-108002109))... 

The regex, however, was added later in https://github.com/rubygems/rubygems/pull/1387 by segiddins and hsbt...  I have no idea why they chose the regex approach, but they are certainly Ruby experts...

I agree that the check should really only be made at build.  I suppose the problem with that is that people using the `rake release` bundler task won't get a chance to see the warning until it's too late.

Lemme know when you're back in Chicago :)

brixen:
@bf4 you may have a different definition for "acceptable" than I do, but this is rubygems output

```
$ gem build rubysl-openssl.gemspec
WARNING:  WARNING: license value 'BSD' is invalid.  Use a license identifier from
http://spdx.org/licenses or 'Nonstandard' for a nonstandard license.
```

"invalid" doesn't sound very acceptable. "unrecognized", "did you mean XYZ", or something similar might be, well, acceptable. More relevant, why is this tool dictating this? And why not in a linting step? Deep questions.

Anyway, the Rubinius issue is fixed. No plans to be in Chicago. Come visit Portland. :)

bf4:
> "invalid" doesn't sound very acceptable. "unrecognized", "did you mean XYZ", or something similar might be, well, acceptable. More relevant, why is this tool dictating this? And why not in a linting step? Deep questions.

Ok, the words could probably use some reworking :)  'Someone should do something' :)

Portland... love the city, would love to go, but I just about never fly west.

PR #1402, opened 2015-12-27
Extract caching into its own module.

beauby:


lib/active_model/serializer/caching.rb
```diff
@@ -0,0 +1,106 @@
...
+        (?=      # stop previous match when
+          :\d+     # a colon is followed by one or more digits
+          :in      # followed by a colon followed by in
+        )
+      /x
+
+      module ClassMethods
+        def inherited(base)
+          super
+          caller_line = caller.second
```
beauby: Note that now that caching is inside a mixin, the actual serializer definition is in the second position in the call stack.

bf4: Could still be

```ruby
def inherited(base)
  caller_line = caller.first
  super
  base._cache_digest = digest_caller_file(caller_line)
end
```

Also, I'd rather use `caller[1]` than second just because I prefer to avoid core extensions. I'm not sure if you have a preference.

bf4: though, we should probably rename the local var to `serializer_source_location`

lib/active_model/serializer/caching.rb
```diff
@@ -0,0 +1,106 @@
...
+          serializer_file_path = caller_line[CALLER_FILE]
+          serializer_file_contents = IO.read(serializer_file_path)
+          Digest::MD5.hexdigest(serializer_file_contents)
+        rescue TypeError, Errno::ENOENT
+          warn <<-EOF.strip_heredoc
+            Cannot digest non-existent file: '#{caller_line}'.
+            Please set `::_cache_digest` of the serializer
+            if you'd like to cache it.
+            EOF
+          ''.freeze
```
bf4: In one of my unpushed branches I did:

```ruby
-        rescue TypeError, Errno::ENOENT
-          warn <<-EOF.strip_heredoc
-            Cannot digest non-existent file: '#{caller_line}'.
-            Please set `::_cache_digest` of the serializer
-            if you'd like to cache it.
-            EOF
-          ''.freeze
+      rescue TypeError, Errno::ENOENT, NoMethodError
+        super_omg_digest = "#{self.object_id.to_s}_#{rand(Time.now.to_i)}"
+        warn <<-EOF.strip_heredoc
+           Cannot digest non-existent file: '#{caller_line}'.
+           Please set `::_cache_digest` of the serializer
+           if you'd like to cache it.
+           Using #{super_omg_digest} (the object_id).
+        EOF
+        super_omg_digest
```

:)

bf4: (not relevant here, but it made me think of it)

beauby: Wasn't the reason why we introduced this "non-existent file" thing because the regexp was broken on paths containing spaces?

lib/active_model/serializer/caching.rb
```diff
@@ -0,0 +1,106 @@
...
+          self._cache_only = options.delete(:only)
+          self._cache_except = options.delete(:except)
+          self._cache_options = (options.empty?) ? nil : options
+        end
+
+        # Used to cache serializer name => serializer class
+        # when looked up by Serializer.get_serializer_for.
+        def serializers_cache
+          @serializers_cache ||= ThreadSafe::Cache.new
+        end
```
bf4: `serializers_cache` should probably be in the same file as `get_serializer_for`, right?

```ruby
     # @api private
     # Find a serializer from a class and caches the lookup.
     # Preferentially retuns:
     #   3. nil
```

beauby: Right, I overlooked that.

bf4:
:100: I'm not big on the extra-a-module pattern, but I really think this helps us better understand the parts of the serializer public api.  Interestingly, there was once a [Serializer::DSL object](https://github.com/rails-api/active_model_serializers/commit/cf91146471fa17899010d255d8b2a1686f19d271), but it was only in the [refactor branch](https://github.com/rails-api/active_model_serializers/compare/refactor).

lib/active_model/serializer/caching.rb
bf4: introduced because the file might not exist.  the spaces was just a bug in the regexp

bf4:
I'm ok to merge this once tests pass

lib/active_model/serializer/caching.rb
beauby: Do actually have any regression test for this? I can't think of a case where this would happen.

lib/active_model/serializer/caching.rb
beauby: Nope, the `Caching#inherited` method is always called from `Serializer#inherited`. I did make the change to `caller[1]` though.

lib/active_model/serializer/caching.rb
bf4: Yeah, it's tested

B mobile phone

beauby:
Tests passed, I'm merging as @bf4 agreed.

lib/active_model/serializer/caching.rb
bf4: And it happened, which is why it came up :). It also happens quite a bit in my benchmarking test :)

beauby: Hmm ok, I'll look it up.

bf4: e.g. in IRB, see https://github.com/rails-api/active_model_serializers/pull/1271

PR #1403, opened 2015-12-28
Conditional attributes/associations (if/unless)

beauby:
My take at conditional attributes/associations. Previous work has been done in this direction (#1266), but I figured it was easily expressed on top of #1370. 
This PR adds support for `if/unless` options to attribute/association definitions in serializers. Those options take a symbol of a method name on the serializer.

Example:
```ruby
  attributes :name, :description, :date_of_birth
  attribute :private_data, if: :is_current_user?

  def is_current_user?
    object.id == current_user.id
  end
end
```

Note: if this gets traction, I'll add tests.

Ref #1245, #1266, #1058, #922, #1141

### DIFF
--- a/lib/active_model/serializer/attribute.rb
+++ b/lib/active_model/serializer/attribute.rb
@@ -1,12 +1,39 @@
 module ActiveModel
   class Serializer
-    Attribute = Struct.new(:name, :block) do
+    Attribute = Struct.new(:name, :options, :block) do
       def value(serializer)
         if block
           serializer.instance_eval(&block)
         else
           serializer.read_attribute_for_serialization(name)
         end
+      end
+
+      def included?(serializer)
+        case condition
+        when :if
+          serializer.public_send(condition)
+        when :unless
+          !serializer.public_send(condition)
+        else
+          true
+        end
+      end
+
+      private
+
+      def condition_type
+        if options.key?(:if)
+          :if
+        elsif options.key?(:unless)
+          :unless
+        else
+          :none
+        end
+      end
+
+      def condition
+        options[condition_type]
       end
     end
   end
--- a/lib/active_model/serializer/attribute.rb
+++ b/lib/active_model/serializer/attribute.rb
@@ -1,40 +1,25 @@
+require 'active_model/serializer/field'
+
 module ActiveModel
   class Serializer
-    Attribute = Struct.new(:name, :options, :block) do
-      def value(serializer)
-        if block
-          serializer.instance_eval(&block)
-        else
-          serializer.read_attribute_for_serialization(name)
-        end
-      end
-
-      def included?(serializer)
-        case condition
-        when :if
-          serializer.public_send(condition)
-        when :unless
-          !serializer.public_send(condition)
-        else
-          true
-        end
-      end
-
-      private
-
-      def condition_type
-        if options.key?(:if)
-          :if
-        elsif options.key?(:unless)
-          :unless
-        else
-          :none
-        end
-      end
-
-      def condition
-        options[condition_type]
-      end
+    # Holds all the meta-data about an attribute as it was specified in the
+    # ActiveModel::Serializer class.
+    #
+    # @example
+    #   class PostSerializer < ActiveModel::Serializer
+    #     attribute :content
+    #     attribute :name, key: :title
+    #     attribute :email, key: :author_email, if: :user_logged_in?
+    #     attribute :preview do
+    #       truncate(object.content)
+    #     end
+    #
+    #     def user_logged_in?
+    #       current_user.logged_in?
+    #     end
+    #   end
+    #
+    class Attribute < Field
     end
   end
 end
--- a/lib/active_model/serializer/field.rb
+++ b/lib/active_model/serializer/field.rb
@@ -1,0 +1,55 @@
+module ActiveModel
+  class Serializer
+    # Holds all the meta-data about a field (i.e. attribute or association) as it was
+    # specified in the ActiveModel::Serializer class.
+    # Notice that the field block is evaluated in the context of the serializer.
+    Field = Struct.new(:name, :options, :block) do
+      # Compute the actual value of a field for a given serializer instance.
+      # @param [Serializer] The serializer instance for which the value is computed.
+      # @return [Object] value
+      #
+      # @api private
+      #
+      def value(serializer)
+        if block
+          serializer.instance_eval(&block)
+        else
+          serializer.read_attribute_for_serialization(name)
+        end
+      end
+
+      # Decide whether the field should be serialized by the given serializer instance.
+      # @param [Serializer] The serializer instance
+      # @return [Bool]
+      #
+      # @api private
+      #
+      def included?(serializer)
+        case condition
+        when :if
+          serializer.public_send(condition)
+        when :unless
+          !serializer.public_send(condition)
+        else
+          true
+        end
+      end
+
+      private
+
+      def condition_type
+        if options.key?(:if)
+          :if
+        elsif options.key?(:unless)
+          :unless
+        else
+          :none
+        end
+      end
+
+      def condition
+        options[condition_type]
+      end
+    end
+  end
+end
--- a/lib/active_model/serializer/field.rb
+++ b/lib/active_model/serializer/field.rb
@@ -25,7 +25,7 @@ module ActiveModel
       # @api private
       #
       def included?(serializer)
-        case condition
+        case condition_type
         when :if
           serializer.public_send(condition)
         when :unless
@@ -38,13 +38,14 @@ module ActiveModel
       private
 
       def condition_type
-        if options.key?(:if)
-          :if
-        elsif options.key?(:unless)
-          :unless
-        else
-          :none
-        end
+        @condition_type ||=
+          if options.key?(:if)
+            :if
+          elsif options.key?(:unless)
+            :unless
+          else
+            :none
+          end
       end
 
       def condition
--- a/lib/active_model/serializer/field.rb
+++ b/lib/active_model/serializer/field.rb
@@ -24,14 +24,14 @@ module ActiveModel
       #
       # @api private
       #
-      def included?(serializer)
+      def excluded?(serializer)
         case condition_type
         when :if
-          serializer.public_send(condition)
-        when :unless
           !serializer.public_send(condition)
+        when :unless
+          serializer.public_send(condition)
         else
-          true
+          false
         end
       end
 
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -1,18 +1,24 @@
+require 'active_model/serializer/field'
+
 module ActiveModel
   class Serializer
     # Holds all the meta-data about an association as it was specified in the
     # ActiveModel::Serializer class.
     #
     # @example
-    #  class PostSerializer < ActiveModel::Serializer
+    #   class PostSerializer < ActiveModel::Serializer
     #     has_one :author, serializer: AuthorSerializer
     #     has_many :comments
     #     has_many :comments, key: :last_comments do
     #       object.comments.last(1)
     #     end
-    #  end
+    #     has_many :secret_meta_data, if: :is_admin?
     #
-    #  Notice that the association block is evaluated in the context of the serializer.
+    #     def is_admin?
+    #       current_user.admin?
+    #     end
+    #   end
+    #
     #  Specifically, the association 'comments' is evaluated two different ways:
     #  1) as 'comments' and named 'comments'.
     #  2) as 'object.comments.last(1)' and named 'last_comments'.
@@ -21,32 +27,13 @@ module ActiveModel
     #    # [
     #    #   HasOneReflection.new(:author, serializer: AuthorSerializer),
     #    #   HasManyReflection.new(:comments)
+    #    #   HasManyReflection.new(:comments, { key: :last_comments }, #<Block>)
+    #    #   HasManyReflection.new(:secret_meta_data, { if: :is_admin? })
     #    # ]
     #
     # So you can inspect reflections in your Adapters.
     #
-    Reflection = Struct.new(:name, :options, :block) do
-      # @api private
-      def value(instance)
-        if block
-          instance.instance_eval(&block)
-        else
-          instance.read_attribute_for_serialization(name)
-        end
-      end
-
-      # @api private
-      def included?(serializer)
-        case condition_type
-        when :if
-          serializer.public_send(condition)
-        when :unless
-          !serializer.public_send(condition)
-        else
-          true
-        end
-      end
-
+    class Reflection < Field
       # Build association. This method is used internally to
       # build serializer's association by its reflection.
       #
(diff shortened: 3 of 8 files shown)


lib/active_model/serializer/attribute.rb
```diff
@@ -0,0 +1,38 @@
...
+          serializer.instance_eval(&block)
+        else
+          serializer.read_attribute_for_serialization(name)
+        end
+      end
+
+      def include?(serializer)
+        case condition_type
+        when :if
+          serializer.send(condition)
```
bf4: `public_send` ?

lib/active_model/serializer/attribute.rb
```diff
@@ -0,0 +1,38 @@
...
+        when :unless
+          !serializer.send(condition)
+        else
+          true
+        end
```
bf4: There's definitely a latent object or interface between the attribute and reflection... to be considered, I think, but doesn't need to be here

bf4:
Looks like a good approach to me.

lib/active_model/serializer/attribute.rb
beauby: Just updated it.

lib/active_model/serializer/attribute.rb
beauby: Agreed, but I'm not too sure what's best currently. I vote for taking care of it later.

bf4: + 100

B mobile phone

lib/active_model/serializer/attribute.rb
```diff
@@ -0,0 +1,38 @@
+module ActiveModel
+  class Serializer
+    Attribute = Struct.new(:name, :block, :key, :condition_type, :condition) do
+      def initialize(name, block, options)
+        super(name, block)
```
bf4: since the `:key, :condition_type, :condition` attributes to an initializer lumped as `options` (which I initially missed), why not just take in options and create our own accessors for key, condition_type, condition?

beauby: That's indeed a possibility. The rationale was to do the work only once.

bf4: By 'work' you mean defining accessors for the options?  Ok, I get that.  I think I'd rather add an attr_accessor than change the Struct definition if I needed to change the options. It took me a couple of reads of the object before I realized you were treating most of the struct attributes as options.  (I'm not new to subclassing the initialize... I just usually use it for setting defaults for unset values, rather than as implicit accessors).

Do you foresee taking advantage of the struct's comparable properties? I'm trying to balance the clever vs. useful in my head.

bf4:
Also ref https://github.com/rails-api/active_model_serializers/issues/1333

lib/active_model/serializer/reflection.rb
```diff
@@ -79,6 +91,20 @@ def build_association(subject, parent_serializer_options)
...
+        elsif options.key?(:unless)
+          :unless
+        else
+          :none
+        end
+      end
+
+      def condition
+        options[condition_type]
+      end
```
bf4: yet another sign there's a common ancestor or module

bf4:
Looking good.. let's finish #1370 and add some tests and docs here :)

lib/active_model/serializer/reflection.rb
beauby: Agreed, we could either make a `Field` class from which `Attribute` and `Reflection` would inheritate, or a `ConditionalField` mixin. Not sure what's best, and naming is obviously up for debate.

bf4:
@beauby #1370 is merged, so let's rebase, add tests, and docs!

lib/active_model/serializer/reflection.rb
bf4: I think regular inheritance makes sense in this case

lib/active_model/serializer/field.rb
```diff
@@ -0,0 +1,40 @@
+module ActiveModel
+  class Serializer
+    Field = Struct.new(:name, :options, :block) do
```
beauby: Kept it as a struct because tests in `test/serializers/association_macros_test.rb` made use of the comparability. Wouldn't mind modifying those in a subsequent PR.

beauby:
PR ready.

lib/active_model/serializer/field.rb
```diff
@@ -0,0 +1,56 @@
+module ActiveModel
+  class Serializer
+    # Holds all the meta-data about a field (i.e. attribute or association) as it was
```
bf4: grammar: `s/i.e./e.g.`

reason: meaning is to change 'that is' to 'for example'

lib/active_model/serializer/field.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+      end
+
+      # Decide whether the field should be serialized by the given serializer instance.
+      # @param [Serializer] The serializer instance
+      # @return [Bool]
+      #
+      # @api private
+      #
+      def included?(serializer)
```
bf4: nitpick, not a blocker:

I think `excluded?` would be better for two reasons

1. included sounds like include
1. the usage e.g. `next unless reflection.included?(self)` and  implementation, which defaults to true, both point to the method's purpose to exclude. 

I'd be ok with just making `def excluded?(serializer); !included?(serializer); end` and using changing the usage code to e.g. `next if attr.excluded?(self)`. I think the `included?` implementation is easier to reason about, but `excluded?` is the better name to use...

lib/active_model/serializer/field.rb
beauby: I actually meant i.e., because I'm just defining what "field" means.

lib/active_model/serializer/reflection.rb
```diff
@@ -1,18 +1,24 @@
...
     #     has_many :comments
     #     has_many :comments, key: :last_comments do
     #       object.comments.last(1)
     #     end
-    #  end
+    #     has_many :secret_meta_data, if: :is_admin?
+    #
+    #     def is_admin?
+    #       current_user.admin?
+    #     end
```
bf4: I like how you slipped the scope in there :) ref: https://github.com/rails-api/active_model_serializers/pull/1252

bf4:
@beauby two small comments, one grammar, one internal api. :100: LGTM

lib/active_model/serializer/field.rb
beauby: Agreed that `excluded?` is better and less confusing. Updating.

lib/active_model/serializer/field.rb
bf4: still think it should be e.g., but since it's intentional on your part and it's super unimportant, ok to merge when green

test/serializers/attribute_test.rb
```diff
@@ -95,6 +95,29 @@ def test_virtual_attribute_block
...
+          def false
+            false
+          end
+        end
+
+        model = ::Model.new
+        hash = serializable(model, serializer: serializer).serializable_hash
+        expected = { if_attribute_included: nil, unless_attribute_included: nil }
+
+        assert_equal(expected, hash)
```
bf4: we should look into if rubocop has a linter for parens around assert.. hashtag lazy comment

lib/active_model/serializer/field.rb
beauby: Agreed that it's not super important, but as I like grammar as well: a field is *defined* as "either an attribute or an association".

bronson:
Is there a plan for updating the docs?  Wondering if there's a reason for not updating them with the code.  (like, maybe they're going to be rewritten anyway so there's no point?)

Just curious.

bronson:
Looks like this doesn't work?

```ruby
attributes Property::PropertyParams, if: :is_property?
```

It does work with singular `attribute`...  so, if I use a conditional, I can only operate one attribute at a time?

bronson:
Also, allowing `if: true` and `if: false` would be nice for debugging and consistency.  Right now they raise `true is not a symbol nor a string`.

beauby:
Hi @bronson. I have "writing docs for conditional fields" on my TODO list, but I'm fairly busy at the moment, and since people seemed eager to benefit from the feature I figured it was worth merging (instead of waiting a few days/weeks for the docs to be written, then rebasing, etc.). I'd absolutely love for someone else to issue a PR though :dancers: 
> So, using conditionals requires me to operate one attribute at a time?
You are right, currently there is no forwarding of options when defining multiple attributes at a time (it happens [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/attributes.rb#L36)).
> Also, allowing `if: true` and `if: false` would be nice for debugging and consistency. Right now they raise `true is not a symbol nor a string`.
I'm not sure about this. I could see the benefit for debugging, but it would add unnecessary complexity IMO. It should be sufficient to test the actual condition methods. Do you have a specific case in mind where this would be necessary?

beauby:
I'd be much more in favor of allowing either a symbol (as now) *or a lambda* as a value for the `if/unless` options, which would allow for `attribute :title, if: -> { true }` in your case.

bronson:
1. Right on, agree with making stuff available soon (as long as it's reasonably thought out).  Glad you have docs in mind.

2. The lack of forwarding for multiple attributes, is that by design or just omission?  Just curious.

3. Well, `if: true` was the first thing I tried when I saw things weren't working.  If it's really difficult to implement then, sure, a lambda should work just fine.  It doesn't really feel like :if elsewhere in Rails though.

Thanks for writing this!  Definitely looking forward to using it.

beauby:
2. There has been talks about it in the past, but there was no option that would really make sense for multiple attributes back then (`key` only applies to one attribute). I wouldn't be opposed to a PR for this (it's just a matter of 2 lines within `AM::Serializer.attributes`).
3. You're right, the name is probably not best, but I couldn't think of anything more suitable. The idea (which will appear clearly in the docs) is that `if/unless` take a symbol referring to a method.

beauby:
@bronson Note that, in the meantime, if you have several conditional attributes, you could do the following:
```ruby
[:name, :birthday, :address, :phone].each { |attr| attribute attr, if: :condition? }
def condition?
  ...
end
```

Mohakjuneja:
Is the support for if/unless already implemented in the master? I don't want my subjects association to be included when the controller method is index
Here's what I tried-

  
```
class Admin::SemesterSerializer < ActiveModel::Serializer

  attributes :_id, :semester, :course_id

  has_many :subjects, serializer: Admin::SubjectSerializer, unless: :index

end
```
But I get the error undefined method `index' for #<Admin::SemesterSerializer

NullVoxPopuli:
@Mohakjuneja the serializer doesn't have access to the controller's methods.

Have you seen these docs? https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#scope

maybe they could help?

Mohakjuneja:
I figured it out! I apologize for being impatient and posting the issue too soon. AMS has consumed a lot of my time upgrading from 0.9.3 to the master. Thank you!

NullVoxPopuli:
what did you do? (for those who may come by this issue in the future with the same problem?)

Mohakjuneja:
I added the `index: true` part to the end of my controller method

`respond_with semesters.where(:semester_num.lte => current_user.current_semester).desc(:semester_num), each_serializer: Admin::SemesterSerializer, index: true
`

And in the serializer, I accessed it using the `instance_options[:index]` method

bf4:
@Mohakjuneja 

>  AMS has consumed a lot of my time upgrading from 0.9.3 to the master. Thank you!

You would win many friends and much karma if you could help us write an updating guide! ref: https://github.com/rails-api/active_model_serializers/issues/1005

Mohakjuneja:
@bf4 I'm starting up my own venture (wish me luck!) but I'll be glad to help you with it! This community has been very supportive :)

jDeppen:
I was able to determine the view by adding this to my controller
```serialization_scope :view_context```
and this to my serializer
```
[:name, :birthday, :address, :phone].each { |attr| attribute attr, if: :show? }
def show?
  view_context.action_name == 'show'
end
```

**Comment by Mohakjuneja** [comment]
It stopped working all of a sudden
I had to make changes to my serializer to make it work again. Are we still not fixed on the syntax here?

2 days ago-
`  attribute :total, :subject_name if: :view_uploads?
`

Now-
`  attributes :total, :subject_name if :view_uploads?
`

attribute to attributes AND if: to if

**Comment by Mohakjuneja** [comment]
It stopped working all of a sudden.
throws error at the if: symbol now on this line

```
SyntaxError (/home/mojo/Documents/Rails/hopin/app/serializers/api/v1/content_serializer.rb:5: syntax error, unexpected ':'
  
attribute :total, :subject_name if: :view_uploads?
                                     ^):
  app/serializers/api/v1/content_serializer.rb:5: syntax error, unexpected ':'

```

What just happened?

**Comment by NullVoxPopuli** [comment]
You need a comma before your `:if` :-)

**Comment by Mohakjuneja** [comment]
Hey @NullVoxPopuli , that doesn't work either.
```

  def view_uploads?
    instance_options[:uploads]
  end
```

The instance_options[:uploads] and view_uploads? is receiving and returning the value I'm passing from the controller.
The if/unless for attribute seems to have broken

**Comment by NullVoxPopuli** [comment]
`attributes` can be used for multiple attributes. :-)

**Comment by Mohakjuneja** [comment]
`  attributes :total, :subject_name, if: :view_uploads?
`
TypeError ({:if=>:view_uploads?} is not a symbol nor a string)

I'm sorry about taking so much of your time but I've tried everything

**Comment by NullVoxPopuli** [comment]
so maybe you do need to define each attribute on it's own line with it's own if condition, so:

```
attribute :total, if: :view_uploads?
attribute :total, if: :view_uploads?
```

**Comment by Mohakjuneja** [comment]
I'll do that for now but I don't find that a very effective solution. We may need to take a look at it. Also I don't know why it worked two days ago but doesn't work anymore. Thank you @NullVoxPopuli :)

**Comment by bf4** [comment]
@Mohakjuneja that's what version control is for :)   Let me know if I'm summarizing this correctly (ignoring syntax errors), then please open a new issue if we need further discussion.

works

```ruby
attribute :total, if: :view_uploads?
attribute :subject_name, if: :view_uploads?
```

desired, doesn't work

```ruby
attributes :total, :subject_name, if: :view_uploads? 
```

Mohakjuneja:
@bf4 I've been thinking about the current syntax and I'm wondering if we should stick to it.
In cases where you want a certain attribute to be present in multiple cases but not all, this makes more sense.
for instance
```

attribute :total, if: [:view_uploads?, edit_content]
attribute :subject_name, if: :view_uploads?
```

What is your opinion about that?

beauby:
I'd much rather allow lambdas as `if/unless` values than introduce a cumbersome "array of method name symbols" syntax (which wouldn't make it obvious whether it is a conjunction or a disjunction).

MajedBojan:
the attributes doesn't work with scope if your shown multiple attributes the correct way put it as arry
```
%w[attribute1 attribute2 attribute3].each { |att| attribute att, if: :condition? }
 ```

PR #1405, opened 2015-12-28
Fix JRruby at 9.0.4.0 since TravisCI is using pre1 for some reason

bf4:


.travis.yml
```diff
@@ -8,8 +8,13 @@ rvm:
   - 2.2.3
   - 2.3.0
   - ruby-head
+  - jruby-9.0.4.0
   - rbx-2
 
+before_install:
+  - '[ "$JRUBY_OPTS" != "" ] && export JRUBY_OPTS="-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug"'
+  - echo "JRUBY_OPTS: '$JRUBY_OPTS'"
```
groyoh: Try:

```yaml
  - 'echo "JRUBY_OPTS: ''$JRUBY_OPTS''"'
```

`:` can't be used in plain style and when using single-quoted style single quote must be escaped with `''`.

bf4: @groyoh Thanks. Any tips on why the build is failing. (Also, I just remembered I should see about adding jruby-9k to the appveyour build)

Issue #1400, opened 2015-12-27
[RFC] Option to decide whether a "link" should be a string or object


groyoh:
JSONAPI spec speficy that a `link` can either a string or an object (http://jsonapi.org/format/#document-links). Currently when using `ActiveModel::Serializer::link` method, we cannot specify whether the link should be serialized as a string or an object.
If we don't use a block, then it is always serialized as a string:
```ruby
link :self, "//a-url" # => { :self => "//an-url" }
```
and if we use the block version, then it is always serialized as an object:
```ruby
link :self do
  href "//a-url"
end # => { :self => { :href => "//an-url" } }
```

I see some cases where we might want to use a block (to have a dynamic href) but still want it to be serialized as a string (since it is allowed by the spec) or not use a block and serialize it as an object.

I think an option parameter to the `Serializer::link` method could fix this e.g.
```ruby
link :self, "//an-url", as_string: false # => { :self => { :href => "//an-url" } }
link :self, as_string: true do
  href "//an-url"
end # => { :self => "//an-url" }
```

beauby:
You make a valid point, although I'm not convinced by the `as_string` option. We could possibly add a `string` method to the `Link` class, which would force the value to be a string.

bf4:
I think a config option would be sufficient.  The default would be: If a link has only one item in it, then the value is a string.  Else, it's an object. And configurable would be: Always use an object.  Also, passing in 'self: href' would always result in an object

groyoh:
@beauby I thought about this too but then realized that it would prevent us from serializing as an object when not using a block e.g. `link :self, "//an-url" # => { :self => { :href => "//an-url" } }`. In my opinion, we should able to decide how to serialize it whether we use a block a not for consistency reason.

beauby:
@groyoh Not really, as you could currently do `link :self, { href: '//an-url' }`.

groyoh:
@beauby you're right. I forgot about this even though I'm not convinced it's the best idea to be able to pass in a hash directly. If I were to choose I'd rather only pass a string and then AMS would take care of serializing as string or a hash. I also find it less error prone (e.g. no chance of typo when writting `href`).

beauby:
@groyoh That's why we introduced the DSL for link objects. I'd like to address as many use-cases as possible but at the same time I don't want to introduce too much complexity and confusion.

groyoh:
@beauby alright, so in the end would it be like this:

```ruby
[...]

PR #1406, opened 2015-12-29
Add support for dynamic string-links in JsonApi adapter.

beauby:
Solves #1400.

### DIFF
--- a/lib/active_model/serializer/adapter/json_api/link.rb
+++ b/lib/active_model/serializer/adapter/json_api/link.rb
@@ -3,29 +3,39 @@ module ActiveModel
     module Adapter
       class JsonApi
         class Link
-          def initialize(serializer)
+          def initialize(serializer, value)
             @object = serializer.object
             @scope = serializer.scope
+
+            # Use the return value of the block unless it is nil.
+            if value.respond_to?(:call)
+              @value = instance_eval(&value)
+            else
+              @value = value
+            end
           end
 
           def href(value)
-            self._href = value
+            @href = value
+            nil
           end
 
           def meta(value)
-            self._meta = value
+            @meta = value
+            nil
           end
 
-          def to_hash
-            hash = { href: _href }
-            hash.merge!(meta: _meta) if _meta
+          def as_json
+            return @value if @value
+
+            hash = { href: @href }
+            hash.merge!(meta: @meta) if @meta
 
             hash
           end
 
           protected
 
-          attr_accessor :_href, :_meta
           attr_reader :object, :scope
         end
       end
--- a/test/adapter/json_api/links_test.rb
+++ b/test/adapter/json_api/links_test.rb
@@ -13,6 +13,10 @@ module ActiveModel
             end
 
             link :other, '//example.com/resource'
+
+            link :yet_another do
+              "//example.com/resource/#{object.id}"
+            end
           end
 
           def setup
@@ -52,7 +56,8 @@ module ActiveModel
                   stuff: 'value'
                 }
               },
-              other: '//example.com/resource'
+              other: '//example.com/resource',
+              yet_another: '//example.com/resource/1337'
             }
             assert_equal(expected, hash[:data][:links])
           end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Breaking changes:
 
 Features:
 
+- [#1406](https://github.com/rails-api/active_model_serializers/pull/1406) Allow for custom dynamic values in JSON API links (@beauby)
 - [#1270](https://github.com/rails-api/active_model_serializers/pull/1270) Adds `assert_response_schema` test helper (@maurogeorge)
 - [#1099](https://github.com/rails-api/active_model_serializers/pull/1099) Adds `assert_serializer` test helper (@maurogeorge)
 - [#1403](https://github.com/rails-api/active_model_serializers/pull/1403) Add support for if/unless on attributes/associations (@beauby)
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -210,15 +210,7 @@ module ActiveModel
 
         def links_for(serializer)
           serializer._links.each_with_object({}) do |(name, value), hash|
-            hash[name] =
-              if value.respond_to?(:call)
-                link = Link.new(serializer)
-                link.instance_eval(&value)
-
-                link.to_hash
-              else
-                value
-              end
+            hash[name] = Link.new(serializer, value).as_json
           end
         end
 


test/adapter/json_api/links_test.rb
```diff
@@ -13,6 +13,10 @@ class LinkAuthorSerializer < ActiveModel::Serializer
             end
 
             link :other, '//example.com/resource'
```
bf4: Off topic: given that [JSON API is undecided on relative links](https://github.com/json-api/json-api/issues/898), I'd rather not fill our examples with them

bf4:
@beauby Made a PR into this suggesting a `Link` interface more similar to what we've been doing in Reflection and  https://github.com/beauby/active_model_serializers/pull/2Attribute

test/adapter/json_api/links_test.rb
beauby: Agreed, however this is not a relative link. `//resource` would be undefined, but `//example.com/resource` is a valid url.

bf4: @beauby words... words.. I meant 'protocol relative'.  Sorry I typo'ed again :(

beauby: I don't feel strongly about this, although it is not clear from the discussion whether they are disallowing relative urls or protocol-relative urls.

bf4: Yeah, I just updated my question to clarify that

lib/active_model/serializer/adapter/json_api/link.rb
```diff
@@ -3,30 +3,43 @@ class Serializer
...
             self._href = value
+            nil
           end
 
           def meta(value)
             self._meta = value
+            nil
           end
 
+          def serializable_hash
```
groyoh: I'm not sure if `serializable_hash` is a good name considering it can be a string. What's your point on this?

beauby: Indeed, it is not a perfect name. The rationale was that `serializable_hash` is already used all around in AMS for objects that are "ready to be transformed into JSON", i.e. no further treatment is required (used in `SerializableResource.new(resource).serializable_hash`, PaginationLink.new(stuff).serializable_resource`).

bf4: @groyoh @beauby What not have a `value` method like our other 'value' objects, `Reflection`, `Attribute`..?

beauby: @bf4 I'm not satisfied with `value` but you're right, it's probably better to keep it consistent for now and possibly update everything in a subsequent commit.

lib/active_model/serializer/adapter/json_api/link.rb
```diff
@@ -3,30 +3,43 @@ class Serializer
...
       class JsonApi
         class Link
-          def initialize(serializer)
+          def initialize(serializer, value)
             @object = serializer.object
             @scope = serializer.scope
+            if value.respond_to?(:call)
+              self._value = instance_eval(&value)
+            else
+              self._value = value
```
bf4: What's our desired behavior if someone does

```ruby
link :foo do
  meta 'thing'
  'https://example.com/doohickie
end
```

?

right now it would only return the block's return value

bf4: We should also probably comment that in the block form, if `meta` and/or `href` are used, the `Link` is an object.  If neither are used, the `Link` is a URI (the return value of the block).  If both are used, only the return value?

beauby: The way I see it, providing a return value to the block should be interpreted as "I'm a big boy and this is my link value". Otherwise, "building" a link object through the DSL should be done, well, via the DSL. What are your thoughts?

beauby: Agreed that it should be stated clearly.

lib/active_model/serializer/adapter/json_api/link.rb
bf4: I think it mainly looks weird because we're passing value into.  Maybe just have `as_json` and `to_json`?

bf4: (but in a follow-up PR)

beauby: `as_json` is the way to go IMO (for the record, `as_json` means build something ready to be directly transformed into JSON, i.e. a Hash mainly, but we could extend the meaning to be "something that will eventually be part of a Hash representing a JSON document", and `to_json` means actually building a JSON string from the object).

bf4: :+1:

lib/active_model/serializer/adapter/json_api/link.rb
bf4: So, let's update and warn against the scenario in https://github.com/rails-api/active_model_serializers/pull/1406/files#r48668245 and ok to merge

bf4: with the `serializable_hash` to `as_json` change in https://github.com/rails-api/active_model_serializers/pull/1406/files#r48702394

beauby:
All concerns taken care of. I'll merge later today unless someone raises other concerns.

lib/active_model/serializer/adapter/json_api/link.rb
```diff
@@ -3,30 +3,45 @@ class Serializer
...
           end
 
+          protected
+
+          attr_accessor :_href, :_meta, :_value
+          attr_reader :object, :scope
+
+          private
+
           def to_hash
```
bf4: `block_value || hash_value` ? `to_hash` doesn't really jive well with how I think of the method

bf4: granted, this is internal api, so just talking about naming

bf4:
@beauby Please add changelog and docs, and it's good to go.  I don't really like the internals of the `Link` object, particularly the naming of `_value` vs. `to_hash`-- I don't find them very descriptive-- but I can live with it.

lib/active_model/serializer/adapter/json_api/link.rb
```diff
@@ -3,30 +3,45 @@ class Serializer
...
       class JsonApi
         class Link
-          def initialize(serializer)
+          def initialize(serializer, value)
             @object = serializer.object
             @scope = serializer.scope
+
+            # Use the return value of the block unless it is nil.
+            if value.respond_to?(:call)
+              self._value = instance_eval(&value)
```
groyoh: Could you please explain why you used an `attr_accessor` for `_value` instead of using a reader and calling `@_value = instance_eval(&value)`? I don't see why `_value` is different than `object` or `scope`.

beauby: You're right, there's no reason. There just seemed to be a trend (which I kinda disapprove actually) in AMS to move towards accessors rather than ivars. I'll update it.

bf4: agreed it should be a naked instance variable here

my general thinking: setters and getters are, all things being equal, a better way to handle accessing instance values, since it protects the code against changes and typos.  However, there's no need to message dispatch when setting the value in an initializer, so a reader-only is fine there.  As to eschewing accessors all together, I'd probably only do that for performance reasons, or when maybe when I'm certain there code won't change often so mis-typing a variable name is unlikely.

bf4:
@beauby squash commits, add changelog and good to merge.  I want to release rc4 soon so am trying to get whatever prs with good features that can be merged, in.

beauby:
PR ready, waiting for Travis.

PR #1410, opened 2015-12-31
fix broken link

georgemillo:
previously if you clicked it while reading the MD doc on github it would send you to https://github.com/rails-api/active_model_serializers/blob/master/docs/jsonapi.org, which obviously doesn't exist. Adding 'http://' fixes it


bf4:
Awesome!

PR #1393, opened 2015-12-23
Setup benchmark testing

bf4:
Based on https://github.com/rails-api/active_model_serializers/pull/872

## Update March 8, 2016

PR now contains:

- Setup dummy app files in `test/dummy`
- Setup dummy test server `bin/serve_dummy
  - Note:  Serializer caching can be completely disabled by passing in
  `CACHE_ON=off bin/serve_dummy start` since Serializer#_cache is only
  set at boot.
- run with
  - ./bin/bench
  - `bin/bench` etc adapted from ruby-bench-suite
  - target files are `test/dummy/bm_*.rb`. Just add another to run it.
  - benchmark cache/no cache
  - remove rake dependency that loads unnecessary files
- remove git gem dependency
  - Running over revisions to be added in subsequent PR

- Removed
  -  [regression runner (across revisions)](https://github.com/rails-api/active_model_serializers/blob/d0845248ced0297341ed4d5cf4a31f2bcfa026fd/bin/bench_regression)
  - [FakeIPS](https://github.com/rails-api/active_model_serializers/blob/d0845248ced0297341ed4d5cf4a31f2bcfa026fd/test/dummy/benchmarking_support.rb#L17-L152) for running without benchmark-ips
  - [derailed benchmarks](https://github.com/rails-api/active_model_serializers/blob/d0845248ced0297341ed4d5cf4a31f2bcfa026fd/tasklib/perf.rake) [gemfile](https://github.com/rails-api/active_model_serializers/blob/d0845248ced0297341ed4d5cf4a31f2bcfa026fd/Gemfile#L39-L45)
  - [rack-mini-profiler](https://github.com/rails-api/active_model_serializers/blob/d0845248ced0297341ed4d5cf4a31f2bcfa026fd/test/dummy/app.rb#L57-L63)

## Update Feb 12, 2016

`bin/bench` runs files in `test/dummy` that start with `bm_`. type `bin/bench --help` for more on the options and see `bm_caching.rb` for an example implementation.  Based on the ruby-bench-suite

```plain
bin/bench -e
caching on: caching serializers: gc off 805.7643471872851/ips   # turning gc on didn't seem to make a difference
caching off: caching serializers: gc off 774.1753998716134/ips
caching on: non-caching serializers: gc off 974.3776823158275/ips
caching off: non-caching serializers: gc off 875.7715343734802/ips
Benchmark results:
{
  "commit_hash": "fb376b7",
  "version": "0.10.0.rc4",
[...]

PR #1412, opened 2016-01-05
Setup dummy test server

bf4:
For use in https://github.com/rails-api/active_model_serializers/pull/1393

### DIFF
--- a/bin/revision_runner
+++ b/bin/revision_runner
@@ -120,6 +120,7 @@ class RevisionRunner
     bundle
     base = Shellwords.shellescape(TMP_DIR_BASE)
     cmd = "BASE=#{base} #{Shellwords.shelljoin(cmd)}"
+    cmd.sub('bin/bench', 'tmp/revision_runner/bench')
     debug { cmd }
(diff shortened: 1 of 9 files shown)


bin/revision_runner
```diff
@@ -0,0 +1,162 @@
...
+require 'fileutils'
+require 'pathname'
+require 'shellwords'
+require 'English'
+require 'net/http'
+require 'json'
+
+############################
+#
+# A wrapper around git-bisect that stashes some code (bin/bench,test/dummy/*) for use in making requests
```
bf4: not true right now, but I think is a good idea

bf4:
Note: this PR has code in it that [extends past its original mandate](https://github.com/bf4/active_model_serializers/compare/65093055d6b47c6a22a1f265c56a4f694eb80d78...af78396f7d55c8143a9b60cac1f26263d071f4d3).. I was just working on the problem and decided to push it up for review, even though it's not really part of it and should be in https://github.com/rails-api/active_model_serializers/pull/1393

PR #1418, opened 2016-01-07
Don't pluralize the CollectionSerializer#root for #json_key

Empact:
One of three constituents is used to provide the CollectionSerializer's #json_key:
1) the :root option - controlled by the caller
2) the #name of the first resource serializer - the root or
   underscored model name
3) the underscored #name of the resources object - generally
   equivalent to the underscored model name of #2

Of the three, only the latter 2 are out of the callers control, and
only the latter two are expected to be singular by default. Not
pluralizing the root gives the caller additional flexibility in
defining the desired root, whether conventionally plural,
unconventionally plural (e.g. objects_received:) or singular.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Breaking changes:
         Adapter functions.
     * named `Base` because it's a Rails-ism.
     * It helps to isolate and highlight what the Adapter interface actually is.
+- [#1418](https://github.com/rails-api/active_model_serializers/pull/1418)
+  serialized collections now use the root option as is; now, only the
+  root derived from the serializer or object is always pluralized.
 
 Features:
 
--- a/lib/active_model/serializer/collection_serializer.rb
+++ b/lib/active_model/serializer/collection_serializer.rb
@@ -23,8 +23,7 @@ module ActiveModel
       end
 
       def json_key
-        key = root || serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
+        root || derived_root
       end
 
       def paginated?
@@ -36,6 +35,13 @@ module ActiveModel
       protected
 
       attr_reader :serializers
+
+      private
+
+      def derived_root
+        key = serializers.first.try(:json_key) || object.try(:name).try(:underscore)
+        key.try(:pluralize)
+      end
     end
   end
 end
--- a/test/action_controller/serialization_test.rb
+++ b/test/action_controller/serialization_test.rb
@@ -171,7 +171,7 @@ module ActionController
         with_adapter :json do
           get :render_array_using_custom_root
         end
-        expected = { custom_roots: [{ name: 'Name 1', description: 'Description 1' }] }
+        expected = { custom_root: [{ name: 'Name 1', description: 'Description 1' }] }
         assert_equal 'application/json', @response.content_type
         assert_equal expected.to_json, @response.body
       end
@@ -181,7 +181,7 @@ module ActionController
           get :render_array_that_is_empty_using_custom_root
         end
 
-        expected = { custom_roots: [] }
+        expected = { custom_root: [] }
         assert_equal 'application/json', @response.content_type
         assert_equal expected.to_json, @response.body
       end
--- a/test/collection_serializer_test.rb
+++ b/test/collection_serializer_test.rb
@@ -62,8 +62,9 @@ module ActiveModel
         assert_equal expected, @serializer.root
       end
 
-      def test_json_key
-        assert_equal 'comments', @serializer.json_key
+      def test_json_key_with_resource_with_serializer
+        singular_key = @serializer.send(:serializers).first.json_key
+        assert_equal singular_key.pluralize, @serializer.json_key
       end
 
       def test_json_key_with_resource_with_name_and_no_serializers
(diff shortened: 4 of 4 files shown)


test/action_controller/serialization_test.rb
```diff
@@ -171,7 +171,7 @@ def test_render_array_using_custom_root
         with_adapter :json do
           get :render_array_using_custom_root
         end
-        expected = { custom_roots: [{ name: 'Name 1', description: 'Description 1' }] }
+        expected = { custom_root: [{ name: 'Name 1', description: 'Description 1' }] }
```
bf4: Technically this is a breaking change.  I think this is ok since we still haven't released 0.10 but a note in the changelog would be good

test/collection_serializer_test.rb
```diff
@@ -47,50 +47,53 @@ def test_serializer_option_not_passed_to_each_serializer
...
         resource.define_singleton_method(:name) { nil }
         serializer = collection_serializer.new(resource)
-        assert_equal serializer.json_key, nil
+        assert_nil serializer.json_key
       end
 
       def test_json_key_with_resource_without_name_and_no_serializers
         serializer = collection_serializer.new([])
-        assert_equal serializer.json_key, nil
+        assert_nil serializer.json_key
```
bf4: Why this change? Besides being out of scope, it's a better failure failure message 'expected foo to be nil but what bar' over  'expected foo to be nil',no?

bf4: (though you are right that a lot of the assertions are backwards, should be assert expected, actual)

bf4:
You wrote:

> expected to be singular

Would you mind pointing me to where the JSON adapter collection root is expected to be singular?

Indeed for JSON API this is configurable, but not for the JSON adapter.

test/collection_serializer_test.rb
Empact: Yeah, I made a commit fixing the assertions because most of the issues I found were in the test file I updated. `assert_nil` is equivalent to `assert_equal` with nil as the expectation, so the failure message should be the same. http://apidock.com/ruby/Test/Unit/Assertions/assert_nil

Doing the changes first ensured there wouldn't be conflicts between the two sets of changes, but I'm happy split them out for easier consideration.

Empact:
@bf4 I said

> Of the three, only the latter 2 are out of the callers control, and
> only the latter two are expected to be singular by default.

the latter 2 are expected to be singular (unless overriden), because they are drawn from the classes that are not collection-oriented - either the `json_key` of a non-collection serializer, or the name of the collection object - which is singular assuming this is geared toward rails where:

```ruby
>> Profile.limit(10).name
=> "Profile"
```

I'm not sure what you mean by `JSON adapter collection root`.

test/collection_serializer_test.rb
Empact: #1423

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -23,8 +23,11 @@ def initialize(resources, options = {})
       end
 
       def json_key
-        key = root || serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
+        root || begin
+          singular_key = serializers.first.try(:json_key) ||
+            object.try(:name).try(:underscore)
+          singular_key.try(:pluralize)
+        end
```
bf4: what a mess this code is.  maybe have this be something like `root || derived_root` so it's a little easier to see what's going on?  I initially missed that the fundamental change was no longer modifying `root` when present

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ Breaking changes:
         Adapter functions.
     * named `Base` because it's a Rails-ism.
     * It helps to isolate and highlight what the Adapter interface actually is.
+- [#1418](https://github.com/rails-api/active_model_serializers/pull/1418)
+  When generating a collection serialization's `json_key`, the `root:`
+  option is no longer pluralized. This enables you to generate any
```
bf4: Maybe

> serialized collections now use the `root` option as is; now, only the root derived from the serializer or object is always pluralized.

(Makes me think we should allow this to be configured like in JSON API)

bf4:
Looks good. Just some small comments

Empact:
@bf4 Should be good to go

test/action_controller/serialization_test.rb
bf4: I can see this breaking a lot of apps.. I'm wondering how we can communicate that?  @rails-api/ams

natematykiewicz:
@bf4, I know you said this could break a lot of apps. Do you think it's worth creating a setting for it, or just let it be part of upgrading to 0.10?

bf4:
@natematykiewicz good question.  @joaomdmoura how about after we merge this we immediately cut rc4?

Empact:
Incidentally the other changes I've been putting together so far have been additive only. I think an RC to socialize this would be good.

beauby:
LGTM. For a future PR, it might be worth extracting most of the `root` / `json_key` logic into the `Attributes` adapter, as it is the only one actually making use of it (`JsonApi` uses a custom `type` option, which could very well be just an alias to root actually).

Empact:
Interesting - incidentally I also have a branch adding the `:except` option to `Attributes` only.

https://github.com/brigade/active_model_serializers/tree/attributes-except

test/action_controller/serialization_test.rb
kurko: It's in the changelog under `breaking changes` so I think it's good now.

PR #1419, opened 2016-01-07
Don't require serializable objects to extend ActiveModel::Naming

Empact:
The aspects of `#model_name` used here are all dependent on the class
alone, we don't need the full namespaced model name to produce
proper results.

This change takes advantage of `#model_name` if defined on the object,
but otherwise provides a name based on the class alone. This saves
us a bunch of:

```ruby
  class Foo
    extend ActiveModel::Naming
    include ActiveModel::Serialization
  end
```

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -135,7 +135,7 @@ module ActiveModel
 
     # Used by adapter as resource root.
     def json_key
-      root || object.class.model_name.to_s.underscore
+      root || model_name.to_s.underscore
     end
 
     def read_attribute_for_serialization(attr)
@@ -152,6 +152,16 @@ module ActiveModel
     # Used by JsonApi adapter to build resource links.
     def links
       self.class._links
+    end
+
+    # @api private
+    # Returns an ActiveModel::Name for the serializer's object
+    def model_name
+      if object.class.respond_to?(:model_name)
+        object.class.model_name
+      else
+        ActiveModel::Name.new(object.class)
+      end
     end
 
     protected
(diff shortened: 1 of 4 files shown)


lib/active_model/serializer.rb
```diff
@@ -154,6 +154,16 @@ def links
       self.class._links
     end
 
+    # @api private
+    # Returns an ActiveModel::Name for the serializer's object
+    def model_name
```
bf4: This is a really interesting change. Would you mind adding some tests to show how serializing an object without a `model_name` behaves?

bf4: btw, have you seen https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/model.rb and that supporting `[ActiveModel::Errors](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/errors.rb)` (ref #1004) would require the model extend `ActiveModel::Name`?  thoughts on that (after adding tests to show the effect of this change)?

Empact:
I"m going to drop this because I think you're right that Errors, etc will make the requirement necessary. Thanks for taking a look.

bf4:
@Empact thanks for this.  I hope you stay involved and even join the [slack](https://github.com/rails-api/active_model_serializers/blob/master/CONTRIBUTING.md)!  Would you be interested in working on anything in particular?

Empact:
@bf4 Currently I'm working on getting an existing site outputting the same json as under 0.8.0, which is driving most of the changes. Should be very helpful in terms of giving people an easier upgrade path without dropping legacy clients or starting over.

bf4:
Yes yes yes!

B mobile phone

PR #1420, opened 2016-01-09
Adds tests and documentation for polymorphism

marcgarreau:
#### What's this PR do?
Adds baseline documentation for using polymorphic relationships within AMS.
EDIT: Added adapter tests, because there were none.
#### Any background context you want to provide?
I was unable to find any documentation within the docs about polymorphic relationships. Happy to move this documentation to another location or add to it if its inadequate.

### DIFF
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -76,6 +76,18 @@ def blog
 end
 ```
 
+### Polymorphic Relationships
+
+Polymorphic relationships are serialized by specifying the relationship, like any other association. For example:
+
+```ruby
+class PictureSerializer < ActiveModel::Serializer
+  has_one :imageable
+end
+```
+
+For more context, see the [tests](../../test/adapter/polymorphic_test.rb) for each adapter.
+
 ### Caching
 
 #### ::cache
--- a/test/adapter/polymorphic_test.rb
+++ b/test/adapter/polymorphic_test.rb
@@ -1,0 +1,72 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
+      class PolymorphicTest < ActiveSupport::TestCase
+        setup do
+          @employee = Employee.new(id: 42, name: 'Zoop Zoopler', email: 'zoop@example.com')
+          @picture = @employee.pictures.new(id: 1, title: 'headshot-1.jpg')
+          @picture.imageable = @employee
+
+          @attributes_serialization = serializable(@picture, serializer: PolymorphicBelongsToSerializer) # uses default adapter: attributes
+          @json_serialization = serializable(@picture, adapter: :json, serializer: PolymorphicBelongsToSerializer)
+          @json_api_serialization = serializable(@picture, adapter: :json_api, serializer: PolymorphicBelongsToSerializer)
+        end
+
+        def test_attributes_serialization
+          expected =
+            {
+              id: 1,
+              title: 'headshot-1.jpg',
+              imageable: {
+                id: 42,
+                name: 'Zoop Zoopler'
+              }
+            }
+
+          assert_equal(expected, @attributes_serialization.as_json)
+        end
+
+        def test_json_serializer
+          expected =
+            {
+              picture: {
+                id: 1,
+                title: 'headshot-1.jpg',
+                imageable: {
+                  id: 42,
+                  name: 'Zoop Zoopler'
+                }
+              }
+            }
+
+          assert_equal(expected, @json_serialization.as_json)
+        end
+
+        def test_json_api_serializer
+          expected =
+            {
+              data: {
+                id: '1',
+                type: 'pictures',
+                attributes: {
+                  title: 'headshot-1.jpg'
+                },
+                relationships: {
+                  imageable: {
+                    data: {
+                      id: '42',
+                      type: 'employees'
+                    }
+                  }
+                }
+              }
+            }
+
+          assert_equal(expected, @json_api_serialization.as_json)
+        end
+      end
+    end
+  end
+end
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -72,6 +72,14 @@ Comment = Class.new(Model) do
   end
 end
 
+class Employee < ActiveRecord::Base
+  has_many :pictures, as: :imageable
+end
+
+class Picture < ActiveRecord::Base
+  belongs_to :imageable, polymorphic: true
+end
+
 module Spam; end
 Spam::UnrelatedLink = Class.new(Model)
 
@@ -231,6 +239,16 @@ VirtualValueSerializer = Class.new(ActiveModel::Serializer) do
 
   def maker
   end
+end
+
+PolymorphicHasManySerializer = Class.new(ActiveModel::Serializer) do
+  attributes :id, :name
+end
+
+PolymorphicBelongsToSerializer = Class.new(ActiveModel::Serializer) do
+  attributes :id, :title
+
+  has_one :imageable, serializer: PolymorphicHasManySerializer
 end
 
 Spam::UnrelatedLinkSerializer = Class.new(ActiveModel::Serializer) do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Misc:
 - [#1535](https://github.com/rails-api/active_model_serializers/pull/1535) Move the adapter and adapter folder to
   active_model_serializers folder and changes the module namespace. (@domitian @bf4)
 - [#1497](https://github.com/rails-api/active_model_serializers/pull/1497) Add JRuby-9000 to appveyor.yml(@corainchicago)
+- [#1420](https://github.com/rails-api/active_model_serializers/pull/1420) Adds tests and documentation for polymorphism(@marcgarreau)
 
 ### v0.10.0.rc4 (2016/01/27 11:00 +00:00)
 Breaking changes:
--- a/test/fixtures/active_record.rb
+++ b/test/fixtures/active_record.rb
@@ -18,6 +18,17 @@ ActiveRecord::Schema.define do
     t.references :post
     t.timestamp null: false
   end
+  create_table :employees, force: true do |t|
+    t.string :name
+    t.string :email
+    t.timestamp null: false
+  end
+  create_table :pictures, force: true do |t|
+    t.string :title
+    t.string :imageable_type
+    t.string :imageable_id
+    t.timestamp null: false
+  end
 end
 
 module ARModels


beauby:
Hi @marcgarreau, thanks for the PR! Could you specify which version of AMS this applies to? I believe in 0.10/master polymorphic associations are handled automatically, and the `polymorphic` option does not exist.

marcgarreau:
Didn't realize that! (I am using 0.10/master.) You're correct that it is not required, but it appears that it is still an option. With the `polymorphic` option, the payload is:

```javascript
{
   picture: {
      id: 1,
      imageable: {
         type: "product",
         product: {
            id: 3,
            username: "Product 3"
         }
      }
   }
}
```

and without the option:

```javascript
{
   picture: {
      id: 1,
      imageable: {
         id: 3,
         title: "Product 3"
      }
   }
}
```

I don't think this is the place to debate the merits of each, but this would be nice stuff to know via README. I will update my changes to reflect this and get your feedback again.

beauby:
Oh, you're right. This is an adapter-specific option (specific to the `Attributes` and `Json` adapters).

docs/general/getting_started.md
```diff
@@ -80,6 +80,55 @@ end
...
+      type: "product",
+      product: {
+        id: 3,
+        title: "Product 3"
+      }
+    }
+  }
+}
+```
+
```
bf4: :100: Should this doc change be in this file?  Maybe in `serializers` or `layouts and rendering`? Just a thought, this might be the best spot, as well.

Is there a test for this behavior? Would be great to add.

marcgarreau: Busy week! Writing tests now, and will relocate these docs to a more appropriate location in next push.

test/adapter/json/polymorphic_test.rb
```diff
@@ -0,0 +1,25 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
+      class Json
+        class PolymorphicTest < ActiveSupport::TestCase
+          def setup
```
bf4:
```ruby
setup do 
 #...
end
```

removes the need to call `super` in def setup

docs/general/serializers.md
```diff
@@ -74,6 +74,55 @@ def blog
...
+    id: 1,
+    imageable: {
+      id: 3,
+      title: "Product 3"
+    }
+  }
+}
+```
+
+In prior releases (and optionally in `0.10`), a `polymorphic: true` parameter is required:
```
bf4: where is `polymorphic: true` used in 0.10 for serialization? Is it an untested behavior?

docs/general/serializers.md
```diff
@@ -74,6 +74,55 @@ def blog
...
+
+```ruby
+class PictureSerializer < ActiveModel::Serializer
+  has_one :imageable, polymorphic: true
+end
+```
+
+In this case, an example payload might be:
```
bf4: This is the JSON API adapter?

docs/general/getting_started.md
bf4: Awesome, thanks!

docs/general/serializers.md
beauby: After investigation, there is no trace of any `polymorphic` option in 0.10-rc1+.

docs/general/serializers.md
beauby: Nope, due to the very nature of the JSON API spec, such a parameter would never be necessary.

marcgarreau: Pardon the ambiguity. I originally uncovered this behavior with the default adapter, Attributes. I went back and confirmed on 0.9.4 that the option actually is not required there either. I will remove the content about the option.

test/adapter/json/polymorphic_test.rb
marcgarreau: I'm sorry, I don't know what you mean. This file pattern is taken directly from the other JSON adapter tests (i.e. https://github.com/rails-api/active_model_serializers/blob/master/test/adapter/json/belongs_to_test.rb).

beauby: @marcgarreau Nothing wrong with the way you did it. It just happens that you could also use the block syntax as @bf4 suggested in order to avoid an explicit call to `super`.

bf4: @beauby nothing wrong, but we're taking advantage of activesupport::testcase which has this feature (just replace `s/def setup/setup do`)

beauby: @bf4 Agreed, I also favor it. I just wanted to make it clear to @marcgarreau that it was an improvement suggestion rather than error correction.

marcgarreau: Now I'm with you. Thanks for the hand-holding (and learning me something new)! Updated.

marcgarreau:
Checking in; How does this look? Is there anything I can do?

beauby:
@marcgarreau It looks good to me, but I'd like somebody with more experience (than me, cc @joaomdmoura) with polymorphic associations to chip in, to make sure the way we handle them actually makes sense.
Other than that, great job, I'll merge it soon unless somebody raises their voice.

docs/general/serializers.md
```diff
@@ -74,6 +74,30 @@ def blog
...
+Polymorphic relationships are serialized just by specifying the relationship. For example:
+
+```ruby
+class PictureSerializer < ActiveModel::Serializer
+  has_one :imageable
+end
+```
+
+If the instance of `Picture` belongs to a `Product`, an example payload might be:
+
```
bf4: Specify adapter

beauby: Oh you're right @bf4. For some reason, I thought the whole text was inside the `Json` section of the adapters' doc.

test/adapter/json/polymorphic_test.rb
```diff
@@ -0,0 +1,25 @@
...
+    module Adapter
+      class Json
+        class PolymorphicTest < ActiveSupport::TestCase
+          setup do
+            @employee = Employee.new(id: 42, name: 'Zoop Zoopler', email: 'zoop@example.com')
+            @picture = @employee.pictures.new(id: 1, title: 'headshot-1.jpg')
+            @picture.imageable = @employee
+
+            @serializer = PictureSerializer.new(@picture)
+            @adapter = ActiveModel::Serializer::Adapter::Json.new(@serializer)
```
bf4: Use serailization = serializable(@picture, adapter: :json)

Then assert_equal ,.., serialization.as_json


(In transit, sorry for brief)

docs/general/serializers.md
bf4: Yeah, it's really a serializer concern, eg serializer.associations that is presented consistently by adapter

bf4: Consistent with adapter

docs/general/serializers.md
```diff
@@ -74,6 +74,30 @@ def blog
 end
 ```
 
+### Polymorphic Relationships
+
+Polymorphic relationships are serialized just by specifying the relationship. For example:
```
bf4: Ie like any other relationships since the object itself resolves the associated object, which ams treats no sifferently from any other object

beauby: My question is: should it, though? I lack experience with polymorphic associations, but is it expected that the client needs not know the type of associated resources in that case?

test/fixtures/poro.rb
```diff
@@ -61,6 +61,14 @@ def cache_key
   end
 end
 
+class Employee < ActiveRecord::Base
+  has_many :pictures, as: :imageable
+end
```
bf4: Maybe seializer name can describe what it's for testing?  Polymorhipichasmany polymorphic belongsto

docs/general/serializers.md
bf4: Only possible issue is with serializable type in a has many... Eg has_many -> [user, post] what is root in json and tyoe in json api with explicit and implicit association serializer

docs/general/serializers.md
```diff
@@ -74,6 +74,30 @@ def blog
...
+end
+```
+
+If the instance of `Picture` belongs to a `Product`, an example payload might be:
+
+```javascript
+{
+  picture: {
+    id: 1,
+    imageable: {
```
bf4: Should root key br imagable here? Otherwise don't know what type of object it is...

test/adapter/json/polymorphic_test.rb
marcgarreau: Understood. That is nicer, thanks.

docs/general/serializers.md
bf4: Does 0.10 pass same tests as previous impl that required polymorphic option?

test/adapter/json/polymorphic_test.rb
bf4: Yeah, sorry for late comments... This is great pr and thanks for work... Just saw with fresh eyes, I guess... On phone todayy then traveking tomorow sonwanted to take a look

marcgarreau: No issue at all - thanks for taking a peek while on the move!

marcgarreau:
Maybe this will answer multiple questions at once (cc/ @bf4):

I went back and double checked all the return values. On `0.10.0.rc3`, an example payload for both default and `json` adapter is the following (regardless of polymorphic option):

```javascript
{
   picture: {
      id: 1,
      title: "sdf",
      imageable: {
         id: 1,
         name: "Nerb Flerger"
      }
   }
}
```

With the `json_api` adapter (also regardless of polymorphic option):

```javascript
{
   data: {
      id: "1",
      type: "pictures",
      attributes: {
         title: "sdf"
      },
      relationships: {
         imageable: {
            data: {
               id: "1",
               type: "employees"
            }
         }
      }
   }
}
```

On release `0.9.4`, polymorphic documentation was lacking and I couldn't figure out how to add other adapters, but default payloads without the option return:

```javascript
{
   picture: {
      id: 1,
      title: "sdf",
      imageable: {
         id: 1,
         name: "Nerb Flerger"
      }
   }
}
```

and payloads with the option return:

```javascript
{
   picture: {
      id: 1,
      title: "sdf",
      imageable: {
         type: 'employee',
         employee: {
            id: 1,
            name: "Nerb Flerger"
         }
      }
   }
}
```

bf4:
Awsome!

0.9 has no adapters... And both 0.8 and 0.9 use pre-1.0 json api which varied a bit

B mobile phone

marcgarreau:
@bf4 I made the suggested change to the test and removed the example payload from the docs, because of the question about the payload key (i.e. how to identify which resource type you're receiving). That smells like an issue for more discussion and another PR.

test/adapter/json/polymorphic_test.rb
```diff
@@ -0,0 +1,24 @@
...
+    module Adapter
+      class Json
+        class PolymorphicTest < ActiveSupport::TestCase
+          setup do
+            @employee = Employee.new(id: 42, name: 'Zoop Zoopler', email: 'zoop@example.com')
+            @picture = @employee.pictures.new(id: 1, title: 'headshot-1.jpg')
+            @picture.imageable = @employee
+
+            @serialization = serializable(@picture, adapter: :json)
+            ActionController::Base.cache_store.clear
```
bf4: Did you find this cache clear neceassry or it was just copied?  It shouldn't be necessary

test/adapter/json/polymorphic_test.rb
```diff
@@ -0,0 +1,24 @@
...
+          end
+
+          def test_polymorphic_serializer
+            assert_equal({ id: 42, name: 'Zoop Zoopler' }, @serialization.as_json[:picture][:imageable])
+          end
```
bf4: I think you can do 'as_json(only: {picture: :imageable})' not 100% sure... 


In any case, 
Would you mind adding a test for the other two adapters? And then probably just ref these tests from the docs?

bf4: Thanks for being responsive... I know thr async of this can be a bit annoying when there's a lot of back and forth... i really appreciate your brjngjng this up and working on is documenting the behavior and even making us thing about it

marcgarreau: Thank you for the same! tests for `json_api` and `attributes` added, as well as links to them from the docs.

docs/general/serializers.md
```diff
@@ -74,6 +74,18 @@ def blog
...
+
+Polymorphic relationships are serialized by specifying the relationship, like any other association. For example:
+
+```ruby
+class PictureSerializer < ActiveModel::Serializer
+  has_one :imageable
+end
+```
+
+For more context, see the tests for each adapter: [attributes](https://github.com/rails-api/active_model_serializers/blob/master/test/adapter/attributes/polymorphic_test.rb), [json_api](https://github.com/rails-api/active_model_serializers/blob/master/test/adapter/json_api/polymorphic_test.rb), [json](https://github.com/rails-api/active_model_serializers/blob/master/test/adapter/json/polymorphic_test.rb)
```
bf4: Absolute links not necessary.  Relative is preferred for this kind of thing

bf4: Or ref a link witha conmit and highlighted line numbers (hint, when viewing a file, hit 'y' and adjust the url to point to this repo)

test/adapter/json/polymorphic_test.rb
bf4: Yeah, thanks.  Do you think we coukd make it more clear by having one test file tests against all adapters and make the expected hash differences stand out more?

test/adapter/json_api/polymorphic_test.rb
```diff
@@ -0,0 +1,25 @@
...
+          setup do
+            @employee = Employee.new(id: 42, name: 'Zoop Zoopler', email: 'zoop@example.com')
+            @picture = @employee.pictures.new(id: 1, title: 'headshot-1.jpg')
+            @picture.imageable = @employee
+
+            @serialization = serializable(@picture, adapter: :json_api)
+          end
+
+          def test_polymorphic_serializer
+            expected = { data: { id: '42', type: 'employees' } }
```
bf4: This is great.  I'd love to see a has_many polymorphic... Followip pr ok if that's too much

marcgarreau:
Consolidated tests into one and updated to the relative path. Happy to follow up with a has_many PR. This feels like a well-encapsulated explanation for this use-case (thanks to you guys' feedback), so I'd like to get this over the line first. As usual, happy to accommodate any other feedback.

beauby:
@marcgarreau Could you give #1453 a try?

test/adapter/polymorphic_test.rb
```diff
@@ -0,0 +1,72 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    module Adapter
```
joaomdmoura: IMHO we should split this into different files under the adapters namespace for that sake of organization

marcgarreau: Thanks for the feedback @joaomdmoura. My first implementation was split out like you suggested, but I followed @bf4's suggestion to combine these, which has a benefit of being able to compare the payloads side by side. If you'd like, I'll revert the changes. Let me know the final verdict.

joaomdmoura: np, you can keep it as it is, it a non blocking comment for sure, I can talk with @bf4 later to know if there any other reason to keep it this way, but it's just a minor detail

marcgarreau:
@beauby nice work on #1453! i saw that joão picked it up, but if you still need help testing anything, I can get around to it this evening.

test/adapter/polymorphic_test.rb
bf4: I hope to work with @remear to do some of these in cucumber so they're easier to compare, in general.  I suggested one file since yhe differences are subtle '. I'd be for separate files if we were more consistent in test file organization 

B mobile phone

beauby: Yeah, the tests are currently a bloody mess. We really need to do something about it.

beauby:
@marcgarreau Awesome, testing would be great, and if you have a bit of spare time, writing AMS tests for it would be "even awesomer"!

marcgarreau:
@beauby: :+1:

all: anything stopping this from getting over the line or are we considering adding scope? i'm happy to make follow up PR(s) for related work.

docs/general/serializers.md
```diff
@@ -74,6 +74,18 @@ def blog
 end
 ```
 
+### Polymorphic Relationships
+
+Polymorphic relationships are serialized by specifying the relationship, like any other association. For example:
```
beauby: If we merge #1453, this will have to change.

marcgarreau: Understood. Will wait for the word then and update docs/tests if necessary.

beauby:
@marcgarreau Would you mind submitting a PR with your tests to my branch?

marcgarreau:
@beauby will do shortly! didn't make it back online last night.

marcgarreau:
@beauby @bf4 did we decide to change anything with the `polymorphic` option? If so, I'll update this. If not, is this PR ready to go?

remear:
What remains to be done here?

marcgarreau:
@remear based on the label, maybe adding a little something to the changelog? The changes are just adding tests and bit of documentation, and I believe that's all good to go. Happy to accommodate if that's not the case.

remear:
@marcgarreau https://github.com/rails-api/active_model_serializers/pull/1420/files#r50025311, perhaps?

remear:
I think if you can address that comment and squash these commits I'm good with merging this.

marcgarreau:
Great @remear, thanks for the attention. Will get to it shortly.

remear:
@marcgarreau Before you squash, could you also add this to the [CHANGELOG](https://github.com/rails-api/active_model_serializers/blob/master/CHANGELOG.md)

marcgarreau:
Missed that comment prior to squashing @remear, will address.

remear:
Thanks, @marcgarreau!

PR #1423, opened 2016-01-11
Tidy up the tests

Empact:
* Use assert_nil where appropriate - assert_nil is equivalent to assert_equal with nil as the expectation, so the failure message should be the same. http://apidock.com/ruby/Test/Unit/Assertions/assert_nil
* Lead with the expected value in collection_serializer_test.rb, etc
 so that expected/actual in test failure messages are not reversed


bf4:
Great!

PR #1425, opened 2016-01-12
Extract links and type-related methods to their own module.

beauby:


lib/active_model/serializer.rb
```diff
@@ -17,32 +19,10 @@ class Serializer
     include Associations
     include Attributes
     include Caching
+    include Links
+    include Type
```
bf4: this pattern will likely bite us if we continue it indefinitely, but I like how it's helping us understand the code better right now

lib/active_model/serializer.rb
```diff
@@ -148,12 +128,6 @@ def read_attribute_for_serialization(attr)
       end
     end
 
-    # @api private
-    # Used by JsonApi adapter to build resource links.
```
bf4: comment and scope got lost

lib/active_model/serializer.rb
beauby: Well it's quite a common pattern. In any case, it does not make sense to have a 300+ lines `serializer.rb` file.

lib/active_model/serializer.rb
```diff
@@ -17,32 +19,10 @@ class Serializer
...
     require 'active_model/serializer/adapter'
 
-    with_options instance_writer: false, instance_reader: false do |serializer|
-      serializer.class_attribute :_type, instance_reader: true
-      serializer.class_attribute :_links # @api private : links definitions, @see Serializer#link
-      self._links ||= {}
-    end
-
-    # Serializers inherit _attribute_mappings, _reflections, and _links.
-    # Generates a unique digest for each serializer at load.
```
bf4: I wonder if there's a way to document what's inherited without looking for all the hooks. maybe register our own? dunno

lib/active_model/serializer.rb
beauby: This method actually got replaced by an instance reader on `_links`.

beauby: https://github.com/rails-api/active_model_serializers/pull/1425/files#diff-e0ecc03de43b196cb38915afa26cd602R8

lib/active_model/serializer/links.rb
```diff
@@ -0,0 +1,33 @@
+module ActiveModel
+  class Serializer
+    module Links
+      extend ActiveSupport::Concern
+
+      included do
+        with_options instance_writer: false, instance_reader: true do |serializer|
+          serializer.class_attribute :_links # @api private
```
bf4: So, just added as an instance reader, ok

lib/active_model/serializer/links.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      module ClassMethods
+        def inherited(base)
+          super
+          base._links = _links.dup
+        end
+
+        # Define a link on a serializer.
+        # @example
+        #   link :self { "/posts/#{object.id}" }
```
bf4: missing comma in comment. Also, we should use absolute urls, right?

beauby: No missing comma (this is a block, not a hash). As far as absolute urls are concerned, yes :)

lib/active_model/serializer.rb
beauby: In my point of view it makes sense that each feature handles inheritance for its own data, no?

lib/active_model/serializer/links.rb
bf4: > No missing comma (this is a block, not a hash). 

:sadtrombone:

lib/active_model/serializer.rb
bf4: we should just have a reference conversation we write about extracting to modules vs. classes to link to when this comes up :)

lib/active_model/serializer.rb
bf4: yeah.. just makes it harder to chase down by inspecting the code... really just thinking about readability

lib/active_model/serializer.rb
```diff
@@ -17,32 +19,10 @@ class Serializer
     include Associations
     include Attributes
     include Caching
+    include Links
```
bf4: maybe `include Links # Used by JsonApi adapter to build resource links.`

PR #1426, opened 2016-01-12
Add a default_include_tree config variable to ActiveModel::Serializer

Empact:
This is useful to set application-wide default behavior - e.g. in
previous versions of AMS the default behavior was to serialize the
full object graph by default - equivalent to the '**' include tree.

Currently just the global setting, but I think this could also work
on a per-serializer basis, with more attention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Breaking changes:
 
 Features:
+- [#1426](https://github.com/rails-api/active_model_serializers/pull/1426) Add ActiveModelSerializers.config.default_includes (@empact)
 
 Fixes:
 - [#1710](https://github.com/rails-api/active_model_serializers/pull/1710) Prevent association loading when `include_data` option
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -52,9 +52,11 @@ Each adapter has a default key transform configured:
 `config.key_transform` is a global override of the adapter default. Adapters
 still prefer the render option `:key_transform` over this setting.
 
+##### default_includes
+What relationships to serialize by default.  Default: `'*'`, which includes one level of related
+objects. See [includes](adapters.md#included) for more info.
 
 ## JSON API
-
 
 ##### jsonapi_resource_type
 
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -10,8 +10,6 @@ module ActiveModel
     module Associations
       extend ActiveSupport::Concern
 
-      DEFAULT_INCLUDE_TREE = ActiveModel::Serializer::IncludeTree.from_string('*')
-
       included do
         with_options instance_writer: false, instance_reader: true do |serializer|
           serializer.class_attribute :_reflections
@@ -80,10 +78,11 @@ module ActiveModel
         end
       end
 
-      # @param [IncludeTree] include_tree (defaults to all associations when not provided)
+      # @param [IncludeTree] include_tree (defaults to the
+      #   default_includes config value when not provided)
       # @return [Enumerator<Association>]
       #
-      def associations(include_tree = DEFAULT_INCLUDE_TREE)
+      def associations(include_tree = ActiveModelSerializers.default_include_tree)
         return unless object
 
         Enumerator.new do |y|
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -19,6 +19,7 @@ module ActiveModel
           collection_serializer
         end
 
+        config.default_includes = '*'
         config.adapter = :attributes
         config.jsonapi_resource_type = :plural
         config.jsonapi_version = '1.0'
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -3,8 +3,13 @@ module ActiveModelSerializers
     class Attributes < Base
       def initialize(serializer, options = {})
         super
-        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
         @cached_attributes = options[:cache_attributes] || {}
+        @include_tree =
+          if options[:include]
+            ActiveModel::Serializer::IncludeTree.from_include_args(options[:include])
+          else
+            ActiveModelSerializers.default_include_tree
+          end
       end
 
       def serializable_hash(options = nil)
--- a/test/action_controller/json/include_test.rb
+++ b/test/action_controller/json/include_test.rb
@@ -4,6 +4,10 @@ module ActionController
   module Serialization
     class Json
       class IncludeTest < ActionController::TestCase
+        INCLUDE_STRING = 'posts.comments'.freeze
+        INCLUDE_HASH = { posts: :comments }.freeze
+        DEEP_INCLUDE = 'posts.comments.author'.freeze
+
         class IncludeTestController < ActionController::Base
           def setup_data
             ActionController::Base.cache_store.clear
@@ -38,17 +42,28 @@ module ActionController
 
           def render_resource_with_include_hash
             setup_data
-            render json: @author, include: { posts: :comments }, adapter: :json
+            render json: @author, include: INCLUDE_HASH, adapter: :json
           end
 
           def render_resource_with_include_string
             setup_data
-            render json: @author, include: 'posts.comments', adapter: :json
+            render json: @author, include: INCLUDE_STRING, adapter: :json
           end
 
           def render_resource_with_deep_include
             setup_data
-            render json: @author, include: 'posts.comments.author', adapter: :json
+            render json: @author, include: DEEP_INCLUDE, adapter: :json
+          end
+
+          def render_without_recursive_relationships
+            # testing recursive includes ('**') can't have any cycles in the
+            # relationships, or we enter an infinite loop.
+            author = Author.new(id: 11, name: 'Jane Doe')
+            post = Post.new(id: 12, title: 'Hello World', body: 'My first post')
+            comment = Comment.new(id: 13, body: 'Commentary')
+            author.posts = [post]
+            post.comments = [comment]
+            render json: author
           end
         end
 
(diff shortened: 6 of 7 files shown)


lib/active_model/serializer/adapter/attributes.rb
```diff
@@ -4,7 +4,9 @@ module Adapter
       class Attributes < Base
         def initialize(serializer, options = {})
           super
-          @include_tree = IncludeTree.from_include_args(options[:include] || '*')
+          @include_tree = IncludeTree.from_include_args(
+            options[:include] || ActiveModel::Serializer.config.default_include_tree
```
bf4: s//ActiveModelSerializers.config

test/action_controller/json/include_test.rb
```diff
@@ -4,6 +4,10 @@ module ActionController
   module Serialization
     class Json
       class IncludeTest < ActionController::TestCase
+        INCLUDE_STRING = 'posts.comments'
+        INCLUDE_HASH = { posts: :comments }
+        DEEP_INCLUDE = 'posts.comments.author'
```
bf4: :100:

lib/active_model/serializer/configuration.rb
```diff
@@ -19,6 +19,7 @@ def config.array_serializer
           collection_serializer
         end
 
+        config.default_include_tree = '*'
```
bf4: Can you think of a more descriptive name for this?

Future PR: would be a reasonable serializer config

bf4:
rubocop failures

lib/active_model/serializer/configuration.rb
beauby: How about `default_nesting_level`?
A per-serializer config would raise issues (for instance parent has `'**'` and child has `'*'`).

beauby:
I :+1: the idea.

lib/active_model/serializer/configuration.rb
bf4: > How about default_nesting_level?

- `default_included_associations`? 
- `default_included_relationships`?
- `default_includes`?
- `default_included_resources`?

ref: 

- https://github.com/rails-api/active_model_serializers/blob/v0.5.0/lib/active_model/serializer.rb#L402-L424
- https://github.com/rails-api/active_model_serializers/blob/v0.7.0/lib/active_model/serializer.rb#L287-L297
- https://github.com/rails-api/active_model_serializers/blob/v0.8.0/lib/active_model/serializer.rb#L364-L369
- https://github.com/rails-api/active_model_serializers/blob/1e0c9ef93b96c640381575dcd30be07ac946818b/lib/active_model/serializer.rb#L240-L245

> A per-serializer config would raise issues (for instance parent has '**' and child has '*').

well, if the AuthorSerializer has a default includes of 'posts.comments' (which is crazy), then it would presumably set the includes of the PostSerializer to 'comments' and the includes of CommentSerializer to empty, right?

noahsilas: I've pushed up an updated patch that shows what this looks like with `default_includes`, which feels like a fairly reasonable name to set the default value used for the `include` option.

I don't really love any of the options we've talked about so far though, so I thought that addressing the awkward naming with documentation was probably a better approach than bikeshedding the name too much.

If someone has strong opinions, I'm totally open to changing the name and submitting again.

docs/general/configuration_options.md
```diff
@@ -9,6 +9,8 @@ preferably inside an initializer.
 
 - `adapter`: The [adapter](adapters.md) to use. Possible values: `:attributes, :json, :json_api`. Default: `:attributes`.
 - `serializer_lookup_enabled`: When `false`, serializers must be explicitly specified. Default: `true`
+- `default_includes`: What relationships to serialize by default.  Default: `'*'`, which includes one level of related
+  objects. See [includes](adapters.md#included) for more info.
```
groyoh: Shouldn't it be `See [included]` since the anchor is called `included`?

noahsilas: Hmm. I wrote this to match the config key, although I can see it being a little disconcerting to click through and find a different header. Of course, the thing actually being described is the `include` option, which doesn't match either of these 100%. And, of course, it's a little weird to be linking to docs for the JSON API adapter to show behavior that is also applicable to adapters like `Attributes`. :disappointed:

groyoh: You're right. What we're actually missing here is a general documentation of the `include` option for all the adapters.

noahsilas:
Fixed conflicts, added changelog entry.

bf4:
@noahsilas looks good.  I don't see the changelog...

noahsilas:
D'oh. Updated again with the changelog entry.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Features:
 - [#1454](https://github.com/rails-api/active_model_serializers/pull/1454) Add support for
   relationship-level links and meta attributes. (@beauby)
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
+- [#1426](https://github.com/rails-api/active_model_serializers/pull/1426) Add ActiveModelSerializers.config.default_includes (@empact)
```
bf4: @Empact @noahsilas put at top of features plz (sorry)

lib/active_model/serializer/associations.rb
```diff
@@ -80,10 +78,11 @@ def associate(reflection)
         end
       end
 
-      # @param [IncludeTree] include_tree (defaults to all associations when not provided)
+      # @param [IncludeTree] include_tree (defaults to the
+      #   default_includes config value when not provided)
       # @return [Enumerator<Association>]
       #
-      def associations(include_tree = DEFAULT_INCLUDE_TREE)
+      def associations(include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(ActiveModel::Serializer.config.default_includes))
```
bf4: would be better if the default were memoized so it doesn't need to be recalculated all the time.  I can't imagine changing the default after initialization.

```ruby
def self.default_include_tree
  @default_include_tree ||= ActiveModel::Serializer::IncludeTree.from_include_args(ActiveModel::Serializer.config.default_includes)
end

def associations(include_tree = self.class.default_include_tree)
```

would be a little better, assuming we're concerned about the file being evaluated before the `default_includes` is set in an initializer.

Also, I missed this, but it should be `s/ActiveModel::Serializer.config/ActiveModelSerializers.config` as we're moving away from the `Serializer` class being the primary namespace.  For now, it's just a reference, so, not wrong, just not the pattern we're moving towards.

test/action_controller/json/include_test.rb
bf4: A followup PR with a deep include hash would be good, but no need to do more work in this pr

test/action_controller/json/include_test.rb
```diff
@@ -74,37 +78,58 @@ def test_render_without_include
           assert_equal(expected, response)
         end
 
+        def test_render_with_default_includes_hash
+          with_default_includes INCLUDE_HASH do
```
bf4: this is a bit misleading, right, because you'd never have a default includes of `posts.comments`, right? the more common, scenarios, cc @beauby would be `'*'`, `''`, `'*.*'`, I think?

lib/active_model/serializer/associations.rb
noahsilas: Ah, that's a really good call. Fixing now.

noahsilas: Ah, that's a really good call. Fixing now.

CHANGELOG.md
noahsilas: Done.

remear:
Any progress on this?

test/action_controller/json/include_test.rb
beauby: That's how I see it as well.

noahsilas:
I've rebased to fix merge conflicts, and updated tests to match feedback that we weren't covering the most common cases to use this config option. How do y'all feel now?

noahsilas:
I'm surprised to see this build failing. It seems like this failure is related to the RuboCop 0.40 release. (https://github.com/bbatsov/rubocop/releases/tag/v0.40.0)

Running with 0.39 locally, there are no errors.

beauby:
You have the full list of rubocop offences [here](https://travis-ci.org/rails-api/active_model_serializers/jobs/128999246#L413).

remear:
Would you please rebase this again to pull in the latest master that pins rubocop to 0.39.0.

hut8:
This is awesome - any chance of it being merged shorrtly?

lib/active_model/serializer/associations.rb
```diff
@@ -67,6 +65,15 @@ def has_one(name, options = {}, &block)
...
         end
 
+        # Memoized default include tree (shared across all instances)
+        # @return [IncludeTree]
+        # @api private
+        def default_include_tree
+          @default_include_tree ||= ActiveModel::Serializer::IncludeTree.from_include_args(
+            ActiveModelSerializers.config.default_includes
+          )
+        end
```
bf4: nice 'cause easy to override.  should this be a private api?  I'd consider making it public, but no need to delay this PR

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -3,7 +3,9 @@ module Adapter
     class Attributes < Base
       def initialize(serializer, options = {})
         super
-        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
+        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(
+          options[:include] || ActiveModelSerializers.config.default_includes
+        )
```
bf4: 👍

lib/active_model/serializer/configuration.rb
```diff
@@ -19,6 +19,7 @@ def config.array_serializer
           collection_serializer
         end
 
+        config.default_includes = '*'
```
bf4: 👍

test/action_controller/json/include_test.rb
```diff
@@ -4,6 +4,10 @@ module ActionController
   module Serialization
     class Json
       class IncludeTest < ActionController::TestCase
+        INCLUDE_STRING = 'posts.comments'.freeze
+        INCLUDE_HASH = { posts: :comments }.freeze
+        DEEP_INCLUDE = 'posts.comments.author'.freeze
```
bf4: 👍

test/action_controller/json/include_test.rb
```diff
@@ -158,8 +224,14 @@ def test_render_resource_with_deep_include
               ]
             }
           }
+        end
 
-          assert_equal(expected, response)
+        def with_default_includes(include_tree)
+          original = ActiveModel::Serializer.config.default_includes
```
bf4: should be `ActiveModelSerializers.config`

also this has no effect on the memoized default, right?

test/action_controller/json/include_test.rb
```diff
@@ -38,17 +42,28 @@ def render_without_include
...
 
           def render_resource_with_deep_include
             setup_data
-            render json: @author, include: 'posts.comments.author', adapter: :json
+            render json: @author, include: DEEP_INCLUDE, adapter: :json
+          end
+
+          def render_without_recursive_relationships
+            # testing recursive includes ('**') can't have any cycles in the
+            # relationships, or we enter an infinite loop.
```
bf4: do you have a test case that demonstrates this?

bf4:
Looks good to me, though I have some followup questions

test/action_controller/json/include_test.rb
noahsilas: This is really to support the test on line 131. If you try to use the '**' includes for the actions above, you get spec failures from infinite recursion.

I'm not sure how to resolve that without getting into the weeds. #1538 is meant to help provide some nice ways to avoid this in the presence of reference cycles in serializers, but needs to be brought up to date with master.

test/action_controller/json/include_test.rb
noahsilas: You're right. I just had a head scratching moment trying to figure out what is happening here.

It turns out that in normal usage we don't seem to use the default value, but are instead creating a new include tree (see lib/active_model_serializers/adapter/attributes.rb line 6). 

This makes me think that we should either remove the memoized class method (simpler), or ensure that we are using the memoized include tree on instances too (potentially more performant, harder to test as we need to do something like clear the cache on all serializers).

Do you have any preferences? I'm leaning towards removing the memoized version, since it is equivalent to the existing code, but I can implement either version.

noahsilas:
I've updated the memoized default include tree to be used in more places. I made it a public API on `ActiveModelSerializers`, which has a couple of benefits:
- Only one instance, instead of one instance per serializer class
- Easy to clear this cache, for instance when testing different configurations (see test/action_controller/json/include_test.rb).

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -3,8 +3,13 @@ module Adapter
     class Attributes < Base
       def initialize(serializer, options = {})
         super
-        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
         @cached_attributes = options[:cache_attributes] || {}
+        @include_tree = \
```
beauby: No need for `\` here.

NullVoxPopuli:
needs rebase :-(

noahsilas:
Rebased :-)

lib/active_model_serializers/adapter/attributes.rb
NullVoxPopuli: @Empact, can you address this?

noahsilas: Done.

NullVoxPopuli:
LGTM

hut8:
Thanks so much for this, @noahsilas and others! Awesome job :+1:

Empact:
Same, thanks @noahsilas for finishing it off. 👍

PR #1435, opened 2016-01-14
Fix Rails 5 warnings (uses Rails5Shim)

bf4:
```plain
DEPRECATION WARNING: ActionController::TestCase HTTP request methods will accept only
keyword arguments in future Rails versions.

Examples:

get :show, params: { id: 1 }, session: { user_id: 1 }
process :update, method: :post, params: { id: 1 }
 (called from test_deserialization at ~/repos/active_model_serializers/test/action_controller/json_api/deserialization_test.rb:42)
 ```

### DIFF
--- a/test/support/rails5_shims.rb
+++ b/test/support/rails5_shims.rb
@@ -3,17 +3,35 @@ module Rails5Shims
     # https://github.com/rails/rails/blob/b217354/actionpack/lib/action_controller/test_case.rb
     REQUEST_KWARGS = [:params, :session, :flash, :method, :body, :xhr]
 
+    def get(path, *args)
+      fold_kwargs!(args)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/support/rails5_shims.rb [hunk 1]
```diff
@@ -3,17 +3,35 @@ module ControllerTests
...
+    def patch(path, *args)
+      fold_kwargs!(args)
+      super
+    end
+
+    def put(path, *args)
+      fold_kwargs!(args)
+      super
+    end
+
```
remear: Wasn't sure if you want to get crazy with metaprogramming just for this.

bf4: nah, I like easy to read

PR #1436, opened 2016-01-14
Controller namespace lookup.

beauby:
Just took a stab at it. Basically, it allows for inferring the serializer in the following scenario:
```ruby
class Api::V1::PostController < ActionController::Base
  def create
    @post = Post.create(params)
    render json: @post
  end
end

class Api::V1::PostSerializer < ActiveModel::Serializer
  ...
end
```

Ref #1265, #886.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -56,7 +56,7 @@ module ActionController
 
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
-        options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(self) unless options.key?(:serialization_context)
+        options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(controller: self) unless options.key?(:serialization_context)
         serializable_resource = get_serializer(resource, options)
         super(serializable_resource, options)
       end
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -56,7 +56,7 @@ module ActionController
 
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
-        options.fetch(:serialization_context) { options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request) }
+        options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(self) unless options.key?(:serialization_context)
         serializable_resource = get_serializer(resource, options)
         super(serializable_resource, options)
       end
--- a/test/serializers/serializer_for_test.rb
+++ b/test/serializers/serializer_for_test.rb
@@ -115,18 +115,54 @@ module ActiveModel
           assert_equal nil, serializer
         end
 
-        def test_serializer_for_nested_resource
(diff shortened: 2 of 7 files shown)


lib/action_controller/serialization.rb
```diff
@@ -56,7 +56,7 @@ def use_adapter?
 
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
-        options.fetch(:serialization_context) { options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request) }
+        options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(controller: self) unless options.key?(:serialization_context)
```
bf4: :-1: I'd rather do it like 0.9 and not drag an enormous object around

see https://github.com/rails-api/active_model_serializers/blob/0-9-stable/lib/action_controller/serialization.rb#L61-L93

bf4: let's pass `view_context` if we really want an enourmous object

bf4:
:-1: on passing the `controller` around. I really, *really* don't want that in our codebase.

Can we discuss other impls

- 0.9 impl https://github.com/rails-api/active_model_serializers/blob/0-9-stable/lib/action_controller/serialization.rb#L61-L93
- https://github.com/skalee/active_model_serializers-namespaces
- https://github.com/hookercookerman/active_model_version_serializers

bf4:
sorry for all the :-1: screaming baby making me a little frustrated...

lib/action_controller/serialization.rb
beauby: Not sure I understand: the `serialization_context` option is **already** being passed around.

beauby:
Note that we don't pass the controller around. We just pass it to `SerializationContext` so that it extracts information from it.

beauby:
Note also that #1196 solves a different problem.

lib/action_controller/serialization.rb
bf4: sorry, should have deleted the comment.. was being braindead

NullVoxPopuli: I'd like to see this broken up in to multiple longs

```ruby
unless options.key?(:serialization_context)
  options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(controller: self)
end

```

test/serializers/serializer_for_test.rb
```diff
@@ -115,12 +115,6 @@ def test_serializer_for_namespaced_resource_with_lookup_disabled
           assert_equal nil, serializer
         end
 
-        def test_serializer_for_nested_resource
-          comment = ResourceNamespace::Comment.new
-          serializer = ResourceNamespace::PostSerializer.serializer_for(comment)
-          assert_equal ResourceNamespace::PostSerializer::CommentSerializer, serializer
-        end
-
```
beauby: This test did not make sense anymore with the new lookup convention.

jordanell:
Any progress here?

NullVoxPopuli:
@beauby, could you rebase? 

we have another request for this feature, by @dgdosen

beauby:
Closing this in favor of RFC #1442.

PR #1443, opened 2016-01-18
Update pagination links test to verify bug exists in generating pagination links

andrewhavens:
This PR simply adds a failing test to confirm that there is a bug. I found this bug while trying to generate pagination links in my own app. When paginating a list, in which the last page has less results than the previous, the last page will change the size of the links to match the result. It should always use the same  number.

For example:
* Given a total of 5 items
* When specifying 2 items per page
* There should be 3 pages (2 on each page, with a final page with 1 item)

When fetching the last page (and specifying 2 items per page), the `self`, `prev`, and `first` links should all have `page[size]=2` and assume the correct number of pages.

The failing tests include:
* `test_render_only_prev_and_first_pagination_links`
* `test_render_only_prev_and_first_pagination_links_with_additional_params`

### DIFF
--- a/test/action_controller/json_api/pagination_test.rb
+++ b/test/action_controller/json_api/pagination_test.rb
@@ -13,11 +13,7 @@ module ActionController
 
         class PaginationTestController < ActionController::Base
           def setup
-            @array = [
-              Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1' }),
-              Profile.new({ name: 'Name 2', description: 'Description 2', comments: 'Comments 2' }),
-              Profile.new({ name: 'Name 3', description: 'Description 3', comments: 'Comments 3' })
-            ]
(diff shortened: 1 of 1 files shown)


bf4:
Awesome! Would you like to hell work on a PR to fix it?

andrewhavens:
Unfortunately, I can't guarantee any availability to work on a fix for this. I'm currently working on unrelated projects.

beauby:
Is this an AMS or a Kaminari related issue?

dmons:
No, its not a AMS or a Kaminari issues. I assume I created duplicated issue #1452.
A can suggest do not generate ```page[size]``` param here https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json_api/pagination_links.rb#L17

NullVoxPopuli:
This looks to have been resolved.
I fetched andrewhavens/active_model_serializers, and cherry-picked the commit for this PR. All tests pass. And I just now see that #1570 addressed the failing tests in this PR.

Great work everyone!

PR #1300, opened 2015-10-26
Refactor JsonApi adapter in order to avoid recomputation of hashes fo…

beauby:
…r included resources.
Ref: #1239

Plus make the method names closer to the JSON API terminology.

Main changes:
 + introduced the `attributes_for` method (for symmetry with `relationships_for`)
 + renamed `relationship_value_for` into `linkage_for`
 + computation of all hashes at once, with an annotation on each one saying whether it is a `primary` or `included` resource, so that no hash is ever computed twice, and there is not need for the `serializable_hash_for_single_resource` / `serializable_hash_for_collection` distinction. The adapter no longer calls itself recursively.
 + introduced `process_resource` which does the job of `add_included_resources_for`, `included_for`, `primary_data_for`, `serializable_hash_for_single_resource`, `serializable_hash_for_collection` (and removed those)

PR #1447, opened 2016-01-19
Refactor JsonApi adapter to avoid redundant computations.

beauby:
This is a rewrite of #1300, stripped down to the bare minimum.

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -52,18 +52,24 @@ module ActiveModel
         def serializable_hash(options = nil)
           options ||= {}
 
-          hash =
-            if serializer.respond_to?(:each)
-              serializable_hash_for_collection(options)
-            else
-              serializable_hash_for_single_resource
-            end
+          is_collection = serializer.respond_to?(:each)
+          serializers = is_collection ? serializer : [serializer]
+          primary_data, included = resource_objects_for(serializers)
+
+          hash = {}
+          hash[:data] = is_collection ? primary_data : primary_data[0]
+          hash[:included] = included if included.any?
 
           ApiObjects::JsonApi.add!(hash)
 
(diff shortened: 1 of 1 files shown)


anonova:
```
rails-api/active_model_serializers:master: [active_model_serializers] Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModel::Serializer::Adapter::JsonApi (2186.42ms)
rails-api/active_model_serializers:master: [active_model_serializers] Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModel::Serializer::Adapter::Attributes (515.21ms)
beauby/active_model_serializers:jsonapi-refactor-2016: [active_model_serializers] Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModel::Serializer::Adapter::JsonApi (1094.52ms)
```

Nice gains! Seeing about a 2x speedup from master with 1457 objects (250 root + 55 references + 1152 included resources). I may do some profiling at some point since this still seems quite slow, but it's definitely better than before.

It's also about 2x slower, rather than 4x, than the attributes adapter now.

beauby:
@anonova Thanks for those numbers! We really need to get that benchmarking PR done soon. In general, I would expect JsonApi to be slower than Attributes, because in case of `has_many` or `has_one` associations, JsonApi will have to access the related models to fill the `relationships` part of the base object with the resource identifiers, even if the associated models are not included, whereas the Attributes adapter will simply skip them. So in the end, the JsonApi adapter is slightly richer, and it would seem more fair to compare JsonApi with nesting level `n` against JsonApi with nesting level `n+1`.
But in any case, we'll make sure the JsonApi adapter gets as fast as possible!

beauby:
Any objection to this @rails-api/ams?

bf4:
Looks good.  @anonova gist or paste bm code w results?

kurko:
LGTM. +2

PR #1450, opened 2016-01-20
Update serializers.md

vahid-almasi:
Fix docs for overriding attribute and association methods

### DIFF
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -198,6 +198,18 @@ class PostSerializer < ActiveModel::Serializer
 end
 ```
 
+Or if you want to override association in block format, you can use:
+
+```ruby
+class PostSerializer < ActiveModel::Serializer
+  attributes :id, :body
+
+  has_many :comments do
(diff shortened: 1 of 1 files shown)


docs/general/serializers.md
```diff
@@ -198,6 +198,18 @@ class PostSerializer < ActiveModel::Serializer
...
 ```
 
+Or if you want to override association in block format, you can use:
+
+```ruby
+  attributes :id, :body
+
+  has_many :comments do
+    "#{object.comments.active}"
```
beauby: This will return a `String`. You probably meant `object.comments.active`.

beauby:
Thanks @vahid-almasi! I think we should remove the overriding by method declaration from the docs altogether, as we plan to drop support for it for 0.10-release.

beauby:
Thanks a lot for your work @vahid-almasi. I believe the docs were updated by #1462. Feel free to issue a new PR to improve them! Closing this in the meantime.

PR #1458, opened 2016-01-22
	Preserve the serializer type when fragment caching

bdmac:
We were not previously cloning the type setting into the dynamically
generated cached/non-cached serializers for a given fragment-cached
serializer. This led to the type generated for JsonApi having the wrong
value when fragment caching is enabled by adding either :except or :only
options to cache.

This pulls the type setting from the fragment-cached serializer forward
onto the dynamic caching classes so it is preserved in the output.

### DIFF
--- a/lib/active_model/serializer/adapter/cached_serializer.rb
+++ b/lib/active_model/serializer/adapter/cached_serializer.rb
@@ -24,7 +24,7 @@ module ActiveModel
         end
 
         def fragment_cached?
-          @klass._cache && (@klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except)
+          @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
         end
 
         def cache_key
--- a/lib/active_model/serializer/adapter/cached_serializer.rb
+++ b/lib/active_model/serializer/adapter/cached_serializer.rb
@@ -24,7 +24,7 @@ module ActiveModel
         end
 
         def fragment_cached?
-          @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
+          @klass._cache && (@klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except)
         end
 
         def cache_key
--- a/test/adapter/fragment_cache_test.rb
+++ b/test/adapter/fragment_cache_test.rb
@@ -31,8 +31,13 @@ module ActiveModel
           }
           assert_equal(@spam_hash.fetch, expected_result)
         end
+
+        def test_fragment_fetch_with_type_override
+          type_serializer = TypedRoleSerializer.new(Role.new(name: 'Another Author'))
+          type_hash       = ActiveModel::Serializer::Adapter::JsonApi.new(type_serializer).serializable_hash
+          assert_equal(type_hash[:data][:type], TypedRoleSerializer._type)
+        end
       end
     end
   end
 end
-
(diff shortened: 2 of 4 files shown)


lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -24,7 +24,7 @@ def cached?
         end
 
         def fragment_cached?
-          @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
+          @klass._cache && (@klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except)
```
bf4: this can be it's own PR. would you mind? https://github.com/CheckMateIO/active_model_serializers/commit/e9e2b5712df0eda12399bac31a5d126b8b040082

bf4:
Awesome, thanks! :+1: some comments above

test/adapter/fragment_cache_test.rb
```diff
@@ -31,8 +31,13 @@ def test_fragment_fetch_with_namespaced_object
           }
           assert_equal(@spam_hash.fetch, expected_result)
         end
+
+        def test_fragment_fetch_with_type_override
+          type_serializer = TypedRoleSerializer.new(Role.new(name: 'Another Author'))
+          type_hash       = ActiveModel::Serializer::Adapter::JsonApi.new(type_serializer).serializable_hash
+          assert_equal(type_hash[:data][:type], TypedRoleSerializer._type)
```
bf4: better to use the serializable test helper (below may contain typos)

```ruby
serialization = serializable(Role.new(name: 'Another Author'), serializer: TypedRoleSerializer, adapter: :json_api).serializable_hash
assert_equal(TypedRoleSerializer._type, serialization.fetch(:data).fetch(:type))
```

test/fixtures/poro.rb
```diff
@@ -222,6 +222,18 @@ def maker
...
+  type 'my-roles'
+  cache only: [:name], skip_digest: true
+  attributes :id, :name, :description, :slug
+
+  def slug
+    "#{object.name}-#{object.id}"
+  end
+
+  belongs_to :author
+end
```
bf4: would you mind putting this in the test file?  We're try to no longer add files to poro unless they are for general usage

test/fixtures/poro.rb
```diff
@@ -222,6 +222,18 @@ def maker
...
 end
 
+TypedRoleSerializer = Class.new(ActiveModel::Serializer) do
+  type 'my-roles'
+  cache only: [:name], skip_digest: true
+  attributes :id, :name, :description, :slug
+
+  def slug
+    "#{object.name}-#{object.id}"
+  end
```
bf4: noting that in `rc4` slug won't be call here. you'd have to do `attribute :slug { "#{object.name}-#{object.id}" }`.  can be removed from the poro here, though, since it's not being used

bdmac:
@bf4 comments addressed. Will open a secondary PR with the fix for `fragment_cached?`

bf4:
Rebase, merged, and changelog added in https://github.com/rails-api/active_model_serializers/commit/527c2aed98a9b8687726b691620334f4c4e0bb9e

bf4:
Closed by https://github.com/rails-api/active_model_serializers/commit/527c2aed98a9b8687726b691620334f4c4e0bb9e

bf4:
Closed by a40998273d1eb6594e1344cae41603f9a5966386

bf4:
This was merged, but GitHub didn't pick it up. I'm not sure why

PR #1459, opened 2016-01-23
updates Readme to reflect AMS is no longer default in Rails 5

joyvuu-dave:
As of https://github.com/rails/rails/commit/f8edd2043e864aff0bde9289da550709532268a6 AMS is no longer included in Rails by default when the `--api` option is used.


beauby:
Thanks @joyvuu-dave!

prusswan:
some background on this issue (at least for Rails 5.0): https://github.com/rails/rails/commit/f8edd2043e864aff0bde9289da550709532268a6

PR #1471, opened 2016-01-28
[Cleanup] Serializer caching is its own concern

bf4:
## Update 2016-03-13

Rebased off of master.

## Original Description

I was looking at the caching code and realized that although it is called
from the adapter, it is it's own thing.  In fact, if it weren't for
fragment caching, which uses the adapter instance, the caching code could
be used entirely by the serializer!

I wrote this PR in [two](https://github.com/rails-api/active_model_serializers/commit/25e85bf280b4e693863aed5cb92996880567085a) [commits](https://github.com/rails-api/active_model_serializers/commit/38de293ce4f323f5ab1638e87538ad386b35962e) so that the moving down of the CachedSerializer
and FragmentCache classes from the Adapter namespace and the moving to the
ActiveModelSerializers namespace can be more easily seen.  I intend to squash
them before merge, but would be happy to do it now.


lib/active_model_serializers/fragment_cache.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      cached_adapter     = adapter.class.new(cached_serializer, instance_options)
+      non_cached_adapter = adapter.class.new(non_cached_serializer, instance_options)
+
+      # Get serializable hash from both
+      cached_hash     = cached_adapter.serializable_hash
+      non_cached_hash = non_cached_adapter.serializable_hash
+
+      # Merge both results
+      adapter.fragment_cache(cached_hash, non_cached_hash)
```
bf4: This is an adapter method since the adapter might want to customize how the two hashes are merged, e.g. see JSON API adapter.


```diff
-          non_cached_hash.merge cached_hash
+          root = false if instance_options.include?(:include)
+          core_cached       = cached_hash.first
+          core_non_cached   = non_cached_hash.first
+          no_root_cache     = cached_hash.delete_if { |key, _value| key == core_cached[0] }
+          no_root_non_cache = non_cached_hash.delete_if { |key, _value| key == core_non_cached[0] }
+          cached_resource   = (core_cached[1]) ? core_cached[1].deep_merge(core_non_cached[1]) : core_non_cached[1]
+          hash = (root) ? { root => cached_resource } : cached_resource
+
+          hash.deep_merge no_root_non_cache.deep_merge no_root_cache
```

lib/active_model_serializers/fragment_cache.rb
```diff
@@ -0,0 +1,107 @@
...
+      non_cached_attributes.each do |attribute|
+        options = serializer.class._attributes_keys[attribute]
+        options ||= {}
+        # Add non-cached attributes to non-cached Serializer
+        serializers[:non_cached].constantize.attribute(attribute, options)
+      end
+    end
+
+    # Given a resource name and its serializer's class
+    # 1. Dyanmically creates a CachedSerializer and NonCachedSerializer
```
bf4: typo :frowning:

lib/active_model_serializers/cached_serializer.rb
```diff
@@ -0,0 +1,42 @@
+require 'active_model_serializers/fragment_cache'
+module ActiveModelSerializers
+  class CachedSerializer
+    def initialize(serializer)
+      @cached_serializer = serializer
+      @klass             = @cached_serializer.class
```
groyoh: I think using `serializer.class` here is less typo-prone. A typo on `serializer` would result in `NameError` whereas a typo in `@cached_serializer` would result in `nil`.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -71,7 +70,14 @@ def serializable_hash(options = nil)
 
         def fragment_cache(cached_hash, non_cached_hash)
           root = false if instance_options.include?(:include)
-          ActiveModel::Serializer::Adapter::JsonApi::FragmentCache.new.fragment_cache(root, cached_hash, non_cached_hash)
+          core_cached       = cached_hash.first
+          core_non_cached   = non_cached_hash.first
+          no_root_cache     = cached_hash.delete_if { |key, _value| key == core_cached[0] }
+          no_root_non_cache = non_cached_hash.delete_if { |key, _value| key == core_non_cached[0] }
+          cached_resource   = (core_cached[1]) ? core_cached[1].deep_merge(core_non_cached[1]) : core_non_cached[1]
+          hash = (root) ? { root => cached_resource } : cached_resource
```
groyoh: I know this code is pre-existing but could you give an explanation of what's going on here. I can't seem to understand what this block of code is supposed to do. I especially don't get why we would have a `root` here.

NullVoxPopuli:
needs rebase :-)

PR #1477, opened 2016-01-28
Fixed fragment_cached? method to check if caching

bdmac:
I noticed that fragment caching does not actually check if caching is
enabled as it seemingly should.

The way CachedSerializer#fragment_cached? worked previously would return
true even in an environment where caching was disabled as defined by
`ActiveModelSerializers.config.perform_caching`.

Added check for `_cache` like in the `cached?` method before checking
whether `_cache_only` or `_cache_except` is set.

There were no existing tests for any of these methods but it's a pretty
trivial change.

### DIFF
--- a/lib/active_model/serializer/adapter/cached_serializer.rb
+++ b/lib/active_model/serializer/adapter/cached_serializer.rb
@@ -24,7 +24,7 @@ module ActiveModel
         end
 
         def fragment_cached?
-          @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
+          @klass._cache && (@klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except)
         end
 
         def cache_key


bdmac:
BTW this happens because the implementation of the `cache` DSL sets the various options (in this case specifically `_cache_only` and `_cache_except` regardless of whether or not caching is actually enabled. Relevant code for reference:

```ruby
        def cache(options = {})
          self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
          self._cache_key = options.delete(:key)
          self._cache_only = options.delete(:only)
          self._cache_except = options.delete(:except)
          self._cache_options = (options.empty?) ? nil : options
        end
```

lib/active_model/serializer/adapter/cached_serializer.rb
```diff
@@ -24,7 +24,7 @@ def cached?
         end
 
         def fragment_cached?
-          @klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except
+          @klass._cache && (@klass._cache_only && !@klass._cache_except || !@klass._cache_only && @klass._cache_except)
```
bf4: I guess the idea was to check if fragment_caching was available independent of whether or not caching is available

bf4: really, the block should check for `@klass._cache ` first, I think. `return yield unless @klass._cache` guard clause or something

bdmac: Right maybe except I don't think you can fragment cache if caching isn't available so that seems moot. I think the answer to that question shouldn't be relevant if caching is disabled.

bf4:
@bdmac Would you mind adding a test for this and a changelog entry at the top of  'fixes'?

bf4:
thanks https://github.com/rails-api/active_model_serializers/pull/1477#issuecomment-176521410

bf4:
Merged with changelog and tests added in https://github.com/rails-api/active_model_serializers/commit/2789cc5221478a357225811356d815f4004255fc

PR #1478, opened 2016-01-28
[FIX] Serializers can now be defined *before* Rails initializes and cache store will be correctly set

bf4:
Update: https://github.com/rails-api/active_model_serializers/pull/1478#issuecomment-197703871 and https://github.com/rails-api/active_model_serializers/pull/1478#issuecomment-197706572 https://github.com/rails-api/active_model_serializers/pull/1478#issuecomment-201111318

This is kind of a dumb fix for caching that demonstrates the
need in Rails for Serializer::_cache to be set after the controller is loaded
when Serializer::cache has already been called.

## Notes on the issue

```ruby
# Initialize app before any serializers are defined, for sanity's sake.
# Otherwise, you have to manually set perform caching.
#
# Details:
#
# 1. Upon load, when AMS.config.perform_caching is true,
#    serializers inherit the cache store from ActiveModelSerializers.config.cache_store
# 1. If the serializers are loaded before Rails is initialized (`Rails.application.initialize!`),
#    these values are nil, and are not applied to the already loaded serializers
# 1. If Rails is initialized before any serializers are loaded, then the configs are set,
#    and are used when serializers are loaded
# 1. In either case, `ActiveModelSerializers.config.cache_store`, and
#    `ActiveModelSerializers.config.perform_caching` can be set at any time before the serializers
#    are loaded,
#    e.g.  `ActiveModel::Serializer.config.cache_store ||=
#      ActiveSupport::Cache.lookup_store(ActionController::Base.cache_store ||
#      Rails.cache || :memory_store)`
#    and `ActiveModelSerializers.config.perform_caching = true`
# 1. If the serializers are loaded before Rails is initialized, then,
#    you can set the `_cache` store directly on the serializers.
#    `ActiveModel::Serializer._cache ||=
#      ActiveSupport::Cache.lookup_store(ActionController::Base.cache_store ||
#      Rails.cache || :memory_store`
#    is sufficient.
#    Setting `_cache` to a truthy value will cause the CachedSerializer
#    to consider it cached, which will apply to all serializers (bug? :bug: )
#
# This happens, in part, because the cache store is set for a serializer
# when `cache` is called, and cache is usually called when the serializer is defined.
#
# So, there's now a 'workaround', something to debug, and a starting point.
Rails.application.initialize!

# HACK: Serializer::cache depends on the ActionController-dependent configs being set.
  require_relative 'fixtures'
end
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -43,7 +43,8 @@ Rake::TestTask.new do |t|
   t.libs << 'test'
   t.libs << 'lib'
   t.test_files = FileList['test/**/*_test.rb']
-  t.ruby_opts = ['-w -r./test/test_helper.rb']
+  t.ruby_opts = ['-r./test/test_helper.rb']
+  t.ruby_opts << ' -w' unless ENV['NO_WARN'] == 'true'
   t.verbose = true
 end
 
--- a/lib/active_model/serializer/caching.rb
+++ b/lib/active_model/serializer/caching.rb
@@ -99,34 +99,18 @@ module ActiveModel
           self._cache_options = options.empty? ? nil : options
         end
 
+        # Value is from ActiveModelSerializers.config.perform_caching. Is used to
+        # globally enable or disable all serializer caching, just like
+        # Rails.configuration.action_controller.perform_caching, which is its
+        # default value in a Rails application.
         # @return [true, false]
-        # We're using class variables here because it is a class attribute
-        # that is globally true for the `ActiveModel::Serializer` class; i.e. neither per subclass nor inherited.
-        #
-        # We're not using a class_attribute because of the special behavior in
-        # `perform_caching` setting itself to `ActiveModelSerializers.config.perform_caching`
-        # when first called if it wasn't first set.
-        #
-        # This is to allow us to have a global config that can be set any time before
-        # `perform_caching` is called.
-        #
-        # One downside of this, is that subsequent setting of the global config will not change
-        # `ActiveModel::Serializer.perform_caching`, but that should be an edge case that
-        # is easily handled.
-        #
-        # If you, reading this, can figure out how to have ActiveModel::Serializer always delegate
-        # `perform_caching` and `perform_caching=` to the global config, that would make a nice PR.
+        # Memoizes value of config first time it is called with a non-nil value.
         # rubocop:disable Style/ClassVars
         def perform_caching
-          return @@perform_caching if defined?(@@perform_caching)
-          self.perform_caching = ActiveModelSerializers.config.perform_caching
+          return @@perform_caching if defined?(@@perform_caching) && !@@perform_caching.nil?
+          @@perform_caching = ActiveModelSerializers.config.perform_caching
         end
         alias perform_caching? perform_caching
-
-        # @param [true, false]
-        def perform_caching=(perform_caching)
-          @@perform_caching = perform_caching
-        end
         # rubocop:enable Style/ClassVars
 
         # The canonical method for getting the cache store for the serializer.
--- a/lib/active_model/serializer/caching.rb
+++ b/lib/active_model/serializer/caching.rb
@@ -89,6 +89,10 @@ module ActiveModel
         # https://github.com/rails-api/active_model_serializers/pull/1249#issuecomment-146567837
         def cache(options = {})
           self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
+          serializer = self
+          ActiveSupport.on_load(:action_controller) do
+            serializer._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
+          end
           self._cache_key = options.delete(:key)
           self._cache_only = options.delete(:only)
           self._cache_except = options.delete(:except)
--- a/lib/active_model/serializer/caching.rb
+++ b/lib/active_model/serializer/caching.rb
@@ -5,7 +5,7 @@ module ActiveModel
 
       included do
         with_options instance_writer: false, instance_reader: false do |serializer|
-          serializer.class_attribute :_cache         # @api private : the cache object
+          serializer.class_attribute :_cache         # @api private : the cache store
           serializer.class_attribute :_fragmented    # @api private : @see ::fragmented
           serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key
           serializer.class_attribute :_cache_only    # @api private : when fragment caching, whitelists cached_attributes. Cannot combine with except
@@ -71,6 +71,7 @@ module ActiveModel
         #   when Rails.configuration.action_controller.perform_caching
         #
         # @params options [Hash] with valid keys:
+        #   cache_store    : @see ::_cache
         #   key            : @see ::_cache_key
         #   only           : @see ::_cache_only
         #   except         : @see ::_cache_except
@@ -88,15 +89,73 @@ module ActiveModel
         # @todo require less code comments. See
         # https://github.com/rails-api/active_model_serializers/pull/1249#issuecomment-146567837
         def cache(options = {})
-          self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
-          serializer = self
-          ActiveSupport.on_load(:action_controller) do
-            serializer._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
-          end
+          self._cache =
+            options.delete(:cache_store) ||
+            ActiveModelSerializers.config.cache_store ||
+            ActiveSupport::Cache.lookup_store(:null_store)
           self._cache_key = options.delete(:key)
           self._cache_only = options.delete(:only)
           self._cache_except = options.delete(:except)
           self._cache_options = options.empty? ? nil : options
+        end
+
+        # @return [true, false]
+        # We're using class variables here because it is a class attribute
+        # that is globally true for the `ActiveModel::Serializer` class; i.e. neither per subclass nor inherited.
+        #
+        # We're not using a class_attribute because of the special behavior in
+        # `perform_caching` setting itself to `ActiveModelSerializers.config.perform_caching`
+        # when first called if it wasn't first set.
+        #
+        # This is to allow us to have a global config that can be set any time before
+        # `perform_caching` is called.
+        #
+        # One downside of this, is that subsequent setting of the global config will not change
+        # `ActiveModel::Serializer.perform_caching`, but that should be an edge case that
+        # is easily handled.
+        #
+        # If you, reading this, can figure out how to have ActiveModel::Serializer always delegate
+        # `perform_caching` and `perform_caching=` to the global config, that would make a nice PR.
+        # rubocop:disable Style/ClassVars
+        def perform_caching
+          return @@perform_caching if defined?(@@perform_caching)
+          self.perform_caching = ActiveModelSerializers.config.perform_caching
+        end
+        alias perform_caching? perform_caching
+
+        # @param [true, false]
+        def perform_caching=(perform_caching)
+          @@perform_caching = perform_caching
+        end
+        # rubocop:enable Style/ClassVars
+
+        # The canonical method for getting the cache store for the serializer.
+        #
+        # @return [nil] when _cache is not set (i.e. when `cache` has not been called)
+        # @return [._cache] when _cache is not the NullStore
+        # @return [ActiveModelSerializers.config.cache_store] when _cache is the NullStore.
+        #   This is so we can use `cache` being called to mean the serializer should be cached
+        #   even if ActiveModelSerializers.config.cache_store has not yet been set.
+        #   That means that when _cache is the NullStore and ActiveModelSerializers.config.cache_store
+        #   is configured, `cache_store` becomes `ActiveModelSerializers.config.cache_store`.
+        # @return [nil] when _cache is the NullStore and ActiveModelSerializers.config.cache_store is nil.
+        def cache_store
+          return nil if _cache.nil?
+          return _cache if _cache.class != ActiveSupport::Cache::NullStore
+          if ActiveModelSerializers.config.cache_store
+            self._cache = ActiveModelSerializers.config.cache_store
+          else
+            nil
+          end
+        end
+
+        def cache_enabled?
+          perform_caching? && cache_store && !_cache_only && !_cache_except
+        end
+
+        def fragment_cache_enabled?
+          perform_caching? && cache_store &&
+            (_cache_only && !_cache_except || !_cache_only && _cache_except)
         end
       end
     end
--- a/test/active_model_serializers/adapter_for_test.rb
+++ b/test/active_model_serializers/adapter_for_test.rb
@@ -1,5 +1,7 @@
+require 'test_helper'
+
 module ActiveModelSerializers
-  class AdapterForTest < ActiveSupport::TestCase
+  class AdapterForTest < ::ActiveSupport::TestCase
     UnknownAdapterError = ::ActiveModelSerializers::Adapter::UnknownAdapterError
 
     def test_serializer_adapter_returns_configured_adapter
--- a/test/serializers/caching_configuration_test_isolated.rb
+++ b/test/serializers/caching_configuration_test_isolated.rb
@@ -1,0 +1,168 @@
+# Execute this test in isolation
+require 'support/isolated_unit'
+
+class CachingConfigurationTest < ActiveSupport::TestCase
+  include ActiveSupport::Testing::Isolation
+
+  setup do
+    require 'rails'
+    # AMS needs to be required before Rails.application is initialized for
+    # Railtie's to fire in Rails.application.initialize!
+    # (and make_basic_app initializes the app)
+    require 'active_model_serializers'
+    # Create serializers before Rails.application.initialize!
+    # To ensure we're testing that the cache settings depend on
+    # the Railtie firing, not on the ActionController being loaded.
+    create_serializers
+  end
+
+  def create_serializers
+    @cached_serializer = Class.new(ActiveModel::Serializer) do
+      cache skip_digest: true
+      attributes :id, :name, :title
+    end
+    @fragment_cached_serializer = Class.new(ActiveModel::Serializer) do
+      cache only: :id
+      attributes :id, :name, :title
+    end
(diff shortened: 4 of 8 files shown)


bf4:
e.g. work done in https://github.com/rails-api/active_model_serializers/pull/1412/files#diff-8c03da6a006ecb5ab61b873ef8f5cba3R76

bf4:
Fixes https://github.com/rails-api/active_model_serializers/issues/923

lib/active_model/serializer/caching.rb
```diff
@@ -89,6 +89,10 @@ def fragmented(serializer)
         # https://github.com/rails-api/active_model_serializers/pull/1249#issuecomment-146567837
         def cache(options = {})
           self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
+          serializer = self
+          ActiveSupport.on_load(:action_controller) do
```
bf4: duplicate since we can't rely on an app loading a controller

lib/active_model/serializer/caching.rb
```diff
@@ -89,6 +89,10 @@ def fragmented(serializer)
         # https://github.com/rails-api/active_model_serializers/pull/1249#issuecomment-146567837
         def cache(options = {})
           self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
```
bf4: The real problem is that calling this method should enable the cache via sentinel value, rather than setting the cache store dependent on an controller attribute.  Each serializer could still have its own cache_store (though I'm not sure how often that would be used, and it causes extra complexity here).

groyoh:
What about something like?

```ruby
def _cache
  return @cache if @cache_set
  @cache_set = true
  @cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
end
def _cache=(cache)
  @cache_set = true
  @cache = cache
end
```

AFAIK, the `_cache` method would never be called before the first resource is serialized. I think we can safely assume that at that point, the `ActiveModelSerializers.config.cache_store` and `ActiveModelSerializers.config.perform_caching` config should be already set. Thus, if we set the `@cache` at that point, we wouldn't need anykind of work-around. Adding `_cache=` then, we would still be able to set the cache manually if needed.

bf4:
Also a problem that _cache doubled as the serializer cache and a sentinel
to use caching

lib/active_model/serializer/caching.rb
bf4: Older versions of AMS used a method `perform_caching` named same as rails.  I think that might be a good idea

bf4: easiest impl, have a class attribute `perform_caching` that acts as the sentinel. That was we don't have to change the api much... caching can still be turning on/off per serializer, but configured at load and disabled/enabled before use

bf4:
ignore 4d6082f0917c7bd6ad2000e0bd5fcd7cc980b4c5.. more objects and loses performance, oops

NullVoxPopuli:
@bf4, could you rebase this and fix the rubocop warnings?

bf4:
I thought this was broken and I gave up for now

B mobile phone

NullVoxPopuli:
Does the issue still need to be resolved?

bf4:
@NullVoxPopuli absolutely yes

bf4:
@NullVoxPopuli @remear @groyoh I believe I fixed it!  There's definitely a bit of jankyness, so please comment if you see a better way.

A summary of our needs:

1. Each serializer can have its own cache store.  `_cache` is a class attribute
1. We only want the serializer cache to be enabled when `cache` is called and `perform_caching` is true.

Summary of problems:

1. Right now, `_cache` serves as both the cache_store and the sentinel value that caching is enabled for the serializer.
1. If the serializer is loaded before Rails initializes or if the Rails initializes before AMS is required, the railtie won't fire, `perform_caching`  will be falsy, and the cache store will never be set

Summary of solution:

1. See comments in diff. Basically, when `cached` is called, we now mark the serializer as 'cacheable', and instead of `_cache`, call `cache_store` to get the configure cache store. And when `cache_store` is called on a 'cacheable' serializer, it checks to see if the global cache store is set, and uses that. Phew

bf4:
## Benchmarking

Need to confirm if there's a performance hit.

### Before

Master HEAD db87f8d85e7a46ef195a555cb7e2509f1affc622

```bash
{
  "commit_hash": "db87f8d",
  "version": "0.10.0.rc4",
  "rails_version": "4.0.13",
  "benchmark_run[environment]": "2.2.3p173",
  "runs": [
    {
      "benchmark_type[category]": "caching on: caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 646.036,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1854
    },
    {
      "benchmark_type[category]": "caching off: caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 608.395,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1854
    },
    {
      "benchmark_type[category]": "caching on: non-caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 747.475,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1392
    },
    {
      "benchmark_type[category]": "caching off: non-caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 676.216,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1392
    }
  ]
}
```

### After

PR HEAD a7eca8160f629cdaf34a8ab41776d448b4b892ff

```bash
caching on: caching serializers: gc off 578.3836770108551/ips; 1851 objects
caching off: caching serializers: gc off 537.3985284808116/ips; 1851 objects
caching on: non-caching serializers: gc off 731.2126917485139/ips; 1392 objects
caching off: non-caching serializers: gc off 707.7475437630438/ips; 1392 objects
Benchmark results:
{
  "commit_hash": "a7eca81",
  "version": "0.10.0.rc4",
  "rails_version": "4.0.13",
  "benchmark_run[environment]": "2.2.3p173",
  "runs": [
    {
      "benchmark_type[category]": "caching on: caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 578.384,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1851
    },
    {
      "benchmark_type[category]": "caching off: caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 537.399,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1851
    },
    {
      "benchmark_type[category]": "caching on: non-caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 731.213,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1392
    },
    {
      "benchmark_type[category]": "caching off: non-caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 707.748,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1392
    }
  ]
}
```

lib/active_model/serializer/caching.rb
```diff
@@ -88,12 +89,74 @@ def fragmented(serializer)
...
           self._cache_only = options.delete(:only)
           self._cache_except = options.delete(:except)
           self._cache_options = options.empty? ? nil : options
         end
+
+        # @return [true, false]
+        # We're using class variables here because it is a class attribute
+        # that is globally true for the `ActiveModel::Serializer` class; i.e. neither per subclass nor inherited.
+        #
+        # We're not using a class_attribute because of the special behavior in
```
NullVoxPopuli: thanks for the explanation! I like to read the thoughts behind decisions

lib/active_model/serializer/caching.rb
```diff
@@ -88,12 +89,74 @@ def fragmented(serializer)
...
+        # `perform_caching` setting itself to `ActiveModelSerializers.config.perform_caching`
+        # when first called if it wasn't first set.
+        #
+        # This is to allow us to have a global config that can be set any time before
+        # `perform_caching` is called.
+        #
+        # One downside of this, is that subsequent setting of the global config will not change
```
NullVoxPopuli: just because the class var is memoized?

lib/active_model/serializer/caching.rb
```diff
@@ -88,12 +89,74 @@ def fragmented(serializer)
...
+        # `ActiveModel::Serializer.perform_caching`, but that should be an edge case that
+        # is easily handled.
+        #
+        # If you, reading this, can figure out how to have ActiveModel::Serializer always delegate
```
NullVoxPopuli: what was the issue with AMS.config not being delegable? is that the race condition you mention in the discussion side of this PR?

test/serializers/caching_configuration_test_isolated.rb
```diff
@@ -0,0 +1,168 @@
...
+    @non_cached_serializer = Class.new(ActiveModel::Serializer) do
+      attributes :id, :name, :title
+    end
+  end
+
+  class PerformCachingTrue < CachingConfigurationTest
+    setup do
+      # Let's make that Rails app and initialize it!
+      make_basic_app do |app|
+        app.config.action_controller.perform_caching = true
```
NullVoxPopuli: So this sets the class variable.

Do have this delegate to the global config, to we need to have the global config loaded first when AMS loads? (I have no idea what loads first currently)

NullVoxPopuli: why is this tied in to action_controller?

test/active_model_serializers/adapter_for_test.rb
```diff
@@ -1,5 +1,7 @@
+require 'test_helper'
+
 module ActiveModelSerializers
-  class AdapterForTest < ActiveSupport::TestCase
```
NullVoxPopuli: why is this change needed?

NullVoxPopuli:
Maybe something to look at, not sure if this'll work.

But, re: delegating to the config,

in active_model_serializers.rb#config, the Serializer class should autoload when the config method is called on, yeah?

in lib/ams/serializers.rb:

```ruby
require 'active_model/serializer/collection_serializer'
require 'active_model/serializer/array_serializer'
require 'active_model/serializer/error_serializer'
require 'active_model/serializer/errors_serializer'
require 'active_model/serializer/include_tree'
require 'active_model/serializer/associations'
require 'active_model/serializer/attributes'
require 'active_model/serializer/caching'
```
what if configuration was at the top? would that help?

or is the loading race condition elsewhere?

lib/active_model/serializer/caching.rb
bf4: No, class variables are considered an anti-pattern in Ruby.  This is distinct from class instance variables. Go and learn :)

lib/active_model/serializer/caching.rb
bf4: I didn't figure out how to get the serializer class to delegate to the config.

It's not really a race condition; is more of load order.

test/serializers/caching_configuration_test_isolated.rb
bf4: > So this sets the class variable.

no, `app.config.action_controller.perform_caching = true` sets that Rails config

> why is this tied in to action_controller?
 
That's where Rails puts controller-specific configs.  Many objects in Rails can be configured differently.  ActionController::Base.cache_store need not equal Rails.cache.  ActionController::Base.logger need not equal Rails.logger.

The cache that we want in AMS is the controller cache.  Whether it is the same as the Rails cache is not a detail we need to worry about.

test/active_model_serializers/adapter_for_test.rb
bf4: See commit.  Doesn't need to be in this PR, but doesn't really deserved its own.  Just something I saw when running files individually

bf4:
@NullVoxPopuli 

> in lib/ams/serializers.rb:
> what if configuration was at the top? would that help?

nope. we're talking about when Rails is initialized vs. when the serializers are loaded vs. when the controller is first loaded.  That's not a race condition. It's strictly determined by load order.  More details in the PR description

lib/active_model/serializer/caching.rb
```diff
@@ -88,12 +89,74 @@ def fragmented(serializer)
...
+          return nil if _cache.nil?
+          return _cache if _cache.class != ActiveSupport::Cache::NullStore
+          if ActiveModelSerializers.config.cache_store
+            self._cache = ActiveModelSerializers.config.cache_store
+          else
+            nil
+          end
+        end
+
+        def cache_enabled?
```
groyoh: :+1:

lib/active_model/serializer/caching.rb
groyoh: So what happens exactly if you have:

```ruby
def perform_caching?
  !!ActiveModelSerializers.config.perform_caching
end
```
? What does not work in such case?

lib/active_model/serializer/caching.rb
```diff
@@ -88,12 +89,74 @@ def fragmented(serializer)
         # @todo require less code comments. See
         # https://github.com/rails-api/active_model_serializers/pull/1249#issuecomment-146567837
         def cache(options = {})
-          self._cache = ActiveModelSerializers.config.cache_store if ActiveModelSerializers.config.perform_caching
+          self._cache =
+            options.delete(:cache_store) ||
```
groyoh: I'm not sure if the test cover this case.

bf4: which case? which test?

groyoh: I don't think there is a test for "Using the `cache_store` options" e.g.

```ruby
  cache cache_store: some_cache_store
end
```

Correct me if I'm wrong.

bf4: Nah, I just added that as undocumented feature that needs test :)

bf4: @groyoh rebased with test added

lib/active_model/serializer/caching.rb
bf4: Done @groyoh

bf4:
Benchmark remains consistent with https://github.com/rails-api/active_model_serializers/pull/1478#issuecomment-197706572:

```
bin/bench -r 5
caching on: caching serializers: gc off 636.6898235724902/ips; 1851 objects
caching off: caching serializers: gc off 583.3875333741602/ips; 1851 objects
caching on: non-caching serializers: gc off 740.5398081283838/ips; 1392 objects
caching off: non-caching serializers: gc off 720.725015737152/ips; 1392 objects
caching on: caching serializers: gc off 633.8812177805845/ips; 1851 objects
caching off: caching serializers: gc off 587.4968492563474/ips; 1851 objects
caching on: non-caching serializers: gc off 785.4658865833786/ips; 1392 objects
caching off: non-caching serializers: gc off 738.8961641332226/ips; 1392 objects
caching on: caching serializers: gc off 645.0143440480517/ips; 1851 objects
caching off: caching serializers: gc off 612.072503541359/ips; 1851 objects
caching on: non-caching serializers: gc off 792.3932952308679/ips; 1392 objects
caching off: non-caching serializers: gc off 748.0955311576615/ips; 1392 objects
caching on: caching serializers: gc off 633.5915591875073/ips; 1851 objects
caching off: caching serializers: gc off 606.7087955734825/ips; 1851 objects
caching on: non-caching serializers: gc off 797.7486292204281/ips; 1392 objects
caching off: non-caching serializers: gc off 752.7270149276842/ips; 1392 objects
caching on: caching serializers: gc off 649.7175402003957/ips; 1851 objects
caching off: caching serializers: gc off 605.4492794365647/ips; 1851 objects
caching on: non-caching serializers: gc off 788.2106786368261/ips; 1392 objects
caching off: non-caching serializers: gc off 746.5016796342492/ips; 1392 objects
Benchmark results:
{
  "commit_hash": "e38b362",
  "version": "0.10.0.rc4",
  "rails_version": "4.0.13",
  "benchmark_run[environment]": "2.2.3p173",
  "runs": [
    {
      "benchmark_type[category]": "caching on: caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 649.718,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1851
    },
    {
      "benchmark_type[category]": "caching off: caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 612.073,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1851
    },
    {
      "benchmark_type[category]": "caching on: non-caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 797.749,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1392
    },
    {
      "benchmark_type[category]": "caching off: non-caching serializers: gc off",
      "benchmark_run[result][iterations_per_second]": 752.727,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 1392
    }
  ]
}
```

lib/active_model/serializer/caching.rb
groyoh: Thanks :+1:

Rakefile
```diff
@@ -43,7 +43,8 @@ Rake::TestTask.new do |t|
   t.libs << 'test'
   t.libs << 'lib'
   t.test_files = FileList['test/**/*_test.rb']
-  t.ruby_opts = ['-w -r./test/test_helper.rb']
+  t.ruby_opts = ['-r./test/test_helper.rb']
+  t.ruby_opts << ' -w' unless ENV['NO_WARN'] == 'true'
```
groyoh: This should be already on master? Weird that it still shows in the diff.

test/serializers/caching_configuration_test_isolated.rb
```diff
@@ -0,0 +1,168 @@
...
+    end
+
+    test 'the cached serializer does not have fragment_cache_enabled?' do
+      refute @cached_serializer.fragment_cache_enabled?
+    end
+
+    test 'the non-cached serializer cache_store is nil' do
+      assert_equal nil, @non_cached_serializer._cache
+      assert_equal nil, @non_cached_serializer.cache_store
+      assert_equal nil, @non_cached_serializer._cache
```
groyoh: Almost thought this was a mistake. Maybe these tests require a bit of documentation or a different description such `"the non-cached serializer cache_store remains nil"` ?

groyoh:
Looks good to me :tada:

test/serializers/caching_configuration_test_isolated.rb
bf4: good point.  `cache_store` does not set nil `_cache` to anything

Issue #1487, opened 2016-01-30
How to get to_json to behave without render?


dudo:
In my controller, I can convince the app to use my serializers:
```
def show
  @user = User.find_by_id(params[:id])
  render json: @user
end
```
But for the life of me, I can't seem to figure out how to get the object to use the serializer on command. I've tried the following to no avail:
```
@user.to_json
@user.as_json
@user.serializable_hash

UserSerializer.new(@user)
UserSerializer.new(@user).attributes (this almost works, but I have a has_many that I can't get to show up)
```
What am I missing?? This is driving me nuts.

domitian:
You can't, as it is mentioned in this document [https://github.com/domitian/active_model_serializers/blob/design-docs-from-textile-to-md/docs/DESIGN.md#the-most-basic-serializer](Basic Serializer) in design docs.

> Rails will transparently use your serializer when you use render :json in your controller.
So it won't use your serializers `as_json` and `serializable_hash` methods when you are calling from the console, it will by default use the active record methods.

bf4:
@dudo What version of AMS are you on? https://github.com/rails-api/active_model_serializers/blob/master/CONTRIBUTING.md

The docs describe how to serialize outside a controller https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/outside_controller_use.md

@domitian the file you linked is the original design doc from 2011

dudo:
0.10.0.rc4

```
UserSerializer.new(User.first).as_json
  User Load (0.9ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> {"object"=>{"id"=>1, "name"=>"Brett Dudo", "plaid_public_token"=>nil, "created_at"=>Thu, 21 Jan 2016 05:28:45 UTC +00:00, "updated_at"=>Wed, 03 Feb 2016 05:27:23 UTC +00:00, "email"=>"brett@dudo.io", "public_token"=>"..."}, "instance_options"=>{}, "root"=>nil, "scope"=>nil}
```

```
class UserSerializer < ActiveModel::Serializer
  attributes :id, :name, :email, :monthly_income_amount, :monthly_available_balance

  has_many :current_transactions
end
```

Hopefully that describes the mismatch. I'm using Devise, not sure if that messes with serialization. Also, `plaid_public_token` is on there, which I'm not sure how... the column is actually just `public_token`.

bf4:
[...]

PR #1490, opened 2016-02-01
Design docs file format changed from textfile to github md

domitian:
`design.textile` file in docs folder is now `design.md` with github flavoured markdown syntax. Also fixed some  typos and grammatical mistakes in that file.


domitian:
Not able to add label to the pull request.

bf4:
Thanks for this @domitian 

I'm not sure there's much value to changing it to `md`, since it's really just an archival doc, but since you did the work, I'll take a look at it.

Based on your comment in #1487 it looks like a useful change would be make it more clear this not the current design.  The current design is the architecture doc.

domitian:
@bf4 got your point, I changed the name of the doc to `LEGACY-DESIGN.md` which should make it clear that it is not current design doc.
Added to the start of this doc that this is not current design in bold case and to have a look at Architecture doc.

bf4:
Thinking about the file names .... Original_design and rename architecture to design? Initial_design?

domitian:
yeah, for renaming architecture to design is good.
legacy_design if feel will give more meaning to the other doc than initial/original design, we could also think of `old_design`

bf4:
initial_? legacy and old don't really convey that it was the original design doc. Maybe in architecture just link to [serializers.textile in Rails (before reverted)](https://github.com/rails/rails/blob/fcacc6986ab60f1fb2e423a73bf47c7abd7b191d/railties/guides/source/serializers.textile) or [README.textile in first 0.0.1 commit](https://github.com/rails-api/active_model_serializers/commit/d72b66d4c5355b0ff0a75a04895fcc4ea5b0c65e?short_path=8aa641a#diff-8aa641a619b8ebeca0116d41774daf45) Or the [DESIGN.textline form 0-8-stable](https://github.com/rails-api/active_model_serializers/blob/0-8-stable/DESIGN.textile)? which was the source of this?

bf4:
while you're in here or in a new PR, we also need to change the yard docs links in the readme to http://www.rubydoc.info/github/rails-api/active_model_serializers/v0.10.0.rc4 since master doesn't appear to be updating.. and maybe also http://www.rubydoc.info/github/rails-api/active_model_serializers/0-9-stable and http://www.rubydoc.info/github/rails-api/active_model_serializers/0-8-stable

domitian:
@bf4 created a new PR for the Yard docs links #1491

domitian:
As per this [commit](https://github.com/rails-api/active_model_serializers/commit/7d678844ae6900fa6cd8d1615f338d905c6dd73a) from you in the past, the source is 0.8-stable.
My suggestion is create a design folder and keep design documents for each version.

bf4:
Well, the doc was copied from there but it originated in rails before 0.0.1. The readme references the readme's of 0.8 and 0.9 which already serve the purpose of their design

B mobile phone

domitian:
So should we just remove the file and add references in architecture doc about the past design?

bf4:
That's what I'm thinking now

B mobile phone

domitian:
So should we link it and remove the file then?

domitian:
Can I remove the file and add link in architecture.md then?

bf4:
> Can I remove the file and add link in architecture.md then?

@domitian Yeah, that sounds good.

bf4:
@domitian Are you still interested in finishing this PR? https://github.com/rails-api/active_model_serializers/pull/1490#issuecomment-179321838

Issue #1380, opened 2015-12-16
Add 'why doesn't the Serializer in 0.10 have to_json/as_json/serializable_hash' to common questions


bf4:
see [common issues and resolutions section in contributing](https://github.com/rails-api/active_model_serializers/blob/master/CONTRIBUTING.md#common-issues-and-resolutions)

So, tl;dr the serializer has a different interface because it is different.

You could monkey patch the serializer with a call to the serializable resource and the attributes adapter?

The serializer in the 0.10 series does have `attributes` and `associations` methods.

and see https://github.com/rails-api/active_model_serializers/blob/master/docs/ARCHITECTURE.md

and perhaps https://github.com/rails-api/active_model_serializers/pull/1371 (until it is merged, then ask me to update this issue).

and ref https://github.com/rails-api/active_model_serializers/issues/1321#issuecomment-165177508

bf4:
There's no demand for this. Closing as stale.

PR #1494, opened 2016-02-03
Make serializers serializable, step 1.

bf4:
I think this is a first step in a good idea...

This comes up not infrequently, especially as it was an interface in previous releases.

Now, a serializer responds to `serializable_hash`, `as_json`, `to_json`, `to_h`, and `to_hash` as is expected from experience, and sorta makes sense, as well. (to_h and to_hash are sentinels Ruby uses for generating json etc)

e.g. https://github.com/rails-api/active_model_serializers/issues/1487

Obviously it needs tests, but I wanted to just get the idea out there.

re step 2, it's non-trivial.. I futzed a bit in https://github.com/bf4/active_model_serializers/compare/serializable_serializer...serializable_serializers and https://github.com/bf4/active_model_serializers/compare/serializable_serializer...serializable_serializers_take1

ref:

- https://github.com/rails/rails/blob/v5.0.0.beta2/activemodel/lib/active_model/serialization.rb
- https://github.com/rails/rails/blob/v5.0.0.beta2/activemodel/lib/active_model/serializers/json.rb
- https://github.com/rails/rails/blob/v5.0.0.beta2/activerecord/lib/active_record/serialization.rb#L11-L17


lib/active_model/serializer.rb
```diff
@@ -113,6 +113,56 @@ def initialize(object, options = {})
...
     end
 
+    # @return [Hash] containing the attributes and first level
+    # associations, similar to how ActiveModel::Serializers::JSON is used
+    # in ActiveRecord::Base.
+    #
+    # TODO: Move to here the Attributes adapter logic for
+    # +serializable_hash_for_single_resource(options)+
+    # and include <tt>ActiveModel::Serializers::JSON</tt>.
+    # So that the below is true:
```
bf4: > and include <tt>ActiveModel::Serializers::JSON</tt>. So that the below is true:

this is really a new feature that probably be implemented

lib/active_model/serializer.rb
```diff
@@ -113,6 +113,56 @@ def initialize(object, options = {})
...
+      adapter_opts ||= {}
+      adapter = ActiveModel::Serializer::Adapter.create(self, { include: '*' }.merge!(adapter_opts))
+      adapter.serializable_hash
+    end
+    alias_method :to_hash, :serializable_hash
+    alias_method :to_h,    :serializable_hash
+
+    # @see #serializable_hash
+    # TODO: When moving attributes adapter logic here, @see #serializable_hash
+    # So that the below is true:
```
bf4: >  So that the below is true:

Same comment as https://github.com/rails-api/active_model_serializers/pull/1494/files#r51977412

lib/active_model/serializer.rb
```diff
@@ -113,6 +113,56 @@ def initialize(object, options = {})
       end
     end
 
+    # @return [Hash] containing the attributes and first level
```
bf4: Arguably these two methods should now be in one of the modules.. or in their own 'serializable' module..

lib/active_model/serializer.rb
```diff
@@ -113,6 +113,56 @@ def initialize(object, options = {})
...
+    #     # To include the result of some method calls on the model use :methods:
+    #     serializer.as_json(methods: :permalink)
+    #
+    #     # To include associations use :include:
+    #     serializer.as_json(include: :posts)
+    #     # Second level and higher order associations work as well:
+    #     serializer.as_json(include: { posts: { include: { comments: { only: :body } }, only: :title } })
+    def serializable_hash(adapter_opts = nil)
+      adapter_opts ||= {}
+      adapter = ActiveModel::Serializer::Adapter.create(self, { include: '*' }.merge!(adapter_opts))
```
bf4: I'm somewhat interested in looking at the behavior of the adapter opts here.. the `include` doesn't seem to do anything

bf4:
ref: https://github.com/rails-api/active_model_serializers/issues/1380

Issue #1409, opened 2015-12-30
Run all branches against JRuby on CI


bf4:
ref: https://github.com/rails-api/active_model_serializers/commit/8bddef6

Diff should be something like:

```diff
rvm:
  - 2.0.0
  - 2.1
  - 2.2.3
  - 2.3.0
  - ruby-head
+  - jruby-9.0.4.0
    - rbx-2
  
+ before_install:
+  - '[ "$JRUBY_OPTS" != "" ] && export JRUBY_OPTS="-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug"'
+  - 'echo "JRUBY_OPTS: ''$JRUBY_OPTS''"'
+
  install: bundle install --path=vendor/bundle --retry=3 --jobs=3
  cache:
    directories:
 @@ -33,8 +38,6 @@ matrix:
    include:
    - rvm: 2.2
      env: CAPTURE_STDERR=true
-  - rvm: jruby-9.0.4.0
-    env: JRUBY_OPTS='-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug'
```

PR #1496, opened 2016-02-05
Run all branches against JRuby on CI

nadavshatz:
Meant to resolve https://github.com/rails-api/active_model_serializers/issues/1409

- Added recent JRuby versions to `.travis.yml`
- Added OracleJDK8 and OracleJDK7 (Travis default) builds
- Added Changelod line under misc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,14 @@ rvm:
   - 2.3.0
   - ruby-head
   - rbx-2
+  - jruby-9.0.4.0
+  - jruby-head
+
+jdk:
+  - oraclejdk8
+
+before_install:
+  - '[ "$JRUBY_OPTS" != "" ] && export JRUBY_OPTS="--dev -Xcli.debug=true --debug"'
 
 install: bundle install --path=vendor/bundle --retry=3 --jobs=3
 cache:
@@ -30,10 +38,16 @@ matrix:
     env: RAILS_VERSION=master
   - rvm: 2.1
     env: RAILS_VERSION=master
-  include:
   - rvm: jruby-9.0.4.0
-    env: JRUBY_OPTS='-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug'
+    env: RAILS_VERSION=master
+  - rvm: jruby-9.0.4.0
+    env: RAILS_VERSION=4.0
+  - rvm: jruby-head
+    env: RAILS_VERSION=master
+  - rvm: jruby-head
+    env: RAILS_VERSION=4.0
   allow_failures:
     - rvm: ruby-head
+    - rvm: jruby-head
     - rvm: rbx-2
   fast_finish: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Fixes:
 - [#1488](https://github.com/rails-api/active_model_serializers/pull/1488) Require ActiveSupport's string inflections (@nate00)
 
 Misc:
+- [#1496](https://github.com/rails-api/active_model_serializers/pull/1496) Run all branches against JRuby on CI (@nadavshatz)
 - [#1559](https://github.com/rails-api/active_model_serializers/pull/1559) Add a deprecation DSL. (@bf4 @groyoh)
 - [#1543](https://github.com/rails-api/active_model_serializers/pull/1543) Add the changes missing from #1535. (@groyoh)
 - [#1535](https://github.com/rails-api/active_model_serializers/pull/1535) Move the adapter and adapter folder to


bf4:
@nadavshatz Nice! :+1:  do you think all these runs are worth the cost to the CI provider and time of the run?  I've never tested against so many JDK's or point releases of JRuby.

.travis.yml
```diff
@@ -9,6 +9,13 @@ rvm:
   - 2.3.0
   - ruby-head
   - rbx-2
+  - jruby-9.0.4.0
+  - jruby-9.0.5.0
```
bf4: are they different enough that it's worth having both?

bf4:
It also seems the Rails 4.0 is missing from the jruby 9k series... (only run on jruby-head). This is a version that often fails differently, so...

nadavshatz:
@bf4 First off thanks for the quick response!

Regarding the many JDK:
We use OracleJDK8, Travis has for default 7, which i think some ubuntu versions do as well, which means many people will use. I can change it to only use the latest (8) if you prefer.

Re many point releases:
I was only going to use the latest stable release (9.0.5.0) but we already encountered regression issues with it and I wanted the tests to represent at least 2 versions. Once Jruby 9.1.0.0 is release we can change it to that and the last 9.0... version.

Rails 4.0 was failing, I thought that considering it is so old there is no need to add it, but I can put it back if you'd like. should it not be allowed to fail? Also - Jruby 9.0.x.0 is ruby 2.2 compatible and I wasn't sure if that is the issue for Rails 4.0.

Let me know what you think and I'll update the PR!

bf4:
Well, we still support 4.0, so that needs to remain under test... I honestly don't know the value of testing the different jdks... I was going to ref Rails but it looks like they don't test JRuby at all anymore https://github.com/rails/rails/commit/0e8c04529159522aadbfe4fe71ea3d326df07d52

nadavshatz:
I removed OracleJDK7 and Added back Rails 4.0, is it possible that Rails 4.0 doesn't work with Ruby 2.2? I'm not sure why the tests are failing.

.travis.yml
groyoh: One `jruby-9xxx` should be enough IMO. About having `jruby-9xxx` and `jruby-head`, I think it's generally a good idea to keep both. On the other side, I wonder if keeping `2.0.0` is necessary.

groyoh: We could also maybe consider using an `include` so that we don't run jrubies against all rails version.

nadavshatz: @groyoh 
`jruby-9xxx` and `jruby-91xx` (coming soon) are two different `ruby` versions. but for now, having just one `jruby-9xxx` should be enough, i agree.

regarding the `include` - you mean that we won't run `jruby` against Rails 4.0 ?

bf4: Well, Rails doesn't ;) https://github.com/rails/rails/pull/23499

bf4: So, diff will be something like

```diff
   - 2.3.0
   - ruby-head
   - rbx-2
+  - jruby-9.0.5.0
+
+jdk:
+  - oraclejdk8
+
+before_install:
+  - '[ "$JRUBY_OPTS" != "" ] && export JRUBY_OPTS="-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug" && echo "JRUBY_OPTS: \'$JRUBY_OPTS\'"'
 
 install: bundle install --path=vendor/bundle --retry=3 --jobs=3
 cache:
@@ -30,10 +40,12 @@ matrix:
     env: RAILS_VERSION=master
   - rvm: 2.1
     env: RAILS_VERSION=master
   - rvm: jruby-9.0.4.0
-    env: JRUBY_OPTS='-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug'
+    env: RAILS_VERSION=5.0.0.beta2
+  - rvm: jruby-head
   allow_failures:
     - rvm: ruby-head
+    - rvm: jruby-head
     - rvm: rbx-2
+   exclude:
+     - rvm: jruby-9.0.5.0
+       env: RAILS_VERSION=4.0
   fast_finish: true
```

nadavshatz:
@bf4 I've updated the PR and i think its now how you wanted it ?

nadavshatz:
I'll pay attention and update to add/change to the latest stable versions as they come out with new PR's

.travis.yml
```diff
@@ -30,10 +38,18 @@ matrix:
...
-  include:
-  - rvm: jruby-9.0.4.0
-    env: JRUBY_OPTS='-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug'
+  - rvm: jruby-9.0.5.0
+    env: RAILS_VERSION=master
+  - rvm: jruby-9.0.5.0
+    env: RAILS_VERSION=4.0
+  - rvm: jruby-head
+    env: RAILS_VERSION=4.0
+  - rvm: jruby-9.0.5.0
```
groyoh: This is a duplicate from above at line 41.

groyoh:
This look good to me (apart the duplicate). @bf4 any objection?

.travis.yml
nadavshatz: Good catch, thanks!

nadavshatz:
@groyoh fixed the duplicate, thanks for that!

.travis.yml
```diff
@@ -9,6 +9,14 @@ rvm:
...
   - ruby-head
   - rbx-2
+  - jruby-9.0.5.0
+  - jruby-head
+
+jdk:
+  - oraclejdk8
+
+before_install:
+  - '[ "$JRUBY_OPTS" != "" ] && export JRUBY_OPTS="-Xcompat.version=2.0 --server -Xcompile.invokedynamic=false -Xcli.debug=true --debug"'
```
bf4: plz change to `9.0.4.0` since travis caches it, faster, and JRUBY_OPTS to `'--dev --debug --Xcli.debug=true'`. see https://github.com/rails/rails/pull/23499#discussion_r55161731

nadavshatz: I pushed a fix for these, thanks!

Let me know if you want me to squash the commits into one.

bf4: Squash plz :)

nadavshatz: Done!

NullVoxPopuli:
LGTM :+1: I'll merge if no one has objections.

bf4:
@NullVoxPopuli please wait for rebase

.travis.yml
bf4: @nadavshatz you say done, but maybe you didn't push it yet?

NullVoxPopuli:
I need to make a checklist for reviewing.

nadavshatz:
@bf4 @NullVoxPopuli Rebased and squashed. Commit should be clean now.

Please let me know if there is anything else missing/to be fixed.

Cheers :beers:

NullVoxPopuli:
cool, thanks, @nadavshatz -- just gotta wait for travis/appveyor

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Misc:
 - [#1535](https://github.com/rails-api/active_model_serializers/pull/1535) Move the adapter and adapter folder to
   active_model_serializers folder and changes the module namespace. (@domitian @bf4)
 - [#1497](https://github.com/rails-api/active_model_serializers/pull/1497) Add JRuby-9000 to appveyor.yml(@corainchicago)
+- [#1496](https://github.com/rails-api/active_model_serializers/pull/1496) Run all branches against JRuby on CI
```
bf4: @nadavshatz would you mind adding this to the top of the list and give yourself credit? `on CI (@nadavshatz)`. Thanks.

nadavshatz: Done! thanks!

PR #1503, opened 2016-02-09
[0.8-stable] additional documentation for POROs

feministy:
hey pals! :wave: 

i know, i know, it's not documentation for master :disappointed:, but still, it's documentation! :smile:

it's mostly clarification on when you need to take additional steps to use AM:S with POROs/alternative ORMs, along with a small example to show *how* to implement the required steps to use it. (ok, i also removed the apostrophes, you caught me)

well, you can just read it on the next tab :relaxed: 

let me know if you'd like any changes or adjustments here and i'm happy to do so!

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,17 +45,34 @@ the serializer generator:
 $ rails g serializer post
 ```
 
-### Support for PORO's and other ORM's.
+### Support for POROs and other ORMs.
 
 Currently `ActiveModel::Serializers` adds serialization support to all models
-that descend from `ActiveRecord` or include `Mongoid::Document`. If you are
-using another ORM, or if you are using objects that are `ActiveModel`
-compliant but do not descend from `ActiveRecord` or include
-`Mongoid::Document`, you must add an include statement for
-`ActiveModel::SerializerSupport` to make models serializable. If you
-also want to make collections serializable, you should include
+that descend from `ActiveRecord` or include `Mongoid::Document`. If you are:
+
+- using another ORM, or
+- using objects that are `ActiveModel` compliant but do not descend from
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,18 +45,35 @@ the serializer generator:
 $ rails g serializer post
 ```
 
-### Support for PORO's and other ORM's.
```
bf4: Hi @feministy I'm curious if you have a reference for this grammatical change or if it's just what you're used to.   I prefer the current state, myself, but I believe there's no consensus, right?

bf4:
@feministy Thanks! :+1: Looks great! Just one comment and would you like to add yourself to the changelog?

README.md
feministy: it's what i'm used to - apostrophe's signal possession or a contraction, not a plural. i can swap back if it would be confusing to users!

feministy:
@bf4 can do! sorry i missed that - will update this evening and push back up. thanks for the quick review! :smile:

README.md
bf4: Apparently, it's a matter of some debate, but your emendation is 'more correct'.  https://english.stackexchange.com/questions/55970/plurals-of-acronyms-letters-numbers-use-an-apostrophe-or-not

feministy: i love that you can be _more correct_ about something so tiny :joy: grammar is so cooool

feministy:
@bf4 updated the changelog as requested under "unreleased" - let me know if that needs to be moved to a different header

bf4:
Thanks... If you want a few more prs the changelog for 08 is lacking info that's in mastee iirc, also 09 - onmephone

PR #1454, opened 2016-01-21
Add support for relationship-level links and meta attributes.

beauby:
Based on top of #1447.
Allows the following:
```ruby
has_many :posts do
  link :self do
    href '//example.com/link_author/relationships/posts'
    meta stuff: 'value'
  end
  link :related do
    href '//example.com/link_author/posts'
    meta count: object.posts.count
  end
  include_data false
end
```

PR #1504, opened 2016-02-10
Follow up to #1454

groyoh:
PR #1454 was merged with some missing fixes. All these fixes are addressed by this PR:
- Rename `ActiveModel::Serializer::Adapter::JsonApi::Association` to `ActiveModel::Serializer::Adapter::JsonApi::Relationship`
- `Move ActiveModel::Serializer::Adapter:: JsonApi::Relationship` and `ActiveModel::Serializer::Adapter::JsonApi::ResourceIdentifier` to `ActiveModel::Serializer::Adapter::JsonApi::ApiObjects` module
- Add unit test for `ActiveModel::Serializer::Adapter::JsonApi::Relationship`
- Add unit test for `ActiveModel::Serializer::Adapter::JsonApi::ResourceIdentifier`

@bf4 here you go ;)


test/adapter/json_api/api_objects/relationship_test.rb
```diff
@@ -0,0 +1,168 @@
...
+                    id: '2',
+                    type: 'posts'
+                  }
+                ]
+              }
+              test_relationship(expected, options: { include_data: true })
+            end
+
+            def test_relationship_data_not_included
+              test_relationship({}, options: { include_data: false })
```
groyoh: This case should probably raise an error since JSONAPI does not allow the relationship to be empty (http://jsonapi.org/format/#document-resource-object-relationships). I'll let you guys decide how to procede with this one.

bf4: Depends if the conditions are programmer error.  Generated nil values aren't included... So? In this case, is it always programmer error? Should it create an error response?

bf4:
Awesome, thanks! Let's keep up the momentum

bf4:
@groyoh Can you add a Changelog entry at the top of whichever section makes sense to you?

test/adapter/json_api/api_objects/relationship_test.rb
groyoh: In this case, it would most likely be a programmer mistake who would do something like:

```ruby
has_one :author do
  include_data false
end
```

which is kind of useless but that may happen. This should be the concern of AMS to make sure that the serializer either does not serializer the whole relationship in such case (not usefull in my opinion) or raise an error to let the programmer know that this is a misuse of the gem.

groyoh:
@bf4 I'll add it to the changelog ;)

groyoh:
@bf4 I added the changelog. Should I also update the documentation or create another PR for this?

bf4:
If you're offering docs now is good :)

B mobile phone

groyoh:
@bf4 since this change mainly concerns JSONAPI, should it go into `docs/serializers.md`  next to the associations documentation or somewhere else?

bf4:
@groyoh right now the serializers include some jsonapi concerns.  As long as it's part of the serializer, I think it makes sense to put in that file, but include more usage info on it in the jsonapi or adapter section

groyoh:
@bf4 @beauby  I updated the PR because I figured out that #1454 actually had a bug. When using a block for a relationship like:

```ruby
has_one :bio do
  link :self, "whatever"
end
# => { data: nil, links: { self: "whatever" } }
```
would always render `data` as nil because we would take the block return value as the relationship value. I added some specs for it too.

I am unsure about the solution though, let me know what you think.

lib/active_model/serializer/reflection.rb
```diff
@@ -42,17 +42,17 @@ def initialize(*)
 
       def link(name, value = nil, &block)
         @_links[name] = block || value
-        nil
+        :nil
```
groyoh: Not sure if this should be called `:nil` or something else. I actually wanted to have your input on this as I was not too creative here.

lib/active_model/serializer/reflection.rb
```diff
@@ -60,7 +60,12 @@ def value(serializer)
         @scope = serializer.scope
 
         if block
-          instance_eval(&block)
+          block_value = instance_eval(&block)
+          if block_value == :nil
```
groyoh: This assumes that a relationship value MUST never be a `Symbol`. I guess this is a safe choice but let me know if you disagree.

test/adapter/json_api/relationship_test.rb
```diff
@@ -0,0 +1,173 @@
+require 'test_helper'
```
groyoh: I added this file to add proper test for each relationship feature. I think some features are currently missing (e.g. defining the relationship using the block return value) and will add them ASAP.

test/adapter/json_api/relationship_test.rb
```diff
@@ -0,0 +1,173 @@
...
+              link :related do
+                ids = object.locations.map!(&:id).join(',')
+                href "//example.com/locations/#{ids}"
+              end
+            end
+
+            has_many :posts do
+              link :related do
+                ids = object.posts.map!(&:id).join(',')
+                href "//example.com/posts/#{ids}"
```
groyoh: I also figured out that there is a bug here. Using:

```ruby
has_many :posts do
  link :related do
    ids = object.posts.map!(&:id).join(',')
    meta ids: ids
  end
end
# => { links: { posts: { related: { href: nil, meta: { ids: [1337] } }  } } }
```
results in a `nil` href. According to JSONAPI, we should be able to use `meta` without `href` (http://jsonapi.org/format/#document-links).

groyoh:
Also please let me know if I should squash the commits.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -6,10 +6,9 @@ class JsonApi < Base
         autoload :PaginationLinks
         autoload :FragmentCache
         autoload :Link
-        autoload :Association
-        autoload :ResourceIdentifier
         autoload :Meta
         autoload :Deserialization
+        require 'active_model/serializer/adapter/json_api/api_objects'
```
bf4: So, in https://github.com/rails-api/active_model_serializers/pull/1004/files#r52563837 I noticed that we've already started a pattern of sticking the json_api objects in the json_api folder. e.g. above we have Meta and Link which are also json_api object.

In code I removed I move that object into `json_api/jsonapi.rb`.  What do you think about just sticking also the api objects in the `json_api` folder as we've essentially already been doing?

groyoh: It's fine by me. Should I still keep the `ApiObjects` namespace?

scharfie:
Hello.  Apologies if this should be reported somewhere else, but I just tried this new feature and it seems to be breaking the `data` attribute inside the relationship any time a block is passed to the relationship definition.

For example, if you do:

```
has_many :comments

def comments
  # return an array of Comment objects
end
```

Then the `relationships / comments / data` object will have the expected array of type and id attributes for associated comments.

However, if you change the relationship to the following:

```
has_many :comments do
  include_data true # or even just an empty block here
end
```

Then `relationships / comments / data` is always `null`.  In essence, the very presence of a block on the relationship seems to be breaking the resulting payload.

groyoh:
@scharfie have you tried with master or with my [branch](https://github.com/groyoh/active_model_serializers/tree/association-blocks)? #1454 indeed contained a bug and this PR should fix it.

scharfie:
@groyoh Sorry, I just realized this - your branch does resolve the issue.  Nice work :+1:

lib/active_model/serializer/reflection.rb
bf4: this is to signify something like  `:block_returned_nil_value`?

lib/active_model/serializer/reflection.rb
bf4: I think that's right

test/adapter/json_api/relationship_test.rb
bf4: awesome!

test/adapter/json_api/relationship_test.rb
bf4: :exclamation: :scream: :100:

test/adapter/json_api/api_objects/relationship_test.rb
bf4: **FOLLOWUP**

lib/active_model/serializer/adapter/json_api.rb
bf4: **NEEDS FOLLOWUP**

bf4:
Merged!

lib/active_model/serializer/reflection.rb
groyoh: Yes. For more details, there are two use cases to consider for the relationship block:
* Where the block return the value we want to use for the relationship:
```ruby
has_one :author do
  meta some: :thing
  object.own_author
end
```
* Where the block return some value to define that we want to use the default relationship:
```ruby
has_one :author do
  meta some: :thing
end
```

For the latest, I defined the value to be `:nil` but it could be change to anything else if a better suggestion pops up since this is the internal API.

bf4: Maybe create an issue or start a pr to discuss any improvements 

B mobile phone

PR #1511, opened 2016-02-11
Add benchmarks for adapters

ianks:
I was curious how the adapters stacked up against each other performance wise. While doing this, I realized it would be useful to maintain some simple benchmark tests that can be used during development. This one is a simple benchmark which tests the performance of each adapter. If you are interested in the results from my machine: [here they are](https://gist.github.com/ianks/7c61f7ce767663a9560a).

### DIFF
--- a/bench/adapters.rb
+++ b/bench/adapters.rb
@@ -1,0 +1,52 @@
+#!/usr/bin/env ruby
+
+require 'rubygems'
+require 'bundler'
+Bundler.setup
+
+require 'active_model_serializers'
+require 'active_support/all'
+require 'benchmark/ips'
+require_relative './../test/fixtures/poro'
+
+def bench_adapters(resource)
+  serializer = PostWithCustomKeysSerializer.new(resource)
+
+  Benchmark.ips do |x|
+    ActiveModel::Serializer::Adapter.adapter_map.each do |adapter_name, adapter_class|
+      x.report(adapter_name) { adapter_class.new(serializer).to_json } unless adapter_name == 'null'
+    end
+
+    x.compare!
+  end
+end
+
+def bench_with_associations
+  post = Post.new(id: 1, title: 'New Post', body: 'Body')
+  author = Author.new(id: 1, name: 'Ian K')
+  first_comment = Comment.new(id: 1, body: 'ZOMG A COMMENT')
+  second_comment = Comment.new(id: 2, body: 'ZOMG ANOTHER COMMENT')
+  post.comments = [first_comment, second_comment]
+  first_comment.post = post
+  second_comment.post = post
+  post.author = author
+  blog = Blog.new(id: 1, name: 'My Blog!!')
+  post.blog = blog
+
+  bench_adapters(post)
+end
+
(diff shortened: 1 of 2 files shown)


bf4:
@ianks Thanks for this!  Very awesome.  :+1:  Do you have a particular philosophy about writing benchmark code for an app, on where to put it, add to it, and when and how to run it?  I have a PR to add a benchmark for performance regressions that I've kind of stalled on... If that interests you, I'd love your input, especially since I'd want it to co-exist with this.

bf4:
ref: https://github.com/rails-api/active_model_serializers/pull/872 and my mess of https://github.com/rails-api/active_model_serializers/pull/1393 and https://github.com/rails-api/active_model_serializers/pull/1412

bench/adapters.rb
```diff
@@ -0,0 +1,52 @@
...
+require 'active_model_serializers'
+require 'active_support/all'
+require 'benchmark/ips'
+require_relative './../test/fixtures/poro'
+
+def bench_adapters(resource)
+  serializer = PostWithCustomKeysSerializer.new(resource)
+
+  Benchmark.ips do |x|
+    ActiveModel::Serializer::Adapter.adapter_map.each do |adapter_name, adapter_class|
```
bf4: oh fun

ianks:
Hey @bf4 I checked out your PR. It is really good. 

I don't really have a 'philosophy' on benchmarking per se; other than it is something one should strive to do when coding. I think the approach you take in your PR is very important, as it documents perf regressions from changes in the code _automatically_. It is almost like an integration test suite for perf. 

The goal of this PR, is to simply have a place where people can write benchmarks for small sub-sets of the code they are working on. 

For example, I wanted to see if I could increase the perf of the JSONAPI adapter, and benchmarking only the adapters makes sense in this case. Almost like a unit-test of perf. 

TL;DR -- Your approach works better for documenting perf and catching regressions. The approach of this PR is more geared towards creating a place for contributors to write benchmarks for specific pieces of code. I hope that makes sense? Ha :smile:

bf4:
@ianks you've inspired me.. it's better now.. I focused on just the benchmark before worrying about regression testing across revisions.. and now there's a structure for this PR.. test/dummy/bm_adapter.rb (or should it be test/benchmark/... not sure 'cause I want the dummy app easily accessible).  Other thought is to update the https://github.com/rails-api/active-model-serializers-example and use that...

groyoh:
Just a quick thought on this: shouldn't there be a benchmark using the `include` option to test performance when including associations?

bf4:
@groyoh some many things should be... :)

bf4:
@ianks So, https://github.com/rails-api/active_model_serializers/issues/1393 has been merged.. wanna use that structure? you'd be the first, so a good time to help see how well it works, easy to use, and write up usage docs :)

remear:
Benchmarking should use and iterate on the structure introduced by #1393. Closing this, but I'd really like to see the work done in this PR done in a new PR with the new benchmarking structure.

PR #1515, opened 2016-02-13
Add symbol support for Serializer.type method

groyoh:
The `Serializer.type` method now accepts symbol as paremeter:
```ruby
  type :profile
end
```

I don't know what you guys think about this but I usually prefer to use symbol as much as possible and thus prefer to write `type :profile` than `type 'profile'`. Furthermore this change should not have any runtime impact.

I also refactored the tests to make it more DRY.


lib/active_model/serializer/type.rb
```diff
@@ -17,7 +17,7 @@ module ClassMethods
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     type 'authors'
         def type(type)
-          self._type = type
+          self._type = type ? type.to_s : nil
```
bf4: why check if type is nil?

test/adapter/json_api/type_test.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          private
+
+          def assert_type(resource, expected_type, opts = {})
+            opts = opts.reverse_merge(adapter: :json_api)
+            hash = serializable(resource, opts).serializable_hash
+            assert_equal(expected_type, hash.fetch(:data).fetch(:type))
+          end
+
+          def serializable(resource, options = {})
```
bf4: this already exists in test/support

bf4: but I see it's in the original, so, not on you :)

groyoh: I'll remove it then.

lib/active_model/serializer/type.rb
groyoh: Because of [this](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/fragment_cache.rb#L98), we might call `serializer.type(nil)` and then `nil.to_s` results in `""`. I tried `self._type = type if type` at first but then two specs were failling out of nowhere. It looked like something related to the cache but I could not really figure it out.

bf4: type && type.to_s

B mobile phone

groyoh: Ah right, it didn't come to my mind.

groyoh: Changed it.

joaomdmoura:
Nice @groyoh I like this, @bf4 made some great comments, glad that you took care of it.
Can you also update [our docs](https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#type) to reflect this? I also prefer symbols.

groyoh:
@joaomdmoura could you check if the doc I added sounds good to you?

docs/general/serializers.md
```diff
@@ -107,12 +107,30 @@ end
 
 #### ::type
 
-e.g.
+The `::type` method defines the JSONAPI [type](http://jsonapi.org/format/#document-resource-object-identification) that will be rendered for this serializer.
+It either takes a `String` or `Symbol` as parameter.
+
+Note: This method is useful only when using the `:json_api` adapter.
```
bf4: though, TODO: it should replace the `json_key` method and `root` option..

test/adapter/json_api/type_test.rb
```diff
@@ -0,0 +1,61 @@
...
+          end
+
+          class SymbolTypeSerializer < ActiveModel::Serializer
+            attributes :name
+            type :profile
+          end
+
+          def setup
+            @author = Author.new(id: 1, name: 'Steve K.')
+          end
```
bf4: better to use `setup do @author = Author.new(id: 1, name: 'Steve K.') end` to avoid the need to call `super` (which is lacking here)

test/adapter/json_api/type_test.rb
```diff
@@ -0,0 +1,61 @@
...
+
+          def with_jsonapi_resource_type inflection
+            old_inflection = ActiveModelSerializers.config.jsonapi_resource_type
+            ActiveModelSerializers.config.jsonapi_resource_type = inflection
+            yield
+          ensure
+            ActiveModelSerializers.config.jsonapi_resource_type = old_inflection
+          end
```
bf4: the inflection helper should probably be i the private section at the bottom. I know, pre-existing

bf4:
:+1:

docs/general/serializers.md
groyoh: Is there an open PR for this?

test/adapter/json_api/type_test.rb
groyoh: I'll move it ;)

test/adapter/json_api/type_test.rb
groyoh: Thanks for the info. I never use `ActiveSupport::TestCase` so I did not know.

joaomdmoura:
LGTM :+1:

docs/general/serializers.md
bf4: Not that I recall

Issue #1514, opened 2016-02-13
[RFC] How to handle improper usage error


groyoh:
In #1504, I figured out that in some part of the gem we do not make sure that user use the gem properly to match jsonapi specification.

The main example I ran into is when using a relationship that does not include `data` and do not use `links` and `meta` (Ref. https://github.com/rails-api/active_model_serializers/pull/1504#discussion_r52399639): 

``` ruby
has_one :author do
  include_data false
end
# => { relationships: { author: {} } }
```

This results in an empty relationship and is not allowed by jsonapi specification (http://jsonapi.org/format/#document-resource-object-relationships) but the gem does not prevent the user from doing so.

Should the gem raise an error in such case or let the user figure out on its own that he's using the gem wrongly?

remear:
@groyoh Was this resolved by #1516?

groyoh:
@remear no this was not resolved.

PR #1516, opened 2016-02-13
Fix bug displaying nil for relationship link href

groyoh:
Follow up to https://github.com/rails-api/active_model_serializers/pull/1504#discussion_r52518656. When using the relationship link with a block, calling "meta" without "href", e.g.

```ruby
has_one :bio do
  link :related do
    meta id: 1
  end
end
# => { links: { posts: { related: { href: nil, meta: { id: 1 } }  } } }.
```
results in in a nil "href".
According to JSONAPI, we should be able to use meta without href (http://jsonapi.org/format/#document-links).

Depending on the ouput from #1514, this PR might need some follow up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Features:
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
 
 Fixes:
+- [#1516](https://github.com/rails-api/active_model_serializers/pull/1501) No longer return a nil href when only
+  adding meta to a relationship link. (@groyoh)
 - [#1458](https://github.com/rails-api/active_model_serializers/pull/1458) Preserve the serializer
   type when fragment caching. (@bdmac)
 - [#1477](https://github.com/rails-api/active_model_serializers/pull/1477) Fix `fragment_cached?`
--- a/test/adapter/json_api/relationship_test.rb
+++ b/test/adapter/json_api/relationship_test.rb
@@ -19,16 +19,22 @@ module ActiveModel
 
             has_many :locations do
               link :related do
-                ids = object.locations.map!(&:id).join(',')
+                ids = object.locations.map(&:id).join(',')
                 href "//example.com/locations/#{ids}"
               end
             end
 
             has_many :posts do
               link :related do
-                ids = object.posts.map!(&:id).join(',')
+                ids = object.posts.map(&:id).join(',')
                 href "//example.com/posts/#{ids}"
                 meta ids: ids
+              end
+            end
+
+            has_many :comments do
+              link :self do
+                meta ids: [1]
               end
             end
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Features:
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
 
 Fixes:
+- [#1516](https://github.com/rails-api/active_model_serializers/pull/1501) Using only meta for relationship links would result
+  in a `nil` href. (@groyoh)
```
bf4: > No longer return a `nil` href when only adding meta to a relationship link.

test/adapter/json_api/relationship_test.rb
```diff
@@ -27,7 +27,6 @@ class RelationshipAuthorSerializer < ActiveModel::Serializer
             has_many :posts do
               link :related do
                 ids = object.posts.map!(&:id).join(',')
-                href "//example.com/posts/#{ids}"
```
bf4: Could you comment on this change as an assertion for the bug fix?

groyoh: In #1504, I originally wrote the test to test the `meta` call and there was no call to `href`. But then found out that we had this bug and added a call to `href` to prevent having a `nil` href. This change simply remove the `href` to test the behavior I wanted to test from the beginning. Thinking about it again,  I could also had another test instead.

groyoh: I added another spec instead to make it more clear.

CHANGELOG.md
groyoh: Fixed it ;)

test/adapter/json_api/relationship_test.rb
```diff
@@ -19,19 +19,25 @@ class RelationshipAuthorSerializer < ActiveModel::Serializer
 
             has_many :locations do
               link :related do
-                ids = object.locations.map!(&:id).join(',')
+                ids = object.locations.map(&:id).join(',')
```
groyoh: I was using `map!` instead of `map` with actually change the array and might cause some issues.

test/adapter/json_api/relationship_test.rb
```diff
@@ -19,19 +19,25 @@ class RelationshipAuthorSerializer < ActiveModel::Serializer
...
             has_many :posts do
               link :related do
-                ids = object.posts.map!(&:id).join(',')
+                ids = object.posts.map(&:id).join(',')
                 href "//example.com/posts/#{ids}"
                 meta ids: ids
               end
             end
 
+            has_many :comments do
```
groyoh: This actually does a call to `meta` without calling `href`.

test/adapter/json_api/relationship_test.rb
```diff
@@ -164,7 +176,14 @@ def test_relationship_with_everything
               },
               meta: { liked: true }
             }
-            assert_equal(expected, hash[:data][:relationships][:likes])
+            assert_relationship(:likes, expected)
+          end
+
+          private
+
+          def assert_relationship(relationship_name, expected)
```
groyoh: Created this helper method to DRY the code.

bf4: nice.

bf4:
Looks good.

Issue #600, opened 2014-08-23
New 0.9 gem does not work out of the box with rails-api


sachse:
I'm not sure whether this issue should be opened here or on the `rails-api` repository so let me know if this is the wrong place.

Anyway, `active_model_serializers` does not work out of the box with a new `rails-api` project, as the JSON responses aren't being overridden. I found [this Stack Overflow question](https://stackoverflow.com/questions/20811994/activemodelserializer-not-being-called-when-rendering-json), which resulted in one solution: making `ApplicationController` inherit from `ActionController::Base` instead of `ActionController::API`. With this I found out that `ApplicationController` needs `ActionController::Serialization` for AMS to work.

For those of you that are experiencing the issue, that means that you can just add the following to `ApplicationController` to work around the issue for now:

    include ActionController::Serialization

If this behaviour is intended, shouldn't it be mentioned in the README?

I'm on Rails 4.1.5, rails-api 0.2.1 and active_model_serializers 0.9.0.

sbycrosz:
I could confirm this happens for Rails 5 with ActiveModelSerializer 0.9.4,  and adding the include works.
+1 for adding this to readme :+1:

steveklabnik:
Sure, please send a PR and I'm happy to merge :smile:

oyeanuj:
@MarcusSky @sbycrosz While generating scaffolds with a Rails 5 API project, I ran into the following issue on 0.9x which goes away on 0.10x - did you guys see the same issue?

```ruby
→ rails g scaffold User
Running via Spring preloader in process 20090
      invoke  active_record
      create    db/migrate/20160911170034_create_users.rb
      create    app/models/user.rb
      invoke    test_unit
      create      test/models/user_test.rb
      create      test/fixtures/users.yml
      invoke  resource_route
       route    resources :users
      create  app/serializers/user_serializer.rb
      invoke  scaffold_controller
Could not find "api_controller.rb" in any of your source paths. Your current source paths are:
/Users/anuj/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/active_model_serializers-0.9.5/lib/active_model/serializer/generators/serializer/templates
```

PR #1517, opened 2016-02-13
Add documentation for usage with Rails 5 API mode

sbycrosz:
https://github.com/rails-api/active_model_serializers/issues/600

### DIFF
--- a/docs/general/getting_started.md
+++ b/docs/general/getting_started.md
@@ -96,3 +96,12 @@ class PostsController < ApplicationController
 
 end
 ```
+
+Do note when using Rails 5 API mode (or when you're not using `ActionController::Base` as the base controller class), you'll need to include `ActionController::Serialization` in your `ApplicationController`
+
+```ruby
+class ApplicationController < ActionController::API
+  include ActionController::Serialization
+end
+```
+


docs/general/getting_started.md
```diff
@@ -96,3 +96,12 @@ class PostsController < ApplicationController
 
 end
 ```
+
+Do note when using Rails 5 API mode (or when you're not using `ActionController::Base` as the base controller class), you'll need to include `ActionController::Serialization` in your `ApplicationController`
+
+```ruby
+class ApplicationController < ActionController::API
+  include ActionController::Serialization
```
bf4: Thanks for this!

However, I don't believe this is still true.  In Rails 5 I see the [ActionController::API  uses the `:action_controller` load hook](https://github.com/rails/rails/blob/v5.0.0.beta2/actionpack/lib/action_controller/api.rb#L144), so [our railtie should include it there, as well](https://github.com/rails-api/active_model_serializers/blob/v0.10.0.rc4/lib/active_model_serializers/railtie.rb#L13-L15).

Have you confirmed this is still a problem on AMS rc4 and Rails 5 beta2?

In any case, I agree that our test suite needs to add ActionController::API testing!

joaomdmoura:
After I read the original issue it seems it's related to 0.9.4, so this should not be merge into master but 0.9.x branch.
@sbycrosz can you confirm it? Meanwhile I'm closing this, but I'll be glad to reopen it once we clear it out :smile:

Issue #1520, opened 2016-02-18
Sparse fieldsets usage suppresses relationships


felixbuenemann:
When using sparse fieldsets in combination with includes on 0.10.0rc4 (and master), the serialized json api document is missing the `relationships` top level key, but the `included` key is still present.

**Test Case:**

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  # gem 'active_model_serializers', '0.10.0.rc4'
  gem 'active_model_serializers', github: 'rails-api/active_model_serializers'
end

require 'pp'

class Node < ActiveModelSerializers::Model
  attr_accessor :id, :name, :children
end

class NodeSerializer < ActiveModel::Serializer
  attribute :name
  has_many :children, serializer: NodeSerializer
end

root = Node.new(id: 1, name: 'root')
root.children = [
  Node.new(id: 2, name: 'child1', children: []),
  Node.new(id: 3, name: 'child2', children: [])
]

serializer = NodeSerializer.new(root)

normal = ActiveModel::Serializer::Adapter::JsonApi.new(serializer)
pp normal.as_json
# {:data=>
#  {:id=>"1",
#   :type=>"nodes",
#   :attributes=>{:name=>"root"},
#   :relationships=>{:children=>{:data=>nil}}}}
# It seems has_many doesn't work properly with poros, but the :relationships key is present

sparse = ActiveModel::Serializer::Adapter::JsonApi.new(serializer, fields: { nodes: [:name] })
pp sparse.as_json
# {:data=>{:id=>"1", :type=>"nodes", :attributes=>{:name=>"root"}}}
```

groyoh:
Your example contains an error but the behavior you state is indeed a bug since excluded relationships fields should not appear in `included` (http://jsonapi.org/format/#document-compound-documents).

The corrected example:
```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  # gem 'active_model_serializers', '0.10.0.rc4'
  gem 'active_model_serializers', github: 'rails-api/active_model_serializers'
end

require 'pp'

class Node < ActiveModelSerializers::Model
  attr_accessor :id, :name, :children
end

class NodeSerializer < ActiveModel::Serializer
  attribute :name
  has_many :children, serializer: NodeSerializer
end

root = Node.new(
  id: 1,
  name: 'root',
  children: [
    Node.new(id: 2, name: 'child1', children: []),
[...]

PR #1528, opened 2016-02-20
Fix bug happening when using fields and include

groyoh:
See #1520. When the :fields and :include options were used to filter out
and include the same attribute, the attribute would not appear in the
relationship but would still be included in the "included" array.

### DIFF
--- a/lib/active_model/serializer/adapter/json_api.rb
+++ b/lib/active_model/serializer/adapter/json_api.rb
@@ -113,7 +113,10 @@ module ActiveModel
         end
 
         def process_relationships(serializer, include_tree)
+          resource_identifier    = ApiObjects::ResourceIdentifier.new(serializer).as_json
+          requested_associations = fieldset.fields_for(resource_identifier[:type])
           serializer.associations(include_tree).each do |association|
+            next unless requested_associations.nil? || requested_associations.include?(association.name)
             process_relationship(association.serializer, include_tree[association.key])
           end
         end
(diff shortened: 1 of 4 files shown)


test/adapter/json_api/fields_test.rb
```diff
@@ -62,7 +62,7 @@ def test_fields_relationships
           end
 
           def test_fields_included
-            fields = { posts: [:author], comments: [:body] }
+            fields = { posts: [:author, :comments], comments: [:body] }
```
groyoh: With this fix, the included comments would be kicked out if we don't include them in the `fields` and thus the spec would fail.

lib/active_model/serializer/adapter/json_api.rb
```diff
@@ -113,7 +113,10 @@ def process_resource(serializer, primary)
         end
 
         def process_relationships(serializer, include_tree)
+          resource_identifier    = ApiObjects::ResourceIdentifier.new(serializer).as_json
+          requested_associations = fieldset.fields_for(resource_identifier[:type])
           serializer.associations(include_tree).each do |association|
+            next unless requested_associations.nil? || requested_associations.include?(association.name)
```
groyoh: Probably not the most beautiful solution but it is already done in a similar way in the [Attributes](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/attributes.rb#L21) class and finding a cleaner solution would have implied too much refactoring for this small fix.

caJaeger:
Is there an ETA on when this will be rolled in? Quite an important thing for our application, and I imagine many others.

Issue #1536, opened 2016-02-24
Setting root key strategy globally for json adapter


RomanKapitonov:
After updating to `v0.10.0.rc4` tag I had to specify `ActiveModelSerializers.config.adapter = :json` in an initializer for a rails application in order to make active_model_serializer to add root key to the the response. However if the model is nested the key is built using the whole class name. For example if I specify a model:

```ryby
class Users::Profile < ActiveRecord::Base
end
```

and then in controller:

```ruby
class Users::ProfilesController < ApplicationController
  def show
    @profile = Users::Profile.first
     respond_with @profile, serializer: UserProfileSerializer
  end
end
```

It will build json containing the root key as follows: `{'users/profile': {'name': 'test'}}`. Is there any way to make active_model_serializer to use the class itself ignoring nesting? I've looked through the source code and found a method `json_key` which builds the root element of the json. But overriding that method is not always working. Moving root element build method to a overridable strategy may be a good feature request candidate

RomanKapitonov:
I've found a way to reproduce the issue when `json_key` method is not called. If you respond_with empty collection of a given type (fetched by `cancancan` method `load_and_authorize_resource`) in `index` action it will ignore the `json_key` and render an empty collection with wrong root key.

groyoh:
What about using the `:root` options like so:
```ruby
render json: @profile, serializer: UserProfileSerializer, root: "profile"
```
?

groyoh:
I see two other solutions here:
* Create your own `ProfileCollectionSerializer` like so:

```ruby
class ProfileCollectionSerializer < ActiveModel::
  def json_key
    "profiles"
  end
end
```
```ruby
render json: @profiles, serializer: ProfileCollectionSerializer
```

* define a wrapper around the cancan collection that responds to `#name`:

```ruby
class CollectionWrapper
  include Enumerable
  delegate :each, to: :@collection
  attr_reader :name

  def initialize(collection, name)
    @collection = collection
    @name = name
  end
end
```
```ruby
render json: CollectionWrapper.new(@profiles, :profiles), serializer: ProfileCollectionSerializer
```

RomanKapitonov:
@groyoh Those are all valid options, the only concern is that they force you to be explicit about how you define the root element name. I've created a [pull requests](https://github.com/rails-api/active_model_serializers/pull/1537) which should allow you to implicitly use `json_key` from item serializer when resource is an empty relation

bf4:
[...]

PR #1537, opened 2016-02-24
[FIX] Fetch json key from item serializer if empty collection is pass…

RomanKapitonov:
…ed to collection serializer and each_searializer is specified.

### DIFF
--- a/lib/active_model/serializer/collection_serializer.rb
+++ b/lib/active_model/serializer/collection_serializer.rb
@@ -10,8 +10,14 @@ module ActiveModel
       def initialize(resources, options = {})
         @root = options[:root]
         @object = resources
+
+        serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)
+
+        if resources.blank? && options[:serializer]
+          @each_serializer = options[:serializer]
+        end
+
         @serializers = resources.map do |resource|
-          serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)
           serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
 
           if serializer_class.nil?
@@ -23,7 +29,7 @@ module ActiveModel
       end
 
       def json_key
-        root || derived_root
+        root || derived_root || guess_root || default_root
       end
 
       def paginated?
@@ -39,8 +45,15 @@ module ActiveModel
       private
 
       def derived_root
-        key = serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
+        serializers.first.try(:json_key).try(:pluralize)
+      end
+
+      def default_root
+        object.try(:name).try(:underscore).try(:pluralize)
+      end
+
+      def guess_root
+        @each_serializer.try(:allocate).try(:json_key).try(:pluralize)
       end
     end
   end
--- a/test/fixtures/poro.rb
+++ b/test/fixtures/poro.rb
@@ -164,6 +164,12 @@ PaginatedSerializer = Class.new(ActiveModel::Serializer::CollectionSerializer) d
   end
 end
 
+MessagesSerializer = Class.new(ActiveModel::Serializer) do
+  def json_key
+    'messages'
+  end
+end
+
 AlternateBlogSerializer = Class.new(ActiveModel::Serializer) do
   attribute :id
   attribute :name, key: :title
--- a/test/collection_serializer_test.rb
+++ b/test/collection_serializer_test.rb
@@ -84,6 +84,12 @@ module ActiveModel
         assert_nil serializer.json_key
       end
 
+      def test_json_key_with_empty_resources_with_serializer
+        resource = []
+        serializer = collection_serializer.new(resource, serializer: MessagesSerializer)
+        assert_equal 'messages', serializer.json_key
+      end
+
       def test_json_key_with_root
         expected = 'custom_root'
         serializer = collection_serializer.new(@resource, root: expected)


RomanKapitonov:
Fixes https://github.com/rails-api/active_model_serializers/issues/1536

test/fixtures/poro.rb
```diff
@@ -164,6 +164,12 @@ def json_key
   end
 end
 
+MessagesSerializer = Class.new(ActiveModel::Serializer) do
+  def json_key
+    'messages'
+  end
+end
```
bf4: would you mind moving this to the test file? poro's is a little cluttered

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -39,8 +45,15 @@ def paginated?
       private
 
       def derived_root
-        key = serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
+        serializers.first.try(:json_key).try(:pluralize)
```
bf4: we really need to replace `Serializer#json_key` with `Serializer._type`...

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -10,8 +10,14 @@ class CollectionSerializer
       def initialize(resources, options = {})
         @root = options[:root]
```
bf4: I'm thinking this should really be `@root = collection_root(options[:root], options[:serializer])`

and something like (assuming root and json_key should behave the same... I don't remember)

```ruby
private

def collection_root(explicit_root, explicit_each_serializer)
  explicit_root || derived_root(explicit_each_serializer)
end

def derived_root(explicit_each_serializer)
 serializers.first.try!(:json_key).try!(:pluralize) ||
    (resources.empty && explicit_each_serializer.try!(:allocate).try!(:json_key).try!(:pluralize)) ||
    object.try!(:name).try!(:underscore).try!(:pluralize)
end
```

aside: all these `try` make my head hurt

groyoh: :+1: I also don't really enjoy all these `try`.

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -39,8 +45,15 @@ def paginated?
       private
 
       def derived_root
-        key = serializers.first.try(:json_key) || object.try(:name).try(:underscore)
-        key.try(:pluralize)
+        serializers.first.try(:json_key).try(:pluralize)
+      end
+
+      def default_root
+        object.try(:name).try(:underscore).try(:pluralize)
```
groyoh: What about:
```ruby
def default_root
  return unless object.respond_to?(:name)
  object.name.to_s.underscore.pluralize
end

File: lib/active_model/serializer/collection_serializer.rb (hunk 2 above)
Comment (groyoh) [inline lib/active_model/serializer/collection_serializer.rb]: Here what about:

```ruby
def derived_root
  return unless serializers.empty?
  serializers.first.json_key.to_s.pluralize
end
```

Makes it more readable IMO.

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -10,8 +10,14 @@ class CollectionSerializer
       def initialize(resources, options = {})
         @root = options[:root]
         @object = resources
+
+        serializer_context_class = options.fetch(:serializer_context_class, ActiveModel::Serializer)
+
+        if resources.blank? && options[:serializer]
+          @each_serializer = options[:serializer]
```
groyoh: `@each_serializer` is too much linked to the controller option name IMO. I'd rather call it `@resource_serializer` or simply `@serializer`.

bf4: Or explicit _collection_serializer

lib/active_model/serializer/collection_serializer.rb
bf4: Right now, the only scenario tested is essentially:

```diff
@root = collection_root(options)

# condenses `root || derived_root`
def collection_root(options)
  if root = options[:root]
    root
 else
-   key = serializers.first.try(:json_key) || object.try(:name).try(:underscore)
+   key = serializers.first.try(:json_key) || options[:serializer].try!(:allocate).try!(:json_key).try!(pluralize) || object.try(:name).try(:underscore)
    key.try(:pluralize)
  end
end
```

(or a more thorough impl would be something like)

```ruby
@root = collection_root(options)

# condenses `root || derived_root`
def collection_root(options)
  if root = options[:root]
    root
 else
   key = object.empty? && (serializer_class = options[:serializer]) && (serializer_class.allocate.json_key.try!(:pluralize) || serializer_class._type) # shouldn't need the allocate here
    key ||= object.any? && object.name.underscore
    key ||= '' # oh no! we should never be here. empty collection and no serializer option
    key.pluralize
  end
end
```
I'm also realizing we're inconsistent with object vs. resources.  We should probably always refer to it as object (which is the serializer api for the thing it encapsulates).

bf4:
@RomanKapitonov are you planning on coming back to this?

test/fixtures/poro.rb
NullVoxPopuli: :+1:

NullVoxPopuli:
@RomanKapitonov are you planning on helping resolve the PR Review comments? :-\

lib/active_model/serializer/collection_serializer.rb
groyoh: @bf4 @NullVoxPopuli  @remear  I really think that `allocate` is not the right solution here. `json_key` by default use `root` or the `object` (https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L185). Neither of these would be available and might result in a bug if the `json_key` method is not overidden. That might also break for polymorphic collections.

I stay on my position that the prefered solution here would be the ones described at https://github.com/rails-api/active_model_serializers/issues/1536#issuecomment-188434945. Or even subclass the collection serializer and add its own root handling.

bf4:
Followed up with #1618

Issue #1333, opened 2015-11-17
Include option for associations in serializers


oyeanuj:
I am migrating from 0.9.3 to 0.10.0-rc3 to take advantage of json-api. Based on the current docs, it seems if I want the association's data as part of the response, I will have to specify it as part of 'include' option when calling render in the controller. 

Is there a way to specify this in the serializer instead of the controller, so that I can specify it once and everytime I render that serializer, those associations are included in the json-api by default (instead of having to specify it every time I render that object)? That seemed to be the default behavior in the 0.9.x 

TLDR: Instead of 
```ruby
# posts_controller.rb
render @post, include: ['author', 'metrics', 'media']
```
something like this:
```ruby
# post_serializer.rb
has_one :author, include: true 
has_many :metrics #or included by default
has_many :media
```

Maybe it is already there and I missed it. Either way, I would love some clarity. 

Thanks!

beauby:
Not at the moment, but there is discussion about it.

oyeanuj:
Fwiw, I think having them included by default would make the upgrade path from 0.8 and 0.9 to 0.10 much easier and more backwards compatible (and not needing to include associations every time or refactor code to prevent repetition).

iMacTia:
@oyeanuj, PR #1127 does exactly what you're asking, but unfortunately have been merged the 21st of September, only 5 days after 0.10.0-rc3 was released.
@joaomdmoura, any plan for the rc4? this feature have been requested by many and rc3 have been out for almost 2 months now.

oyeanuj:
@iMacTia Thats really good to know, thank you for that. Couple of questions with your PR: 

1. Without the include, does it only get the type and id? And include specific fields if we want more of that object? Ideally to maintain backward compatibility, there would be an `include: false` option which only then shows the id & type but by default associations would be included.

2. Does your PR also support `except` and `only` options on associations too for backwards compatibility? 

Thanks again!

iMacTia:
@oyeanuj unfortunately I can't take the merit for that PR because is all @NullVoxPopuli work :)
So it's quite hard for me to correctly answer your question, also because I've never used the json-api adapter so far, but only the standard json one.
I'm simply really interested as much as you are on getting this new features out od the door ;) So sometimes I open the project page to see what's the status.

beauby:
@oyeanuj 

[...]

PR #1538, opened 2016-02-24
Add :except as an option to Adapter::Attributes

noahsilas:
This supports passing `except:`  as an option to `render` or as an option to `has_many`, etc. declrations. This is useful when avoiding circular includes.

The `except` name was chosen to mirror behavior found in the 0.8x branch (which will help our upgrade path), and partly addresses #1333.

### DIFF
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -14,10 +14,12 @@ module ActiveModel
 
         # Return the +attributes+ of +object+ as presented
         # by the serializer.
-        def attributes(requested_attrs = nil, reload = false)
+        def attributes(options = {}, reload = false)
+          requested_attrs = options[:only]
+          excepts = Array(options[:except])
           @attributes = nil if reload
           @attributes ||= self.class._attributes_data.each_with_object({}) do |(key, attr), hash|
-            next if attr.excluded?(self)
+            next if attr.excluded?(self) || excepts.include?(key)
             next unless requested_attrs.nil? || requested_attrs.include?(key)
             hash[key] = attr.value(self)
           end
--- a/docs/general/adapters.md
+++ b/docs/general/adapters.md
@@ -180,6 +180,19 @@ It could be combined, like above, with other paths in any combination desired.
   render json: @posts, include: 'author.comments.**'
 ```
 
+#### Excluded
+
+Sometimes you want to omit a specific field or association during serialization.
+You can use the `except` option for this:
+
+```ruby
+  render json: @posts, include: '*', except: :author
+```
+
+This is particularly helpful if you are using the recursive include wildstar
+(`**`), as it can lead to infinite recursion when you have associations that
+can be traversed in a cycle.
+
 ##### Security Considerations
 
 Since the included options may come from the query params (i.e. user-controller):
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -64,6 +64,7 @@ Where:
   - `unless:`
   - `virtual_value:`
   - `polymorphic:` defines if polymorphic relation type should be nested in serialized association.
+  - `except`: Specify attributes or associations to be ommitted from serialization
 - optional: `&block` is a context that returns the association's attributes.
   - prevents `association_name` method from being called.
   - return value of block is used as the association value.
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -33,7 +33,9 @@ module ActiveModelSerializers
 
       def resource_relationships(options)
         relationships = {}
+        excepts = Array(options[:except])
         serializer.associations(@include_tree).each do |association|
+          next if excepts.include?(association.key)
           relationships[association.key] ||= relationship_value_for(association, options)
         end
 
@@ -45,7 +47,8 @@ module ActiveModelSerializers
         return unless association.serializer && association.serializer.object
 
         opts = instance_options.merge(include: @include_tree[association.key])
-        relationship_value = Attributes.new(association.serializer, opts).serializable_hash(options)
+        hash_opts = options.merge(except: association.options[:except])
+        relationship_value = Attributes.new(association.serializer, opts).serializable_hash(hash_opts)
 
         if association.options[:polymorphic] && relationship_value
           polymorphic_type = association.serializer.object.class.name.underscore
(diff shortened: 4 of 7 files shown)


lib/active_model/serializer/attributes.rb
```diff
@@ -14,10 +14,12 @@ module Attributes
 
         # Return the +attributes+ of +object+ as presented
         # by the serializer.
-        def attributes(requested_attrs = nil, reload = false)
+        def attributes(options = {}, reload = false)
+          requested_attrs = options[:only]
+          excepts = Array(options[:except])
```
bf4: usually only and except are mutually exclusive. If I'm reading this correctly, they're not, here, right?

bf4:
Nice

lib/active_model/serializer/attributes.rb
noahsilas: Right, here both are applied as filters. `only: [:a, :b], except: [:b]` is equivalent to `only: [:a]`. This seems reasonable to me, although I can imagine how someone could be confused by it. Should this raise if we receive both?

bf4: Do you think this can be made to work with the if/unless in https://github.com/rails-api/active_model_serializers/pull/1403 ? That seems to cover this case except for having options to pass in from the renderer.

noahsilas: I'm not sure I follow; if/unless seems like it can prevent the entire association from being included/rendered, where `except` will still render the association but exclude some specific fields.

bf4: I mean, I like the idea of accepting requested options and excluded options, but this could be done, with a little more effort, via attribute `unless` and `instance_options`.  

The PR is great.  I'm just uncertain if I want it as a new feature to maintain, or if it's a first class behavior... Perhaps some other @rails-api/ams @rails-api/commit maintainers could chime in

NullVoxPopuli: @noahsilas if you can add a test using only/except and if/unless, I'll accept the PR.

I see this being used for excluding attributes more often than including them. Such as excluding the created_at and updated_at fields.

bf4: Except that the serializer already functions as a whitelist

This feature is only useful when you opt in to more data than you want AND
then choose not to use existing features to exclude data. This is really
the job of the scope and instance options along with attrribte if/unless
and multiple serializers. My gut is this is not a primary use case we want
to support or encourage, and it can probably also use the existing 'fields'
option AND this doesn't apply to jsonapi

A good feature pr isn't free unless op will stick around to maintain it.

NullVoxPopuli: That's a good point. Do we have docs explaining how to achieve that with the scope and instance options?

bf4: It can be inferred, might be nice to make it explicit

NullVoxPopuli: yeah, at least, to prevent PRs for things that can already be achieved. 
Is that something you want to do, @bf4 ?

AmrN: @bf4 I was really looking for something like this and I have no idea how to achieve it otherwise, could you please show an example of how to do it using attribute `unless` and `instance_options`

bf4: Sure amr

Could you share your ams vesion from the gemfile.lock, your serializer,
your render command, and your desired output?

AmrN: @bf4 
active_model_serializers (0.10.0.rc4)
Say I have a `Chatroom` serializer: 

```
class ChatroomSerializer < ActiveModel::Serializer
  attributes :id, :name, :description, :room_type, :subscribed
  has_many :subscribers
  has_many :visitors

  def subscribed
    object.subscribers.include? current_user
  end
end
```
Now sometimes the client may only want basic attributes for the chatroom like `name`, `description`, `room_type`...etc. and doesn't care about the associations, so I'd like to do something like this in my actions:
`render json: @chatroom, except: [:subscribers, :visitors]`

or even something like this: 
`render json: @chatroom, except: params[:except_attrs]`

I know I can specify a specific serializer to render, but I don't want to create a separate serializer for each small variation, am I making sense?

NullVoxPopuli:
Nice, @bf4 -- so, we are in agreement that a PR should be made with those docs and that this PR should be closed?

noahsilas:
@bf4 I wonder if it might help if I describe how we're using this patch.

We've been using AMS 0.8, and want to begin migrating our API towards JSON API. To do that, we plan on being an early adopter of AMS 0.10 - we will be adding a feature to our API to allow clients to opt-in to the new format via an `Accept` header, and swap the request from the Attributes adapter onto the JsonApi adapter.

To make this change with as little breakage as possible, we are using #1426 to change the default include behavior to match 0.8x (using `'**'` as the default includes setting), but that leaves us with the problem of resolving nested cycles. In AMS 0.8, we did this with the `except` option, which we add to 0.10 here.

Given a reasonable amount of time, we would like to move all our clients onto JsonApi and be able to remove most of these `except`s in our code, but until we get there, this is a very helpful tool for us. (If it's not merged, we will likely use a version of this patch ourselves while we migrate). I am open to including documentation marking this as a deprecated feature, since it does feel like it shouldn't be necessary for anyone who is starting with AMS 0.10, and then we could remove it in 0.11 (1.0?).

Given that context, do you think there is a path forward to merge this (or something similar)?

Would it be helpful if I added some documentation or a wiki page on upgrading from 0.8 that describes our experience?

bf4:
@noahsilas So, one plus in your corner is that `except` is one of the valid options passed to [`ActiveModel::Serialization`](https://github.com/rails/rails/blob/master/activemodel/lib/active_model/serialization.rb)

bf4:
ref https://github.com/rails-api/active_model_serializers/issues/1643

beauby:
I can totally see the usefulness of this for `Json`/`Attributes` adapters, not so much for `JsonApi`.
I'm personally not against this feature.

remear:
Closing this due to age.

I think a more interesting approach would be to make this ability function more like how the `json_api` adapter does `fieldset`.

NullVoxPopuli:
@remear, I don't know if you've seen -- but the plan is to extract all the features to the "serializer", so that all adapters can have all features, but just differ in how they are rendered

Issue #1532, opened 2016-02-22
Deserializing Empty HasMany Relationship Array Params in JSONAPI Adapter


namack:
When using the JSONAPI Adapter to deserialize incoming parameters, `has_many` relationships with empty `data` arrays are deserialized incorrectly.

For example, in the `articles_controller.rb`, the `article_params` are deserialized using `ActiveModelSerializers::Deserialization.jsonapi_parse(params, only: [:title, :src, :images])`

When an incoming payload includes empty `has_many` relationships, the deserializer changes the key to it's singular (`belongs_to`/`has_one`) form.
For example, the following payload will cause an error:
```json
{
  "data": {
    "type": "articles",
    "attributes": {
      "title": "Ember Hamster",
      "src": "http://example.com/images/productivity.png"
    },
    "relationships": {
      "images": {
        "data": []
      }
    }
  }
}
``` 
The resulting `article_params` will parse `image_id: nil`, instead of `images: []`. Since an `article` `has_many :images`, this will cause an error and the `POST` will fail. 

#1248
cc: @matt-morris

beauby:
As you can see [here](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json_api/deserialization.rb#L185), this should get picked up. I'll investigate your payload tomorrow. Are you certain that's what your client is issuing?

janvarljen:
@namack 

I had the same issue. The thing is Rails converts empty arrays to nils (security issues). Then the AMS deserializer gets "data": nil insted of "data": [] and that's why it's deserialized to image_id instead of image_ids.

This issue in Rails is fixed but only for Rails 5 version: 

https://github.com/rails/rails/pull/16924

atm I'm manually fixing my params until I migrate to Rails 5 like this:

```ruby
params[:images_ids] = [] if params.key?(:image_id)
```

namack:
@janvarljen

Thanks for the explanation - I am seeing that exact behavior. The empty array is converted to `nil` before the deserializer even gets the chance to deserialize, which is why [line 185](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/adapter/json_api/deserialization.rb#L185) never gets the chance to change the key correctly.

@beauby - Should I close this since the issue is solved in Rails 5?

namack:
https://github.com/rails-api/active_model_serializers/pull/1541

beauby:
[...]

PR #1541, opened 2016-02-26
Deserialize empty hasmany

namack:
In < Rails 5, all incoming params with [empty arrays are converted to `nil`](https://github.com/rails/rails/pull/16924). When a JSONAPI `has_many` relationships payload is empty (which is very common for Ember Data `POST`), the deserializer is given `nil` as the associated data, therefore assuming that the relationship is `has_one` or `belongs_to` and pluralizes the key incorrectly.

https://github.com/rails-api/active_model_serializers/issues/1532

### DIFF
--- a/lib/active_model/serializer/adapter/json_api/deserialization.rb
+++ b/lib/active_model/serializer/adapter/json_api/deserialization.rb
@@ -180,12 +180,13 @@ module ActiveModel
           #
           # @api private
           def parse_relationship(assoc_name, assoc_data, options)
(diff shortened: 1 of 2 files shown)


namack:
This probably isn't the best solution and is more of a band-aid for < Rails 5 projects, but I thought I'd throw it out there just in case.

namack:
Issue is avoided by [overriding default parser](https://github.com/rails-api/active_model_serializers/issues/1532#issuecomment-190277868).

PR #1543, opened 2016-02-28
Follow up to #1535

groyoh:
#### Purpose
Follow up to #1535.
Closes #1529.

#### Changes
- One test was renamed because it contained `__`
- The ActiveModelSerializers::Deserialization module now call `Adater` instead of `ActiveModelSerializers::Adapter`.
- The changelog was added for #1535

PR #1558, opened 2016-03-07
Add a deprecation DSL

bf4:
Follows https://github.com/rails-api/active_model_serializers/pull/1543
and 6b4c8df6fb6bc142ee6a74da51bb26c42a025b3c

### DIFF
--- a/lib/active_model/serializer/adapter.rb
+++ b/lib/active_model/serializer/adapter.rb
@@ -3,40 +3,37 @@ module ActiveModel
     # @deprecated Use ActiveModelSerializers::Adapter instead
     module Adapter
       class << self
+        extend ActiveModelSerializers::Deprecate
+
         def create(resource, options = {})
-          warn_deprecation
           ActiveModelSerializers::Adapter.create(resource, options)
         end
+        deprecate :create, 'ActiveModelSerializers::Adapter.'
 
         def adapter_class(adapter)
-          warn_deprecation
           ActiveModelSerializers::Adapter.adapter_class(adapter)
         end
+        deprecate :adapter_class, 'ActiveModelSerializers::Adapter.'
 
         def adapter_map
-          warn_deprecation
           ActiveModelSerializers::Adapter.adapter_map
         end
+        deprecate :adapter_map, 'ActiveModelSerializers::Adapter.'
 
         def adapters
-          warn_deprecation
           ActiveModelSerializers::Adapter.adapters
         end
+        deprecate :adapters, 'ActiveModelSerializers::Adapter.'
 
         def register(name, klass = name)
-          warn_deprecation
           ActiveModelSerializers::Adapter.register(name, klass)
         end
+        deprecate :register, 'ActiveModelSerializers::Adapter.'
 
         def lookup(adapter)
-          warn_deprecation
           ActiveModelSerializers::Adapter.lookup(adapter)
         end
-
-        def warn_deprecation
-          warn "Calling deprecated #{name} (#{__FILE__}) from #{caller[1..3].join(', ')}. Please use ActiveModelSerializers::Adapter"
-        end
-        private :warn_deprecation
+        deprecate :lookup, 'ActiveModelSerializers::Adapter.'
       end
 
       require 'active_model/serializer/adapter/base'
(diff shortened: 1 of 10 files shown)


bf4:
cc @groyoh

lib/active_model/serializer/adapter.rb
```diff
@@ -3,40 +3,37 @@ class Serializer
     # @deprecated Use ActiveModelSerializers::Adapter instead
     module Adapter
       class << self
+        extend ActiveModelSerializers::Deprecate
+
         def create(resource, options = {})
```
groyoh: Would `delegate :create, to: ActiveModelSerializers::Adapter` work here?

groyoh:
Awesome work :100:

lib/active_model/serializer/adapter.rb
bf4: y'know, I tried it with ruby's delegate and forwardable, but not with rails' delegate.  Feel free to try. I think you can push to my branch.

bf4:
@groyoh to be honest, I'm not sure if I like this as a 'another thing in AMS. OTOH, it does take most of the drudgery out of deprecating. OTOH (three hands), we can remove it when we hit 0.10.0.

groyoh:
I also think that it would be better to get rid of it once we hit 0.10.0. But right now, it definitely makes things cleaner and easier to handle, especially once we move `ActiveModel::Serializer` to `ActiveModelSerializers` module.

lib/active_model/serializer/adapter.rb
```diff
@@ -3,40 +3,37 @@ class Serializer
...
         def lookup(adapter)
-          warn_deprecation
           ActiveModelSerializers::Adapter.lookup(adapter)
         end
-
-        def warn_deprecation
-          warn "Calling deprecated #{name} (#{__FILE__}) from #{caller[1..3].join(', ')}. Please use ActiveModelSerializers::Adapter"
-        end
-        private :warn_deprecation
+        deprecate :lookup, 'ActiveModelSerializers::Adapter.'
```
NullVoxPopuli: I really like this implementation, good thinkin'

lib/active_model_serializers/deprecate.rb
```diff
@@ -0,0 +1,49 @@
...
+#       end
+#     end
+#
+# Adapted from https://github.com/rubygems/rubygems/blob/1591331/lib/rubygems/deprecate.rb
+module ActiveModelSerializers
+  module Deprecate
+    ##
+    # Simple deprecation method that deprecates +name+ by wrapping it up
+    # in a dummy method. It warns on each call to the dummy method
+    # telling the user of +replacement+ (unless +replacement+ is :none) that it is planned to go away.
```
NullVoxPopuli: good docs!

lib/active_model_serializers/deprecate.rb
```diff
@@ -0,0 +1,49 @@
...
+#       end
+#
+#       extend ActiveModelSerializers::Deprecate
+#       deprecate :instance_method, "ActiveModelSerializers::NewPlace#new_method"
+#
+#       class << self
+#         extend ActiveModelSerializers::Deprecate
+#         deprecate :klass_method, :none
+#       end
+#     end
```
NullVoxPopuli: I like the example!

NullVoxPopuli:
I actually think we should keep the deprecation module when we hit 0.10. That way when can use the same code if nothing better comes around by the time we need to start deprecating stuff again.

Unless there are any more TODOs, I'm good with merging this.

groyoh:
@NullVoxPopuli Wouldn't deprecation lead to a breaking change at that point?

NullVoxPopuli:
yeah, I mean, remove the deprecations when we hit 0.10, but keep the ability to deprecate with `ActiveModelSerializers::Deprecate`

NullVoxPopuli:
Actually, we still need a changelog in here -> then I'm ok with merging.

NullVoxPopuli:
closed in favor of #1559

lib/active_model_serializers/deprecate.rb
bf4: Copied

PR #1563, opened 2016-03-08
Fix regression where '?' is not removed from boolean syntactic sugar on attributes.

NullVoxPopuli:
#### Purpose

Pre 0.10, `?` was removed from attributes.

#### Changes

Added Test

#### Caveats

What about methods with `?` -- the ones that aren't just syntactic sugar?
This may result in checking if the object responds_to a method without the question mark -- which may lead to unexpected behavior.

#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/issues/605

### DIFF
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -55,7 +55,15 @@ module ActiveModel
         #     end
         def attribute(attr, options = {}, &block)
           key = options.fetch(:key, attr)
+          key = _remove_question_mark(key)
           _attributes_data[key] = Attribute.new(attr, options, block)
+        end
+
+        # @api private
+        # replace ? at the end of an attribute,
+        # as JSON API disallows the use of ? at the end of keys.
+        def _remove_question_mark(key)
+          key.to_s.sub(/\?\z/, '').to_sym
         end
 
         # @api private
--- a/test/serializers/attribute_test.rb
+++ b/test/serializers/attribute_test.rb
@@ -81,6 +81,17 @@ module ActiveModel
         assert_equal('custom', hash[:blog][:id])
       end
 
+      def test_id_with_question_mark_has_question_mark_removed
+        serializer = Class.new(ActiveModel::Serializer) do
+          attribute :boolean?
+        end
+
+        blog = Blog.new(boolean?: true)
+        hash = ActiveModel::SerializableResource.new(blog, adapter: :json, serializer: serializer).serializable_hash
+
+        assert_equal(true, hash[:blog][:boolean])
+      end
+
       PostWithVirtualAttribute = Class.new(::Model)
       class PostWithVirtualAttributeSerializer < ActiveModel::Serializer
         attribute :name do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Features:
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
 
 Fixes:
+- [#1563](https://github.com/rails-api/active_model_serializers/pull/1563) Remove '?' character from attribute keys' suffix. (@NullVoxPopuli)
 - [#1570](https://github.com/rails-api/active_model_serializers/pull/1570) Fixed pagination issue with last page size. (@bmorrall)
 - [#1516](https://github.com/rails-api/active_model_serializers/pull/1516) No longer return a nil href when only
   adding meta to a relationship link. (@groyoh)
--- a/lib/active_model/serializer/caching.rb
+++ b/lib/active_model/serializer/caching.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module ActiveModel
   class Serializer
     module Caching
--- a/lib/active_model/serializer/version.rb
+++ b/lib/active_model/serializer/version.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module ActiveModel
   class Serializer
     VERSION = '0.10.0.rc4'.freeze
--- a/lib/active_model_serializers/adapter.rb
+++ b/lib/active_model_serializers/adapter.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module ActiveModelSerializers
   module Adapter
     UnknownAdapterError = Class.new(ArgumentError)
--- a/lib/active_model_serializers/adapter/base.rb
+++ b/lib/active_model_serializers/adapter/base.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module ActiveModelSerializers
   module Adapter
     class Base
--- a/lib/active_model_serializers/json_pointer.rb
+++ b/lib/active_model_serializers/json_pointer.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module ActiveModelSerializers
   module JsonPointer
     module_function
--- a/lib/active_model_serializers/logging.rb
+++ b/lib/active_model_serializers/logging.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 ##
 # ActiveModelSerializers::Logging
 #
--- a/test/action_controller/json_api/pagination_test.rb
+++ b/test/action_controller/json_api/pagination_test.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'test_helper'
 require 'will_paginate/array'
 require 'kaminari'
--- a/test/action_controller/serialization_scope_name_test.rb
+++ b/test/action_controller/serialization_scope_name_test.rb
@@ -27,7 +27,7 @@ class DefaultScopeNameTest < ActionController::TestCase
 
   def test_default_scope_name
     get :render_new_user
-    assert_equal '{"data":{"id":"1","type":"users","attributes":{"admin?":false}}}', @response.body
+    assert_equal '{"data":{"id":"1","type":"users","attributes":{"admin":false}}}', @response.body
   end
 end
 
@@ -58,6 +58,6 @@ class SerializationScopeNameTest < ActionController::TestCase
 
   def test_override_scope_name_with_controller
     get :render_new_user
-    assert_equal '{"data":{"id":"1","type":"users","attributes":{"admin?":true}}}', @response.body
+    assert_equal '{"data":{"id":"1","type":"users","attributes":{"admin":true}}}', @response.body
   end
 end
--- a/test/active_model_serializers/adapter_for_test.rb
+++ b/test/active_model_serializers/adapter_for_test.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module ActiveModelSerializers
   class AdapterForTest < ActiveSupport::TestCase
     UnknownAdapterError = ::ActiveModelSerializers::Adapter::UnknownAdapterError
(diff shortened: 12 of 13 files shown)


NullVoxPopuli:
I guess the ? could just be removed in the rendered output, and if there are collisions, write to the log?

like if someone did 

```ruby
attribute :boolean, :boolean?
```

test/serializers/attribute_test.rb
```diff
@@ -81,6 +81,23 @@ def id
         assert_equal('custom', hash[:blog][:id])
       end
 
+      def test_id_with_question_mark_has_question_mark_removed
+        serializer = Class.new(ActiveModel::Serializer) do
+          # TODO: if we were to remove the question mark automatically,
+          #       what do we do if we want to call a method that has a
+          #       question mark.
+          #       e.g.:
+          #               def my_question?; ....; end
```
bf4: In rails boolean fields work w and w out the ?. String inquirer takes care of it

bf4: Best practice is just don't use a ? In your serializer

? Is def verboten in json api

NullVoxPopuli: yeah, but I mean if someone wanted to use a method in their model rather than just a db-backed field.

e.g.:
```ruby
2.2.3 :001 > def test?; 'hi'; end
 => :test? 
2.2.3 :002 > test
ArgumentError: wrong number of arguments (0 for 2..3)
	from (irb):2:in `test'
	from (irb):2
	from /home/preston/.rvm/rubies/ruby-2.2.3/bin/irb:11:in `<main>'
2.2.3 :003 > test?
 => "hi" 
2.2.3 :004 > 
```

bf4: > like if someone did
> attribute :boolean, :boolean?

Don't do that.  When would that ever make sense?

bf4: If your ? Method doesn't return a boolean that's a bug waiting to happen

NullVoxPopuli: well, if that's the case, then I guess we should remove it, and just note in our docs, about the possibility of a collision, if someone were silly enough to try rendering both of `:boolean, :boolean?` where both of those point to different data.

maybe it's just an edge case, and it won't be that big of a deal.

bf4: I would say they are asking for trouble and should get it.  They can always set a key:

NullVoxPopuli: good point.

Ok, I'll implement the fix for this then.

bf4: Still is a good test to have to demonstrate behavior.  Need to confirm json api adapter strips it

NullVoxPopuli: just json api adapter? or all adapters?

NullVoxPopuli:
Fixed

NullVoxPopuli:
squashed

remear:
Looks good to me

lib/active_model/serializer/attributes.rb
```diff
@@ -17,6 +17,8 @@ module Attributes
         def attributes(requested_attrs = nil, reload = false)
           @attributes = nil if reload
           @attributes ||= self.class._attributes_data.each_with_object({}) do |(key, attr), hash|
+            key = key.to_s.gsub(/\?\z/, '').to_sym
```
groyoh: Would it be possible to move that at https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/attributes.rb#L57? That way it's done at the serializer definition and not everytime we call the `#attributes` method.

NullVoxPopuli: yes! good catch. Pushed :-)

lib/active_model/serializer/attributes.rb
```diff
@@ -55,6 +55,9 @@ def attributes(*attrs)
         #     end
         def attribute(attr, options = {}, &block)
           key = options.fetch(:key, attr)
+          # replace ? at the end of an attribute, as JSON API disallows the use
+          # of ? at the end of keys.
+          key = key.to_s.gsub(/\?\z/, '').to_sym
```
groyoh: I'll be picky here but I usually tend to avoid using `gsub` when `sub` can be used: ` key = key.to_s.sub(/\?\z/, '').to_sym` :grin:

groyoh:
Looks good :+1: feel free to listen to my picky comment :wink:

lib/active_model/serializer/attributes.rb
NullVoxPopuli: I didn't know sub could handle regex -- that's why I use it. 

http://ruby-doc.org/core-2.2.0/String.html#method-i-sub
well, whaddayaknow.

oh, it looks like gsub replaces all occurrences, and sub replaces just the first occurrence.

that leads me to believe sub would be slightly faster. I'll swap it out.

NullVoxPopuli:
@groyoh just swapped out gsub for sub. thanks for the tip!

bf4:
'tr' is actually faster for single char and '?' Is only legal once and at the end

B mobile phone

lib/active_model/serializer/attributes.rb
bf4: Gem install fasterer

B mobile phone

NullVoxPopuli:
@bf4, actually, it looks like sub is faster.

I made a benchmark
```ruby
require 'benchmark'

attribute = 'some_boolean?'
n = 1000000
Benchmark.bm do |benchmark|
  benchmark.report { n.times{ attribute.sub(/\?\z/, '') }}
  benchmark.report { n.times{ attribute.gsub(/\?\z/, '') }}
  benchmark.report { n.times{ attribute.tr('?', '') }}
end
```
results:
```
⬡ v4.3.1 💎  2.2.3 $ ruby test.rb
   0.530000   0.000000   0.530000 (  0.550814)
   1.450000   0.010000   1.460000 (  1.472772)
   0.550000   0.000000   0.550000 (  0.557974)

```

NullVoxPopuli:
just did another one with the mutating versions of each of those:

```ruby
require 'benchmark'

attribute = 'some_boolean?'
n = 1000000
Benchmark.bm do |x|
  x.report('sub  ') { n.times{ attribute.sub(/\?\z/, '') }}
  x.report('gsub ') { n.times{ attribute.gsub(/\?\z/, '') }}
  x.report('tr   ') { n.times{ attribute.tr('?', '') }}
  x.report('sub! ') { n.times{ attribute.sub!(/\?\z/, '') }}
  x.report('gsub!') { n.times{ attribute.gsub!(/\?\z/, '') }}
  x.report('tr!  ') { n.times{ attribute.tr!('?', '') }}
end

```


results:
```
⬡ v4.3.1 💎  2.2.3 $ ruby test.rb
sub    0.530000   0.000000   0.530000 (  0.545544)
gsub   1.430000   0.000000   1.430000 (  1.478049)
tr     0.540000   0.000000   0.540000 (  0.563370)
sub!   0.350000   0.040000   0.390000 (  0.390127)
gsub!  0.430000   0.000000   0.430000 (  0.431214)
tr!    0.480000   0.000000   0.480000 (  0.486672)

```

bf4:
Try w benchmark-ips

NullVoxPopuli:
ok, new results:

```
⬡ v4.3.1 💎  2.2.3 $ ruby test.rb
               sub      37.765k i/100ms
               gsub     24.482k i/100ms
               tr       40.036k i/100ms
               sub!     42.971k i/100ms
               gsub!    41.631k i/100ms
               tr!      42.918k i/100ms
-------------------------------------------------
               sub        1.832M (± 8.4%) i/s -      9.064M
               gsub       1.858M (± 8.9%) i/s -      9.156M
               tr         1.589M (± 6.1%) i/s -      7.927M
               sub!       2.122M (± 6.6%) i/s -     10.528M
               gsub!      2.071M (± 8.7%) i/s -     10.241M
               tr!        1.756M (± 7.9%) i/s -      8.712M

Comparison:
               sub! :  2122277.9 i/s
               gsub!:  2070864.3 i/s - 1.02x slower
               gsub :  1857651.1 i/s - 1.14x slower
               sub  :  1832473.0 i/s - 1.16x slower
               tr!  :  1756340.4 i/s - 1.21x slower
               tr   :  1589408.6 i/s - 1.34x slower

```

via

```ruby
require 'benchmark/ips'

attribute = 'some_boolean?'

Benchmark.ips do |x|
  # 5 seconds, warm up for 2 seconds
  x.config(time: 5, warmup: 2)

  x.report('sub  ') { attribute.sub(/\?\z/, '') }
  x.report('gsub ') { attribute.gsub(/\?\z/, '') }
  x.report('tr   ') { attribute.tr('?', '') }
  x.report('sub! ') { attribute.sub!(/\?\z/, '') }
  x.report('gsub!') { attribute.gsub!(/\?\z/, '') }
  x.report('tr!  ') { attribute.tr!('?', '') }

  x.compare!
end
```

lib/active_model/serializer/attributes.rb
```diff
@@ -55,6 +55,9 @@ def attributes(*attrs)
         #     end
         def attribute(attr, options = {}, &block)
           key = options.fetch(:key, attr)
+          # replace ? at the end of an attribute, as JSON API disallows the use
+          # of ? at the end of keys.
+          key = key.to_s.sub(/\?\z/, '').to_sym
```
bf4: better if this were in a method someone could monkey patch if they absolutely hate it

better if it only applied in json api

best if it were part of the 'transform keys' logic for json api (which doesn't really exist right now)

NullVoxPopuli: > better if this were in a method someone could monkey patch if they absolutely hate it

I agree with it being in its own method (just pushed)

> better if it only applied in json api

do we have a mechanism where this can only apply to JSON API?

> best if it were part of the 'transform keys' logic for json api (which doesn't really exist right now)
 
for later! :-)

NullVoxPopuli:
Closing this in favor of custom key transform.

See: https://github.com/rails-api/active_model_serializers/blob/9e992358d87bb69b52006caf08b6b1093db2b53d/lib/active_model_serializers/key_transform.rb
```ruby
require 'active_support/core_ext/hash/keys'

module ActiveModelSerializers
  module KeyTransform
    module_function

    # Transforms keys to UpperCamelCase or PascalCase.
    #
    # @example:
    #    "some_key" => "SomeKey",
    # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
    def dashed_without_question(hash)
      hash.deep_transform_keys! { |key| key.to_s.sub(/\?\z/, '').dasherize.to_sym }
    end
    ...
end
```

and then in an initializer:
```ruby
ActiveModelSerializers.config.key_transform = :dashed_without_question
```

/cc @dangalipo, @sija, @pedromtavares from #605

dangalipo:
cool :+1:

Sija:
looks cool with me too! :+1: I wonder tho' is there/would be a possibility of defining `ActiveModelSerializers.config.key_transform` as an array — with transformers executed in a row? I could image it might come helpful at times.

NullVoxPopuli:
@Sija you could just write a custom transform that does your chained operations?

Sija:
@NullVoxPopuli :-1: for modularity and another set of future/possible monkey-patches

NullVoxPopuli:
not all monkey patches are bad. :-\

bf4:
>  you could just write a custom transform that does your chained operations?

@NullVoxPopuli that's a really good idea.  It'd be pretty easy to write and document.

Sija:
@NullVoxPopuli chaining IMO allows doing that in moar modular fashion tho' I agree that it's possible using one method with chained operations as you said

PR #1567, opened 2016-03-09
Remove rbx from CI. Allowing failures is just wasting resources

bf4:
#### Purpose

Allowing all RBX versions to fail is basically wasting resources.

Unless there's someone willing to keep it green, there's not reason
to keep it running on CI.

#### Changes

Remove rbx from CI


NullVoxPopuli:
LGTM

PR #1572, opened 2016-03-09
Never mutate controller options

bf4:
#### Purpose

Pagination links incorrectly passing options to `serializable_hash`.

Only the renderer should be passing options to `serializable_hash`.

It's only working now because we are mutating the controller options which it
later passes to `to_json` which calls `serializable_hash` on the adapter.

This PR ensures that we do not mutate controller options, which
exposes failures *only* in the pagination links.

 #### Changes

1. Make a (deep) dup of rendering options to pass to SerializableResource
1. Freezes rendering options (might be undesirable)

 #### Caveats

- freezing rendering options might be undesirable in general, and probably is
not necessary for this fix.

 #### Related GitHub issues

AMS 0.10.0.rc4 / master

https://github.com/rails-api/active_model_serializers/blob/6b4c8df6fb6bc142ee6a74da51bb26c42a025b3c/lib/active_model_serializers/adapter/json_api.rb

```ruby
def pagination_links_for(serializer, options)
  PaginationLinks.new(serializer.object, options[:serialization_context]).serializable_hash(options)
end
```

Those options come from the call to `adapter.serializable_hash(options)`
which is wrong.  They should be passed into the Adapter initializer.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -26,13 +26,13 @@ module ActionController
         respond_to?(_serialization_scope, true)
     end
 
-    def get_serializer(resource, options = {})
+    def get_serializer(resource, serializer_options = {})
       if !use_adapter?
         warn 'ActionController::Serialization#use_adapter? has been removed. '\
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
-        options[:adapter] = false
+        serializer_options[:adapter] = false
       end
-      serializable_resource = ActiveModel::SerializableResource.new(resource, options)
+      serializable_resource = ActiveModel::SerializableResource.new(resource, serializer_options)
       if serializable_resource.serializer?
         serializable_resource.serialization_scope ||= serialization_scope
         serializable_resource.serialization_scope_name = _serialization_scope
@@ -56,10 +56,12 @@ module ActionController
 
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
-        options.fetch(:serialization_context) do
-          options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, options)
+        options.freeze
+        serializer_options = options.deep_dup
+        serializer_options.fetch(:serialization_context) do
+          serializer_options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, serializer_options)
         end
-        serializable_resource = get_serializer(resource, options)
+        serializable_resource = get_serializer(resource, serializer_options)
         super(serializable_resource, options)
       end
     end
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -26,13 +26,13 @@ module ActionController
         respond_to?(_serialization_scope, true)
     end
 
-    def get_serializer(resource, serializer_options = {})
+    def get_serializer(resource, serialization_options = {})
       if !use_adapter?
         warn 'ActionController::Serialization#use_adapter? has been removed. '\
           "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
-        serializer_options[:adapter] = false
+        serialization_options[:adapter] = false
       end
-      serializable_resource = ActiveModel::SerializableResource.new(resource, serializer_options)
+      serializable_resource = ActiveModel::SerializableResource.new(resource, serialization_options)
       if serializable_resource.serializer?
         serializable_resource.serialization_scope ||= serialization_scope
         serializable_resource.serialization_scope_name = _serialization_scope
@@ -57,11 +57,14 @@ module ActionController
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
         options.freeze
-        serializer_options = options.deep_dup
-        serializer_options.fetch(:serialization_context) do
-          serializer_options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, serializer_options)
+        serialization_options = options.deep_dup
+        if options.key?(:serializer)
+          serialization_options[:serializer] = options[:serializer]
         end
-        serializable_resource = get_serializer(resource, serializer_options)
+        unless serialization_options.key?(:serialization_context)
+          serialization_options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, serialization_options)
+        end
+        serializable_resource = get_serializer(resource, serialization_options)
         super(serializable_resource, options)
       end
     end
--- a/lib/active_model/serializable_resource.rb
+++ b/lib/active_model/serializable_resource.rb
@@ -2,7 +2,7 @@ require 'set'
 require 'active_model_serializers/adapter'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :serialization_context])
     include ActiveModelSerializers::Logging
 
     delegate :serializable_hash, :as_json, :to_json, to: :adapter
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -128,7 +128,7 @@ module ActiveModelSerializers
 
         if is_collection && serializer.paginated?
           hash[:links] ||= {}
-          hash[:links].update(pagination_links_for(serializer, options))
+          hash[:links].update(pagination_links_for(serializer))
         end
 
         hash
@@ -498,8 +498,8 @@ module ActiveModelSerializers
       #   end
       # prs:
       #   https://github.com/rails-api/active_model_serializers/pull/1041
-      def pagination_links_for(serializer, options)
-        PaginationLinks.new(serializer.object, options[:serialization_context]).serializable_hash(options)
+      def pagination_links_for(serializer)
+        PaginationLinks.new(serializer.object, instance_options[:serialization_context]).as_json(instance_options)
       end
 
       # {http://jsonapi.org/format/#document-meta Docment Meta}
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # {http://jsonapi.org/format/ JSON API specification}
 # rubocop:disable Style/AsciiComments
 # TODO: implement!
@@ -43,14 +44,13 @@ module ActiveModelSerializers
 
       # {http://jsonapi.org/format/#crud Requests are transactional, i.e. success or failure}
       # {http://jsonapi.org/format/#document-top-level data and errors MUST NOT coexist in the same document.}
-      def serializable_hash(options = nil)
-        options ||= {}
+      def serializable_hash(_options = nil)
         document = if serializer.success?
-                     success_document(options)
+                     success_document
                    else
                      failure_document
                    end
-        transform_key_casing!(document, options[:serialization_context])
+        transform_key_casing!(document, instance_options[:serialization_context])
       end
 
       # {http://jsonapi.org/format/#document-top-level Primary data}
@@ -68,7 +68,7 @@ module ActiveModelSerializers
       #    links: toplevel_links,
       #    jsonapi: toplevel_jsonapi
       #  }.reject! {|_,v| v.nil? }
-      def success_document(options)
+      def success_document
         is_collection = serializer.respond_to?(:each)
         serializers = is_collection ? serializer : [serializer]
         primary_data, included = resource_objects_for(serializers)
--- a/test/adapter/json/key_case_test.rb
+++ b/test/adapter/json/key_case_test.rb
@@ -21,21 +21,19 @@ module ActiveModelSerializers
         def setup
           ActionController::Base.cache_store.clear
           @blog = Blog.new(id: 1, name: 'My Blog!!', special_attribute: 'neat')
-          serializer = CustomBlogSerializer.new(@blog)
-          @adapter = ActiveModelSerializers::Adapter::Json.new(serializer)
         end
 
         def test_key_transform_default
           mock_request
           assert_equal({
             blog: { id: 1, special_attribute: 'neat', articles: nil }
-          }, @adapter.serializable_hash(@options))
+          }, adapter.serializable_hash)
         end
 
         def test_key_transform_global_config
           mock_request
           result = with_config(key_transform: :camel_lower) do
-            @adapter.serializable_hash(@options)
+            adapter.serializable_hash
           end
           assert_equal({
             blog: { id: 1, specialAttribute: 'neat', articles: nil }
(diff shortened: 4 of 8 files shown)


lib/action_controller/serialization.rb
```diff
@@ -26,13 +26,13 @@ def serialization_scope
         respond_to?(_serialization_scope, true)
     end
 
-    def get_serializer(resource, options = {})
+    def get_serializer(resource, serializer_options = {})
```
NullVoxPopuli: thankyou for specifying which options these are. omg

NullVoxPopuli:
:+1:

groyoh:
Rebased and fixed the tests.

beauby:
:+1:

lib/action_controller/serialization.rb
```diff
@@ -56,10 +56,15 @@ def use_adapter?
 
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
-        options.fetch(:serialization_context) do
-          options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, options)
+        options.freeze
+        serialization_options = options.deep_dup
```
groyoh: @bf4 I'm not sure if it is good to `deep_dup` here as it also `dup` the serializer when present and that made the tests fail. That's why I had to do the workaround below. Is this really needed is your opinion?

bf4: wouldn't the serializer be a class at this point?  How would that be a problem.

I suppose an alternative would be something like strong params where we basically specific which params we want, but that would kill the `instance_options` behavior.

lib/action_controller/serialization.rb
```diff
@@ -56,10 +56,15 @@ def use_adapter?
...
+        if options.key?(:serializer)
+          serialization_options[:serializer] = options[:serializer]
         end
-        serializable_resource = get_serializer(resource, options)
+        unless serialization_options.key?(:serialization_context)
+          serialization_options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, serialization_options)
+        end
```
bf4: using fetch would be the more idiomatic way to do this, and is also easier to read.  @groyoh is there a reason you changed this?

lib/active_model/serializable_resource.rb
```diff
@@ -2,7 +2,7 @@
 require 'active_model_serializers/adapter'
 module ActiveModel
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :serialization_context])
```
bf4: I suppose the problem here will be if a serializer can use url helper to make links without the serialization context.. but that's not anything anyone is working on now

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
bf4: Was there a problem not having this?

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -43,14 +44,13 @@ def default_key_transform
 
       # {http://jsonapi.org/format/#crud Requests are transactional, i.e. success or failure}
       # {http://jsonapi.org/format/#document-top-level data and errors MUST NOT coexist in the same document.}
-      def serializable_hash(options = nil)
-        options ||= {}
+      def serializable_hash(_options = nil)
```
bf4: So, we do need to take options into `serializable_hash from the renderer. I'm willing to be convinced, but when someone passing in options from the `ActiveModel::Serializers::JSON` API to `as_json`, I would expect them to be honored, rather than silently ignored.

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -43,14 +44,13 @@ def default_key_transform
...
         document = if serializer.success?
-                     success_document(options)
+                     success_document
                    else
                      failure_document
                    end
-        transform_key_casing!(document, options[:serialization_context])
+        transform_key_casing!(document, instance_options[:serialization_context])
```
bf4: @groyoh So, there's no reason the adapter can't use `instance_options` for serialization context (or, really, the options to `initialize` would be the proper place to handle this), but putting the serialization_context in the adapter options isn't incompatible with https://github.com/rails-api/active_model_serializers/pull/1572/files#r56778382

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -498,8 +498,8 @@ def links_for(serializer)
       #   end
       # prs:
       #   https://github.com/rails-api/active_model_serializers/pull/1041
-      def pagination_links_for(serializer, options)
-        PaginationLinks.new(serializer.object, options[:serialization_context]).serializable_hash(options)
+      def pagination_links_for(serializer)
+        PaginationLinks.new(serializer.object, instance_options[:serialization_context]).as_json(instance_options)
```
bf4: I actually changed the method signature here from `serializable_hash` to `as_json` to distinguish its usage from `ActiveModel::Serialization`. See https://github.com/rails-api/active_model_serializers/pull/1572/files#r56778413

bf4: (which is what was wrong with the tests that were trying to pass in AMS options to `serializable_hash` in the tests changed in this diff).

Also, not having controller tests made this misunderstanding a bit easier to happen

test/adapter/json/key_case_test.rb
```diff
@@ -64,28 +62,33 @@ def test_key_transform_dashed
...
 
         def test_key_transform_camel_lower
           mock_request(:camel_lower)
           assert_equal({
             blog: { id: 1, specialAttribute: 'neat', articles: nil }
-          }, @adapter.serializable_hash(@options))
+          }, adapter.serializable_hash)
+        end
+
+        def adapter
```
bf4: this should just be `serializable(@blog, serializer: CustomBlogSerializer, adapter: :json, options: whatever)`

test/adapter/json_api/key_case_test.rb
```diff
@@ -36,16 +36,6 @@ class CommentSerializer < ActiveModel::Serializer
...
 
-        def mock_request(key_transform = nil)
-          context = Minitest::Mock.new
-          context.expect(:request_url, URI)
-          context.expect(:query_parameters, {})
-          context.expect(:key_transform, key_transform)
-          context.expect(:url_helpers, Rails.application.routes.url_helpers)
-          @options = {}
-          @options[:serialization_context] = context
-        end
```
bf4: @groyoh This is a pretty big change. Would you mind commenting on it?

test/adapter/json_api/key_case_test.rb
```diff
@@ -312,187 +294,156 @@ def test_success_document_key_transform_camel_lower
...
 
-          expected_errors_object =
-            { :errors =>
-              [
-                { :source => { :pointer => '/data/attributes/published_at' }, :detail => 'must be in the future' },
-                { :source => { :pointer => '/data/attributes/title' }, :detail => 'must be longer' }
-              ]
-          }
-          assert_equal expected_errors_object, result
+        def serialize_resource(resource, options)
```
bf4: since the method calls `as_json`, should really be `resource_serialization`. Also, good to be consistent and use the test helper: `serializable(resource, options).as_json`

test/adapter/json_api/key_case_test.rb
```diff
@@ -312,187 +294,156 @@ def test_success_document_key_transform_camel_lower
...
 
-          resource = ModelWithErrors.new
-          resource.errors.add(:published_at, 'must be in the future')
-          resource.errors.add(:title, 'must be longer')
+        private
 
-          serializer = ActiveModel::Serializer::ErrorSerializer.new(resource)
-          adapter = ActiveModelSerializers::Adapter::JsonApi.new(serializer)
-          result = adapter.serializable_hash(@options)
+        def options(key_transform = nil, extra_options = {})
```
bf4: I have to say, @groyoh I was pretty surprised to find this method here.. https://github.com/rails-api/active_model_serializers/pull/1572/files#r56778552

bf4:
@groyoh Thanks for helping clean this up.  Great work!

In the future, would you mind doing me a favor and if there's substantial changes, as there are here, to add a commit, rather than amend the existing one? It make it easier to see what's changed.  Thanks!

I left a lot of comments here.  I've been super swamped lately, and beggars can't be choosers, so I was wondering how you feel about the direction the comments go, code-wise, and if that's something you'd rather talk to me about, wait for me to come back to, or attend to yourself.

Thanks, again! :rainbow: :sparkles:

test/adapter/json_api/key_case_test.rb
groyoh: Could you explain why `resource_serialization` is better `serialize_resource` in this case? And you're totally right about the helper, I always forget it exists :grin:

test/adapter/json_api/key_case_test.rb
groyoh: Yeah sure. I was a bit overwhelm afterward and didn't have time to comment on the changes.

test/adapter/json_api/key_case_test.rb
groyoh: This method replace `mock_request`. That issue `mock_request` was that it sets the `@options` instance variable. Using it this way made us rely on `@options` when serializing and this sometimes created error and bugs when it was not set. Therefore, I think returning the options are easier to handle and understand rather than depending on the instance variable.

I think that changing the method name to `options_with_serialization_context` or `options_with_mock` might make it easier to understand what it does.

test/adapter/json_api/key_case_test.rb
```diff
@@ -62,14 +52,14 @@ def setup
                            publish_at: @publish_at)
           @comment1.post = @post
           @comment2.post = @post
+          @error_resource = ModelWithErrors.new
```
groyoh: Moved here because it was used by many tests.

test/adapter/json/key_case_test.rb
```diff
@@ -21,21 +21,19 @@ class PostSerializer < ActiveModel::Serializer
...
-          serializer = CustomBlogSerializer.new(@blog)
-          @adapter = ActiveModelSerializers::Adapter::Json.new(serializer)
         end
 
         def test_key_transform_default
           mock_request
           assert_equal({
             blog: { id: 1, special_attribute: 'neat', articles: nil }
-          }, @adapter.serializable_hash(@options))
+          }, adapter.serializable_hash)
```
groyoh: The `serialization_context` is no more passed via `as_json` options but via the adapter options.

test/adapter/json/key_case_test.rb
groyoh: Indeed, I just move the call without thinking too much about it. Here I move the this part of the code in a method, to make sure that `@options` was set before using it.

test/adapter/json_api/key_case_test.rb
```diff
@@ -62,14 +52,14 @@ def setup
...
+          @error_resource.errors.add(:published_at, 'must be in the future')
+          @error_resource.errors.add(:title, 'must be longer')
         end
 
         def test_success_document_key_transform_default
-          mock_request
```
groyoh: There are three main changes in this file:

* Replaced `mock_request` by `options`. ref. https://github.com/rails-api/active_model_serializers/pull/1572/files#r56795398
* Replaced `ActiveModelSerializers::Adapter::JsonApi.new(serializer)`  by `serialize_resource(@post, options(nil, serializer: PostSerializer))` so that the options are passed via the constructor and no more via the `as_json` options.
* Updated the syntax for the hash e.g. `:symbol =>` to `symbol:`, indentation, etc.

lib/action_controller/serialization.rb
groyoh: Isn't fetch main use case supposed to be "retrieving an object"? I changed it because for me using fetch when we don't use the return value feels weird. But I can revert this change if needed.

lib/action_controller/serialization.rb
groyoh: Somehow there was some issues with the `type` not being duped properly. Therefore the test failed cause the `type` method return the wrong value. I didn't dig to much into it though. I can try and spend some more time to figure out why it was duped properly.

lib/active_model_serializers/adapter/json_api.rb
groyoh: Looks like emacs automatically added it. Will fix it.

lib/active_model_serializers/adapter/json_api.rb
groyoh: Which options would they pass to `as_json` that could be used by the `JsonApi` adapter?

Do you mean we should add `instance_options.merge!(options)` here?

lib/active_model_serializers/adapter/json_api.rb
groyoh: I'm not sure I understand your comment.

groyoh:
@bf4 I actually created multiple commit locally but then squashed them up. I admit that it would have been better to leave it as it was for understanding the changes. I'll try to split them up in multiple commits if possible. I also commented on most of the changes :wink:

bf4:
@groyoh I restored my original commits, 8fe6354543d7ace65c3c7698fac2a9f38f344374 6bd777e29a4c140212e3efc77dd52ee680a80456, and cherry-picked 094af45 for easier review, and rebased on master

groyoh:
Sorry about this :disappointed:

lib/action_controller/serialization.rb
```diff
@@ -57,11 +57,14 @@ def use_adapter?
     [:_render_option_json, :_render_with_renderer_json].each do |renderer_method|
       define_method renderer_method do |resource, options|
         options.freeze
-        serializer_options = options.deep_dup
-        serializer_options.fetch(:serialization_context) do
-          serializer_options[:serialization_context] = ActiveModelSerializers::SerializationContext.new(request, serializer_options)
+        serialization_options = options.deep_dup
+        if options.key?(:serializer)
+          serialization_options[:serializer] = options[:serializer]
```
bf4: you said there was some weird bug caused by duping the `serializer` that this fixes?

lib/action_controller/serialization.rb
```diff
@@ -26,13 +26,13 @@ def serialization_scope
         respond_to?(_serialization_scope, true)
     end
 
-    def get_serializer(resource, serializer_options = {})
+    def get_serializer(resource, serialization_options = {})
       if !use_adapter?
```
bf4: I like the name change

lib/action_controller/serialization.rb
groyoh: Yes, I still haven't figured it out but there were three bugs failling. The resulting JSON API types looked like `"some_module_some_class"` instead of the values specified via `type "post"`. So I assume when the class was duped the `_type` wasn't duped properly.

groyoh: You could probably remove these lines and run the tests to trigger it.

lib/action_controller/serialization.rb
groyoh: Do we actually need to dup and freeze the options at all?

bf4: I'm thinking not

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -128,7 +128,7 @@ def success_document(options)
 
         if is_collection && serializer.paginated?
           hash[:links] ||= {}
-          hash[:links].update(pagination_links_for(serializer, options))
+          hash[:links].update(pagination_links_for(serializer))
         end
```
bf4: apparently in the commit i lost that this was @groyoh 's commit

bf4:
Closed by https://github.com/rails-api/active_model_serializers/issues/1647

PR #1566, opened 2016-03-09
Add resource-level meta docs. Update top-level meta docs.

remear:
#### Purpose

Improve meta documentation.

#### Changes

- Add resource-level meta docs
- Clarify top-level meta docs

#### Related GitHub issues

#1235

PR #1574, opened 2016-03-10
Provide key transformation

remear:
#### Purpose

Provide key case translation on a per-request basis.

```render json: posts, each_serializer: PostSerializer, key_transform: :camel_lower```

A global config option may be set to configure the key transform. ```ActiveModelSerializers.config.key_transform = :camel_lower```

#### Changes

- Adds a `key_transform` attribute to `SerializationContext`.
- Adds a `key_transform` option to `Configuration`.
- Translates keys for documents in the `JsonApi` adapter.
- Translates keys for serialized hash in the `Json` adapter.
- Provides the following casing options:
  - `camel` - `SpecialAttribute`
  - `camel_lower` - `specialAttribute`
  - `dashed` - `special-attribute` (default as per JSON API recommendation)
  - `unaltered` - the original, unaltered key
- Key translation precedence is as follows:
  - `SerializationContext` - The `key_transform` is provided as an option via render.
  - Global config - The `key_transform is set in `ActiveModelSerializers.config.key_transform`.
  - Adapter default - The default `key_transform` specified in the adapter.

### DIFF
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -2,26 +2,95 @@
 
 # Configuration Options
 
-The following configuration options can be set on `ActiveModelSerializers.config`,
-preferably inside an initializer.
+The following configuration options can be set on
+`ActiveModelSerializers.config`, preferably inside an initializer.
 
 ## General
 
-- `adapter`: The [adapter](adapters.md) to use. Possible values: `:attributes, :json, :json_api`. Default: `:attributes`.
-- `serializer_lookup_enabled`: When `false`, serializers must be explicitly specified. Default: `true`
+##### adapter
+
+The [adapter](adapters.md) to use.
+
+Possible values:
+
+- `:attributes` (default)
+- `:json`
+- `:json_api`
+
+##### serializer_lookup_enabled
+
+Enable automatic serializer lookup.
+
+Possible values:
+
+- `true` (default)
+- `false`
+
+When `false`, serializers must be explicitly specified.
+
+##### key_transform
+
+The [key transform](key_transform.md) to use.
+
+Possible values:
+
+- `:camel` - ExampleKey
+- `:camel_lower` - exampleKey
+- `:dashed` - example-key
+- `:unaltered` - the original, unaltered key
+- `nil` - use the adapter default
+
+Each adapter has a default key transform configured:
+
+- `Json` - `:unaltered`
+- `JsonApi` - `:dashed`
+
+`config.key_transform` is a global override of the adapter default. Adapters
+still prefer the render option `:key_transform` over this setting.
+
 
 ## JSON API
 
-- `jsonapi_resource_type`: Whether the `type` attributes of resources should be singular or plural. Possible values: `:singular, :plural`. Default: `:plural`.
-- `jsonapi_include_toplevel_object`: Whether to include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
-   in the response document.
-   Default: `false`.
-- Used when `jsonapi_include_toplevel_object` is `true`:
-  - `jsonapi_version`: The latest version of the spec the API conforms to.
-    Default: `'1.0'`.
-  - `jsonapi_toplevel_meta`: Optional metadata. Not included if empty.
-    Default: `{}`.
+
+##### jsonapi_resource_type
+
+Sets whether the [type](http://jsonapi.org/format/#document-resource-identifier-objects)
+of the resource should be `singularized` or `pluralized` when it is not
+[explicitly specified by the serializer](https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#type)
+
+Possible values:
+
+- `:singular`
+- `:plural` (default)
+
+##### jsonapi_include_toplevel_object
+
+Include a [top level jsonapi member](http://jsonapi.org/format/#document-jsonapi-object)
+in the response document.
+
+Possible values:
+
+- `true`
+- `false` (default)
+
+##### jsonapi_version
+
+The latest version of the spec to which the API conforms.
+
+Default: `'1.0'`.
+
+*Used when `jsonapi_include_toplevel_object` is `true`*
+
+##### jsonapi_toplevel_meta
+
+Optional top-level metadata. Not included if empty.
+
+Default: `{}`.
+
+*Used when `jsonapi_include_toplevel_object` is `true`*
+
 
 ## Hooks
 
-To run a hook when ActiveModelSerializers is loaded, use `ActiveSupport.on_load(:action_controller) do end`
+To run a hook when ActiveModelSerializers is loaded, use
+`ActiveSupport.on_load(:action_controller) do end`
--- a/docs/general/key_transform.md
+++ b/docs/general/key_transform.md
@@ -1,0 +1,34 @@
+[Back to Guides](../README.md)
+
+# Key Transforms
+
+Key transforms modify the keys in serialized responses.
+
+Provided key transforms:
+
+- `:camel` - ExampleKey
+- `:camel_lower` - exampleKey
+- `:dashed` - example-key
+- `:unaltered` - the original, unaltered key
+- `nil` - use the adapter default
+
+Key translation precedence is as follows:
+
+##### SerializableResource option
+
+`key_transform` is provided as an option via render.
+
+```render json: posts, each_serializer: PostSerializer, key_transform: :camel_lower```
+
+##### Configuration option
+
+`key_transform` is set in `ActiveModelSerializers.config.key_transform`.
+
+```ActiveModelSerializers.config.key_transform = :camel_lower```
+
+##### Adapter default
+
+Each adapter has a default key transform configured:
+
+- `Json` - `:unaltered`
+- `JsonApi` - `:dashed`
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -26,6 +26,7 @@ module ActiveModel
         # Make JSON API top-level jsonapi member opt-in
         # ref: http://jsonapi.org/format/#document-top-level
         config.jsonapi_include_toplevel_object = false
+        config.key_transform = nil
 
         config.schema_path = 'test/support/schemas'
       end
--- a/lib/active_model_serializers/adapter/base.rb
+++ b/lib/active_model_serializers/adapter/base.rb
@@ -1,3 +1,5 @@
+require 'active_model_serializers/key_transform'
+
 module ActiveModelSerializers
   module Adapter
     class Base
@@ -50,6 +52,27 @@ module ActiveModelSerializers
       def include_meta(json)
         json[meta_key] = meta unless meta.blank?
         json
+      end
+
+      def default_key_transform
+        :unaltered
+      end
+
+      # Determines the transform to use in order of precedence:
+      #   serialization context, global config, adapter default.
+      #
+      # @param serialization_context [Object] the SerializationContext
+      # @return [Symbol] the transform to use
+      def key_transform(serialization_context)
+        serialization_context.key_transform ||
+        ActiveModelSerializers.config.key_transform ||
+        default_key_transform
+      end
+
+      def transform_key_casing!(value, serialization_context)
+        return value unless serialization_context
+        transform = key_transform(serialization_context)
+        KeyTransform.send(transform, value)
       end
     end
   end
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -37,15 +37,20 @@ module ActiveModelSerializers
         @fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(options.delete(:fields))
       end
 
+      def default_key_transform
+        :dashed
+      end
+
       # {http://jsonapi.org/format/#crud Requests are transactional, i.e. success or failure}
       # {http://jsonapi.org/format/#document-top-level data and errors MUST NOT coexist in the same document.}
       def serializable_hash(options = nil)
         options ||= {}
-        if serializer.success?
-          success_document(options)
-        else
-          failure_document
-        end
+        document = if serializer.success?
+                     success_document(options)
+                   else
+                     failure_document
+                   end
+        transform_key_casing!(document, options[:serialization_context])
       end
 
       # {http://jsonapi.org/format/#document-top-level Primary data}
--- a/lib/active_model_serializers/key_transform.rb
+++ b/lib/active_model_serializers/key_transform.rb
@@ -1,0 +1,40 @@
+require 'active_support/core_ext/hash/keys'
+
+module ActiveModelSerializers
+  module KeyTransform
+    module_function
+
+    # Transforms keys to UpperCamelCase or PascalCase.
+    #
+    # @example:
+    #    "some_key" => "SomeKey",
+    # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
+    def camel(hash)
+      hash.deep_transform_keys! { |key| key.to_s.camelize.to_sym }
+    end
+
+    # Transforms keys to camelCase.
+    #
+    # @example:
+    #    "some_key" => "someKey",
+    # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
+    def camel_lower(hash)
+      hash.deep_transform_keys! { |key| key.to_s.camelize(:lower).to_sym }
+    end
+
+    # Transforms keys to dashed-case.
+    # This is the default case for the JsonApi adapter.
+    #
+    # @example:
+    #    "some_key" => "some-key",
+    # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L185-L187 ActiveSupport::Inflector.dasherize}
+    def dashed(hash)
+      hash.deep_transform_keys! { |key| key.to_s.dasherize.to_sym }
+    end
+
+    # Returns the hash unaltered
+    def unaltered(hash)
+      hash
+    end
+  end
+end
--- a/lib/active_model_serializers/serialization_context.rb
+++ b/lib/active_model_serializers/serialization_context.rb
@@ -4,13 +4,14 @@ module ActiveModelSerializers
       attr_writer :url_helpers, :default_url_options
     end
 
-    attr_reader :request_url, :query_parameters
+    attr_reader :request_url, :query_parameters, :key_transform
 
     def initialize(request, options = {})
       @request_url = request.original_url[/\A[^?]+/]
       @query_parameters = request.query_parameters
       @url_helpers = options.delete(:url_helpers) || self.class.url_helpers
       @default_url_options = options.delete(:default_url_options) || self.class.default_url_options
+      @key_transform = options.delete(:key_transform)
     end
 
     def self.url_helpers
--- a/test/adapter/json_api/key_case_test.rb
+++ b/test/adapter/json_api/key_case_test.rb
@@ -1,0 +1,500 @@
+require 'test_helper'
+
+module ActiveModelSerializers
+  module Adapter
+    class JsonApi
+      class KeyCaseTest < ActiveSupport::TestCase
+        Post = Class.new(::Model)
+        class PostSerializer < ActiveModel::Serializer
+          type 'posts'
+          attributes :title, :body, :publish_at
+          belongs_to :author
+          has_many :comments
+
+          link(:self) { post_url(object.id) }
+          link(:post_authors) { post_authors_url(object.id) }
+          link(:subscriber_comments) { post_comments_url(object.id) }
+
+          meta do
+            {
+              rating: 5,
+              favorite_count: 10
+            }
+          end
+        end
+
+        Author = Class.new(::Model)
+        class AuthorSerializer < ActiveModel::Serializer
+          type 'authors'
+          attributes :first_name, :last_name
+        end
+
+        Comment = Class.new(::Model)
+        class CommentSerializer < ActiveModel::Serializer
+          type 'comments'
+          attributes :body
+          belongs_to :author
+        end
+
+        def mock_request(key_transform = nil)
+          context = Minitest::Mock.new
+          context.expect(:request_url, URI)
+          context.expect(:query_parameters, {})
+          context.expect(:key_transform, key_transform)
+          context.expect(:url_helpers, Rails.application.routes.url_helpers)
+          @options = {}
+          @options[:serialization_context] = context
+        end
+
+        def setup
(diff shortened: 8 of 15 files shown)

### CODE REVIEW

**Comment by remear** [comment]
@rails-api/ams Thoughts on this approach?

**Comment by bf4** [comment]
ref: JSON API

- dasherize/kebab-case, and for namespaces https://github.com/json-api/json-api/issues/850

File: test/adapter/json_api/links_test.rb [hunk 1]
```diff
@@ -40,7 +40,7 @@ def test_toplevel_links
                   stuff: 'value'
                 }
               }
-            }).serializable_hash
+            }).serializable_hash(@options)
```
bf4: :question: I don't think this should be here. Not defined anywhere and we haven't been passing anything interesting into the renderer options

bf4: there's a bug in this right now https://github.com/rails-api/active_model_serializers/pull/1572

bf4:
Looks good.  you up for benchmarking it? maybe https://github.com/bf4/book_code/blob/master/adrpo/code/chp8/assert_performance.rb or something.. can go in `test/dummy` now.. which I think does need to be renamed to `test/benchmark`

remear:
I added a `dashed` case option and make it the default.

Also, note that rubocop wanted a lot of changes that weren't actually within the scope of this PR. I went ahead and fixed them in a separate commit.

remear:
The rubocop failures I see in travis right now look like the inverse of what it wanted locally, rubocop-0.38.0.

bf4:
https://github.com/rails-api/active_model_serializers/pull/1577 I have a PR to fix the JRuby's.. should be green soon

bf4:
```plain
Offenses:
lib/active_model/serializer/collection_serializer.rb:17:11: C: Style/GuardClause: Use a guard clause instead of wrapping the code inside a conditional expression. (https://github.com/bbatsov/ruby-style-guide#no-nested-conditionals)
          if serializer_class.nil?
          ^^
lib/active_model/serializable_resource.rb:36:5: C: Style/Alias: Use alias instead of alias_method in a class body. (https://github.com/bbatsov/ruby-style-guide#alias-method)
    alias_method :adapter_instance, :adapter
    ^^^^^^^^^^^^
lib/active_model/serializable_resource.rb:57:5: C: Style/Alias: Use alias instead of alias_method in a class body. (https://github.com/bbatsov/ruby-style-guide#alias-method)
    alias_method :serializer_class, :serializer
    ^^^^^^^^^^^^
lib/active_model_serializers/adapter.rb:4:19: C: Style/MutableConstant: Freeze mutable objects assigned to constants.
    ADAPTER_MAP = {}
                  ^^
test/support/serialization_testing.rb:24:3: C: Style/Alias: Use alias instead of alias_method in a module body. (https://github.com/bbatsov/ruby-style-guide#alias-method)
  alias_method :with_configured_adapter, :with_adapter
  ^^^^^^^^^^^^
156 files inspected, 5 offenses detected
RuboCop failed!
```

`Style/Alias` is a recent update to rubocop that was lagging the ruby-style-guide

That guard clause needs an inline `rubocop:disable Style/GuardClause` iirc

lib/active_model_serializers/adapter/base.rb
```diff
@@ -51,6 +51,20 @@ def include_meta(json)
         json[meta_key] = meta if meta
         json
       end
+
+      def translate_key_casing!(value, serialization_context)
```
NullVoxPopuli: thoughts on writing these transforms in the style of active support inflections? https://github.com/rails-api/active_model_serializers/pull/1566

at least for the sake of examples in comments, 

maybe the key transforms could be in their own module?

```ruby
module WhateverNamespace::KeyTransforms
  module_function
  # transforms keys to UpperCamelCase or PascalCase (this is what through me down this train of thought in the first place, though, this could be because I've been in non-ruby land at work)
  #
  # @example:
  #    "some_key" => "SomeKey",
  # @see: active_support/inflector#camelize
  def camel(hash)
    hash.deep_transform_keys! { |key| key.to_s.camelize.to_sym }
  end
  ...
end
```

and then this method would become:
```ruby
return value unless serialization_context
transform = serialization_context.key_case
KeyTransforms.send(transform)
```

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -22,11 +22,12 @@ def initialize(serializer, options = {})
...
-          success_document(options)
-        else
-          failure_document
-        end
+        document = if serializer.success?
+                     success_document(options)
+                   else
+                     failure_document
+                   end
+        translate_key_casing!(document, options[:serialization_context])
```
NullVoxPopuli: is this method mutating? or does it return a different hash?

like,

does `document.object_id == translate_key_casing!(document, {...}).object_id`?

remear: Should be mutating.

lib/active_model_serializers/adapter/base.rb
remear: I'm not necessarily opposed to it. Are there other benefits to this approach over what's in the PR?

NullVoxPopuli: it would give a place for users to add their own transformations if they wish. Rather than overwriting translate_key_casing, they write their own transform method -- without losing or having to re-implement the 3 you have. I could see this being especially useful, for APIs that have clients that don't all expect the same casing.

remear:
@NullVoxPopuli Implemented. Let me know what you think.

remear:
Also note that now only the `JsonApi` adapter has the `:dashed` default key case transform. The `Json` defaults to `:unaltered`.

lib/active_model_serializers/key_transform.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    # Returns the hash unaltered
+    def unaltered(hash)
+      hash
+    end
+
+    module_function :camel, :camel_lower, :dashed, :unaltered
+
+    def self.method_missing(m, *args, &block)
+      warn "\"#{m}\" is not a defined key transform. Defaulting to :unaltered."
```
NullVoxPopuli: nice!

NullVoxPopuli: can you add a test for this functionality, to make sure we don't get infinite method_missing loops?

test/adapter/json_api/key_case_test.rb
```diff
@@ -0,0 +1,323 @@
...
+          resource.errors.add(:title, 'must be longer')
+
+          serializer = ActiveModel::Serializer::ErrorSerializer.new(resource)
+          adapter = ActiveModelSerializers::Adapter::JsonApi.new(serializer)
+          result = adapter.serializable_hash(@options)
+
+          expected_errors_object =
+            { :errors =>
+              [
+                { :source => { :pointer => '/data/attributes/published_at' }, :detail => 'must be in the future' },
```
NullVoxPopuli: is published_at still supposed to be underscored here?

remear: This one is tricky because we're only modifying keys at the end of the document building and this is in a value. I figured we could look more into it when the error document gets links and meta support.

NullVoxPopuli: right, so... maybe it would be the responsibility of the error serializer to change the keying on the attribute?

remear: It's sounding ideal.

NullVoxPopuli: that could be a separate PR, unless you are feeling adventurous. :-)

lib/active_model_serializers/key_transform.rb
remear: Good call. I guess I forgot to add it.

NullVoxPopuli:
can you rebase, so you get the new JRuby modifications from @bf4 ?

lib/active_model/serializable_resource.rb
```diff
@@ -33,7 +33,7 @@ def serialization_scope_name=(scope_name)
     def adapter
       @adapter ||= ActiveModelSerializers::Adapter.create(serializer_instance, adapter_opts)
     end
-    alias adapter_instance adapter
+    alias_method :adapter_instance, :adapter
```
bf4: :exclamation: revert, revert!

lib/active_model_serializers/adapter/base.rb
```diff
@@ -1,6 +1,9 @@
+require 'active_model_serializers/key_transform'
+
 module ActiveModelSerializers
   module Adapter
     class Base
+      include ActiveModelSerializers::KeyTransform
```
bf4: why a module?

lib/active_model_serializers/key_transform.rb
```diff
@@ -0,0 +1,43 @@
+module ActiveModelSerializers
+  module KeyTransform
```
bf4: perhaps `module_function` here to allow public calling on the module but private methods on inclusion. I like me some utility function modules

lib/active_model_serializers/adapter/base.rb
NullVoxPopuli: what are you thinking for those set of methods?

bf4:
@remear changeset should include `config.jsonapi_resource_type = :plural` right?

lib/active_model_serializers/key_transform.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Transforms keys to UpperCamelCase or PascalCase.
+    #
+    # @example:
+    #    "some_key" => "SomeKey",
+    # @see: active_support/inflector#camelize
+    def camel(hash)
+      hash.deep_transform_keys! { |key| key.to_s.camelize.to_sym }
+    end
```
bf4: I think here you might mean to use a link like `# @see {https://github.com/rails/rails/blob/v5.0.0.beta3/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}`

NullVoxPopuli: I didn't know you could use that syntax for an actual link -- legit

remear:
I'm not sure how `config.jsonapi_resource_type = :plural` relates to these changes. Could you explain further?

lib/active_model_serializers/key_transform.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Returns the hash unaltered
+    def unaltered(hash)
+      hash
+    end
+
+    module_function :camel, :camel_lower, :dashed, :unaltered
+
+    def self.method_missing(m, *args, &block)
+      warn "\"#{m}\" is not a defined key transform. Defaulting to :unaltered."
+      send(:unaltered, *args, &block)
```
bf4: this is super dangerous to eat a method missing entirely.  I think the proper behavior here is to raise.  I mean, if I say `KeyTransform.micky_mouse` it should fail, right?

lib/active_model_serializers/serialization_context.rb
```diff
@@ -4,13 +4,14 @@ class << self
       attr_writer :url_helpers, :default_url_options
     end
 
-    attr_reader :request_url, :query_parameters
+    attr_reader :request_url, :query_parameters, :key_case
```
bf4: why `key_case` and not `key_transform`? or `key_format`?

lib/active_model_serializers/key_transform.rb
NullVoxPopuli: Do you mean you don't want to silently fail with a warning and default behavior? and would rather have it barf?

test/adapter/json_api/key_case_test.rb
bf4: only question is this:  what are valid key formats for JSON API? we should limit options to those, no?

bf4:
@remear `config.jsonapi_resource_type = :plural` isn't that key casing concern?

test/adapter/json_api/key_case_test.rb
bf4: also see `config.jsonapi_resource_type = :plural`

lib/active_model_serializers/serialization_context.rb
remear: It should be key_transform now that we're using this KeyTransform interface.

lib/active_model_serializers/key_transform.rb
remear: It just seemed nicer to just give back what was supplied and warn if they called an undefined method.

remear:
I think it can be considered a key transformation concern, not a key casing one. The same goes with the change for removing `?` from keys. I would like to see things of this nature make their way into key transformation over time. This PR has been focused specifically on key casing but proposes the beginnings of the interface capable of handling all the key transformation. Do we want this PR scope to increase to accommodate the larger picture, or do we prefer to iterate over a few more PRs?

NullVoxPopuli:
My vote is for iterating over a few PRs.

lib/active_model_serializers/key_transform.rb
bf4: @remear @NullVoxPopuli So, before I asked:

> If I say `KeyTransform.micky_mouse` it should fail, right?

Here's a better example, I think:

`KeyTransform.caml`.  Person mistyped `camel` and now instead of a NoMethodError they're getting back `:unaltered`.  That's surprising.

I feel pretty strongly this should fail fast.  The warning message is good, but should be a failure message.  

Also, since you're defining the method missing on the module, are you also eating up all method missing calls on the `Adapter::Base`.  Just seems like asking for trouble with only benefit is a silent failure with a warning somewhere in the logs.

NullVoxPopuli: oooo, that's a good point. that is a bad idea then.

NullVoxPopuli:
@remear can you rebase?

docs/general/key_transform.md
```diff
@@ -0,0 +1,34 @@
...
+
+##### Configuration option
+
+`key_transform` is set in `ActiveModelSerializers.config.key_transform`.
+
+```ActiveModelSerializers.config.key_transform = :camel_lower```
+
+##### Adapter default
+
+Each adapter has a default key transform configured:
```
NullVoxPopuli: nice, I was going to ask about that

NullVoxPopuli:
LGTM :+1:

docs/general/configuration_options.md
```diff
@@ -2,26 +2,84 @@
...
+##### key_transform
+
+The [key transform](key_transform.md) to use.
+
+Possible values:
+
+- `:camel` - ExampleKey
+- `:camel_lower` - exampleKey
+- `:dashed` - example-key
+- `:unaltered` - the original, unaltered key
```
bf4: missing default. perhaps intentional, but the others have it so, should mention something

docs/general/configuration_options.md
```diff
@@ -2,26 +2,84 @@
...
+Set the `type` attributes of resources to singular or plural.
+
+Possible values:
+
+- `:singular`
+- `:plural` (default)
+
+##### jsonapi_include_toplevel_object
+
+Include a [top level JSON API member](http://jsonapi.org/format/#document-jsonapi-object)
```
bf4: as long as we're here `s/JSON API/jsonapi`

docs/general/key_transform.md
```diff
@@ -0,0 +1,34 @@
...
+
+- `Json` - `:unaltered`
+- `JsonApi` - `:dashed`
```
bf4: `Attributes`? 

`Null`? lulz

bf4: I like calling dashed `kebab-case` like ember does... maybe I should make a PR to the ActiveSupport::Inflector to support that...

remear: As far as I can tell, those are marked as deprecated, thus, on their way out.

lib/active_model/serializer/configuration.rb
```diff
@@ -26,6 +26,7 @@ def config.array_serializer
         # Make JSON API top-level jsonapi member opt-in
         # ref: http://jsonapi.org/format/#document-top-level
         config.jsonapi_include_toplevel_object = false
+        config.key_transform = nil
```
bf4: should be `:unaltered` no?

docs/general/configuration_options.md
remear: Are these docs intended to refer to JSON API the spec? The adapter? Something else? Isn't the spec officially named `JSON API`?

lib/active_model/serializer/configuration.rb
bf4: Then there's no need for `default_key_transform`

docs/general/configuration_options.md
remear: It was intentional because it's mentioned elsewhere, but a mention would be good.

lib/active_model/serializer/configuration.rb
remear: Seems like that would preclude the ability to have a per-adapter default.

bf4: oh n/m, that let's you configure per adapter. but this raise a point..

if this value is set to anything truthy, it will affect all adapters.  Is that the intention?

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -37,15 +37,20 @@ def initialize(serializer, options = {})
...
-          success_document(options)
-        else
-          failure_document
-        end
+        document = if serializer.success?
+                     success_document(options)
+                   else
+                     failure_document
+                   end
+        transform_key_casing!(document, options[:serialization_context])
```
bf4: seeing as we're inheriting from a common base, I'm thinking we should just pass up the options?

remear: Maybe. Right now it only needs the serialization context.

lib/active_model_serializers/adapter/base.rb
bf4: `include` can now be removed as we're calling the function on KeyTransform directly

bf4: and maybe alias it `KeyTransform = ActiveModelSerializers::KeyTransform`

lib/active_model/serializer/configuration.rb
remear: For now. Did you have something else in mind?

lib/active_model_serializers/key_transform.rb
```diff
@@ -0,0 +1,38 @@
...
+    def dashed(hash)
+      hash.deep_transform_keys! { |key| key.to_s.dasherize.to_sym }
+    end
+
+    # Returns the hash unaltered
+    def unaltered(hash)
+      hash
+    end
+
+    module_function :camel, :camel_lower, :dashed, :unaltered
```
bf4: Might as well just put the `module_function` as the top scope of the Module. There doesn't seem to be any reason to exclude methods

lib/active_model/serializer/configuration.rb
bf4: Just we should document that. Could be surprising. Or make it a required method in every adapter by just not defining it in the Base

bf4:
@remear :100: we need to add a controller test.  We've gotten in trouble before when not writing those :)

docs/general/configuration_options.md
bf4: Yeah, but the member name is `jsonapi`

docs/general/key_transform.md
bf4: good call.

lib/active_model_serializers/key_transform.rb
```diff
@@ -0,0 +1,38 @@
+module ActiveModelSerializers
+  module KeyTransform
+    # Transforms keys to UpperCamelCase or PascalCase.
+    #
+    # @example:
+    #    "some_key" => "SomeKey",
+    # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
+    def camel(hash)
+      hash.deep_transform_keys! { |key| key.to_s.camelize.to_sym }
```
bf4: at top of file we should add `require 'active_support/core_ext/hash/keys'` to ensure the extension is loaded.

lib/active_model_serializers/key_transform.rb
NullVoxPopuli: It used to be in this file. not sure what happened to it.

docs/general/configuration_options.md
bf4: ok, I see that the missing default option is `nil: use the adapter default`

docs/general/configuration_options.md
```diff
@@ -2,26 +2,84 @@
...
+Possible values:
+
+- `true` (default)
+- `false`
+
+When `false`, serializers must be explicitly specified.
+
+##### key_transform
+
+The [key transform](key_transform.md) to use.
```
bf4: `config.key_transform` is a global override of the adapter default.  Adapters still prefer the render option `:key_transform` over this setting

docs/general/configuration_options.md
```diff
@@ -2,26 +2,93 @@
...
-   Default: `false`.
-- Used when `jsonapi_include_toplevel_object` is `true`:
-  - `jsonapi_version`: The latest version of the spec the API conforms to.
-    Default: `'1.0'`.
-  - `jsonapi_toplevel_meta`: Optional metadata. Not included if empty.
-    Default: `{}`.
+
+##### jsonapi_resource_type
+
+Set the `type` attributes of resources to singular or plural.
```
groyoh: > Set the `type` attributes

Using `attributes` here seems inappropriate. What about something like this:
```md
Sets whether the [type](http://jsonapi.org/format/#document-resource-identifier-objects) of the JSON API resource should be `singularized` or `pluralized` when it is not [explicitly specified by the serializer](https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#type):
```

remear: It could use more clarification. I didn't go out of my way to refactor the content here.

This is already in the JSON API section so the copy should read within that context.

```
Sets whether the [type](http://jsonapi.org/format/#document-resource-identifier-objects) of the resource should be `singularized` or `pluralized` when it is not [explicitly specified by the serializer](https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#type)
```

groyoh: Sounds good.

rthbound:
@bf4, @remear 
It looks like there's some problems with this... The changes in [this branch](https://github.com/rails-api/active_model_serializers/compare/master...rthbound:bug-in-1574) should highlight what I'm talking about.

Basically, we're transforming keys only... `bad_comments` where it would appear in keys is properly transformed, however `bad_comments` in values (e.g. `type: "bad_comments"`) is not transformed.

This also suggests the action controller tests written for this were incomplete.

remear:
It doesn't cover everything yet. For example, [this](https://github.com/rails-api/active_model_serializers/blob/master/test/adapter/json_api/key_case_test.rb#L408), ideally, would have been transformed from `published_at` to `published-at`. Transforming all values is too heavy-handed, so more finesse is needed for these cases. We'll get it fixed up soon.

dgdosen:
Still - this was greatly needed.  Thanks!

PR #1579, opened 2016-03-11
changes plural to pluralize in json_api resource identifier

longh957:
#### Purpose
Fixes typo that was failing using Json_api adapter

#### Changes
Changes plural to pluralize


remear:
@longh957 Could you elaborate on what exactly is failing, and perhaps provide a test that demonstrates the issue?

This currently fails with `NoMethodError:   undefined method 'pluralize'`.

NullVoxPopuli:
can you provide a test that proves that this was failing before (I believe you, but I'm lead to believe that it wasn't covered by a test before)

bf4:
@longh957 Closing as unable to reproduce. Please re-open if you can share more information.

PR #1580, opened 2016-03-11
Raise better exception when no Serializer is defined or found via lookup

rodrigosaito:
#### Purpose
When calling ```ActiveModel::SerializableResource.new(resource, {})``` and no Serializer can be found it was raising the following error:

```
ERROR["test_serializable_resource_when_no_serializer_found", ActiveModel::SerializableResourceTest::SerializableResourceErrorsTest, 5.4571671259909635]
 test_serializable_resource_when_no_serializer_found#ActiveModel::SerializableResourceTest::SerializableResourceErrorsTest (5.46s)
NoMethodError:         NoMethodError: undefined method `new' for nil:NilClass
            /app/lib/active_model/serializable_resource.rb:39:in `serializer_instance'
            /app/lib/active_model/serializable_resource.rb:34:in `adapter'
            /app/lib/active_model_serializers/logging.rb:84:in `notify_render_payload'
            /app/lib/active_model_serializers/logging.rb:78:in `notify_render'
            /app/lib/active_model_serializers/logging.rb:21:in `block (2 levels) in instrument_rendering'
            /app/lib/active_model_serializers/logging.rb:94:in `tag_logger'
            /app/lib/active_model_serializers/logging.rb:20:in `block in instrument_rendering'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:441:in `instance_exec'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:441:in `block in make_lambda'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:342:in `block in simple'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:497:in `block in around'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:505:in `call'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:92:in `__run_callbacks__'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:778:in `_run_render_callbacks'
            /usr/local/bundle/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:81:in `run_callbacks'
            /app/lib/active_model_serializers/logging.rb:68:in `block (2 levels) in notify'
            /app/test/serializable_resource_test.rb:78:in `test_serializable_resource_when_no_serializer_found'
```

#### Changes
Changed code to raise a better error with some useful explanation. Please tell me what error and message to raise and I will update this PR.

### DIFF
--- a/test/serializable_resource_test.rb
+++ b/test/serializable_resource_test.rb
@@ -71,6 +71,13 @@ module ActiveModel
         }
         assert_equal serializable_resource.as_json(options), expected_response_document
       end
+
+      def test_serializable_resource_when_no_serializer_found
+        resource = ModelWithErrors.new
+        serializable_resource = ActiveModel::SerializableResource.new(resource)
+        # TODO define error to raise here
+        assert_raises { serializable_resource.as_json }
+      end
     end
   end
 end
--- a/lib/active_model/serializable_resource.rb
+++ b/lib/active_model/serializable_resource.rb
@@ -36,7 +36,13 @@ module ActiveModel
     alias adapter_instance adapter
 
     def serializer_instance
-      @serializer_instance ||= serializer.new(resource, serializer_opts)
+      if @serializer_instance
+        @serializer_instance
+      else
+        # TODO define error to raise here
+        raise "No Serializer found" unless serializer
+        @serializer_instance = serializer.new(resource, serializer_opts)
+      end
     end
 
     # Get serializer either explicitly :serializer or implicitly from resource


bf4:
Hi, thanks for this.  I'm wondering if you can share some background on how your usage led to this failure.  

In the controller we first check serializer? before deciding if ams applies to the model.

I never considered outside controller usage where a model without a serializer were to be passed in.

I'm not sure what the desired behavior should be.  NoMethodError is certainly not the right exception, though the message and stacktrace seem clear enough.

In the controller passing in an object without a serializer skips ams.  I'm not sure if here we'd want a noop or original object when json is called.  Maybe jusy call as_json on it?

rodrigosaito:
@bf4 Hi, I am not using ams with rails or rails api so I followed this https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/outside_controller_use.md, thats how this error came up to me.

test/serializable_resource_test.rb
```diff
@@ -71,6 +71,13 @@ def test_serializable_resource_with_collection_containing_errors
         }
         assert_equal serializable_resource.as_json(options), expected_response_document
       end
+
+      def test_serializable_resource_when_no_serializer_found
+        resource = ModelWithErrors.new
+        serializable_resource = ActiveModel::SerializableResource.new(resource)
+        # TODO define error to raise here
+        assert_raises { serializable_resource.as_json }
```
NullVoxPopuli: can you also assert what message was thrown here? otherwise this catches all exceptions, which isn't exactly helpful, worst case scenario.

NullVoxPopuli: alternatively, AMS could have its own set of exception classes, but that should probably be another PR / issue for discussion.

rodrigosaito: about the assertion, yes I will assert what message/error was thrown. Just need the decision of what to be raised.

NullVoxPopuli: maybe "No serializer found for #{model_name}"


Though, I wonder, do we have tests for no-serializer? cause it should just use the default json renderer shouldn't it?

or are we wanting to explicity force the use of serializers? I could see that being useful for JSON API.. but... idk.

bf4: I'm thinking in terms of the principle of least surprise.  Would you be more surprised if we just call to_json etc on the object or if it blows up?

I really don't like the idea of it blowing up, unless the message were: ' did you check serializer? first? Did you mean to pass in the serializer name? Did you forget to make a serializer?'

I think the it may be least surprising if it behaves like the controller, and just didn't wrap the object in a serializer/adapter, or at least a 'pass through' adapter.

i.e. The controller says: can I serialize you ? Let's do this: pass through, as you wre

B mobile phone

NullVoxPopuli: @bf4, agreed -- that means instead of @rodrigosaito  throwing an exception, he should log those messages?

bf4: @NullVoxPopuli I'm thinking maybe not we should just change the interface locally and in the controller to be more forgiving.

```diff

    def get_serializer(resource, options = {})
      if !use_adapter?
        warn 'ActionController::Serialization#use_adapter? has been removed. '\
          "Please pass 'adapter: false' or see ActiveSupport::SerializableResource.new"
        options[:adapter] = false
      end
+      ActiveModel::SerializableResource.new(resource, options)
-      serializable_resource = ActiveModel::SerializableResource.new(resource, options)
-      if serializable_resource.serializer?
-        serializable_resource.serialization_scope ||= serialization_scope
-        serializable_resource.serialization_scope_name = _serialization_scope
-        begin
-          # Necessary to ensure we have an adapter for the serializable resource
-          # after it has been figured.
-          # TODO: This logic should be less opaque and probably moved into the SerializableResource.
 -         serializable_resource.tap(&:adapter)
 -       rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
 -         resource
 -       end
 -     else
 -       resource
 -     end
    end
```

and then:

bf4: serializable resource

```diff
    def adapter
      @adapter ||= get_adapter
    end
    alias adapter_instance adapter

   def get_adapter
     if serializer?
        begin
          # Necessary to ensure we have an adapter for the serializable resource
          # after it has been figured.
          ActiveModelSerializers::Adapter.create(serializer_instance, adapter_opts)
        rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
          resource
       end
   end
```

NullVoxPopuli: @bf4, is this soemthing you want to make a PR for?

I 'think', I like this, but I'd like to see an actual implementation to see how it behaves.

esp with the case of no serializer defined, and under what real-life condition would throw that NoSerializerError error

bf4: oh, the `NoSerializerError` is just an unfortunate part of our api right now. It's real.

I figure OP might want to rebase this with changes like that and force push.  I just wrote that up in the browser, might not be precise

bf4:
I'm thinking this PR needs to be adopted.  https://github.com/rails-api/active_model_serializers/pull/1580#discussion_r55843774 wanna take over?

rodrigosaito:
@bf4 @NullVoxPopuli I am willing to finish this, sorry for the delay. Just haven't had the time to move ahead yet.

bf4:
Made the pr https://github.com/rails-api/active_model_serializers/pull/1616

bf4:
Closed by https://github.com/rails-api/active_model_serializers/pull/1616

PR #1581, opened 2016-03-11
Omit meta when blank

remear:
#### Purpose

Given:

```ruby
meta do
  {
    status_details: object.status_details
  }.reject { |k,v| v.nil? }
end
```

When `object.status_details` is null, an empty `meta` member is added to the response, `meta: {}`. These changes alter the behavior to not include `meta` if it's blank. This applies to the `JsonApi` and `Json` adapters.

### DIFF
--- a/test/adapter/json_api/resource_meta_test.rb
+++ b/test/adapter/json_api/resource_meta_test.rb
@@ -17,6 +17,20 @@ module ActiveModel
             end
           end
 
+          class MetaBlockPostBlankMetaSerializer < ActiveModel::Serializer
+            attributes :id
+            meta do
+              {}
+            end
+          end
+
+          class MetaBlockPostEmptyStringSerializer < ActiveModel::Serializer
+            attributes :id
+            meta do
+              ''
+            end
+          end
+
           def setup
             @post = Post.new(id: 1337, comments: [], author: nil)
           end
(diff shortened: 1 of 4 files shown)


remear:
Again, my local rubocop has a lot of offenses listed and I don't know which ones are false positives. So expect this to fail for now.

test/adapter/json_api/resource_meta_test.rb
```diff
@@ -61,6 +75,24 @@ def test_meta_object_resource_in_array
...
             assert_equal(expected, hash)
           end
+
+          def test_meta_object_blank_omitted
+            hash = ActiveModel::SerializableResource.new(
+              @post,
+              serializer: MetaBlockPostBlankMetaSerializer,
+              adapter: :json_api
+            ).serializable_hash
+            refute hash[:data].has_key? :meta
```
NullVoxPopuli: refute?! woah, look who's good at minitest. :-)


srsly, I didn't know this was a thing

NullVoxPopuli:
LGTM! :-)

beauby:
So now explicitly setting `meta` to an empty hash makes it disappear?

NullVoxPopuli:
I believe so.

In what situation would you want it to be an empty hash?

groyoh:
Should this also be the case for relationship and link's meta in order to be consistent?

remear:
Probably. I forgot about those other two cases.

beauby:
@NullVoxPopuli When your client expects it. More generally, it is confusing when you explicitly set a value and it is automagically modified.

jwg2s:
@NullVoxPopuli @beauby also running into this - our client expects meta to always be set, so we're looking for a quick way to override this behavior and render an explicitly passed hash even if it's `blank?`.  Definitely agree with @beauby that it's surprising when you explicitly set a value and it's ripped out when rendering

bf4:
@jwg2s Thanks for offering a PR.  I think that whether meta is set to `nil` or `{}`, meta should be included, and it should be up to AMS internals to not add meta when none has been given.  That puts the ultimate behavioral control in the dev's hands, unless I'm forgetting some super hairy case.  If someone passes `meta do end`, the implicit nil is good enough for me.

PR #1583, opened 2016-03-13
Issue stats don't actually update anymore

bf4:


NullVoxPopuli:
I think I noticed that too, but I wasn't sure if they were broken for everyone.

PR #1593, opened 2016-03-15
Always be warning

bf4:
Turn on warnings when running `rake test` so devs always see them.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -41,8 +41,9 @@ require 'rake/testtask'
 
 Rake::TestTask.new do |t|
   t.libs << 'test'
+  t.libs << 'lib'
   t.test_files = FileList['test/**/*_test.rb']
-  t.ruby_opts = ['-r./test/test_helper.rb']
+  t.ruby_opts = ['-w -r./test/test_helper.rb']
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -41,8 +41,9 @@ require 'rake/testtask'
 
 Rake::TestTask.new do |t|
   t.libs << 'test'
+  t.libs << 'lib'
```
groyoh: Is this actually needed? I thought was in there by default.

groyoh:
Looks good to me  :+1:

Rakefile
bf4: it doesn't seem to need it. in fact, I removed it earlier.  But, most rake test tasks had it so I decided to cargo cult.  I'm find either way. :shrug:

remear: I'm fine with leaving it.

groyoh:
After trying it out, it actually creates more noise than anything else as it also shows warning from grape and action_pack. Not sure if we should keep it afterall.

remear:
It's noisy. I'd like to mention that I have a PR coming to fix the AMS warnings.

PR #1601, opened 2016-03-17
ActiveModel::Serializer _fast_attributes - raise NameError without ba…

remear:
#### Purpose

_fast_attributes - raise NameError without backtrace

Rebased version of #1305

#### Related GitHub issues

#1305

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -107,9 +107,8 @@ module ActiveModel
         #  attributes are added later in a classes lifecycle
         # poison the cache
         define_method :_fast_attributes do
-          raise NameError
+          raise NameError, nil, []
         end
-
       end
 
       def associate(klass, attrs) #:nodoc:


groyoh:
I just would like to point out that the benchmark described in #1305 is not that valid in real-life:

```ruby
serializers = []

10_000.times do
  serializers << Class.new(ActiveModel::Serializer) do
    attributes :id, :name
  end
end

Benchmark.bm do |x|
  x.report("serializable_hash") do
    serializers.each do |serializer|
      serializer.new(u).serializable_hash
    end
  end
end
``` 

First of all, it creates `10_000` serializer classes (is that actually done by anybody?). Then it only call `serializable_hash` once per serializer class. But the exception in `fast_attributes` is only raised the first times `attributes` is called AFAIK and thus should not be raised in subsequent calls.

With the following benchmark:

```ruby
serializers = []

50.times do
  serializers << Class.new(ActiveModel::Serializer) do
    attributes :id, :name
  end
end

Benchmark.ips do |x|
  x.warmup = 40
  x.report("serializable_hash") do
    serializers.each do |serializer|
      serializer.new(u).serializable_hash
    end
  end
end
```

we get the following result:

* Before changes
```shell
Warming up --------------------------------------
   serializable_hash   507.000  i/100ms
   serializable_hash      5.163k (± 2.8%) i/s -     25.857k
```

* After changes

```shell
Warming up --------------------------------------
   serializable_hash   478.000  i/100ms
   serializable_hash      4.859k (± 1.2%) i/s -     24.378k
```

I've ran it many times and it actually seems slower everytime, even though I'm not sure why. There fore I'm not sure if we should merge this.

remear:
@groyoh Using which Ruby version(s)?

groyoh:
JRuby 1.7.15 and JRuby 1.7.23 as the original PR was refering to JRuby.

groyoh:
@remear I'm closing it for now. If you do not agree with my point of view, feel free to re-open. I am open for discussion on this.

PR #1609, opened 2016-03-20
Namespace separator setting

youroff:
#### Purpose
Introduces a global setting to control namespace separator for generating `type` attribute when using `json-api`.
`Cool::Author` becomes `cool--authors` with the default value `--`.
Using custom separator:
```
ActiveModelSerializers.config.jsonapi_namespace_separator = '/'
```
This will make for `cool/authors` which is supported by Ember by default.

#### Changes
`ActiveModelSerializers::Adapter::JsonApi::ResourceIdentifier.type_for`

#### Caveats
As a proposal, we may need to introduce type_transform setting, like key_transform. To provide ways to chose between undersored and dasherized types.

#### Related GitHub issues
Reimplementation of:
https://github.com/rails-api/active_model_serializers/pull/1216

### DIFF
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -63,6 +63,21 @@ Possible values:
 - `:singular`
 - `:plural` (default)
 
+##### jsonapi_namespace_separator
+
+Sets separator string for namespaced models to render `type` attribute. Default value is `--`.
+
+##### jsonapi_type_transform
+
+Provides transform for `type` attribute. Class name `NicePost` gets converted into `nice_post`, `nice-post`, `NicePost` or `nicePost` depending on selected setting.
+
+Possible values:
+
+- `:underscore` (default)
+- `:dashed`
+- `:camel`
+- `:snake`
+
 ##### jsonapi_include_toplevel_object
 
 Include a [top level jsonapi member](http://jsonapi.org/format/#document-jsonapi-object)
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -21,6 +21,8 @@ module ActiveModel
 
         config.adapter = :attributes
         config.jsonapi_resource_type = :plural
+        config.jsonapi_namespace_separator = '--'.freeze
+        config.jsonapi_type_transform = :underscore
         config.jsonapi_version = '1.0'
         config.jsonapi_toplevel_meta = {}
         # Make JSON API top-level jsonapi member opt-in
--- a/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
+++ b/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
@@ -20,11 +20,13 @@ module ActiveModelSerializers
 
         def type_for(serializer)
           return serializer._type if serializer._type
-          if ActiveModelSerializers.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
-          else
-            serializer.object.class.model_name.plural
+          type = serializer.object.class.to_s.split('::')
+          type.map! { |t| KeyTransform.transform(t, ActiveModelSerializers.config.jsonapi_type_transform) }
+          type = type.join ActiveModelSerializers.config.jsonapi_namespace_separator
+          if ActiveModelSerializers.config.jsonapi_resource_type == :plural
+            type = type.pluralize
           end
+          type
         end
 
         def id_for(serializer)
--- a/test/adapter/json_api/resource_identifier_test.rb
+++ b/test/adapter/json_api/resource_identifier_test.rb
@@ -22,62 +22,60 @@ module ActiveModelSerializers
         end
 
         def test_defined_type
-          test_type(WithDefinedTypeSerializer, 'with_defined_type')
+          assert_identifier(WithDefinedTypeSerializer.new(@model), type: 'with_defined_type')
         end
 
         def test_singular_type
-          test_type_inflection(AuthorSerializer, 'author', :singular)
+          assert_with_confing(AuthorSerializer.new(@model), type: 'author', inflection: :singular)
         end
 
         def test_plural_type
-          test_type_inflection(AuthorSerializer, 'authors', :plural)
+          assert_with_confing(AuthorSerializer.new(@model), type: 'authors', inflection: :plural)
+        end
+
+        def test_type_with_namespace
+          spam = Spam::UnrelatedLink.new
+          assert_identifier(Spam::UnrelatedLinkSerializer.new(spam), type: 'spam--unrelated_links')
+        end
+
+        def test_type_with_custom_namespace
+          spam = Spam::UnrelatedLink.new
+          assert_with_confing(Spam::UnrelatedLinkSerializer.new(spam), type: 'spam/unrelated_links', namespace_separator: '/')
         end
 
         def test_id_defined_on_object
-          test_id(AuthorSerializer, @model.id.to_s)
+          assert_identifier(AuthorSerializer.new(@model), id: @model.id.to_s)
         end
 
         def test_id_defined_on_serializer
-          test_id(WithDefinedIdSerializer, 'special_id')
+          assert_identifier(WithDefinedIdSerializer.new(@model), id: 'special_id')
         end
 
         def test_id_defined_on_fragmented
           FragmentedSerializer.fragmented(WithDefinedIdSerializer.new(@model))
-          test_id(FragmentedSerializer, 'special_id')
+          assert_identifier(FragmentedSerializer.new(@model), id: 'special_id')
         end
 
         private
 
-        def test_type_inflection(serializer_class, expected_type, inflection)
-          original_inflection = ActiveModelSerializers.config.jsonapi_resource_type
-          ActiveModelSerializers.config.jsonapi_resource_type = inflection
-          test_type(serializer_class, expected_type)
-        ensure
-          ActiveModelSerializers.config.jsonapi_resource_type = original_inflection
-        end
-
-        def test_type(serializer_class, expected_type)
-          serializer = serializer_class.new(@model)
-          resource_identifier = ResourceIdentifier.new(serializer)
-          expected = {
-            id: @model.id.to_s,
-            type: expected_type
-          }
-
-          assert_equal(expected, resource_identifier.as_json)
-        end
-
-        def test_id(serializer_class, id)
-          serializer = serializer_class.new(@model)
-          resource_identifier = ResourceIdentifier.new(serializer)
+        def assert_with_confing(serializer, opts = {})
           inflection = ActiveModelSerializers.config.jsonapi_resource_type
-          type = @model.class.model_name.send(inflection)
-          expected = {
-            id: id,
-            type: type
-          }
+          namespace_separator = ActiveModelSerializers.config.jsonapi_namespace_separator
+          ActiveModelSerializers.config.jsonapi_resource_type = opts.fetch(:inflection, inflection)
+          ActiveModelSerializers.config.jsonapi_namespace_separator = opts.fetch(:namespace_separator, namespace_separator)
+          assert_identifier(serializer, opts)
+        ensure
+          ActiveModelSerializers.config.jsonapi_resource_type = inflection
+          ActiveModelSerializers.config.jsonapi_namespace_separator = namespace_separator
+        end
 
-          assert_equal(expected, resource_identifier.as_json)
+        def assert_identifier(serializer, opts = {})
+          identifier = ResourceIdentifier.new(serializer)
+          expected = {
+            id: opts.fetch(:id, identifier.as_json[:id]),
+            type: opts.fetch(:type, identifier.as_json[:type])
+          }
+          assert_equal(expected, identifier.as_json)
         end
       end
     end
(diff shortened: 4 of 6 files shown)


docs/general/configuration_options.md
```diff
@@ -63,6 +63,10 @@ Possible values:
 - `:singular`
 - `:plural` (default)
 
+##### jsonapi_namespace_separator
+
+Sets separator string for namespaced models to render `type` attribute. Default value is `--`.
+
```
bf4: An alternative name or perhaps alias would be `dashed_namespace_separator` (which might also make a nice PR into Rails, to make it easier to configure the replacement string from `/` to `--`)

lib/active_model/serializer/configuration.rb
```diff
@@ -21,6 +21,7 @@ def config.array_serializer
 
         config.adapter = :attributes
         config.jsonapi_resource_type = :plural
+        config.jsonapi_namespace_separator = '--'
```
bf4: `.freeze`

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
```diff
@@ -20,11 +20,13 @@ def as_json
 
         def type_for(serializer)
           return serializer._type if serializer._type
-          if ActiveModelSerializers.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
-          else
-            serializer.object.class.model_name.plural
+          type = serializer.object.class.to_s.split('::')
+          type.map! { |t| t.underscore.downcase }
```
bf4: I think this should re-use or add logic in [KeyTransform](https://github.com/rails-api/active_model_serializers/blob/61412d80cabc3900dfbd886bc3f25fb0e0d39433/lib/active_model_serializers/key_transform.rb#L30-L32)

(And for bonus points, there's something in [fragment_cache](https://github.com/rails-api/active_model_serializers/blob/61412d80cabc3900dfbd886bc3f25fb0e0d39433/lib/active_model_serializers/fragment_cache.rb#L112-L113) that also does some class stringify munging that looks similar to the above, and we might want to include)

test/adapter/json_api/resource_identifier_test.rb
```diff
@@ -22,62 +22,60 @@ class FragmentedSerializer < ActiveModel::Serializer; end
...
+          test(FragmentedSerializer.new(@model), id: 'special_id')
         end
 
         private
 
-        def test_type_inflection(serializer_class, expected_type, inflection)
-          original_inflection = ActiveModelSerializers.config.jsonapi_resource_type
-          ActiveModelSerializers.config.jsonapi_resource_type = inflection
-          test_type(serializer_class, expected_type)
+        def test(serializer, opts = {})
```
bf4: Maybe rename to `assert_something`?  given the way test::unit /minitest works, calling something `test`, even if private, triggers my spidey sense. Also, it would make it more clear what the assertion is.

(That said, I see the original code already has such private methods, so, :8ball: :eyes: :question: )

bf4:
@youroff Awesome, thanks for this!

docs/general/configuration_options.md
youroff: Not sure I understand. It's supposed to be a setting for any custom separator, not only dasherized. Why should it be called `dashed_namespace_separator`? The point, that I didn't use `model_name` from rails. Just plain and simple class name processing.

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
youroff: Yeah, as I mentioned in PR description, type name transform should be customizable too, but I'm not sure if that's always the case that it's the same as `key_transform`. Is it? Technically we could apply `key_transform` on type's segments. That would be pretty easy, but is it flexible enough?

test/adapter/json_api/resource_identifier_test.rb
youroff: Yeah, initially I made it `test_with_config` and `test_serializer`, but made it shorter and unified. I'll return it back.

youroff:
Also, don't you mind if I remove this test in this PR?
https://github.com/rails-api/active_model_serializers/blob/master/test/adapter/json_api/type_test.rb
It seems to be a duplication since `ResourceIdentifier` has been introduced.

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
NullVoxPopuli: @bf4 Even though it makes perfect logical sense to have this in `KeyTransform`, cause it is a utility module that transforms keys, would it be too much of a performance hit to add the logic for the separator?

At the very least, this logic could be added as a new method in the `KeyTransform` module.

youroff: It makes perfect sense to use KeyTransform (with according setting), but what if someone needs to keys underscored and type dasherized? Is it unlikely? Or we're gonna use methods from KeyTransform with another setting?

youroff: I just checked KeyTransform and I don't see how it could help here. Correct me if I'm wrong, but it transforms keys in supplied hash, and we're only working with string here. Any suggestions how to apply it here?

youroff: What I would do (and it seems like a pretty big change) is extracting transform on string somewhere and use it from KeyTransform and ResourceIdentifier independently. Would it work?

NullVoxPopuli: oh, I would have just added a new method to KeyTransform and called KeyTransform.whatever_method directly

youroff: Ok, makes sense. How about if we leave simple transforms on strings in KeyTransform and move call on hash (with `deep_transform_keys!`) to where it's actually used/defined? Otherwise there'll be code repetition or misleading method naming... For example camel_lower is already doing that with hash keys, so I will have to come up with some not so clean name for simple method for strings.

youroff:
Just ran into some ambiguity in KeyTransform names. Currently it's:
```
- `:camel` - ExampleKey
- `:camel_lower` - exampleKey
- `:dashed` - example-key
- `:unaltered` - the original, unaltered key
- `nil` - use the adapter default
```
First off, I suggest to rename `:camel_lower` into a `:snake`, I believe it's widely used name for this case.
Second, `:unaltered` leaves no clues what's it. Probably, would be better to call it `:underscore`. I made simple method in KeyTransform that makes cross conversion pretty easy, so we could use it with a parameter, everywhere it's needed.
```
    def transform(string, type)
      string = string.underscore
      case type.to_sym
      when :dashed
        string.dasherize
      when :camel
        string.camelize
      when :snake
        string.camelize(:lower)
      else
        string
      end
    end
```

What do you think, guys?

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
```diff
@@ -20,11 +20,13 @@ def as_json
...
+          type = type.join ActiveModelSerializers.config.jsonapi_namespace_separator
```
bf4: I was thinking this would be a good use-case for gsub, but it looks like that is would be marginally slower and a little harder to read

```ruby
Benchmark.realtime { 1000.times { 'AdminUser::Person::THING'.gsub(/[a-zA-Z]+(::|\z)/) do |sub| sub.underscore.downcase.sub('/','--') end } }
# => 0.07714433899673168
>> Benchmark.realtime { 1000.times { 'AdminUser::Person::THING'.split('::').map! {|t| t.underscore.downcase }.join('--') } }     
# => 0.07239663798827678
```

bf4:
@youroff I think I'm ok to merge this. Do you feel done with it?

youroff:
There were some not committed additions. Please check last commit in this PR. And probably I need to rebase as there's some conflict already.

bf4:
@youroff @remear think we can merge this soon?

youroff:
@bf4 Yup, I'll prepare it today

youroff:
Well, I think this is ready, though CI is failing, not sure why. Any hints?

NullVoxPopuli:
It says spec coverage is not high enough

youroff:
Oh, right, the test for transforms is definitely needed. Will add that soon.

remear:
FYI, I think key_transform will soon become https://github.com/remear/active_model_serializers/blob/transforms/lib/active_model_serializers/key_transform.rb via #1645.

With the changes in #1645, I'm having a hard time seeing the reason for the transform method for KeyTransform. If it's something you still want, could you share some more reasoning.

bf4:
or https://github.com/rails-api/active_model_serializers/blob/0-9-stable/lib/active_model/serializer.rb#L260-L282

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
```diff
@@ -20,11 +20,13 @@ def as_json
 
         def type_for(serializer)
           return serializer._type if serializer._type
-          if ActiveModelSerializers.config.jsonapi_resource_type == :singular
-            serializer.object.class.model_name.singular
-          else
-            serializer.object.class.model_name.plural
+          type = serializer.object.class.to_s.split('::')
+          type.map! { |t| KeyTransform.transform(t, ActiveModelSerializers.config.jsonapi_type_transform) }
+          type = type.join ActiveModelSerializers.config.jsonapi_namespace_separator
```
bf4: basically, this method can just piggy back on #1645 right @remear @youroff

remear: that's what i'm thinking

youroff:
I agree, looks good. So should I wait until that PR merged?
I would just rename `camel_lower` to `snake` and make sure that it accepts any case. For example:
```
2.3.0 :001 > "some-key".camelcase
 => "Some-key"
```
So it may be good idea to underscore string first to make it suitable to cross-conversion in any direction.

youroff:
And I'm not sure what exactly is unaltered. Is there some assumption what is the default format provided by backend? Shouldn't we rely on some particular case per adaptor? Maybe custom lambda in config would do the trick for some weird cases if there are any?

bf4:
`unaltered` is the Rails default which is snake_case, but I suppose it's technically whatever `as_json` or `to_json` return if someone has customized that

remear:
I'm not sure underscored is an assumption we want to make. Therefore, `unaltered` is simply the value as it was supplied with no alterations performed.

bf4:
> I'm not sure underscored is an assumption we want to make. 

Yeah, I just meant that Rails will default to outputting snake case, which we wouldn't alter

youroff:
Ok, then if we don't have any assumption on what is provided, then we need to underscore everything before conversion, as I mentioned earlier, if provided dasherized string it won't be converted into camel case correctly.

PS: Disregard proposal on snake case, seems like I remember it wrong, wiki says it's underscored.

remear:
Sure, that seems necessary.

For those following along, the problem is:

```ruby
[1] pry(main)> value = 'some-value'
=> "some-value"
[2] pry(main)> value.underscore
=> "some_value"
[3] pry(main)> value.camelize
=> "Some-value"
[4] pry(main)> value.underscore.camelize
=> "SomeValue"
```

youroff:
So I remove changes to KeyTransform and wait until that PR is merged to rebase and finish this one?

remear:
I'm adding the underscore changes to KeyTransform now. It would be great if you could look them over in a little bit. We'll then merge in #1645 and this can be rebased to work with those changes.

remear:
@youroff Added -- https://github.com/remear/active_model_serializers/blob/transforms/lib/active_model_serializers/key_transform.rb

youroff:
Hmm.. I think there are couple of potential problems, not severe but they can be easily fixed and code will be shorter. So as far as I see, we don't have configuration validation, so if we call transform through `:send` it may be insecure. Besides there's a code repetition that could be eliminated if we supplied transformation as a parameter. Something like this:
https://gist.github.com/youroff/f0200c24d8fe168455ee4a779064bce0
What do you think?

remear:
A key _(tee hee)_ reason  for the current structure is so users can add custom transforms. Seems like this approach would hinder that ability.

We can continue discussing this using #1645 as a base, which I'll be merging shortly.

bf4:
`so if we call transform through :send it may be insecure`

value comes from a config, so it under control, no?  there shouldn't be anything going into the `send` coming from external sources

Unless you mean `public_send`, which makes no difference since `send` is a public method

youroff:
Yeah, I meant that user can mess things up himself. That's why I said it wasn't really severe. For custom transforms I would use lambda in config. But I'm ok with either way, just thought it could be nicer.

youroff:
And actually, with proposed implementation it can be extended too, quite easily. We just call super, unless value is String and type is new custom type and provide custom implementation otherwise.

youroff:
There are some breaking changes introduced here:
https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api/resource_identifier.rb#L8
Not sure how to deal with it. I would just remove options from resource identifier, to rely on settings again. Any ideas?

NullVoxPopuli:
@youroff could you rebase? :-)

bf4:
Merged into master and fixed tests. Submitted merge-result  for review before merge into master is pushed https://github.com/rails-api/active_model_serializers/pull/1791

bf4:
Closed by https://github.com/rails-api/active_model_serializers/commit/cd9689683ce84aaa449739631a87b50465153ba6

PR #1615, opened 2016-03-24
Singularize and capitalize polymorphic types

dpdawson:
Singularizing polymorphic association types allows you to use plural or singular
type values in your requests as allowed by JSON API.

Because ActiveRecord calls #constantize on the polymorphic association type
when you use the hash to create a new model, you must capitalize the type to
avoid a NameError.


test/adapter/json_api/parse_test.rb
```diff
@@ -125,7 +125,7 @@ def test_polymorphic
               src: 'http://example.com/images/productivity.png',
               author_id: nil,
               photographer_id: '9',
-              photographer_type: 'people',
```
bf4: So, this shouldn't change, as this is the serialization

bf4: n/m I take that back.  You're right. I misread that this was the input being changed, not the output.  Rails needs this format as the deserialized attributes.

bf4:
@dpdawson Awesome!  Would you mind adding a changelog entry at the top of the fixes section?  Also, a test of an ActiveRecord object using this result would be fantastic, if you don't mind the extra effort.

dpdawson:
As I looked into it more, I realized the error only gets thrown when you validate the relationship is present. The fix doesn't cause any of the other tests to fail, so I assume it's ok.

I added a test, but I'm sure it's not the best implementation because I don't use Minitest and the organization of the tests is a bit difficult to discern at first glance.

lib/active_model_serializers/adapter/json_api/deserialization.rb
```diff
@@ -188,7 +188,7 @@ def parse_relationship(assoc_name, assoc_data, options)
             end
 
           polymorphic = (options[:polymorphic] || []).include?(assoc_name.to_sym)
-          hash["#{prefix_key}_type".to_sym] = assoc_data['type'] if polymorphic
+          hash["#{prefix_key}_type".to_sym] = assoc_data['type'].singularize.capitalize if polymorphic
```
bf4: we probably want to encapsulate this singular.capitalize logic in a transformer cc @remear

test/active_record_test.rb
```diff
@@ -5,5 +5,38 @@ class ActiveRecordTest < ActiveSupport::TestCase
...
+              id: @image.id
+            }
+          },
+        },
+        type: 'pictures'
+      }
+    })
+
+    ActiveModelSerializers::Deserialization.jsonapi_parse!(
+      params.to_unsafe_h,
```
bf4: lol

test/active_record_test.rb
```diff
@@ -5,5 +5,38 @@ class ActiveRecordTest < ActiveSupport::TestCase
 
   def setup
     @resource = ARModels::Post.new
+    @image = Image.create
+  end
+
+  def test_active_model_record_with_validated_polymorphic_relationship_creation
+    picture = Picture.create!(picture_params)
+
+    assert_equal(@image, picture.imageable)
```
bf4: looks good

bf4:
With a rebase and a brief look at the Transformer with @remear this should be ready to mege

dpdawson:
I'd be happy to take a look if pointed in the right direction.

remear:
@dpdawson Could you rebase this off master and take a look at adjusting `KeyTransform` to suit the needs of `singularize.capitalize`.

lib/active_model_serializers/adapter/json_api/deserialization.rb
zigzackattack: could we `camelize` instead of `capitalize`? that would add support for snake cased multi word model names, right?

bartiaco:
Any updates on this? Or has the problem been solved in a different way? I'm currently running into this issue.

bartiaco:
One note: I don't think the current implementation will work for multi-word classes, given that the standard wordbreak in jsonapi is a '-'

"blog-posts".singularize.capitalize
=> "Blog-post"

brunowego:
I get this situation here, by now, I solve with hardcoded fix on controller. I [made a SO](http://stackoverflow.com/questions/43714000/polymorphic-association-with-ams-using-jsonapi-specification) because I had not seen this. Please consider merge it.

bf4:
@brunowego @bartiaco I'm reading the comments above and looks like the PR didn't get finished by anyone. Unless @dpdawson wants to pick it up again, it'll need to be reworked against master, and also related #1928  cc @NullVoxPopuli

lib/active_model_serializers/adapter/json_api/deserialization.rb
ClayShentrup: `assoc_data.fetch('type').underscore.classify`

Using [] is dangerous.

PR #1616, opened 2016-03-25
SerializableResource handles no serializer like controller

bf4:
Fixes when passing in a resource to SerializableResource outside of a controller, a
resource without a known Serializer will no longer raise
`NoMethodError: undefined method 'new' for nil:NilClass`, but will behave like the
controller and just use the resource, as is.

Follows https://github.com/rails-api/active_model_serializers/pull/1580#discussion_r55843774

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -33,20 +33,12 @@ module ActionController
         options[:adapter] = false
       end
       serializable_resource = ActiveModel::SerializableResource.new(resource, options)
-      if serializable_resource.serializer?
-        serializable_resource.serialization_scope ||= serialization_scope
-        serializable_resource.serialization_scope_name = _serialization_scope
-        begin
-          # Necessary to ensure we have an adapter for the serializable resource
-          # after it has been figured.
-          # TODO: This logic should be less opaque and probably moved into the SerializableResource.
-          serializable_resource.tap(&:adapter)
-        rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
-          resource
-        end
-      else
-        resource
-      end
+      serializable_resource.serialization_scope ||= serialization_scope
+      serializable_resource.serialization_scope_name = _serialization_scope
+      # For compatibility with the JSON renderer: `json.to_json(options) if json.is_a?(String)`.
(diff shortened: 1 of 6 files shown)


lib/action_controller/serialization.rb
```diff
@@ -33,20 +33,12 @@ def get_serializer(resource, options = {})
...
-        end
-      else
-        resource
-      end
+      serializable_resource.serialization_scope ||= serialization_scope
+      serializable_resource.serialization_scope_name = _serialization_scope
+      # For compatibility with the JSON renderer: `json.to_json(options) if json.is_a?(String)`.
+      # Otherwise, since `serializable_resource` is not a string, the renderer would call
+      # `to_json` on a String and given odd results, such as `"".to_json #=> '""'`
+      serializable_resource.adapter.is_a?(String) ? serializable_resource.adapter : serializable_resource
```
groyoh: Does that mean that someone would try to serialize a string directly?

lib/active_model/serializable_resource.rb
```diff
@@ -30,11 +30,19 @@ def serialization_scope_name=(scope_name)
       serializer_opts[:scope_name] = scope_name
     end
 
+    # NOTE: if no adapter is available, returns the resource itself.
     def adapter
```
groyoh: `adapter` does not seem to be the best name any longer.

lib/action_controller/serialization.rb
bf4: No... as per the comment, this handles `render {foo: :bar}.to_json`

lib/active_model/serializable_resource.rb
bf4: well, it is and it isn't.  It's essentially a no-op adapter but without a special object of that name wrapping it :)

bf4:
@groyoh I don't have these failures locally. It makes it hard for me to debug

groyoh:
@bf4 yeah, I actually tried it it out too earlier. Restarted the build too and couldn't figure out what's going on.

bf4:
@groyoh I believe it's related to the notifications being passed a nil serializer since that can now happen

Issue #1617, opened 2016-03-25
CollectionSerializer tests aren't real


bf4:
#### Expected behavior vs actual behavior

While working on following up to https://github.com/rails-api/active_model_serializers/pull/1537/ I noticed:

The CollectionSerializer should only ever have in it a collection, such as an Array or ActiveRecord::Relation. i.e. something that responds to `:each`

The CollectionSerializer has a fallback for the collection root/json_key `object.try(:name)` (where object is the collection of resources based in).

The tests since #1007 and #1013 have monkey patched `object` to have the method `name`.

If we change `object.try(:name)` to `object.try!(:name)` we raise an exception since a regular Array `[]` does not have a `name` method, and `try!` which checks if the receiver responds to the method.

However, the code does not ensure that the `object` is actually responds to name.  e.g. an `Array` doesn't have a name, but the tests use a helper `build_named_collection` that returns an monkey-patched array `[]` that has a `[].name #=> MeResource`

This failure is hidden by the `try` methods.

```ruby
        key = serializers.first.try(:json_key) || object.try(:name).try(:underscore)
        key.try(:pluralize)
```

To make the interface clear, it would be best to remove the `try` methods altogether

#### Steps to reproduce
*(e.g., detailed walkthrough, runnable script, example application)*

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 

0.10.0.rc4, `git rev-parse master` 61412d80cabc3900dfbd886bc3f25fb0e0d39433

Output of `ruby -e "puts RUBY_DESCRIPTION"`: 

ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]

OS Type & Version: 

Mac 10.10.5 Yosemite

Integrated application and version *(e.g., Rails, Grape, etc)*: 

n/a

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*

#### Additonal helpful information
*(e.g., Gemfile.lock, configurations, PR containing a failing test, git bisect results)*

PR #1618, opened 2016-03-25
Empty collection root key from explicit serializer option

bf4:
Fixes https://github.com/rails-api/active_model_serializers/issues/1617

Closes https://github.com/rails-api/active_model_serializers/pull/1537

- [x] when empty collection, get key from explicit serializer
- [ ] serializer instance `json_key` uses `_type` when root not given
  before falling back on object class
- [ ] remove crazy metaprogramming in collection test that monkey patches
  Array instances.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -182,7 +182,7 @@ module ActiveModel
 
     # Used by adapter as resource root.
     def json_key
-      root || object.class.model_name.to_s.underscore
+      root || _type || object.class.model_name.to_s.underscore
     end
 
     def read_attribute_for_serialization(attr)
--- a/test/collection_serializer_test.rb
+++ b/test/collection_serializer_test.rb
@@ -3,6 +3,10 @@ require 'test_helper'
 module ActiveModel
   class Serializer
     class CollectionSerializerTest < ActiveSupport::TestCase
+      MessagesSerializer = Class.new(ActiveModel::Serializer) do
+        type 'messages'
+      end
+
       def setup
         @comment = Comment.new
         @post = Post.new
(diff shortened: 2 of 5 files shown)


lib/active_model/serializer.rb
```diff
@@ -182,7 +182,7 @@ def as_json(adapter_opts = nil)
 
     # Used by adapter as resource root.
     def json_key
-      root || object.class.model_name.to_s.underscore
+      root || _type || object.class.model_name.to_s.underscore
```
bf4: not required in this PR, but makes sense

lib/active_model/serializer.rb
```diff
@@ -182,7 +182,7 @@ def as_json(adapter_opts = nil)
 
     # Used by adapter as resource root.
     def json_key
```
groyoh: Thinking about it now, shouldn't it be the responsibility of the adapter to determine the json key/root?

groyoh:
Great job! The PR looks good.  I'll be ok to merge it once the rubocop is fixed.

lib/active_model/serializer.rb
bf4: it does, but it uses information from the serializer....

bf4:
Running

lib/active_model/serializer.rb
groyoh: From what I see, we can set the `root` as adapter option and call `serializer._type` and `serializer.object.class.model_name.to_s.underscore` from the adapter. Same goes for collection_serializer. What do you think?

test/collection_serializer_test.rb
```diff
@@ -3,6 +3,10 @@
 module ActiveModel
   class Serializer
     class CollectionSerializerTest < ActiveSupport::TestCase
+      MessagesSerializer = Class.new(ActiveModel::Serializer) do
```
groyoh: We should align on `Class.new(ActiveModel::Serializer)` vs `class Name; end`.

groyoh:
Merging when the build is done :wink:

PR #1620, opened 2016-03-25
Fix warnings

bf4:
CI output needs to be reviewed.

I'm also trying to figure out which tests cause
- `lib/action_dispatch/routing/route_set.rb:429: warning: instance variable @_routes not initialized`
and why.
- If we can silence all the grape warnings `gems/grape-0.15.0/lib/grape`

Also:
- Less hacky test routes setup
- Remove annoying progress reporter broken by warnings output

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -54,7 +54,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'simplecov', '~> 0.11'
   spec.add_development_dependency 'timecop', '~> 0.7'
-  spec.add_development_dependency 'minitest-reporters'
   spec.add_development_dependency 'grape', ['>= 0.13', '< 1.0']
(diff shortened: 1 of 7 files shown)


groyoh:
Looks like the Rails routes are broken.

test/support/rails_app.rb
```diff
@@ -16,12 +16,20 @@ class ActiveModelSerializers::RailsApplication < Rails::Application
 end
 ActiveModelSerializers::RailsApplication.initialize!
 
+Rails.application.routes.draw do
+  get ':controller(/:action(/:id))'
+  get ':controller(/:action)'
+end
 module TestHelper
-  Routes = ActionDispatch::Routing::RouteSet.new
```
groyoh: Tried it out locally, seems like removing this broke the routes.

groyoh:
The routes and grape warning are still quite annoying. Looking forward to get rid of them. Good job!

bf4:
I have a fix for routes

PR #1622, opened 2016-03-25
Use serializer name instead of model name to create fragment cache serializers

### LINKED ISSUE #1621: Superclass absorbs subclass attributes when using inherited serializers with fragment caching

Normally, adding attributes to an inherited serializer does not effect the base serializer. For example

```ruby
  attributes :id
end

class SpecialFooSerializer < FooSerializer
  attribute :name
end
```

In this case, `Foo`s serialized with the `SpecialFooSerializer` would gain the "name" field in the serialization, while `Foo`s serialized with the `FooSerializer` will not. This behavior is broken when using a fragment cache, because the Cached/NonCached serializers are made by model name, not serializer name. If you serialize a `Foo` with the `SpecialFooSerializer`, then internally a `FooCachedSerializer` is made which will be used for subsequent serializations of `Foo`s, whether they use the base serializer or not.

#### Steps to reproduce

Paste this into a Rails terminal. You'll see that both of the elements in the array are the same at the end, even though the second element should not have the name attribute.

```ruby
class Foo 
  include ActiveModel::Model
  attr_accessor :id, :name

    send attr
  end
end

  attributes :id
  cache key: 'foo', only: [:id]
end

class SpecialFooSerializer < FooSerializer
  attribute :name
  cache key: 'special_foo', only: [:id, :name]
end

foo = Foo.new id: 1, name: 'Test'

[
  ActiveModel::SerializableResource.new(foo, serializer: SpecialFooSerializer).serializable_hash,
  ActiveModel::SerializableResource.new(foo).serializable_hash
]
```

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 0.10.0.rc4

Output of `ruby -e "puts RUBY_DESCRIPTION"`: ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]

OS Type & Version: OS X El Capitan

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails

lserman:
Fixes #1621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Features:
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
 
 Fixes:
+- [#1622](https://github.com/rails-api/active_model_serializers/pull/1622) Fix fragment caching inherited serializers to use distinct per-serializer caches. (@lserman)
 - [#1478](https://github.com/rails-api/active_model_serializers/pull/1478) Cache store will now be correctly set when serializers are
   loaded *before* Rails initializes. (@bf4)
 - [#1570](https://github.com/rails-api/active_model_serializers/pull/1570) Fixed pagination issue with last page size. (@bmorrall)
--- a/lib/active_model_serializers/fragment_cache.rb
+++ b/lib/active_model_serializers/fragment_cache.rb
@@ -15,7 +15,7 @@ module ActiveModelSerializers
       object = serializer.object
 
       # It will split the serializer into two, one that will be cached and one that will not
-      serializers = fragment_serializer(object.class.name)
+      serializers = fragment_serializer
 
       # Get serializable hash from both
       cached_hash     = serialize(object, serializers[:cached])
@@ -79,10 +79,12 @@ module ActiveModelSerializers
     #     User_AdminCachedSerializer
     #     User_AdminNonCachedSerializer
     #
-    def fragment_serializer(name)
-      klass      = serializer.class
-      cached     = "#{to_valid_const_name(name)}CachedSerializer"
-      non_cached = "#{to_valid_const_name(name)}NonCachedSerializer"
+    def fragment_serializer
+      klass                 = serializer.class
+      serializer_class_name = to_valid_const_name(klass.name)
+
+      cached     = "Cached#{serializer_class_name}"
+      non_cached = "NonCached#{serializer_class_name}"
 
       cached_serializer     = get_or_create_serializer(cached)
       non_cached_serializer = get_or_create_serializer(non_cached)
--- a/test/cache_test.rb
+++ b/test/cache_test.rb
@@ -1,6 +1,12 @@
 require 'test_helper'
 require 'tmpdir'
 require 'tempfile'
+
+InheritedRoleSerializer = Class.new(RoleSerializer) do
+  cache key: 'inherited_role', only: [:name, :special_attribute]
+  attribute :special_attribute
+end
+
 module ActiveModelSerializers
   class CacheTest < ActiveSupport::TestCase
     def setup
@@ -150,6 +156,14 @@ module ActiveModelSerializers
       assert_equal({ place: 'Nowhere' }, ActionController::Base.cache_store.fetch(@location.cache_key))
     end
 
+    def test_fragment_cache_with_inheritance
+      inherited = render_object_with_cache(@role, serializer: InheritedRoleSerializer)
+      base = render_object_with_cache(@role)
+
+      assert_includes(inherited.keys, :special_attribute)
+      refute_includes(base.keys, :special_attribute)
+    end
+
     def test_uses_file_digest_in_cache_key
       render_object_with_cache(@blog)
       assert_equal(@blog_serializer.attributes, ActionController::Base.cache_store.fetch(@blog.cache_key_with_digest))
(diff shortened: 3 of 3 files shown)


groyoh:
Thanks for this PR! I've been wanting to roll it out myself for quite some time now. Could you add an entry within the CHANGELOG in the `fixes` section?

test/fixtures/poro.rb
```diff
@@ -139,6 +139,11 @@ def slug
   belongs_to :author
 end
 
+InheritedRoleSerializer = Class.new(RoleSerializer) do
```
groyoh: Could you move this class to `test/cache_test.rb` to avoid bloating this file?

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Features:
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
 
 Fixes:
+- [#1622](https://github.com/rails-api/active_model_serializers/pull/1622) Fix inherited serializers when using fragment cache
```
bf4: `Fix fragment caching inheriting serializers to use distinct per-serializer caches. (@Iserman)`

lib/active_model_serializers/fragment_cache.rb
```diff
@@ -79,10 +79,10 @@ def add_attributes_to_serializer(serializer, attributes, attributes_keys)
     #     User_AdminCachedSerializer
     #     User_AdminNonCachedSerializer
     #
-    def fragment_serializer(name)
+    def fragment_serializer
       klass      = serializer.class
-      cached     = "#{to_valid_const_name(name)}CachedSerializer"
-      non_cached = "#{to_valid_const_name(name)}NonCachedSerializer"
+      cached     = "Cached#{to_valid_const_name(klass.name)}"
```
bf4: while in here, would you mind rewriting as

```ruby
serializer_class_name = to_valid_const_name(klass.name)
cached = "Cached#{serializer_class_name}"
non_cached = "NonCached#{serializer_class_name}"
```

bf4:
@lserman awesome work! Please address comments, squash down to one commit, and this is ready to merge! :100:

lserman:
@bf4 thank you, hope I did that squash right... never done this before.

test/cache_test.rb
```diff
@@ -1,6 +1,12 @@
 require 'test_helper'
 require 'tmpdir'
 require 'tempfile'
+
+InheritedRoleSerializer = Class.new(RoleSerializer) do
+  cache key: 'inherited_role', only: [:name, :special_attribute]
+  attribute :special_attribute
+end
```
bf4: Better to include this in the `class CacheTest` so it's not in the global scope, but not a blocker

test/cache_test.rb
```diff
@@ -150,6 +156,14 @@ def test_fragment_fetch_with_virtual_associations
       assert_equal({ place: 'Nowhere' }, ActionController::Base.cache_store.fetch(@location.cache_key))
     end
 
+    def test_fragment_cache_with_inheritance
+      inherited = render_object_with_cache(@role, serializer: InheritedRoleSerializer)
+      base = render_object_with_cache(@role)
+
+      assert_includes(inherited.keys, :special_attribute)
```
bf4: minitest asserts are in the reverse order: expected, actual

bf4:
@lserman :100: you're doing a fantastic job!

test/cache_test.rb
lserman: Not sure what you mean here, should I not use assert_includes? Or am I misunderstanding this method? http://apidock.com/ruby/v1_9_3_392/MiniTest/Assertions/assert_includes

bf4: You're right. That's annoying asser_equal is reverse order

lib/active_model_serializers/fragment_cache.rb
```diff
@@ -79,10 +79,12 @@ def add_attributes_to_serializer(serializer, attributes, attributes_keys)
     #     User_AdminCachedSerializer
```
groyoh: The doc should also be changed accordingly :wink:

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Features:
 - [#1340](https://github.com/rails-api/active_model_serializers/pull/1340) Add support for resource-level meta. (@beauby)
 
 Fixes:
+- [#1622](https://github.com/rails-api/active_model_serializers/pull/1622) Fix fragment caching inherited serializers to use distinct per-serializer caches. (@lserman)
```
groyoh: I'm not too happy with this sentence though. Hard to understand what was the actual change. I'd prefer something closer to the PR title e.g. `Fragment cache now fragments based on the serializer name instead of the model name`.

@bf4 do you agree or am I being too picky here? :smile:

bf4: @groyoh No, not being too picky. The problem with not mentioning inheritance is that without it there fix is unclear.  btw, I've also struggled with what the description should be https://github.com/rails-api/active_model_serializers/pull/1622/files/038f3fc91d103aee0a14dc725b4856bf86a0a0ad#r57530684

groyoh: Yeah, but that's not the only thing it solve actually. If you define two serializers with the same model, you would get the same issue. That's actually what prevented me to properly refactor fragment cache in #1565.

bf4: Good point.  So, `Records are now fragment cached by serializer name,  rather that model name, allowing multiple fragment caches for a model.`

bf4:
Merged! :rainbow: @lserman

PR #1623, opened 2016-03-27
[WIP] Move ActiveModel::Serializer::* to ActiveModelSerializers namespace

groyoh:
#### Purpose

Per https://github.com/rails-api/active_model_serializers/blob/master/docs/rfcs/0000-namespace.md
#### Changes

I moved the following to `ActiveModelSerializers::Resource` namespace:
- `ActiveModel::Serializer::Links`
#### Additional informations

I've updated the PR to only move one file at once based on @bf4 comments. I will either move the remaining file with new commits once the current commit is approved or create a new PR for them.

The original PR description is available below.
# Original PR description
#### Purpose

Per https://github.com/rails-api/active_model_serializers/blob/master/docs/rfcs/0000-namespace.md
#### Changes

The changes seems to be huge but you can easily go through each commit and check them one by one. No deprecation warning needed to be done. I mainly moved the following to `ActiveModelSerializers` namespace:
- `ActiveModel::Serializer::Links`
- `ActiveModel::Serializer::Meta`
- `ActiveModel::Serializer::Type`
- `ActiveModel::Serializer::Field`
- `ActiveModel::Serializer::Attributes`
- `ActiveModel::Serializer::Attribute`
- `ActiveModel::Serializer::Associations`
- `ActiveModel::Serializer::Association`
- `ActiveModel::Serializer::Reflection`
- `ActiveModel::Serializer::HasOneReflection`
- `ActiveModel::Serializer::HasManyReflection`
- `ActiveModel::Serializer::BelongsToReflection`
- `ActiveModel::Serializer::VERSION`
- `ActiveModel::Serializer::Configuration`

`ActiveModel::Serializer::VERSION` now references `ActiveModelSerializers::VERSION`.

Only two tests needed to be changed because they then referenced the wrong class.
### Caveats

From the RFC I'm not sure if they should actually go under `ActiveModelSerializers::Serializer` instead of `ActiveModelSerializers`.

There are still a few files that I didn't move cause I feel it would be more appropriate to handle it their own PR due to the changes needed to move them (e.g. `FieldsSet`)

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -10,7 +10,7 @@ require 'active_model/serializer/caching'
 require 'active_model/serializer/configuration'
 require 'active_model/serializer/fieldset'
 require 'active_model/serializer/lint'
-require 'active_model/serializer/links'
+require 'active_model_serializers/resource/links'
 require 'active_model/serializer/meta'
 require 'active_model/serializer/type'
 
@@ -25,7 +25,7 @@ module ActiveModel
     include Associations
     include Attributes
     include Caching
-    include Links
+    include ActiveModelSerializers::Resource::Links
     include Meta
     include Type
 
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -16,6 +16,7 @@ module ActiveModelSerializers
   autoload :Adapter
   autoload :JsonPointer
   autoload :Deprecate
+  autoload :Resource
 
   class << self; attr_accessor :logger; end
   self.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
--- a/lib/active_model_serializers/resource.rb
+++ b/lib/active_model_serializers/resource.rb
@@ -1,0 +1,4 @@
+module ActiveModelSerializers
+  module Resource
+  end
+end
--- a/lib/active_model_serializers/resource/links.rb
+++ b/lib/active_model_serializers/resource/links.rb
@@ -1,5 +1,5 @@
-module ActiveModel
-  class Serializer
+module ActiveModelSerializers
+  module Resource
     module Links
       extend ActiveSupport::Concern
 


groyoh:
Rebased

groyoh:
> From the RFC I'm not sure if they should actually go under `ActiveModelSerializers::Serializer` instead of `ActiveModelSerializers`.

@bf4 @NullVoxPopuli @remear what do you think?

NullVoxPopuli:
LGTM, I'm surprised so few test files were changed. :-)

Just needs rebase.

groyoh:
Rebased.

lib/active_model/serializer.rb
```diff
@@ -4,15 +4,15 @@
 require 'active_model/serializer/error_serializer'
```
bf4: you can move this also and rename to `ActiveModelSerializers::Resource` per https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185 in the RFC in a followup PR maybe?

lib/active_model/serializer/version.rb
```diff
@@ -1,5 +1,7 @@
+require 'active_model_serializers/version'
+
 module ActiveModel
   class Serializer
-    VERSION = '0.10.0.rc4'.freeze
+    VERSION = ActiveModelSerializers::VERSION
```
bf4: Can just be `ActiveModel::Serializer::VERSION =  ActiveModelSerializers::VERSION`. autoload etc takes care of the rest

bf4: n/m re autoload, we want someone to be able to require just this file

lib/active_model_serializers.rb
```diff
@@ -15,6 +15,14 @@ module ActiveModelSerializers
   autoload :Adapter
   autoload :JsonPointer
   autoload :Deprecate
+  autoload :Attribute
```
bf4: these shouldn't be here. they're serializer/resource concerns. They should be in a serializer/resource namespace

bf4:
@groyoh Maybe when making these 'move everything' prs you can start by just moving one thing for review, so that's there's less of an upkeep burden on you. then, when we're ready to review, we can move all the other files over and merge at once

lib/active_model_serializers.rb
groyoh: Oh didn't see https://github.com/rails-api/active_model_serializers/pull/1301/files#r42963185. It was only a note in the RFC:

> The name can be discussed in a future pull request. For example, we can rename this to Resource following this idea more info about naming in the next section

groyoh:
@bf4 Thanks for the suggestion, I did not realize it would be easier. Should I step back, branch off this branch, rebase this branch to the first commit then cherry-pick each commit incrementally as you suggested?

lib/active_model/serializer.rb
groyoh: Yes, sure. The name change was only defined as a note in the RFC, should I update the RFC too?

groyoh:
@bf4 I followed your advide. I reset the branch to the first commit and updated the commit to namespace with `Resource` as you suggested. Could you please check if it looks better to you now? After that, we can either merge it directly and do the rest in follow up PR or update the PR bit by bit.

bf4:
Specifically for this PR where we may rename Serializer, I wanted to discuss calling it a Resource or Entity. See https://github.com/json-api/json-api/issues/966 for background.  Also, this may be a good opportunity to discuss working with @dgeb and @lgebhardt on integrating https://github.com/cerebris/railsapi-resources/blob/master/lib/railsapi/resource.rb or something like that

remear:
Without knowing much more about it, I wonder if integrating railsapi-resources would be better-suited for after 0.10 final. I'm not sure which version would be ideal for a rename of Serializer.

NullVoxPopuli:
Anything going on with this? / does it relate to:

```
NOTE: ActiveModel::Serializer::Adapter::JsonApi.new is deprecated; use ActiveModelSerializers::Adapter::JsonApi.new instead                 |  ETA: 00:01:06 
ActiveModel::Serializer::Adapter::JsonApi.new called from /home/lprestonsegoiii/.rvm/gems/ruby-2.3.0@aeonvera/bundler/gems/active_model_serializers-019ce559614e/lib/active_model_serializers/adapter.rb:21.
```

NullVoxPopuli:
For what this PR is, I'm good with merging it, as we need to eventually move everything over anyway.

groyoh:
@NullVoxPopuli yes, the goal is to move everything under one single `ActiveModelSerializers` namespace. Nothing needs to be deprecated here as none of these classes are part of the public API. What I started by doing was moving everything at once but it turned out that it was not the best way to do it so I'll move bits by bits instead. Although, @bf4 had some concerns about moving the `Serializer` mixins/concerns into a `Resource` namespace so that we would have:
* `ActiveModel::Serializer` -> `ActiveModelSerializers::Resource`
* `ActiveModel::Serializer::Attribute` -> `ActiveModelSerializers::Resource::Attribute`

So the discussion is kind of on hold to decide if we should use `Resource` instead of `Serializer`. Personally, I like the name change. It fits well with JSON API and avoid redundancy of `ActiveModelSerializers::Serializer`.

NullVoxPopuli:
I like the name change as well. Do we vote now? lol

groyoh:
@bf4 @NullVoxPopuli @remear @beauby any further opinion on this? I'd say we should go on with it if nobody is against it.

NullVoxPopuli:
I'm good with it. Tests pass. no conflicts. Thanks for reminding me about it!

bf4:
I kind of want to hold off any major renamings for now as they slow down other in-progress prs

groyoh:
@bf4 I'm fine with that.

Issue #1243, opened 2015-10-06
Add docs for `include` in Json adapter.


beauby:
The `include` option with Json adapter works the same as with the JsonApi adapter, but it's not written anywhere in the docs.

vasilakisfil:
That's interesting! Do you know if fields or something relevant is also working with JSON adapter? If not is there any interest to send a pull request for that ? Cause I really need it and I am going to implement it anyway..

beauby:
@vasilakisfil Currently you can use the `fields` option with Json/Attributes adapters, but you can only filter the primary resources (`render json: @posts, fields: 'id, title, body'`).

beauby:
I am not sure extending the Json adapter to fit the broader JSON API style `fields` option (where you specify the attributes for each type) is a good thing. The goal is to keep Json/Attributes adapters very simple.

vasilakisfil:
given that's an adapter I don't think it's an issue. Plus there should be
an adapter that opts for compatibility with  0.9.x and  0.8.x (supporting
only, except etc options). Otherwise the way to go is create an external
plugin adapter just like ember? is there support for that?

beauby:
Oh yeah, AMS is deisgned in such a way that you can create your own adapter and/or extend an existing one. I agree the migration from 0.8/0.9 should be made as easy as possible, I'm just not sure what's the best way to make that happen.

NullVoxPopuli:
maybe for the initial 0.10 release, we could copy the old adapters? maybe name them `json_08`, `json_09` etc

NullVoxPopuli:
and then in the next release after 0.10, remove them?

beauby:
First we have to make sure it is actually possible to build such legacy adapters.

phcoliveira:
@beauby, is there any reason to use the option ```include``` when rendering an object? My point is, if that option must be included when of rendering - ```render json: @posts, include: 'comments'``` e.g. -, I expect the serializer to exclude the comments relationship when used without the ```include``` option.

I have a use case where a user can have many sessions, in order to be able to sign them all out in case something bad happens. I have something like this:

``` ruby
class SessionsController
# ...
render json: @session, include: 'user'
# ...
end

class SessionSerializer
# ...

# I am not asking for the "UserSerializer" to include the "sessions" relationship.
# Nevertheless, it does. The "UserSerializer", for its own logical reasons, also
# declares its own relationship with "sessions" and other models.
[...]

PR #1628, opened 2016-03-29
Optional fields in associations (priority on serializer)

vasilakisfil:
#### Purpose
The purpose of this PR is to have a graphql-like support in Attributes/JSON adapters. With this PR, you can filter fields on nested resources on those adapters.

Currently, there is a minor bug in the fields option in master branch. If you have:

```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id],
```

This will apply the fields to all associations too. We don't want that. What we want is:

If there is a fields param, apply it to the main resource. If there are associations, serialize the associations without removing any attributes. For instance:

```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id]
```

Then this should apply the `fields: [:id]` only in the VideoSerializer and not in any association.

If there are associations in the fields array (just like in StrongParams) and these also exist in the main resource, apply the associations fields there. For instance:

```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id, user: [:name]]
```

Then this should apply the `fields: [:id]` only in the VideoSerializer and if there is a user association, apply `fields: [:name]` in the UserSerializer.

Also, a config option was also added (attributes_adapter_include_default) to denote the number of nested level is supported by the adapter. Default value is `'*.*'` which means 2 levels but this can change by the user.

**Options on the serializer level have higher priority.** So whatever you passed in the serializer constructor regarding associations, serializer's associations will render only id:

```ruby
  class VideoSerializer
    attributes :id, :video_id

    belongs_to :user, serializer: UserSerializer, fields: [:id], includes: []
  end
```

#### Changes
A config option was also added (attributes_adapter_include_default) to denote the number of nested level is supported by the adapter. Default value is ('*.*') which means 2 levels but this can change by the user.

#### Caveats
Please note that you can pass an array to fields. If you want to filter a nested resource, then an element of the array must be a hash with key the name of the association and value an array with the fields that you would like to pass.

So for instance, to have only id in the primary resource and in the likes association you should do:

```ruby
render json: @videos.limit(1), each_serializer: VideoSerializer,  fields: [:id, likes: [:id]]
```

**We need a helper function to convert url params to this structure**

Unfortunately, url params do not support this kind of structure due to limitation in the rack parse_nested_query method (https://github.com/rack/rack/issues/524).
Does it mean that this PR is useless ? No at all. Because you can manually exclude fields in the association level from the controller as well as in the serializer level.

And when rack issue is fixed (or someone monkey patches it) we have support for a graphql-like API.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/pull/1287
https://github.com/rails-api/active_model_serializers/pull/1285
https://github.com/rails-api/active_model_serializers/pull/1284
https://github.com/rails-api/active_model_serializers/issues/1243
https://github.com/rails-api/active_model_serializers/pull/1141

Also add default nested level in config for attributes/json adapters

wip

wip

wip

wip

Add default nested level in config for attributes/json adapters

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -103,7 +103,11 @@ module ActiveModel
           begin
             serializer = serializer_class.new(
               association_value,
-              serializer_options(subject, parent_serializer_options, reflection_options)
+              serializer_options(
+                subject,
+                parent_serializer_options,
+                reflection_options.merge(options)
+              )
             )
           rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
             reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -3,7 +3,9 @@ module ActiveModelSerializers
     class Attributes < Base
       def initialize(serializer, options = {})
         super
-        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
+        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(
+          options[:include] || ActiveModel::Serializer.config.attributes_adapter_include_default
+        )
         @cached_attributes = options[:cache_attributes] || {}
       end
 
@@ -60,8 +62,11 @@ module ActiveModelSerializers
 
       def resource_relationships(options)
         relationships = {}
+
         serializer.associations(@include_tree).each do |association|
-          relationships[association.key] = relationship_value_for(association, options)
+          relationships[association.key] = relationship_value_for(
+            association, options
+          )
         end
 
         relationships
@@ -71,8 +76,21 @@ module ActiveModelSerializers
         return association.options[:virtual_value] if association.options[:virtual_value]
         return unless association.serializer && association.serializer.object
 
+        if options[:fields].is_a? Array
+          fields = options[:fields].select do |k| #I would prefer using {...} here but RuboCop fails :/
+            k.is_a? Hash
+          end.try(:first).try(:[], association.name)
+        end
+
         opts = instance_options.merge(include: @include_tree[association.key])
-        Attributes.new(association.serializer, opts).serializable_hash(options)
+        options = options.merge(fields: fields)
+
+        Attributes.new(
+          association.serializer,
+          opts.merge(association.options)
+        ).serializable_hash(
+          options.merge(association.options)
+        )
       end
 
       # no-op: Attributes adapter does not include meta data, because it does not support root.
--- a/test/adapter/json/collection_test.rb
+++ b/test/adapter/json/collection_test.rb
@@ -84,6 +84,72 @@ module ActiveModelSerializers
 
           assert_equal(expected, actual)
         end
+
+        def test_fields_option
+          serializer = ActiveModel::Serializer::CollectionSerializer.new([@first_post, @second_post])
+          adapter = ActiveModelSerializers::Adapter::Json.new(serializer)
+          actual = adapter.serializable_hash(fields: [:id])
+
+          expected = { posts: [{
+            id: 1,
+            comments: [],
+            author: {
+              id: 1,
+              name: 'Steve K.'
+            },
+            blog: {
+              id: 999,
+              name: 'Custom blog'
+            }
+          }, {
+            id: 2,
+            comments: [],
+            author: {
+              id: 1,
+              name: 'Steve K.'
+            },
+            blog: {
+              id: 999,
+              name: 'Custom blog'
+            }
+          }] }
+
+          assert_equal(expected, actual)
+        end
+
+        def test_fields_with_no_associations_include_option
+          serializer = ActiveModel::Serializer::CollectionSerializer.new([@first_post, @second_post])
+          adapter = ActiveModelSerializers::Adapter::Json.new(serializer, include: [])
+          actual = adapter.serializable_hash(fields: [:id])
+
+          expected = { posts: [{
+            id: 1
+          }, {
+            id: 2
+          }] }
+
+          assert_equal(expected, actual)
+        end
+
+        def test_fields_with_associations_fields_option
+          serializer = ActiveModel::Serializer::CollectionSerializer.new([@first_post, @second_post])
+          adapter = ActiveModelSerializers::Adapter::Json.new(serializer, include: [:author])
+          actual = adapter.serializable_hash(fields: [:id, author: [:name]])
+
+          expected = { posts: [{
+            id: 1,
+            author: {
+              name: 'Steve K.'
+            }
+          }, {
+            id: 2,
+            author: {
+              name: 'Steve K.'
+            }
+          }] }
+
+          assert_equal(expected, actual)
+        end
       end
     end
   end
--- a/lib/active_model/serializer/configuration.rb
+++ b/lib/active_model/serializer/configuration.rb
@@ -20,6 +20,7 @@ module ActiveModel
         end
 
         config.adapter = :attributes
+        config.attributes_adapter_include_default = '*'
         config.jsonapi_resource_type = :plural
         config.jsonapi_version = '1.0'
         config.jsonapi_toplevel_meta = {}


bf4:
For reference jsonapi fields specify the object https://github.com/rails-api/active_model_serializers/pull/1285#issuecomment-195880737

test/adapter/json/collection_test.rb
```diff
@@ -84,6 +84,72 @@ def test_include_option
 
           assert_equal(expected, actual)
         end
+
+        def test_fields_option
+          serializer = ActiveModel::Serializer::CollectionSerializer.new([@first_post, @second_post])
+          adapter = ActiveModelSerializers::Adapter::Json.new(serializer)
+          actual = adapter.serializable_hash(fields: [:id])
```
bf4: these options should be passed into the adapter.  most conventional way to do it is

````ruby
actual = serializable([@first_post, @second_post], adapter: :json, fields: [:id]).serializable_hash
```

see https://github.com/rails-api/active_model_serializers/issues/1643

lib/active_model/serializer/reflection.rb
```diff
@@ -103,7 +103,11 @@ def build_association(subject, parent_serializer_options)
           begin
             serializer = serializer_class.new(
               association_value,
-              serializer_options(subject, parent_serializer_options, reflection_options)
+              serializer_options(
+                subject,
+                parent_serializer_options,
+                reflection_options.merge(options)
```
bf4: I'm guessing we should rename this variable to `field_options` so it's clear where it's coming from. It took me a moment.

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -3,7 +3,9 @@ module Adapter
     class Attributes < Base
       def initialize(serializer, options = {})
         super
-        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
+        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(
+          options[:include] || ActiveModel::Serializer.config.attributes_adapter_include_default
+        )
```
bf4: this is a separate feature, no?

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -60,8 +62,11 @@ def serializable_hash_for_single_resource(options)
 
       def resource_relationships(options)
         relationships = {}
+
         serializer.associations(@include_tree).each do |association|
-          relationships[association.key] = relationship_value_for(association, options)
+          relationships[association.key] = relationship_value_for(
+            association, options
+          )
```
bf4: I know this is just cosmetic, am curious what you like about it

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -71,8 +76,21 @@ def relationship_value_for(association, options)
         return association.options[:virtual_value] if association.options[:virtual_value]
         return unless association.serializer && association.serializer.object
 
+        if options[:fields].is_a? Array
+          fields = options[:fields].select do |k| #I would prefer using {...} here but RuboCop fails :/
+            k.is_a? Hash
+          end.try(:first).try(:[], association.name)
```
bf4: > #I would prefer using {...} here but RuboCop fails :/

Yeah, we're using Rails-style block semantics :)

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -71,8 +76,21 @@ def relationship_value_for(association, options)
...
+        end
+
         opts = instance_options.merge(include: @include_tree[association.key])
-        Attributes.new(association.serializer, opts).serializable_hash(options)
+        options = options.merge(fields: fields)
+
+        Attributes.new(
+          association.serializer,
+          opts.merge(association.options)
+        ).serializable_hash(
```
bf4: we need to rename internal apis like `serializable_hash` here to `as_json`. it's just confusing when it's not part of the public serialization api. e.g. https://github.com/rails-api/active_model_serializers/issues/1643

bf4: (just thinking aloud. not actionable here)

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -71,8 +76,21 @@ def relationship_value_for(association, options)
         return association.options[:virtual_value] if association.options[:virtual_value]
         return unless association.serializer && association.serializer.object
 
+        if options[:fields].is_a? Array
```
bf4: this snippet of code is somewhat problematic.. I believe what it says is

1. If options fields is a collection of class Array (bug: should checking if `respond_to?(:to_ary)`)
1. collect all elements of class Hash  (or responds to keys is probably fine)
1. If the collection responds to first, call first (bug: it will always respond to first. and if we only want the first item in the collection, we should just use detect or better lazy and take)
1. If the previous step's result responds to `:[]`, i.e. the collection had a hash in it, get the value keyed by the association name

I think this could be rewritten as

```ruby
fields = Array(options[:fields]).lazy.select {|elem| elem.respond_to?(:keys) }.take(1).map {|elem| association.name }.force.compact.first
```

Would be nice for this to be tested separately or have a descriptive name so it's easier to follow what it's doing, as well

:)

bf4: I think you might also be able to use the `ActiveModel::Serializer::Fieldset.new(options[:fields]).fields_for(association.name)` like the JSON API adapter does

test/adapter/json/collection_test.rb
bf4: Also, since `fields` is part of `JSON API`, we need to make sure fields behave properly there as well. (and the adapter already handles fields)

bf4:
@vasilakisfil The PR has good work in it. Do you plan on coming back to discuss? In particular, there are adapter concerns:

(I'm a little tired, so pardon if this looks jumbled in the morning :zzz: )

I believe you've described the PR as 

| options | result |
|------------|---------------| 
| `fields: [:id]` | primary resource is restricted to `id` field. no associations affected 
| `fields: [:id, user: [:name]]` | primary resource is restricted to `id` field, `user` association is restricted to `name` field. no other associations affected

I also thing this behavior is already true for the JSON API adapter, except that the JSON API only includes resources when specified

and

> Default value is '*.*' which means 2 levels 

This could only apply to the JSON and Attributes adapters. JSON API is by definitions no associations unless included, as if it were `includes: []`

[Regarding `fields` and `include`](https://github.com/rails-api/active_model_serializers/pull/1628#issuecomment-202909524)

Since we're putting our effort behind supporting `JSON API` it's convenient to adopt its nomenclature where appropriate.  In this case, JSON API uses `include` to specify which associations are returned, if and `fields` to restrict the primary resource and any included associations. e.g. `'GET /articles?include=author&fields[articles]=title,body,author&fields[people]=name'` would get the `articles` with fields `title,body,author` and associated `authors` with fields `name`. No other associations would be returned. 

One thing to bring up is to point out that [`fields` is ultimately an adapter concern](https://github.com/rails-api/active_model_serializers/blob/96c5516d217a853bb471bcf215aaebf34414d442/lib/active_model_serializers/serializable_resource.rb#L5) even if the adapter uses it to get attributes and associations from a serializer.

bf4:
Also, how would you say this pr relates to

- Add a default_include_tree config variable to ActiveModel::Serializer #1426
- Pass fields down from constructor #1287
- Constructor does not pass down the options to serializer #1286
- Add regression test for fields in JSON/Attributes adapter #1285

vasilakisfil:
Hello @bf4,

I would like to give some love in those adapters. So I will try to help as much as I can.

The code is in sync with JSONAPI: it already checks includes first and then fields, meaning that if a resource is not included, fields definition on that resource are kind of pointless. The reason fields work without includes now is due to default include level (`*.*`). But if you remove that fields will not work. I will add tests in the weekend.

#1285 points that there is a bug in the fields (this PR fixes it)
#1287 (and #1286 points (in discussion) that fields are looked from the serializer options rather than the adapter options

About #1426 and default include level, we could split it in another feature I agree and I would prefer that to implemented/merged first before merging current PR.

I would work again this weekend to fix some things you mentioned but pretty much I don't have much work to do unless you would like to tell me more what you would like to see from that PR (the info you gave me were interesting but a bit vague as to which direction I should continue working :) )

One question: could option[:fields] be a string ?

bf4:
> the info you gave me were interesting but a bit vague as to which direction

yeah, I know.  Basically, I want to me confident that this PR isn't 

- duplicating existing behavior in ams 
- breaking existing behavior in ams
- changing nesting behavior in unexpected ways or otherwise breaking changes

And sometimes I just don't have the brainpower to figure it out with out OP :)

bf4:
If we could split this PR into two, the nesting and everything else, then close any issues this one supersedes, that would be grewat

remear:
Any progress on this? Was this ever split into multiple PRs?

lib/active_model_serializers/adapter/attributes.rb
beauby: How about the following:
```ruby
fields = options[:fields]
  .select { |k| k.is_a? Hash }
  .try(:[], association.name)
```

remear:
Closing due to inactivity

Issue #1612, opened 2016-03-22
jsonapi_parse should translate key format back again


maletor:
Key formats, introduced in #1248, should play into deserializing params as well.

Typically an application will store keys in underscore format. AMS should then convert them to dashed for JSON API. But, upon jsonapi_parse, they should be converted back to underscores. Currently, they are passed as dashed.

pelted:
I take it this was new in RC4? My JSON API just broke all over the place with a change in key format from underscores. Not seeing a way to configure key format in the docs and these params passed back are also not deserializing.

maletor:
Key formats are new from RC4 on master. The key format defaults are desired and correctly implemented. The reverse part, the deserialization, which is in experimental stage, is not yet implemented correctly. The deserializer should know what the attributes have been transformed to since it knows where they came from.

maletor:
I'm interested to know how this would work with the given `only` and `except` keys. I imagine they would work as they do today and would use underscored names if we're going with the general use case.

rthbound:
@pelted here's [the docs](https://github.com/remear/active_model_serializers/blob/c533d1a7fe202a904af178cd22c54cc14383bab0/docs/general/key_transform.md) on how to configure key format. I use `:unaltered` because my javascript clients are satisfied with underscored keys. We can't use the default (`:dashed`) yet because of the issue @snwfog just referenced ^.

I've closed that issue because I think its resolution will be required in order to close this issue.

PR #1629, opened 2016-03-29
Properly deserialize dasherized keys

lawitschka:
#### Purpose

The JSON API adapater dasherizes every key, but the deserializer left the keys
unaltered. Thus, the client had to send underscored keys in the request body in
order for Rails to properly match sent values to model attributes.

The changes make overwriting `keyForAttribute()` and `keyForRelationship()` in
Ember.js apps obselete, because Active Model Serializer properly deserializes
the request.

#### Changes

This commit adds automatic key transformation on deserialization. Per default the
deserializer transforms the keys to underscore, but this behaviour can also be
changed by including `key_transform` in the deserializer options.

#### Caveats

I did not include tests for the `KeyTransform#underscore`, because the tests for
key transform only tested the transformation in controller context.

#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/issues/1612

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/deserialization.rb
+++ b/lib/active_model_serializers/adapter/json_api/deserialization.rb
@@ -155,7 +155,7 @@ module ActiveModelSerializers
 
         # @api private
         def parse_attributes(attributes, options)
-          attributes
+          transform_keys(attributes, options)
             .map { |(k, v)| { field_key(k, options) => v } }
             .reduce({}, :merge)
         end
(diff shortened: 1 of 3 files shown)


lib/active_model_serializers/adapter/json_api/deserialization.rb
```diff
@@ -182,23 +182,29 @@ def parse_relationship(assoc_name, assoc_data, options)
           prefix_key = field_key(assoc_name, options).to_s.singularize
           hash =
             if assoc_data.is_a?(Array)
-              { "#{prefix_key}_ids".to_sym => assoc_data.map { |ri| ri['id'] } }
+              { "#{prefix_key}_ids".to_sym => assoc_data.map { |ri| ri[:id] } }
```
NullVoxPopuli: why this change?

lib/active_model_serializers/key_transform.rb
```diff
@@ -32,6 +32,16 @@ def dashed(hash)
       hash.deep_transform_keys! { |key| key.to_s.dasherize.to_sym }
     end
 
+    # Transforms keys to underscore.
+    # This is the default case for deserialization in the JsonApi adapter.
+    #
+    # @example:
+    #    "some-key" => "some_key",
+    # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L89-L98 ActiveSupport::Inflector.underscore}
```
NullVoxPopuli: thanks for the docs!

test/action_controller/json_api/deserialization_test.rb
```diff
@@ -20,7 +20,10 @@ def test_deserialization
               'id' => 'zorglub',
               'attributes' => {
                 'title' => 'Ember Hamster',
-                'src' => 'http://example.com/images/productivity.png'
+                'src' => 'http://example.com/images/productivity.png',
+                'image-width' => '200',
```
NullVoxPopuli: thanks for adding multiple key-format scenarios to the test

lib/active_model_serializers/adapter/json_api/deserialization.rb
lawitschka: because `KeyTransform#underscore` (and every other method in `KeyTransform`) transforms all keys to symbols, so string key access does not work here

remear: Just FYI, this is likely going to change very soon. I've been altering the transform stuff to be more flexible. Stay tuned for that.

NullVoxPopuli:
I'm good with  this, anybody ele,? I'd like. To getiit merged

remear:
Seems good to me.

Issue #1500, opened 2016-02-06
Explicit `serializer_lookup_enabled` in controller doesn't retain Adapter for entire render 


trek:
There's some wonky interaction in #1353 that was not present in #1295:

I've isolated this to a integration example: https://github.com/trek/ams-nesting-test

To run the test:

```
git clone https://github.com/trek/ams-nesting-test.git
cd ams-nesting-testing
bundle
bundle exec rake db:migrate
bundle exec rspec spec/controllers
```

The serializer is configured to use JSON API and require an explicit `serializer` option in renders.

https://github.com/trek/ams-nesting-test/blob/e2d590f79a68b76299003e4021499c34544040cc/config/initializers/active_model_serializers.rb

That feature is used here:

https://github.com/trek/ams-nesting-test/blob/e2d590f79a68b76299003e4021499c34544040cc/app/controllers/tweets_controller.rb#L4

You would expect, in [this test case](https://github.com/trek/ams-nesting-test/blob/e2d590f79a68b76299003e4021499c34544040cc/spec/controllers/tweets_controller_spec.rb) to see a JSON response of

```
{
  "data": [
    {
      "id": "1",
      "type": "tweets",
      "attributes": {
        "body": "lolz"
      },
      "relationships": {
        "favorites": {
          "data": [
            {
              "id": "1",
              "type": "favorites"
            },
            {
              "id": "2",
              "type": "favorites"
            }
          ]
        }
      }
    }
  ],
  "included": [
    {
      "id": "1",
      "type": "favorites",
      "attributes": {
        "reason": "one reason"
      }
    },
    {
      "id": "2",
      "type": "favorites",
      "attributes": {
        "reason": "another reason"
      }
    }
  ]
}
```

but instead see

```
{
  "data": [
    {
      "id": "1",
      "type": "tweets",
      "attributes": {
        "body": "lolz"
      },
      "relationships": {
        "favorites": {
          "data": [
            {
              "id": 1,
              "tweet_id": 1,
              "reason": "one reason",
              "created_at": "2016-02-06T17:34:57.611Z",
              "updated_at": "2016-02-06T17:34:57.611Z"
            },
            {
              "id": 2,
              "tweet_id": 1,
              "reason": "another reason",
              "created_at": "2016-02-06T17:34:57.615Z",
              "updated_at": "2016-02-06T17:34:57.615Z"
            }
          ]
        }
      }
    }
  ]
}
```

[...]

PR #1636, opened 2016-03-30
[WIP] Allow users to opt-out from the ActionController extensions

bf4:
#### Purpose
- Make it possible to require active_model_serializers without mixing `ActionController::Serialization` into the controller
#### Changes

Add `ActionController::Serialization.enabled = false` to an initializer and it won't be mixed in.
#### Caveats

This doesn't exactly solve #1500 since it doesn't include _any_ AMS code in the controller, and passing `serializer` or `each_serializer` won't work

Perhaps config should come from Rails, rather than be set on the mixin. e.g. 

`Rails.configuration.action_controller.render_json_with_active_model_serializers = false`
#### Related GitHub issues
- https://github.com/rails-api/active_model_serializers/issues/1500
- https://github.com/rails-api/active_model_serializers/pull/592

And many more

cc @trek

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -13,6 +13,11 @@ module ActionController
       end
     end
 
+    class << self
+      attr_accessor :enabled
+    end
+    self.enabled = true
+
     included do
       class_attribute :_serialization_scope
       self._serialization_scope = :current_user
--- a/lib/active_model_serializers/railtie.rb
+++ b/lib/active_model_serializers/railtie.rb
@@ -9,12 +9,6 @@ module ActiveModelSerializers
       ActiveModel::Serializer.serializers_cache.clear
     end
 
-    initializer 'active_model_serializers.action_controller' do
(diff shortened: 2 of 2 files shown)


NullVoxPopuli:
I like the idea of setting the option in the rails config.

lib/action_controller/serialization.rb
```diff
@@ -13,6 +13,11 @@ def serialization_scope(scope)
       end
     end
 
+    class << self
+      attr_accessor :enabled
+    end
+    self.enabled = true
```
bf4: (pardon the awful pseudo code)

I was actually just thinking about having 
- a configurable default 'ams_ify_controller = true` 
- and then in the controller, by default we only mix in a 'shim' module that
  - has config methods, e.g. ams_ify_controller
  - so that by default ActionController::Serialization would be included in action controller
  - but if the default were set to false, then individual controllers would need to opt in by setting `ams_ify_controller true` .    
  - the method would just extend the controller  with ActionController::Serialization when true

PR #1642, opened 2016-03-31
Prefer object.cache_key when available.

### LINKED ISSUE #1344: Issues with #object_cache_key method

bf4:
Continues work in #1346

> This will prevent objects PORO objects that don't have updated_at defined, from throwing an error.

> Not as big a deal now that PORO objects can inherit ActiveModelSerializers::Model, but still necessary if it's not inherited for whatever reason.

Fixes #1344

Cherry-pick of
https://github.com/bf4/active_model_serializers/commit/040a97b9e9005ee6cb55f5a3d9088098407f54c2
which was a squash of
https://github.com/rails-api/active_model_serializers/commits/f89ed71058322fe7dd35d5c8b209856f8e03ad14

### DIFF
--- a/test/cache_test.rb
+++ b/test/cache_test.rb
@@ -4,6 +4,21 @@ require 'tempfile'
 
 module ActiveModelSerializers
   class CacheTest < ActiveSupport::TestCase
+    UncachedAuthor = Class.new(Author) do
+      # To confirm cache_key is set using updated_at and cache_key option passed to cache
+      undef_method :cache_key
+    end
+
+    Article = Class.new(::Model) do
+      # To confirm error is raised when cache_key is not set and cache_key option not passed to cache
+      undef_method :cache_key
+    end
+
+    ArticleSerializer = Class.new(ActiveModel::Serializer) do
+      cache only: [:place], skip_digest: true
+      attributes :title
+    end
+
     InheritedRoleSerializer = Class.new(RoleSerializer) do
       cache key: 'inherited_role', only: [:name, :special_attribute]
       attribute :special_attribute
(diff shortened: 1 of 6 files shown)


bf4:
I know this is broken, but now it's easier to see what it should be.

bf4:
This code was hard to extract and could use a good review.

@kevintyll Does this work for you? Also see #1644

test/cache_test.rb
```diff
@@ -161,13 +193,20 @@ def test_fragment_cache_with_inheritance
       refute_includes(base.keys, :special_attribute)
     end
 
+    def test_uses_adapter_in_cache_key
```
kevintyll: This test is not proving that the adapter is in the cache key.

kevintyll:
:+1:

test/cache_test.rb
bf4: lulz. Thanks! :+1:

lambda2:
Does it generates the same cache key for different serializers on the same model ?

bf4:
@lambda2 if using object.cache_key, then yes, it will. Please open a new issue with some examples of this

PR #1657, opened 2016-04-04
add require for active_support/json to fix #1656

### LINKED ISSUE #1656: Unintended dependency to active_record

I expect to be able to use AMS without active_record. What I'm seeing is that `to_json` won't serialize the serializer adapter instance correctly if I don't explicitly call `as_json` before hand. If I load active_record and establish a database connection a simple call to `to_json` will work as expected

#### Steps to reproduce

Please have a look at this gist
https://gist.github.com/andreaseger/84edad45b690b142c401a170b742aafb

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: latest master (d30aa4c)

Output of `ruby -e "puts RUBY_DESCRIPTION"`: `ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-linux]`

#### Additonal helpful information
Although I originally found this dependency issue with active_record in one of our apps @groyoh noticed that requiring `active_support/json` is enough to fix the behavior of `to_json`

andreaseger:
#### Purpose
fixes the missing `require "active_support/json"` mentioned in #1656 

#### Changes

- remove explicit `require "active_support/json"` from test_helper.
- add `require "active_support/json"` to `lib/active_model_serializers.rb`

#### Caveats

Given all the other rails stuff required in the tests simply removing the active_support json extension from the test won't make any tests fail.

#### Related GitHub issues

#1656

### DIFF
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -2,6 +2,7 @@ require 'active_model'
 require 'active_support'
 require 'active_support/core_ext/object/with_options'
 require 'active_support/core_ext/string/inflections'
+require 'active_support/json'
 module ActiveModelSerializers
   extend ActiveSupport::Autoload
   autoload :Model
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -14,7 +14,6 @@ require 'rails'
 require 'action_controller'
 require 'action_controller/test_case'
 require 'action_controller/railtie'
-require 'active_support/json'
 require 'active_model_serializers'
 require 'fileutils'
 FileUtils.mkdir_p(File.expand_path('../../tmp/cache', __FILE__))


bf4:
Might adding a changelog entry? I'd also love if you could include code snippet here showing a failure fixed by the require.

bf4:
@andreaseger Sorry, should have been more careful in checking the [original issue where you included a gist](https://gist.github.com/andreaseger/84edad45b690b142c401a170b742aafb)


```ruby
require 'bundler/inline'

gemfile true do
  source 'https://rubygems.org'
  gem 'active_model_serializers', github: 'rails-api/active_model_serializers'
  gem 'activerecord', require: false
  gem 'mysql2'
end
print "==================================================\n\n"

class Shoe
  attr_accessor :id, :name, :vendor
  def initialize(args)
    args.each{ |k,v| send("#{k}=",v)}
  end
  alias read_attribute_for_serialization public_send
end

class ShoeSerializer < ActiveModel::Serializer
  type "shoe"

  attribute :id
  attribute :name
  attribute :vendor
end

shoe = Shoe.new(id: 3, name: "samba", vendor: "adidas")

require 'json'
print ActiveModelSerializers::SerializableResource.new(shoe,
                                                       adapter:   :json_api,
                                                      ).to_json

print "\n==================================================\n"
print ActiveModelSerializers::SerializableResource.new(shoe,
                                                       adapter:   :json_api,
                                                      ).as_json.to_json

require 'active_record'
ActiveRecord::Base.establish_connection(adapter: 'mysql2', database: 'test')
print "\n==================================================\n"
print ActiveModelSerializers::SerializableResource.new(shoe,
                                                       adapter:   :json_api,
                                                      ).to_json

__END__
==================================================

[active_model_serializers] Rendered ShoeSerializer with ActiveModelSerializers::Adapter::JsonApi (0.06ms)
"#<ActiveModelSerializers::Adapter::JsonApi:0x0055ad9b412820>"
==================================================
[active_model_serializers] Rendered ShoeSerializer with ActiveModelSerializers::Adapter::JsonApi (3.09ms)
{"data":{"id":"3","type":"shoe","attributes":{"name":"samba","vendor":"adidas"}}}
==================================================
[active_model_serializers] Rendered ShoeSerializer with ActiveModelSerializers::Adapter::JsonApi (0.34ms)
{"data":{"id":"3","type":"shoe","attributes":{"name":"samba","vendor":"adidas"}}}
```

lib/active_model_serializers.rb
```diff
@@ -2,6 +2,7 @@
 require 'active_support'
 require 'active_support/core_ext/object/with_options'
 require 'active_support/core_ext/string/inflections'
+require 'active_support/json'
```
bf4: is https://github.com/rails/rails/blob/v5.0.0.beta3/activesupport/lib/active_support/json.rb

```ruby
require 'active_support/json/decoding'
require 'active_support/json/encoding'
```

bf4:
I wonder if adding `to_json(options)` to the base adapter would have been sufficient

andreaseger:
I'm not familiar what other stuff is triggered by requiring `active_support/json/encoding` but for what it's worth this updated gist works as well: https://gist.github.com/andreaseger/84edad45b690b142c401a170b742aafb

If you think this would be a better way to solve this I can prepare another PR.

Issue #1660, opened 2016-04-04
Methods defined in serializer are not used 


NullVoxPopuli:
#### Expected behavior vs actual behavior

Possibly related to #1658.

I expect the methods defined in the serializer to be used for the attributes.

What is happening is the attribute is invoked on the object, resulting in an error if the method does not exist.

#### Steps to reproduce
*(e.g., detailed walkthrough, runnable script, example application)*

- clone https://github.com/NullVoxPopuli/aeonvera
- clone https://github.com/NullVoxPopuli/aeonvera-ui
- add swing.vhost to your hosts file
- boot up both apps.
- visit swing.vhost:4200 (the ember-app)
- create an account
- open the network log
- login
- notice the call to current user fails, and gives this error message:
![image](https://cloud.githubusercontent.com/assets/199018/14248613/7eceffe2-fa43-11e5-8a64-27bf22c6be46.png)

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 

Output of `ruby -e "puts RUBY_DESCRIPTION"`: 
```
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
```
OS Type & Version: `Ubuntu 15.10 64bit`

Integrated application and version *(e.g., Rails, Grape, etc)*: 

rails: 4.2.6
AMS: RC4 works
AMS: master has this issue

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*

![image](https://cloud.githubusercontent.com/assets/199018/14248685/e9815df8-fa43-11e5-9dab-c23e2726ff81.png)

#### Additonal helpful information
*(e.g., Gemfile.lock, configurations, PR containing a failing test, git bisect results)*

https://github.com/NullVoxPopuli/aeonvera/blob/master/Gemfile.lock

I tried working on a failing test with the AMS code, but something funky is happening, I'll continue to investigate.

NullVoxPopuli:
This commit works: https://github.com/rails-api/active_model_serializers/commit/5be33afbfb77489db9bd8ea99b20d53b17d2ec9c

NullVoxPopuli:
which is weird, because no code changes have occurred after that commit -- just documentation changes.

groyoh:
@NullVoxPopuli maybe related to #1650

NullVoxPopuli:
@groyoh see these comments: https://github.com/rails-api/active_model_serializers/issues/1658#issuecomment-205295445

[...]

Issue #1658, opened 2016-04-04
Inherited methods are nil


pareeohnos:
#### Expected behavior vs actual behavior
I've created an `ApplicationSerializer` which I want to contain the property `success` which is in all child serializers. My understanding is that by declaring the attribute within the superclass, all child classes will inherit this. I am finding that this is partially true when the attribute declared is a method, rather than a property on the underlying resource.

When I render using an instance of the child serializer, the attribute key is present, however the value is always null. If I override the attribute method and simply call `super` within it, I get the value I expect, however this defeats the purpose of inheritance.

#### Steps to reproduce
I have the following structure

```ruby
class ApplicationSerializer < ActiveModel::Serializer
  attributes :status

  def status
    object.try(:errors).blank? && object.try(:error).blank?
  end

end
```

```ruby
class ErrorResponse < ActiveModelSerializers::Model
  attr_accessor :error
end
```

```ruby
class ErrorResponseSerializer < ApplicationSerializer
  attributes :error
end
```

In my controller, I am rendering the response using:

```ruby
render json: ErrorResponse.new('Error message here'), status: :unauthorized
```

I am expecting to get the JSON response of

```json
{
    "success": false,
    "error": "Error message here"
}
```

However, I instead get the response

```json
{
    "success": null,
    "error": "Error message here"
}
```

To remedy this, I can change the definition of the `ErrorResponseSerializer` class to the following:

```ruby
class ErrorResponseSerializer < ApplicationSerializer
  attributes :error

  def success
    super
  end
end
```

And this then works as expected.

#### Environment

ActiveModelSerializers Version: 0.10.0.rc4 d30aa4c 
Output of `ruby -e "puts RUBY_DESCRIPTION"`: ruby 2.1.3p242 (2014-09-19 revision 47630) [x86_64-darwin13.0]
OS Type & Version: Mac OS X 10.11.3
Integrated application and version: Rails 4.2.1

NullVoxPopuli:
@pareeohnos hey, can you try this commit: 5be33afbfb77489db9bd8ea99b20d53b17d2ec9c

pareeohnos:
hmm that's strange @NullVoxPopuli that's worked. Something after this commit must be breaking something though. I have just installed from master before trying this, and that commit appears to have been added to master 3 days ago?

pareeohnos:
[...]

PR #1650, opened 2016-04-01
Fix serialization scope options

bf4:
Fixes https://github.com/rails-api/active_model_serializers/issues/1509

PR #1661, opened 2016-04-04
Fix read_attribute_for_serialization not seeing parent serializer methods

### LINKED ISSUE #1653: Inherited methods are hidden in serializer
When subclassing a serializer, the parent class's instance methods are not populated in the child class's instance method list, so `_serializer_instance_method_defined?` returns false.

This started breaking with commit a065bc28d1739631c6177c2995ba345ebcb0c7c8.

#### Steps to reproduce
```ruby
class ParentSerializer < ActiveModel::Serializer
  attributes :$id

  define_method(:$id) do
    object.id
  end
end
```

```ruby
class ChildSerializer < ParentSerializer
  attributes :name
end
```

Throws the following error:
`undefined method '$id' for <Child:0x00...>`

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*:  7acbb765bdbf0a6ad8d5bde02aa3648b2b4a48ea

Output of `ruby -e "puts RUBY_DESCRIPTION"`: ruby 2.1.5p273 (2014-11-13 revision 48405) [x86_64-darwin14.0]

OS Type & Version: OSX 10.11.4

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails

bf4:
Fixes #1653, #1658, #1660

Was introduced in https://github.com/rails-api/active_model_serializers/pull/1650

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -96,19 +96,6 @@ module ActiveModel
       end
     end
 
-    # rubocop:disable Style/ClassVars
-    def self.method_added(method_name)
-      @@_serializer_instance_methods ||= Hash.new { |h, k| h[k] = Set.new }
-      @@_serializer_instance_methods[self] << method_name
-    end
-
-    def self._serializer_instance_method_defined?(name)
-      @_serializer_instance_methods ||= (ActiveModel::Serializer.public_instance_methods - Object.public_instance_methods).to_set
-      @_serializer_instance_methods.include?(name) ||
-        @@_serializer_instance_methods[self].include?(name)
-    end
-    # rubocop:enable Style/ClassVars
-
     attr_accessor :object, :root, :scope
 
     # `scope_name` is set as :current_user by default in the controller.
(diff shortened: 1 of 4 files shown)


lib/active_model/serializer.rb
```diff
@@ -189,7 +176,7 @@ def json_key
     end
 
     def read_attribute_for_serialization(attr)
-      if self.class._serializer_instance_method_defined?(attr)
+      if respond_to?(attr)
```
NullVoxPopuli: this is waayyy simpler. interesting. Why was it so complicated before?

NullVoxPopuli:
Thanks, I'll give this a go on my repo, and see how it fairs.

NullVoxPopuli:
I can confirm that this works

bf4:
@NullVoxPopuli 

> Thanks, I'll give this a go on my repo, and see how it fairs.

`s/fairs/fares` which I think makes be a sociopath for pointing out per http://www.npr.org/2016/04/02/472716926/panel-round-one

NullVoxPopuli:
ah, yep: http://grammarist.com/usage/fair-fare/ :-)

NullVoxPopuli:
> does pointing other people's errors make you a helpful person or an ass?

I think it's helpful, depending on the tone. Cause it helps reduce my liklihood of being an idiot in the future

bf4:
Added changelog, rebased off of master, and squashed

NullVoxPopuli:
cool, I'll merge once CI passes

PR #1664, opened 2016-04-05
Fix link to key transforms doc in post install message

rmehner:
Stumbled over this typo, old link (https://github.com/rails-api/active_model_serializers/blob/master/docs/general/key_transform.md) returns 404.


NullVoxPopuli:
LGTM, nice catch!

Issue #1162, opened 2015-09-16
Validate JsonApi serializations against http://jsonapi.org/schema


bf4:
- [ ] Validate JsonApi serializations against http://jsonapi.org/schema

Ref:
- https://github.com/json-api/json-api/issues/867 esp see the distinction in `additionalProperties` when consuming vs. producing

beauby:
:+1:

remear:
What's the intent here? Is this that we should add some tests that ensure the serialized document conforms to the JSON API schema? Or is this for validating every serialized document?

bf4:
@remear the first usage would be to validate that we're generating valid json api.  we already have code that can make an assertion against a schema

PR #1668, opened 2016-04-07
Exclude links without any data

sigmike:
#### Purpose

I couldn't find a way to exclude a link conditionally. So I changed the serializer to exclude links without any data. A link without data is certainly wrong, so it's probably safe to exclude it.

#### Changes

A link without data is serialized as nil, and links serialized as nil are excluded from the output.

It's only implemented in the json-api adapter.

#### Caveats

It may not be the best solution to this problem nor the best implementation. Maybe conditionals would be better.

#### Related GitHub issues

None. Maybe the conditional attributes PR: #1403.

### DIFF
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -464,7 +464,8 @@ module ActiveModelSerializers
       #   }.reject! {|_,v| v.nil? }
       def links_for(serializer)
         serializer._links.each_with_object({}) do |(name, value), hash|
-          hash[name] = Link.new(serializer, value).as_json
+          result = Link.new(serializer, value).as_json
+          hash[name] = result if result
         end
       end
 
--- a/lib/active_model_serializers/adapter/json_api/link.rb
+++ b/lib/active_model_serializers/adapter/json_api/link.rb
@@ -71,6 +71,7 @@ module ActiveModelSerializers
           hash[:href] = @href if defined?(@href)
           hash[:meta] = @meta if defined?(@meta)
 
+          return nil if hash.empty?
           hash
         end
 
--- a/test/adapter/json_api/links_test.rb
+++ b/test/adapter/json_api/links_test.rb
@@ -17,6 +17,7 @@ module ActiveModelSerializers
           link :yet_another do
             "http://example.com/resource/#{object.id}"
           end
+          link(:nil) { nil }
         end
 
         def setup


remear:
With which level of links were you experiencing this?

When last i tested this I recall it omitted links with a nil value by default. Are you saying that it does, in fact, serialize a response similar to:

```json
{
  "data": {
    "attributes": {
      "name": "My Cool Post"
    },
    "links": {
       "comments": "/blogs/4134587815891/comments",
       "author": "/authors/32512",
       "top-comment": null
    },
    "type": "blogs"
  }
}
```

sigmike:
It serializes to an empty hash like that:
```json
{
  "data": {
    "attributes": {
      "name": "My Cool Post"
    },
    "links": {
       "comments": "/blogs/4134587815891/comments",
       "author": "/authors/32512",
       "top-comment": {}
    },
    "type": "blogs"
  }
}
```

groyoh:
Related to #1514.

bf4:
@sigmike would you mind sharing how you'd see a conditional link working?  I believe you could do this using the block form if it's an attribute or association link since it's eval'd at runtime, not at load.

bf4:
(I'd prefer not to give 'nil' a meaning)

sigmike:
> how you'd see a conditional link working?

Maybe like on the attributes: `link(:foo, if: :bar?) { ... }`.

But it would be nice to also be able to pass a block to `if:` (I think it's not possible on attributes right now).

> I believe you could do this using the block form if it's an attribute or association link since it's eval'd at runtime, not at load.

I already use the block form to return the link value. That's why just returning nil to disable to link is very convenient.

bf4:
@sigmike 

sorry, what I meant was

```ruby
has_one :user do 
  link user_url if object.user.admin?
end
```

I don't totally get the use case

sigmike:
I'm not sure I get it. My use case is when I use link directly as documented here: https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#link

I'm using it to provide links to images attached to the resource.

remear:
Here's another example of when this happens:

Given the following in a serializer:

```ruby
link(:foo) { object.bunnies }
```

When `object.bunnies` evaluates to `nil`, the response is:

```json
{
	"data": {
		"id": "5",
		"type": "companies",
		"attributes": {
			"created_at": "2016-03-29T19:09:32.754Z",
			"name": "Northern Bank of the South",
			"updated_at": "2016-03-29T19:09:32.754Z"
		},
		"links": {
			"self": "http://localhost:9010/companies/5",
			"foo": {}
		}
	}
}
```

At minimum, `foo` should probably be `nil`, not `{}`. Ideally, `foo` wouldn't be included in `links`.

bf4:
@remear you've convinced me when you changed 'make a link conditional' to 'handle a nil link'. :)

> link(:foo) { object.bunnies }
> #=>         "links": {  "foo": {} }

is a bug

bf4:
but hmm http://jsonapi.org/format/#document-links

> Each member of a links object is a "link". A link MUST be represented as either:

> - a string containing the link's URL.
> - an object ("link object") which can contain the following members:
>   - href: a string containing the link's URL.
>   - meta: a meta object containing non-standard meta-information about the link.

which says nothing about required object attributes

and the linkage docs imply it should be `null` http://jsonapi.org/format/#document-resource-object-linkage

bf4:
this is an unresolved contradiction in the spec
- https://github.com/json-api/json-api/issues/887

and also
- https://github.com/json-api/json-api/issues/913
- https://github.com/json-api/json-api/issues/911
- https://github.com/json-api/json-api/issues/898

remear:
I was just looking for that. Because there's no clear answer, I'm not sure how we should make this behave.

Here's the link schema for reference:

```json
"link": {
  "description": "A link **MUST** be represented as either: a string containing the link's URL or a link object.",
  "oneOf": [
    {
      "description": "A string containing the link's URL.",
      "type": "string",
      "format": "uri"
    },
    {
      "type": "object",
      "required": [
        "href"
      ],
      "properties": {
        "href": {
          "description": "A string containing the link's URL.",
          "type": "string",
          "format": "uri"
        },
        "meta": {
          "$ref": "#/definitions/meta"
        }
      }
    }
  ]
}
```

Reading this leads me to believe link must either be a string in `uri` format, or an object containing an `href` in `uri` format. I don't see `null` as an option.

bf4:
@remear https://github.com/rails-api/active_model_serializers/pull/1668#issuecomment-209513707 the schema isn't authoritative :)

remear:
My concern is that it might become authoritative depending on the implementation of #1162.

NullVoxPopuli:
has there been a conclusion on this PR?

remear:
@sigmike Sorry for the delay. I'd like to get this merged in. Would you please rebase this against master, add an entry to the CHANGELOG for this under `Features`, and squash back down to a single commit.

bf4:
Merged on command line and added change log

PR #1675, opened 2016-04-11
AMS 0.8: Fix memory leak with :scope_name

Botje:
#### Purpose
Fix a memory leak when :scope_name is used.

#### Changes
Make `ActiveModel::Serializer#initialize` define a method on the serializer's singleton class, not the serializer class. 

#### Caveats
`singleton_class` is not available before Ruby 1.8, I think.

Supplying :scope_name causes `ActiveModel::Serializer#initialize` to
define a method on the class, which retains a reference to the
Serializer instance and any options passed (such as the scope object).

Given the following test script:
```ruby
require 'json'
require 'objspace'

class Canary
  def initialize
    @created_at = Time.now
    @contents = "0123456789" * 1_000_000
  end

  def inspect
    "Canary"
  end
end

class FooSerializer < ActiveModel::Serializer; end

def serialize_something
  ser = FooSerializer.new({id: 99}, scope_name: :current_canary, scope: Canary.new)
  ser.as_json
end

serialize_something
GC.start
sleep 1

puts "Method defined on class: ", FooSerializer.method_defined?(:current_canary)
if p = ObjectSpace.reachable_objects_from(FooSerializer).find{ |pp| Proc === pp }
  puts "Options captured: ", p.binding.local_variable_get(:options)
end
```

Output before applying this patch:
```
Method defined on class: 
true
Options captured: 
{:scope_name=>:current_canary, :scope=>Canary, :hash=>{:foo=>{}}, :unique_values=>{}}
```
Note that the Proc object captures its environment, which includes the `options` hash.

After:
```
Method defined on class: 
false
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,3 +31,8 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: (@windows_platforms + [:jruby])
+
+# JRuby versions before 9.x report their version as "1.9.x" or lower, so lock these to an older version of mime-types
+if defined?(JRUBY_VERSION) and Gem::ruby_version < Gem::Version.new("2.0.0")
+  gem 'mime-types', '< 3'
+end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -294,9 +294,7 @@ module ActiveModel
 
       scope_name = @options[:scope_name]
       if scope_name && !respond_to?(scope_name)
-        self.class.class_eval do
-          define_method scope_name, lambda { scope }
-        end
+        self.singleton_class.send :alias_method, scope_name, :scope
       end
     end
 


lib/active_model/serializer.rb
```diff
@@ -294,9 +294,7 @@ def initialize(object, options={})
 
       scope_name = @options[:scope_name]
       if scope_name && !respond_to?(scope_name)
-        self.class.class_eval do
-          define_method scope_name, lambda { scope }
-        end
+        self.singleton_class.send :alias_method, scope_name, :scope
       end
```
bf4: good call.  In 0.10 I did  `define_singleton_method scope_name, lambda { scope }`, but I like alias_method

bf4:
@Botje are you up for fixing the mime-types dep for jruby19 on Travis?

Botje:
mime-types 3.0+ require ruby 2.0, while the version of JRuby (1.7.19) used in Travis claims to be 1.9.3.
One option would be to pass `JRUBY_OPTS=--2.0` to the travis runners, but that is a non-standard configuration.
The second option is just to lock mime-types to be < 3 for JRuby 1.7x, which I just pushed.

Gemfile
```diff
@@ -31,3 +31,8 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: (@windows_platforms + [:jruby])
+
+# JRuby versions before 9.x report their version as "1.9.x" or lower, so lock these to an older version of mime-types
+if defined?(JRUBY_VERSION) and Gem::ruby_version < Gem::Version.new("2.0.0")
+  gem 'mime-types', '< 3'
```
bf4: could be written, I think, as, but I'm ok with this since 0-8 is being phased out

`gem 'mime-types', ['~> 2.6', < 3'], platform: :jruby if RUBY_VERSION < '2.0'`

ebeigarts:
Can you release this as 0.8.4? We have spent couple of hours to find out about this, our puma workers were using 4GB+ ram.

bf4:
@ebeigarts Are you up to helping update the changelog? I can release once that's done.

ebeigarts:
@bf4 sure, will do that later today.

ebeigarts:
@bf4 I have updated changelog in https://github.com/rails-api/active_model_serializers/pull/2194

PR #1677, opened 2016-04-11
Assert Schema

bf4:
This works, but I'd like some discussionon the interface change, and then I'll
add some tests.

- [x] `assert_request_schema`
  - [ ] Handle missing `type` in create params ([an exception allowed in the spec](http://jsonapi.org/format/#document-resource-objects))
- [x] `assert_schema_schema`
- [ ] validate when no `data` is present, [but just `meta`](http://jsonapi.org/format/#document-top-level)

### DIFF
--- a/lib/active_model_serializers/test/schema.rb
+++ b/lib/active_model_serializers/test/schema.rb
@@ -19,6 +19,7 @@ module ActiveModelSerializers
         assert(matcher.call, matcher.message)
       end
 
+      # May be renamed
       def assert_request_response_schema(schema_path = nil, message = nil)
         assert_request_schema(schema_path, message)
         assert_response_schema(schema_path, message)
@@ -35,6 +36,7 @@ module ActiveModelSerializers
       class AssertSchema
         attr_reader :schema_path, :request, :response, :message, :payload
 
+        # Interface may change.
         def initialize(schema_path, request, response, message, payload = nil)
           require_json_schema!
           @request = request
--- a/lib/active_model_serializers/test/schema.rb
+++ b/lib/active_model_serializers/test/schema.rb
@@ -10,19 +10,36 @@ module ActiveModelSerializers
       #   get :index
       #   assert_response_schema
       def assert_response_schema(schema_path = nil, message = nil)
-        matcher = AssertResponseSchema.new(schema_path, response, message)
+        matcher = AssertResponseSchema.new(schema_path, request, response, message)
+        assert(matcher.call, matcher.message)
+      end
+
+      def assert_request_schema(schema_path = nil, message = nil)
+        matcher = AssertRequestSchema.new(schema_path, request, response, message)
+        assert(matcher.call, matcher.message)
+      end
+
+      def assert_request_response_schema(schema_path = nil, message = nil)
+        assert_request_schema(schema_path, message)
+        assert_response_schema(schema_path, message)
+      end
+
+      def assert_schema(payload, schema_path = nil, message = nil)
+        matcher = AssertSchema.new(schema_path, request, response, message, payload)
         assert(matcher.call, matcher.message)
       end
 
       MissingSchema = Class.new(Minitest::Assertion)
       InvalidSchemaError = Class.new(Minitest::Assertion)
 
-      class AssertResponseSchema
-        attr_reader :schema_path, :response, :message
+      class AssertSchema
+        attr_reader :schema_path, :request, :response, :message, :payload
 
-        def initialize(schema_path, response, message)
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
Is this a WIP?

lib/active_model_serializers/test/schema.rb
```diff
@@ -68,6 +85,10 @@ def response_body
           load_json(response.body)
         end
 
+        def request_params
+          request.env['action_dispatch.request.request_parameters']
```
bf4: this is defined in Rails.  We want it to be only GET params or the POST (raw_params) depending on the request.method, I think. I don't think there's a scenario where they should be merged (as request.params does, IIRC)

lib/active_model_serializers/test/schema.rb
```diff
@@ -41,11 +58,11 @@ def call
         attr_reader :document_store
 
         def controller_path
-          response.request.filtered_parameters[:controller]
+          request.filtered_parameters[:controller]
         end
 
         def action
-          response.request.filtered_parameters[:action]
+          request.filtered_parameters[:action]
```
bf4: I think this was a 🐛

lib/active_model_serializers/test/schema.rb
```diff
@@ -10,19 +10,36 @@ module Schema
...
 
       MissingSchema = Class.new(Errno::ENOENT)
       InvalidSchemaError = Class.new(StandardError)
 
-      class AssertResponseSchema
-        attr_reader :schema_path, :response, :message
+      class AssertSchema
+        attr_reader :schema_path, :request, :response, :message, :payload
 
-        def initialize(schema_path, response, message)
```
bf4: This parameter list is getting rather long

bf4: - we always need a payload
- we always need the request if schema_path is nil
  -  GET, DELETE we don't need the response_body
  - else PUT, POST, PATCH we do need the response body

so a generic assertion would just need the `schema_path, message, and payload`

the request assertion would also need the `request`

and  the response assertion would also need the `request, and response`.. so I guess we can do attribute overloading here...

lib/active_model_serializers/test/schema.rb
```diff
@@ -10,19 +10,36 @@ module Schema
...
+
+      def assert_request_schema(schema_path = nil, message = nil)
+        matcher = AssertRequestSchema.new(schema_path, request, response, message)
+        assert(matcher.call, matcher.message)
+      end
+
+      def assert_request_response_schema(schema_path = nil, message = nil)
+        assert_request_schema(schema_path, message)
+        assert_response_schema(schema_path, message)
+      end
```
bf4: I'm not sure about the naming or practicality of this, so maybe I should ✂️ for now

lib/active_model_serializers/test/schema.rb
```diff
@@ -10,19 +10,36 @@ module Schema
...
       def assert_response_schema(schema_path = nil, message = nil)
-        matcher = AssertResponseSchema.new(schema_path, response, message)
+        matcher = AssertResponseSchema.new(schema_path, request, response, message)
+        assert(matcher.call, matcher.message)
+      end
+
+      def assert_request_schema(schema_path = nil, message = nil)
+        matcher = AssertRequestSchema.new(schema_path, request, response, message)
+        assert(matcher.call, matcher.message)
+      end
```
bf4: These two should really just be sugar for `assert_schema`

bf4:
@NullVoxPopuli yes, it's a WIP, though I am using it. the two other bullet points, on reflection , are really JSON API schema concerns, and I don't think they belong here

bf4:
updated with changelog

NullVoxPopuli:
conflicts :-(

bf4:
@NullVoxPopuli updated

bf4:
I've been using this locally for a few weeks now. no complaints

lib/active_model_serializers/test/schema.rb
```diff
@@ -10,19 +10,38 @@ module Schema
...
+      class AssertSchema
+        attr_reader :schema_path, :request, :response, :message, :payload
 
-        def initialize(schema_path, response, message)
+        # Interface may change.
+        def initialize(schema_path, request, response, message, payload = nil)
           require_json_schema!
+          @request = request
           @response = response
+          @payload = payload
```
maksar, line 44: line 53 should probably be changed, `response_body` should be replaced with `@payload`

bf4: @maksar PRs welcome.. you just commented on a PR that was closed four years ago

PR #1647, opened 2016-04-01
Restrict serializable_hash to accepted options

bf4:
See #1643 for rationale behind these changes.

Some of these failures are addressed in other PRs such as
https://github.com/rails-api/active_model_serializers/pull/1572

PR #1678, opened 2016-04-11
Restrict serializable_hash to accepted options

bf4:
Follows https://github.com/rails-api/active_model_serializers/pull/1647

- [x] Restrict options
- [ ] should root go to the adapter?
- [x] needs tests of ignored options

### DIFF
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -8,7 +8,7 @@ module ActiveModelSerializers
       end
 
       def serializable_hash(options = nil)
-        options ||= {}
+        options = serialization_options(options)
 
         if serializer.respond_to?(:each)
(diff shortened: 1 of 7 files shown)


groyoh:
Root should definetely go to the adapter options IMO.

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -8,7 +8,7 @@ def initialize(serializer, options = {})
       end
 
       def serializable_hash(options = nil)
-        options ||= {}
+        options = serialization_options(options)
```
bf4: also see https://github.com/rails-api/active_model_serializers/pull/965

bf4:
@groyoh on reflection, the `root` to the `ADAPTER_OPTION_KEYS` is an unrelated concern (though I agree)

PR #1681, opened 2016-04-13
Fix the order of pagination links for the json_api adapter

vipin8169:
#### Purpose
Fix the order of pagination links for the json_api adapter

#### Changes
in lib/active_model_serializers/adapter/json_api/pagination_links.rb

In all the pagination links generated by this adapter the self link comes before first and prev, ideally it should come after them.

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/pagination_links.rb
+++ b/lib/active_model_serializers/adapter/json_api/pagination_links.rb
@@ -31,12 +31,12 @@ module ActiveModelSerializers
           return {} if collection.total_pages == FIRST_PAGE
 
           {}.tap do |pages|
-            pages[:self] = collection.current_page
-
             unless collection.current_page == FIRST_PAGE
               pages[:first] = FIRST_PAGE
               pages[:prev]  = collection.current_page - FIRST_PAGE
             end
+
+            pages[:self] = collection.current_page
 
             unless collection.current_page == collection.total_pages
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
why should the `self` link come after first and prev?

what was wrong with it where it was?

vipin8169:
Shouldn't the default order of links be first>>prev>>self>>next>>last instead of self>>first>>prev>>next>>last ??

groyoh:
@vipin8169 Please have a look at http://json.org/:

> An object is an unordered set of name/value pairs.

If this is not an actual bug, please close this PR.

NullVoxPopuli:
Order isn't necessarily guaranteed, as is with all hash-like structures.

bf4:
@vipin8169 

> In all the pagination links generated by this adapter the self link comes before first and prev, ideally it should come after them.

I guess what you really want is to sort the pagination links by size so that 'self' appears in the right spot.

OTOH, what client is so dependent on hash key order that you'd need this?

I think you might have posted these SO questions?

- https://stackoverflow.com/questions/36301450/get-kaminari-pagination-links-in-the-json-generated-by-the-active-model-serializ
- https://stackoverflow.com/questions/36284989/how-do-i-override-pagination-links-generated-by-active-model-serializers

vipin8169:
Yes, I had posted them

PR #1685, opened 2016-04-16
Extract IncludeTree.

beauby:
#### Purpose

Extract the `IncludeTree` class ([new source](https://github.com/beauby/jsonapi_parser/blob/master/lib/json/api/include_directive.rb)) into its own gem ([jsonapi](http://github.com/beauby/jsonapi), in an attempt to share some work with [jsonapi-resources](http://github.com/cerebris/jsonapi-resources) (c.f. https://github.com/cerebris/jsonapi-resources/pull/679).


NullVoxPopuli:
doh, now it needs rebase :-(

beauby:
No worries, should be an easy one. Plus, let's wait for the gem renaming.

NullVoxPopuli:
Removing 'Needs Team Discussion' because we already discussed what to do with this.
To kinda summarize:
 - rename the gem to help out with cliry (pending mr. minecraft gives permission / frees up the jsonapi gem namespace)

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -33,7 +36,7 @@ def serializable_hash_for_single_resource(options)
 
       def resource_relationships(options)
         relationships = {}
-        serializer.associations(@include_tree).each do |association|
+        serializer.associations(@include_directive).each do |association|
           relationships[association.key] ||= relationship_value_for(association, options)
         end
```
beauby: Unrelated to this PR but this method could be rewritten as follows:
```ruby
      def resource_relationships(options)
        serializer.associations(@include_directive)
                  .each_with_object({}) do |association, hash|
          hash[association.key] = relationship_value_for(association, options)
        end
      end
```

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -33,7 +36,7 @@ def serializable_hash_for_single_resource(options)
 
       def resource_relationships(options)
```
beauby: Unrelated to this PR but this method could be rewritten as follows:
```ruby
def resource_relationships(options)
  serializer.associations(@include_directive)
            .each_with_object({}) do |association, hash|
    hash[association.key] = relationship_value_for(association, options)
  end
end
```

beauby:
Rebased and changed the gem's name to `jsonapi` as well as the namespace to `JSONAPI`.

active_model_serializers.gemspec
```diff
@@ -42,6 +42,8 @@ Gem::Specification.new do |spec|
     # 'minitest'
     # 'thread_safe'
 
+  spec.add_runtime_dependency 'jsonapi'
```
NullVoxPopuli: :100:

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -3,8 +3,11 @@ module Adapter
     class Attributes < Base
       def initialize(serializer, options = {})
         super
-        @include_tree = ActiveModel::Serializer::IncludeTree.from_include_args(options[:include] || '*')
-        @cached_attributes = options[:cache_attributes] || {}
+        @include_directive = options.fetch(:include_directive) do
```
NullVoxPopuli: can you document this option (before we all forget about it) ?

NullVoxPopuli:
LGTM, just need to document the new option in the docs, before merging.

bf4:
Yay! though I don't think the jsonapi gem should be implementing the `*` since that's not part of the spec and makes valid params checking a bit weird :)

beauby:
@bf4 The default behavior of the gem is to ignore `*`. One can enable it via the "experimental option" `allow_wildcards`. Moreover, there is [talk](https://github.com/json-api/json-api/issues/940) about adding some kind of wildcard in the spec. I mainly added it not to break backwards compatibility in AMS, but I'm totally open to discussing removing it.

bf4:
Awesome response!

B mobile phone

NullVoxPopuli:
needs rebase :-(

lib/active_model_serializers/adapter/attributes.rb
beauby: This is a private/internal "option". It allows to pass an instance of `JSONAPI::IncludeDirective` directly, instead of a string that will be parsed later. It is only used to avoid unnecessary recomputation of the `IncludeDirective` upon recursive calls of the `Attributes` adapter.

beauby:
Rebased and changelog added!

beauby:
Rebased again. We should probably pin the dependency to `jsonapi` though, as it is in beta.

lib/active_model_serializers/adapter/attributes.rb
NullVoxPopuli: sounds good to me

NullVoxPopuli:
@beauby one last rebase!?

beauby:
Rebased :+1:

NullVoxPopuli:
LGTM! good work @beauby. Your JSON API gem is great! :-)

PR #1687, opened 2016-04-18
Lazify calculating caller file digest until used

bf4:


lib/active_model/serializer/caching.rb
```diff
@@ -42,7 +43,12 @@ module ClassMethods
         def inherited(base)
           super
           caller_line = caller[1]
-          base._cache_digest = digest_caller_file(caller_line)
+          base._cache_digest_file_path = caller_line
+        end
+
+        def _cache_digest
+          return @_cache_digest if defined?(@_cache_digest)
+          @_cache_digest = digest_caller_file(_cache_digest_file_path)
```
beauby: Could be rewritten as `@_cache_digest ||= digest_caller_file(_cache_digest_file_path)` instead of the two lines.

bf4: You're right. It should never be nilnor empty

PR #1693, opened 2016-04-20
Issue warning when no serializer is found for an association.

### LINKED ISSUE #1075: Inconsistent JSONAPI serialization when missing related serializer
Suppose we have a `Post` class with `has_many :comments`, along with a `PostSerializer`. If we forget to define a `CommentSerializer`, rendering a `Post` with the JSONAPI adapter leads to the confusing following result (the comments are actually embedded in the relationship, which breaks JSONAPI spec):
```ruby
{
  data: {
    type: 'posts',
    id: 1,
    attributes: { title: 'Title', body: 'Body' }
    relationships: {
      comments: {
        { 
          data: [{ id: 1, contents: 'Contents', author_id: 1, post_id: 1 },
                 { id: 2, contents: 'Contents', author_id: 2, post_id: 1 }]
        }
      }
    }
  }
}
```
It might be helpful to either have a way to warn the user that a given model does not have a corresponding serializer defined (as suggested in #1068), or have a default behavior of serializing stuff as Resource Identifiers.

beauby:
Fix #1075.

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -119,6 +119,7 @@ module ActiveModel
               serializer_options(subject, parent_serializer_options, reflection_options)
             )
           rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
+            warn "No serializer found for #{association_value}."
             reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
           end
         elsif !association_value.nil? && !association_value.instance_of?(Object)


lib/active_model/serializer/reflection.rb
```diff
@@ -119,6 +119,7 @@ def build_association(subject, parent_serializer_options)
               serializer_options(subject, parent_serializer_options, reflection_options)
             )
           rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
+            warn "No serializer found for #{association_value}."
```
bf4: this really needs to be a throw and catch as it's not exactly an exception, it's flow control.

beauby: Agreed, I always disliked the whole `build_reflection` flow and how it was written. Let's clean this up in a subsequent PR, as this one is simply about adding the warning.

bf4: Reason for it is fine, if it can cause inconsistent serialization.  

We use `warn` elsewhere and not `STDERR`, so that's a good choice.

Maybe also `AMS.logger.warn { "No serializer found for #{association_value}." }` ?

beauby: I wasn't sure about where to warn exactly. Maybe `AMS.logger` is the right place.

NullVoxPopuli:
Maybe pending addition of AMS.logger?

;-)

NullVoxPopuli:
@beauby any chance you could quick make the AMS.logger change?

lib/active_model/serializer/reflection.rb
bf4: Should also include object class name and id

bf4:
Resolved in https://github.com/rails-api/active_model_serializers/issues/1767, I  believe

PR #1695, opened 2016-04-20
Meta no longer handled in Base adapter.

beauby:
Make each relevant adapter (i.e. `Json` and `JsonApi`) handle top level `meta`.

### DIFF
--- a/lib/active_model_serializers/adapter/base.rb
+++ b/lib/active_model_serializers/adapter/base.rb
@@ -26,9 +26,7 @@ module ActiveModelSerializers
       end
 
       def as_json(options = nil)
-        hash = serializable_hash(options)
-        include_meta(hash)
-        hash
+        serializable_hash(options)
       end
 
       def fragment_cache(cached_hash, non_cached_hash)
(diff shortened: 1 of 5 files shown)


lib/active_model_serializers/adapter/base.rb
```diff
@@ -26,9 +26,7 @@ def serializable_hash(_options = nil)
       end
 
       def as_json(options = nil)
-        hash = serializable_hash(options)
-        include_meta(hash)
-        hash
+        serializable_hash(options)
       end
```
bf4: alias method should be fine here, unless we want to keep the `root` option here like in the am:s:j in rails

beauby: It failed because the `alias` would be bound to the pre-inheritance placeholder method.

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -130,6 +130,8 @@ def success_document
           hash[:links].update(pagination_links_for(serializer))
         end
 
+        hash[:meta] = instance_options[:meta] if instance_options[:meta].is_a?(Hash)
```
bf4: ` if instance_options[:meta].is_a?(Hash)`.  maybe ` if instance_options[:meta].respond_to?(:fetch)`

beauby: Actually we kinda want to ensure the value of `meta` is a hash (as per the spec).

beauby:
Tests failing because of cyclomatic complexity of `JsonApi#success_document`.

NullVoxPopuli:
@beauby wow, yeah -- it doesn't even look that bad.

As a whole, LGTM, nice work

beauby:
Now Appveyor is failing for no reason.

NullVoxPopuli:
:+1:

PR #1699, opened 2016-04-21
String/Lambda support for conditional attributes/associations

mtsmfm:
#### Purpose

I want to exclude attribute if its association/attribute is nil.

`:if` option accepts only symbol at this time so we must write condition like following:

```ruby
class UserSerializer < ApplicationSerializer
  has_one :game, if: :game_exists?

  def game_exists?
    object.game
  end
end
```

But it is not bad to support String like [rails](http://guides.rubyonrails.org/active_record_validations.html#conditional-validation):

```ruby
class UserSerializer < ApplicationSerializer
  has_one :game, if: 'object.game'
end
```

I don't have any use case for lambda(proc) at this time but many developers who experienced rails will guess that it accepts lambda.

#### Related GitHub issues

#1403

### DIFF
--- a/lib/active_model/serializer/field.rb
+++ b/lib/active_model/serializer/field.rb
@@ -4,6 +4,12 @@ module ActiveModel
     # specified in the ActiveModel::Serializer class.
     # Notice that the field block is evaluated in the context of the serializer.
     Field = Struct.new(:name, :options, :block) do
+      def initialize(*)
+        super
+
+        validate_condition!
+      end
+
       # Compute the actual value of a field for a given serializer instance.
       # @param [Serializer] The serializer instance for which the value is computed.
       # @return [Object] value
@@ -27,15 +33,43 @@ module ActiveModel
       def excluded?(serializer)
         case condition_type
         when :if
-          !serializer.public_send(condition)
+          !evaluate_condition(serializer)
         when :unless
-          serializer.public_send(condition)
+          evaluate_condition(serializer)
         else
           false
         end
       end
 
       private
+
+      def validate_condition!
+        return if condition_type == :none
+
+        case condition
+        when Symbol, String, Proc
+          # noop
+        else
+          fail TypeError, "#{condition_type.inspect} should be a Symbol, String or Proc"
+        end
+      end
+
+      def evaluate_condition(serializer)
+        case condition
+        when Symbol
+          serializer.public_send(condition)
+        when String
+          serializer.instance_eval(condition)
+        when Proc
+          if condition.arity.zero?
+            serializer.instance_exec(&condition)
+          else
+            serializer.instance_exec(serializer, &condition)
+          end
+        else
+          nil
+        end
+      end
 
       def condition_type
         @condition_type ||=
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -239,27 +239,55 @@ module ActiveModel
           end
         end
 
+        # rubocop:disable Metrics/AbcSize
         def test_conditional_associations
-          serializer = Class.new(ActiveModel::Serializer) do
-            belongs_to :if_assoc_included, if: :true
-            belongs_to :if_assoc_excluded, if: :false
-            belongs_to :unless_assoc_included, unless: :false
-            belongs_to :unless_assoc_excluded, unless: :true
+          model = ::Model.new(true: true, false: false)
 
-            def true
-              true
+          scenarios = [
+            { options: { if:     :true  }, included: true  },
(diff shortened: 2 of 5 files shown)


beauby:
Yeah, I had support for lambda in mind when adding this feature, so :+1: on the idea of this PR.

lib/active_model/serializer/field.rb
```diff
@@ -27,16 +27,29 @@ def value(serializer)
...
           false
         end
       end
 
       private
 
+      def condition_lambda
+        case condition
+        when Symbol
+          -> (serializer) { serializer.public_send(condition) }
```
beauby: I'm not sure building a throwaway lambda here is the best way. How about the following
```ruby
def condition_lambda(serializer)
  case condition
  when Symbol
    serializer.public_send(condition)
  when String
    serializer.instance_eval(condition)
  when Proc
    condition.call(serializer)
...
```

mtsmfm: Yeah, it can be a method :+1: 
Fixed and force pushed :smile:

beauby:
LGTM – I'll merge unless somebody raises concerns.
Would you mind adding corresponding docs for the feature?

lib/active_model/serializer/field.rb
```diff
@@ -27,16 +27,29 @@ def value(serializer)
...
+      def evaluate_condition(serializer)
+        case condition
+        when Symbol
+          serializer.public_send(condition)
+        when String
+          serializer.instance_eval(condition)
+        when Proc
+          condition.call(serializer)
+        else
+          fail ArgumentError
```
bf4: would be good to include a message here

test/serializers/associations_test.rb
```diff
@@ -261,6 +261,36 @@ def false
...
+          serializer = Class.new(ActiveModel::Serializer) do
+            belongs_to :if_assoc_included, if: 'object.true'
+            belongs_to :if_assoc_excluded, if: 'object.false'
+            belongs_to :unless_assoc_included, unless: 'object.false'
+            belongs_to :unless_assoc_excluded, unless: 'object.true'
+          end
+
+          model = ::Model.new(true: true, false: false)
+          hash = serializable(model, serializer: serializer).serializable_hash
+          expected = { if_assoc_included: nil, unless_assoc_included: nil }
```
bf4: took me a moment to figure out the naming is the key to the expectation here :shrug:

test/serializers/associations_test.rb
```diff
@@ -261,6 +261,36 @@ def false
...
+
+          assert_equal(expected, hash)
+        end
+
+        def test_lambda_conditional_associations
+          serializer = Class.new(ActiveModel::Serializer) do
+            belongs_to :if_assoc_included, if: -> (s) { s.object.true }
```
bf4: Isn't the rails way these are used to evaluate in the context of the instance? i.e. `-> { object.true }`?

test/serializers/associations_test.rb
beauby: Agreed, but to be fair I'm the one who introduced those names...

test/serializers/associations_test.rb
mtsmfm: I noticed rails can both way (`-> { object.true }`, `-> (s) { s.object.true }`) :bulb:

lib/active_model/serializer/field.rb
mtsmfm: :+1: 

By the way, it is better if raise an error on evaluate serializer's DSL I think.

https://github.com/rails-api/active_model_serializers/pull/1699/files#diff-689b80d81fb899042cc36668903c9a3dR10

mtsmfm:
Hmm, tests for jruby-9.0.4.0 on appveyor are failed but passed on travis :confused:

beauby:
@mtsmfm Looking into the appveyor thing.

mtsmfm:
@beauby I noticed appveyor setting for JRuby is wrong and fixed (a55502889aa275b359541f26e80450ea7c730e39)

beauby:
@mtsmfm Awesome! Would you mind submitting the AppVeyor settings fix as a separate PR?

appveyor.yml
```diff
@@ -3,21 +3,19 @@ version: '{build}'
 skip_tags: true
```
beauby: Could you rebase?

mtsmfm: rebased!

nicolas-besnard:
This is great for one field, but what if I've a lot a field I wanted to hide ? Do I have to add a predicate after each attribute / association ?

PR #1710, opened 2016-04-29
Prevent loading association when include_data is set to false

### LINKED ISSUE #1707: Bug: include_data loads association

Given the following serializer:

```ruby
  has_many :comments do
    link :related do
      href "/posts/#{object.id}/comments"
    end
    include_data false
  end
end
```

This correctly does not include comment relationship data in the response. However, all the comments are still loaded (ie `select * from comments where post_id = ?`). I would expect this to avoid loading the comments since none of that data is needed in the response.

#### Steps to reproduce

Any serializer with `include_data false`, look at the ActiveRecord queries that fire.

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: master/0165215

Output of `ruby -e "puts RUBY_DESCRIPTION"`: ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]

OS Type & Version: OSX 10.10.5 (Yosemite)

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails 5 beta 3

groyoh:
#### Purpose
Fixes #1707. 

#### Changes
Do not call `read_attribute_for_serialization` when `include_data` is set to `false`.

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -75,10 +75,10 @@ module ActiveModel
 
         if block
           block_value = instance_exec(serializer, &block)
-          if block_value == :nil
-            serializer.read_attribute_for_serialization(name)
-          else
+          if block_value != :nil
             block_value
+          elsif @_include_data
+            serializer.read_attribute_for_serialization(name)
           end
         else
           serializer.read_attribute_for_serialization(name)
--- a/test/adapter/json_api/relationships_test.rb
+++ b/test/adapter/json_api/relationships_test.rb
@@ -5,7 +5,8 @@ module ActiveModel
     module Adapter
       class JsonApi
         class RelationshipTest < ActiveSupport::TestCase
-          RelationshipAuthor = Class.new(::Model)
+          class RelationshipAuthor < ::Model; end
+
           class RelationshipAuthorSerializer < ActiveModel::Serializer
             has_one :bio do
               link :self, '//example.com/link_author/relationships/bio'
@@ -71,7 +72,6 @@ module ActiveModel
 
           def setup
             @post = Post.new(id: 1337, comments: [], author: nil)
-            @blog = Blog.new(id: 1337, name: 'extra')
             @bio = Bio.new(id: 1337)
             @like = Like.new(id: 1337)
             @role = Role.new(id: 'from-record')
@@ -82,7 +82,6 @@ module ActiveModel
             @author = RelationshipAuthor.new(
               id: 1337,
               posts: [@post],
-              blog: @blog,
               reviewer: @reviewer,
               bio: @bio,
               likes: [@like],
@@ -158,10 +157,16 @@ module ActiveModel
           end
 
           def test_relationship_not_including_data
+            @author.define_singleton_method(:read_attribute_for_serialization) do |attr|
+              fail 'should not be called' if attr == :blog
+              super(attr)
+            end
             expected = {
               links: { self: '//example.com/link_author/relationships/blog' }
             }
-            assert_relationship(:blog, expected)
+            assert_nothing_raised do
+              assert_relationship(:blog, expected)
+            end
           end
 
           def test_relationship_including_data_explicit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Breaking changes:
 Features:
 
 Fixes:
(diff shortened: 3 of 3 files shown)


NullVoxPopuli:
gasp! those errors!
didyou/can you rebase? I *think* some jruby fixes were merged recently

groyoh:
Yeah, already rebased to latest master. Not sure what happens here, I can't reproduce it locally.

NullVoxPopuli:
what if you uninstall the sqlite3 adapter? - it's complaining that it's not installed. 

While it's not a dep of ams, it is for the [ActiveRecord tests](https://github.com/rails-api/active_model_serializers/blob/b73b780b7961b7708a40aefd016bb543a4e8137c/test/fixtures/active_record.rb#L3)

groyoh:
I don't have sqlite3 adapter installed locally.

NullVoxPopuli:
even weirder

groyoh:
Actually, sqlite3 adapter comes out of the box with AR. So THAT's even weirder.

lib/active_model/serializer/reflection.rb
```diff
@@ -75,10 +75,10 @@ def value(serializer)
 
         if block
           block_value = instance_exec(serializer, &block)
-          if block_value == :nil
-            serializer.read_attribute_for_serialization(name)
-          else
+          if block_value != :nil
             block_value
+          elsif @_include_data
```
NullVoxPopuli: when is `@_include_data` present?

groyoh: I'm not sure about the question but it is set to `true` by default so includes the relationship `data` by default. The point of the PR is to prevent calling the relationship if `include_data` is set to `false`.

NullVoxPopuli: aight, I guess we'll document _include_data at a later date

NullVoxPopuli:
woo! LGTM :pizza:

NullVoxPopuli:
@groyoh do you want to add a changelog entry?

groyoh:
@NullVoxPopuli will do.

lib/active_model/serializer/reflection.rb
groyoh: It's already documented but the behavior added in this PR is not though.

groyoh:
@NullVoxPopuli added the CHANGELOG.

NullVoxPopuli:
thanks! merging!

YoranBrondsema:
Although small, this PR has a very big impact for the performance of our app. Would it be possible to bump the version so that the latest version includes this? Thanks!

NullVoxPopuli:
you can specify a commit hash in your gemfile. :-)

we aren't ready to bump the version quite yet.

YoranBrondsema:
I know that :-). I was just wondering if a patch update was possible so that everyone can benefit from this change now, but I will resort to specifying the commit.

sandstrom:
This is a great addition, but wouldn't it make more sense if this was an option:

```ruby
has_one :blog, :include_data => false do |serializer|
  serializer.cached_blog
end
```

Instead of being set from within the block? Or have I missed something?

test/adapter/json_api/relationships_test.rb
```diff
@@ -158,10 +157,16 @@ def test_relationship_meta
           end
 
           def test_relationship_not_including_data
+            @author.define_singleton_method(:read_attribute_for_serialization) do |attr|
+              fail 'should not be called' if attr == :blog
```
bf4: this test needs to change as well.  It shouldn't be raising a runtime error  and is hard to read and why do we even need the define_singleton_method?

😦

lib/active_model/serializer/reflection.rb
bf4: @NullVoxPopuli please don't merge code with missing `else` clauses.  I was going to say it was a bug and raise a ❗ but then I realized that it was the feature of the PR. 😱   :(

test/adapter/json_api/relationships_test.rb
bf4: I also don't understand why this test was changed

bf4: In any case, this test file is basically garbage. See  https://github.com/rails-api/active_model_serializers/pull/1857 would be better to do an 'integration' test

lib/active_model/serializer/reflection.rb
NullVoxPopuli: makes sense.

PR #1711, opened 2016-05-01
Support conditional setting of serialization_scope in controllers

yogeshjain999:
#### Purpose
I want to set `serialization_scope` as different object rather than `current_user`, for only specific actions in given controller.

#### Changes
Adds support for setting `serialization_scope` conditionally in controllers, same as calling `before_action` with options.

### DIFF
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -8,8 +8,10 @@ module ActionController
     include ActionController::Renderers
 
     module ClassMethods
-      def serialization_scope(scope)
-        self._serialization_scope = scope
+      def serialization_scope(scope, opts = {})
+        before_action(opts.slice(:only, :except, :if, :unless)) do
+          self._serialization_scope = scope
+        end
       end
     end
 
--- a/test/action_controller/serialization_scope_name_test.rb
+++ b/test/action_controller/serialization_scope_name_test.rb
@@ -34,6 +34,12 @@ module SerializationScopeTesting
   end
   class PostTestController < ActionController::Base
     attr_accessor :current_user
+
+    def render_post
+      self.current_user = User.new(id: 3, name: 'Pete')
+      render json: new_post, serializer: serializer, adapter: :json
+    end
+
     def render_post_by_non_admin
       self.current_user = User.new(id: 3, name: 'Pete', admin: false)
       render json: new_post, serializer: serializer, adapter: :json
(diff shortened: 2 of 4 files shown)


bf4:
I like it.  I need to look at the test changes a bit.

test/action_controller/serialization_scope_name_test.rb
```diff
@@ -71,10 +77,12 @@ class DefaultScopeTest < ActionController::TestCase
     tests PostTestController
 
     def test_default_serialization_scope
+      get :render_post
```
bf4: this is supposed to be missing :)

test/action_controller/serialization_scope_name_test.rb
```diff
@@ -187,14 +233,6 @@ def new_post
...
     tests PostViewContextTestController
 
-    def test_nil_serialization_scope
-      assert_nil @controller._serialization_scope
-    end
-
-    def test_nil_serialization_scope_object
-      assert_nil @controller.serialization_scope
-    end
-
```
bf4: Unrelated change, no?

lib/action_controller/serialization.rb
```diff
@@ -8,8 +8,10 @@ module Serialization
     include ActionController::Renderers
 
     module ClassMethods
-      def serialization_scope(scope)
-        self._serialization_scope = scope
+      def serialization_scope(scope, opts = {})
+        before_action(opts.slice(:only, :except, :if, :unless)) do
```
bf4: Since the method is only called at `render`, if not passed in as an option, I don't think I like having it eagerly evaluated in code that is pretty easy to handle in userland.  Would you be okay with removing the code changes in favor documenting using a `before_action` as an alternative to passing an option to `render`?

bf4: I also would except there to be bugs with `current_user` being called in a `before_action` with all the side-effects authentication-related libs can have

test/action_controller/serialization_scope_name_test.rb
```diff
@@ -34,6 +34,12 @@ def json_key
   end
   class PostTestController < ActionController::Base
     attr_accessor :current_user
+
+    def render_post
+      self.current_user = User.new(id: 3, name: 'Pete')
+      render json: new_post, serializer: serializer, adapter: :json
+    end
```
bf4: this condition is identical to the one below it :)

bf4:
Closing as stale. No response in a month.

lib/action_controller/serialization.rb
yogeshjain999: I agree to your suggestion @bf4 about moving this to userland, created #1824 with added documentation.

PR #1712, opened 2016-05-02
Setting the content_type sets the response Content-Type

bf4:
Otherwise we have two headers, 'Content-Type' and 'CONTENT_TYPE'.
I don't know when Rails decides to use one or the other.


groyoh:
Based on https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/actionpack/lib/action_controller/metal/renderers.rb#L168, this looks good to me! :+1:

PR #1714, opened 2016-05-02
Allow options on multiple attributes at once

nicklandgrebe:
#### Purpose

Add options to `Serializer.attributes` so we can apply conditions to multiple attributes at once.
#### Changes

Check if last item in `attrs` passed into `Serializer.attributes` has condition `class == Hash`, and if so, `pop` it off `attrs`, then pass it into `attribute(attr, options)` every time `attrs` is iterated over.

Before, only `attribute(attr)` would be called every time `attrs` is iterated over.
#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/pull/1403#issuecomment-172044770

### DIFF
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -34,10 +34,11 @@ module ActiveModel
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     attributes :id, :name, :recent_edits
         def attributes(*attrs)
+          options = attrs.pop if attrs.last.class == Hash
           attrs = attrs.first if attrs.first.class == Array
 
           attrs.each do |attr|
-            attribute(attr)
+            attribute(attr, options)
           end
         end
 
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -38,7 +38,7 @@ module ActiveModel
           attrs = attrs.first if attrs.first.class == Array
 
           attrs.each do |attr|
-            attribute(attr, options)
+            attribute(attr, options || {})
           end
         end
 
--- a/lib/active_model/serializer/attributes.rb
+++ b/lib/active_model/serializer/attributes.rb
@@ -34,11 +34,13 @@ module ActiveModel
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     attributes :id, :name, :recent_edits
         def attributes(*attrs)
-          options = attrs.pop if attrs.last.class == Hash
+          options = attrs.last.class == Hash ? attrs.pop : {}
+          options = options.except(:key)
+
           attrs = attrs.first if attrs.first.class == Array
 
           attrs.each do |attr|
-            attribute(attr, options || {})
+            attribute(attr, options)
           end
         end
 
--- a/test/serializers/attributes_test.rb
+++ b/test/serializers/attributes_test.rb
@@ -48,6 +48,7 @@ module ActiveModel
         assert_equal([:id, :title, :body], serializer_class._attributes)
       end
 
+      # rubocop:disable Metrics/AbcSize
       def test_multiple_conditional_attributes
         model = ::Model.new(true: true, false: false)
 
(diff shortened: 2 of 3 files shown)


nicklandgrebe:
Heyo, I added a test for the lines changed. Apologies for not doing this in the first place!

lib/active_model/serializer/attributes.rb
```diff
@@ -34,10 +34,11 @@ def inherited(base)
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     attributes :id, :name, :recent_edits
         def attributes(*attrs)
+          options = attrs.pop if attrs.last.class == Hash
```
groyoh: I guess something like would be better:

```diff
+          options = attrs.last.class == Hash ? attrs.pop : {}
+          options = options.except(:key)
```

At least, we prevent ourselves from having unintended behavior.

test/serializers/attributes_test.rb
```diff
@@ -47,6 +47,16 @@ def test_multiple_calls_with_the_same_attribute
 
         assert_equal([:id, :title, :body], serializer_class._attributes)
       end
+
```
groyoh: I'd rather have a test with multiple attributes and an `if/unless` option since this is the intended use case.

remear:
Note that the CHANGELOG has changed in master recently with the release of 0.10.0. This PR needs to be updated to move the entry to the new master section. The commits also need to be squashed.

zhen639:
Hey, @nicklandgrebe.
I am using your post in my code but it has a problem.
If there are multiple attributes called, duplicated attributes would be ignored in false conditions, for example:

```
class CandidateSerializer < ActiveModel::Serializer
  def self.attributes(*attrs)
    binding.pry
    options = attrs.last.class == Hash ? attrs.pop : {}
    options = options.except(:key)

    attrs.each do |attr|
      attribute(attr, options)
    end
  end

  attributes %I(university resume paper_test_score), if: :in_paper_test?
  attributes %I(university resume homework), if: :in_homework?

  def in_paper_test
    true
  end

  def in_home_work
    false
  end
end
```

It returns 
```
...
:attributes=>{:name=>"zhangsan", :cellphone=>"10000000001", :email=>"a_2@a.com", :gender=>false, :"paper_test_score"=>100},
...
```

Attributes `:university` and `:resume` are ignored since there have false condition in second `attributes` called. In other words, they are overwritten in `attribute(attr, options)` .

Would you like to have a look at this issue?

groyoh:
@zhen639 attributes are not really meant to be (re)defined multiple times (at least that meither tested or implemented AFAIK). I would suggest:

```ruby
class CandidateSerializer < ActiveModel::Serializer
  attribute :paper_test_score, if: :in_paper_test?
  attribute :homework, if: :in_homework?
  attributes %I(university resume), if: :in_paper_test_and_homework?

  def in_paper_test_and_homework?
    in_paper_test? && in_homework?
  end

  def in_paper_test?
    true
  end

  def in_home_work?
    false
  end
end
```

lib/active_model/serializer/attributes.rb
```diff
@@ -34,10 +34,13 @@ def inherited(base)
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     attributes :id, :name, :recent_edits
         def attributes(*attrs)
+          options = attrs.last.class == Hash ? attrs.pop : {}
+          options = options.except(:key)
```
bf4: good call.  I wonder if this could be supported via activesupport `with_options` and no code changes?

bf4:
> attributes are not really meant to be (re)defined multiple times

This is definitely a scenario ams does not protect against

beauby:
@bf4 I think it (AMS) should (protect against attributes redefinition).

adin-ttc:
Anything happening with this? i have a workaround but it would be nice :)

bf4:
basic workaround is if you are

```ruby
attributes :university, :resume, :paper_test_score, if: :in_paper_test?
```

then you can

```ruby
attribute :university, if: :in_paper_test?
attribute :resume, if: :in_paper_test?
attribute :paper_test_score, if: :in_paper_test?
```

other work around is to use composition, inheritance, or options...

bf4:
PR needs to be resubmitted against 0-10-stable.

nicklandgrebe:
Resubmitted as #2148 against 0-10-stable. This totally fell off my radar, thanks for the push!

PR #1716, opened 2016-05-03
Add failing test case for conditional attributes on a shared key

willcosgrove:
#### Purpose

Having a condition on two attributes that share a key, currently makes them share the condition as well.

``` rb
  attribute :other_relation, key: :relationship, unless: :self?
  attribute :self_relation, key: :relationship, if: :self?

  def self?
    object.id == current_user.id
  end

  def self_relation
    "self"
  end

  def other_relation
    "other"
  end
end
```

If `self?` returns true, the `relationship` key will correctly have value `"self"`.  But if `self?` returns false, the `relationship` key will not be on the resulting serialized hash.

This is a contrived example, but hopefully it clearly demonstrates the problem.
#### Changes

This adds a test case to show the problem

### DIFF
--- a/test/serializers/attribute_test.rb
+++ b/test/serializers/attribute_test.rb
@@ -97,7 +97,7 @@ module ActiveModel
       end
 
       # rubocop:disable Metrics/AbcSize
-      def test_conditional_associations
+      def test_conditional_attributes
         model = ::Model.new(true: true, false: false)
 
         scenarios = [
@@ -137,6 +137,26 @@ module ActiveModel
         end
       end
 
+      def test_conditional_attributes_on_shared_key_when_first_attribute_is_ignored
+        model = ::Model.new
+
+        serializer = conditional_attributes_with_shared_key_serializer(skip_first_attribute: true)
+
+        hash = serializable(model, serializer: serializer).serializable_hash
+        assert(hash.key?(:attribute), "attribute key was left out of serializable hash")
+        assert_equal("private", hash[:attribute])
+      end
+
+      def test_conditional_attributes_on_shared_key_when_second_attribute_is_ignored
+        model = ::Model.new
+
+        serializer = conditional_attributes_with_shared_key_serializer(skip_first_attribute: false)
+
+        hash = serializable(model, serializer: serializer).serializable_hash
+        assert(hash.key?(:attribute), "attribute key was left out of serializable hash")
+        assert_equal("public", hash[:attribute])
+      end
+
       def test_illegal_conditional_attributes
         exception = assert_raises(TypeError) do
           Class.new(ActiveModel::Serializer) do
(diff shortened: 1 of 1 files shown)


test/serializers/attribute_test.rb
```diff
@@ -137,6 +137,33 @@ def false
         end
       end
 
+      def test_multiple_conditional_attributes_on_one_key
```
NullVoxPopuli: could you break this up in to two tests?

willcosgrove: Sure, what would you call the two tests?  Since really we're just testing the order of the attribute declarations

NullVoxPopuli: I would do this for my test headers:

```
test 'conditional attributes with shared key when the first attribute is ignored' do ...
test 'conditional attributes with shared key when the second attribute is ignored' do ...
```

NullVoxPopuli: and maybe to reduce complexity, use booleans for if? instead of a variable? idk. Might be too nit-picky. You already found the bug. lol

willcosgrove:
I split those test cases up, and squashed the commits :+1:

beauby:
This is, indeed an edge case, as having two attributes with the same key will have undefined behavior.
The "right way" of doing what you're trying to achieve is to define *one* attribute, that you override depending on your condition. e.g.
```ruby
  attribute :relationship do
    if self?
      self_relation
    else
      other_relation
    end
  end

  def self?
    object.id == current_user.id
  end

  def self_relation
    "self"
  end

  def other_relation
    "other"
  end
end
```

groyoh:
@beauby your snippet has some errors though. The block is not `instance_eval` on the serializer, so you would have to do something like:

```ruby
  attribute :relationship do |serializer|
    if serializer.self?
      serializer.self_relation
    else
      serializer.other_relation
    end
  end

  def self?
    object.id == current_user.id
  end

  def self_relation
    "self"
  end

  def other_relation
    "other"
  end
end
```

beauby:
@groyoh Oh, you changed that, right? ;)

groyoh:
@bf4 changed it but it has never been instance_eval'd on the serializer. You were able to use `serializer` without passing the block param though.

beauby:
I thought it was – my bad!

test/serializers/attribute_test.rb
```diff
@@ -146,6 +166,23 @@ def test_illegal_conditional_attributes
 
         assert_match(/:if should be a Symbol, String or Proc/, exception.message)
       end
+
+      private
+
+      def conditional_attributes_with_shared_key_serializer(skip_first_attribute:)
+        Class.new(ActiveModel::Serializer) do
+          attribute :public_attribute, key: :attribute, unless: skip_first_attribute.to_s
+          attribute :private_attribute, key: :attribute, if: skip_first_attribute.to_s
```
bf4: re: the example I'd prefer to call the key `name` or anything but `attribute`, since that's a macro, it's a little confusing looking.

re: the pattern, looks like this is the job of serialization scope, one of the original primary purposes of the gem.


something like

```ruby
attribute :name do
  scope.skip_first_attribute? ? object.private_name : object.public_name
end
```

test/serializers/attribute_test.rb
```diff
@@ -97,7 +97,7 @@ def test_virtual_attribute_block
       end
 
       # rubocop:disable Metrics/AbcSize
-      def test_conditional_associations
+      def test_conditional_attributes
```
bf4: 👍

PR #1722, opened 2016-05-10
Fix RuboCop 0.40 linter errors

noahsilas:
#### Purpose
Fixing the Travis-CI build, which seems to use [RuboCop 0.40](https://github.com/bbatsov/rubocop/releases/tag/v0.40.0) despite the Gemfile.lock pinning rubocop to 0.38.

#### Changes
- Mostly whitespace changes to satisfy the linters introduced in https://github.com/bbatsov/rubocop/issues/2857.
- One substitution of if/elsif/else for case/when/else.

New lints that I am updating the code style to reflect:

- Style/EmptyCaseCondition: Do not use empty case condition, instead use
  an if expression.

- Style/MultilineArrayBraceLayout: Closing array brace must be on the
  same line as the last array element when opening brace is on the same
  line as the first array element.

- Style/MultilineHashBraceLayout: Closing hash brace must be on the same
  line as the last hash element when opening brace is on the same line
  as the first hash element.

- Style/MultilineMethodCallBraceLayout: Closing method call brace must
  be on the line after the last argument when opening brace is on a
  separate line from the first argument.

#### Related GitHub issues

- Blocking #1426

### DIFF
--- a/lib/active_model/serializer/include_tree.rb
+++ b/lib/active_model/serializer/include_tree.rb
@@ -95,12 +95,11 @@ module ActiveModel
 
       def [](key)
         # TODO(beauby): Adopt a lazy caching strategy for generating subtrees.
-        case
-        when @hash.key?(key)
+        if @hash.key?(key)
           self.class.new(@hash[key])
-        when @hash.key?(:*)
+        elsif @hash.key?(:*)
           self.class.new(@hash[:*])
-        when @hash.key?(:**)
+        elsif @hash.key?(:**)
           self.class.new(:** => {})
         else
           nil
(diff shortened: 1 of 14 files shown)


remear:
The current Gemfile does not pin rubocop to 0.38. #1723 should unblock #1426 until we can resolve issues introduced by 0.40.0. I'm sure this PR will resolve most of them, so thanks very much.

noahsilas:
Ah, I see. I thought that Gemfile.lock was checked in and pinned the version, but it is actually git-ignored. Since it is not there the CI bundles, and gets the less restrictive version in the current Gemfile, which matches 0.4.

Thanks, that explains my befuddlement. about how this was happening.

bf4:
@noahsilas yeah, the best practice for gem dev is not to commit the Gemfile.lock since it's not used when building or installing the gem.

NullVoxPopuli:
I really like these style changes

NullVoxPopuli:
though, I'm worried about these style changes causing conflicts more more pressing PRs.

@bf4 can you think of any?

Issue #1717, opened 2016-05-04
Nested Associations and type declarations in Polymorphic serializations 0.10+ (latest)


cgmckeever:
#### Expected behavior vs actual behavior

The polymorphic test of the master branch leads me to a bit of confusion with the expected behavior of the returned serialization.  Expected, from the example, I would think that imageable would identify the type and then there would be a lower association of 'employee' with the employee data

https://github.com/rails-api/active_model_serializers/blob/db87f8d85e7a46ef195a555cb7e2509f1affc622/test/adapter/polymorphic_test.rb#L31

```
         def test_json_serializer
          expected =
            {
              picture: {
                id: 1,
                title: 'headshot-1.jpg',
                imageable: {
                  id: 42,
                  name: 'Zoop Zoopler'
                }
              }
            }
```

for our current test of 0.8.2, our test expected this behavior, and passed with a and 'asset' key, a 'type' attribute of 'transformed_image' and then a 'transformed_image' key of the subset of data.

```
  it 'creates media specific JSON for a photo in a typical photo album' do
    album = create(:album, :with_photos)
    media = album.active_medias.first
    parsed_json = ActiveModelSerializers::SerializableResource.new(media, {adapter: :json, include: '**', key_transform: true}).as_json
    expect(parsed_json['id']).to eq media.id
    expect(parsed_json['asset']['type']).to eq 'transformed_image'
    expect(parsed_json['asset']['transformed_image']['id']).to eq media.asset.id
    exp

```

when running with latest, all the transformed data is rolled up under asset, and really doesnt properly define the data.

```
   :asset=>
    {:id=>6,
     :parsed_transform_data=>{},
     :width=>4608,
     :height=>4391,

```

```
   class MediaSerializer < ApplicationSerializer
      attributes :id,
             :label,
             :position,
             :status,
             :media_container_id

     has_one :asset, polymorphic: true
 end

```

Switching to json-api does provide a much closer representation

```

{:asset=>{:data=>{:id=>"6", :type=>"transformed-images"}}}},
 :included=>
  [{:id=>"6",
    :type=>"transformed-images",
    :attributes=>

```

Is this the expected data representation for polymorphs?  There was an older thread/issue stating how an earlier version was improperly handing this, and that eventually got changed.  I think that was for a pre 0.10 branch. Ill have to google fu a little more to find that thread.  Any help on this would be much appreciated.  Thanks!

cgmckeever:
[...]

PR #1726, opened 2016-05-15
adds polymorphic option to association definition which includes asso…

cgmckeever:
#### Purpose

Current polymorphic association adds the association serialization directly under the polymorphic name (ie imageable).  This does not allow the client provider a straightforward knowledge of what the polymorphic association is, and thus can not easily determine how to handle the data.

This PR allows for the polymorphic: option flag on the association to nest the attributes under the association type - henceforth allowing the client to know what the polymorphic association is, and handle accordingly.

#### Changes

adds polymorphic: option to the serializer association definition which then determines if serialized association should be nested under association type

#### Caveats

Affected serialization adapters: **attributes** && **json**
**Does not affect json_api **

#### Related GitHub issues

my earlier question/inquiry:
https://github.com/rails-api/active_model_serializers/issues/1717

Earlier Polymorphic testing PR which touched up the serialization package diversion from 0.8x
https://github.com/rails-api/active_model_serializers/pull/1420#issuecomment-170284925

adds polymorphic option to association definition which includes association type in serializer
regen gemlock

### DIFF
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -45,7 +45,14 @@ module ActiveModelSerializers
         return unless association.serializer && association.serializer.object
 
         opts = instance_options.merge(include: @include_tree[association.key])
-        Attributes.new(association.serializer, opts).serializable_hash(options)
+        relationship_value = Attributes.new(association.serializer, opts).serializable_hash(options)
+
+        if association.options[:polymorphic] && relationship_value
+          polymorphic_type = association.serializer.object.class.name.underscore
+          relationship_value = { type: polymorphic_type, polymorphic_type.to_sym => relationship_value }
+        end
+
+        relationship_value
       end
 
       # Set @cached_attributes
--- a/test/adapter/polymorphic_test.rb
+++ b/test/adapter/polymorphic_test.rb
@@ -8,10 +8,17 @@ module ActiveModel
           @employee = Employee.new(id: 42, name: 'Zoop Zoopler', email: 'zoop@example.com')
           @picture = @employee.pictures.new(id: 1, title: 'headshot-1.jpg')
           @picture.imageable = @employee
+        end
 
-          @attributes_serialization = serializable(@picture, serializer: PolymorphicBelongsToSerializer) # uses default adapter: attributes
-          @json_serialization = serializable(@picture, adapter: :json, serializer: PolymorphicBelongsToSerializer)
-          @json_api_serialization = serializable(@picture, adapter: :json_api, serializer: PolymorphicBelongsToSerializer)
+        def serialization(resource, adapter = :attributes)
+          serializable(resource, adapter: adapter, serializer: PolymorphicBelongsToSerializer).as_json
+        end
+
+        def tag_serialization(adapter = :attributes)
+          tag = PolyTag.new(id: 1, phrase: 'foo')
+          tag.object_tags << ObjectTag.new(id: 1, poly_tag_id: 1, taggable: @employee)
+          tag.object_tags << ObjectTag.new(id: 5, poly_tag_id: 1, taggable: @picture)
+          serializable(tag, adapter: adapter, serializer: PolymorphicTagSerializer, include: '*.*').as_json
         end
 
         def test_attributes_serialization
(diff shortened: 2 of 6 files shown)


test/adapter/polymorphic_test.rb
```diff
@@ -35,8 +38,11 @@ def test_json_serializer
                 id: 1,
                 title: 'headshot-1.jpg',
                 imageable: {
-                  id: 42,
-                  name: 'Zoop Zoopler'
+                  type: 'employee',
+                  employee: {
+                    id: 42,
+                    name: 'Zoop Zoopler'
```
cgmckeever: nested assoication

test/fixtures/poro.rb
```diff
@@ -245,7 +245,7 @@ def maker
 PolymorphicBelongsToSerializer = Class.new(ActiveModel::Serializer) do
   attributes :id, :title
 
-  has_one :imageable, serializer: PolymorphicHasManySerializer
+  has_one :imageable, serializer: PolymorphicHasManySerializer, polymorphic: true
```
cgmckeever: optional association polymorphic: flag

cgmckeever:
hat tip to @remear for bouncing ideas and pointing me in the right direction and @groyoh for being awesome!

This is a first pass, seems fairly straightforward and provided the results that I was anticipating of a polymorphic serializer.  Feedback welcome!

beauby:
What's the behavior for `nil` relationships and heterogeneous has-many?

cgmckeever:
@beauby can you elaborate what 'heterogeneous has-many' means?

Im thinking nil will act the same as how it does prior, but Im wondering if at:
https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/attributes.rb#L37

when a relationships[association.key] is nil, that relation is skipped, but now there will be a type with a nil nesting.  

If you outline the desired expectations, I can write tests and code to them

beauby:
For the "heterogeneous has-many" case, I meant the following:
Suppose we have `Articles`, `Products`, and `Tags`, where both `Articles` and `Products` can have many `Tags`. Now, on the `Tag` side, you have many "taggables" (polymorphic). What should be the name of the association when serializing a `Tag` that has both an `Article` and a `Product`?

cgmckeever:
@beauby ... form a pure theoretical perspective, it would act exactly as it does now, plus the effect of this PR which would add the type to the packet.  such as (from the 0.9.1 docs) [http://www.rubydoc.info/gems/active_model_serializers/0.9.1] .. 


```
    "attachments": [
      {
        "type": "image"
        "image": {
          "id": 3
          "name": "logo"
          "url": "http://images.com/logo.jpg"
        }
      },
      {
        "type": "video"
        "video": {
          "id": 12
          "uid": "XCSSMDFWW"
          "source": "youtube"
        }
      }
    ]

```

However, that is a big assumption that it currently acted like this.  Really all this PR does was tack on one extra nest.  

Ill work on a test for both these as I think I know what the desired expectation is now .. thanks

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -46,7 +46,8 @@ def relationship_value_for(association, options)
 
         opts = instance_options.merge(include: @include_tree[association.key])
         relationship_value = Attributes.new(association.serializer, opts).serializable_hash(options)
-        if association.options[:polymorphic]
+
+        if association.options[:polymorphic] && relationship_value
```
cgmckeever: no nesting if empty relationship

cgmckeever: @beauby good call, added check for empty relationship which wont wrap the resultant in a polymorphic type if there is no association (corresponding tests added for :json and :attributes adapters)

test/adapter/polymorphic_test.rb
```diff
@@ -9,9 +9,19 @@ class PolymorphicTest < ActiveSupport::TestCase
...
+          @tag.object_tags << ObjectTag.new(id: 1, poly_tag_id: 1, taggable: @employee)
+          @tag.object_tags << ObjectTag.new(id: 5, poly_tag_id: 1, taggable: @picture)
+        end
+
+        def serialization(resource, adapter = :attribute)
+          serializable(resource, adapter: adapter, serializer: PolymorphicBelongsToSerializer).as_json
+        end
+
+        def tag_serialization(resource, adapter = :attribute)
+          serializable(resource, adapter: adapter, serializer: PolymorphicTagSerializer, include: '*.*').as_json
```
cgmckeever: nested serialization to traverse into the has_many polymorphs

test/adapter/polymorphic_test.rb
```diff
@@ -9,9 +9,19 @@ class PolymorphicTest < ActiveSupport::TestCase
           @picture = @employee.pictures.new(id: 1, title: 'headshot-1.jpg')
           @picture.imageable = @employee
 
-          @attributes_serialization = serializable(@picture, serializer: PolymorphicBelongsToSerializer) # uses default adapter: attributes
-          @json_serialization = serializable(@picture, adapter: :json, serializer: PolymorphicBelongsToSerializer)
-          @json_api_serialization = serializable(@picture, adapter: :json_api, serializer: PolymorphicBelongsToSerializer)
```
cgmckeever: pulled out into a helper method, as serialization variations grew by 2 more for empty relationship testing

test/fixtures/poro.rb
```diff
@@ -248,6 +259,22 @@ def maker
   has_one :imageable, serializer: PolymorphicHasManySerializer, polymorphic: true
 end
 
+PolymorphicSimpleSerializer = Class.new(ActiveModel::Serializer) do
+  attributes :id
+end
```
cgmckeever: simple serializer to demonstrate polymorph inclusions (as they didnt have their own)

test/fixtures/poro.rb
```diff
@@ -70,10 +70,21 @@ def cache_key
...
+  belongs_to :taggable, polymorphic: true
 end
 
 class Picture < ActiveRecord::Base
   belongs_to :imageable, polymorphic: true
+  has_many :object_tags, as: :taggable
+end
+
+class PolyTag < ActiveRecord::Base
+  has_many :object_tags
```
cgmckeever: There were other tests using a non-active record version of Tags.  Didnt want to do to much upheaval to that pattern so made a new Tag Class

test/adapter/polymorphic_test.rb
```diff
@@ -20,31 +30,117 @@ def test_attributes_serialization
...
+            {
+              id: 2,
+              title: 'headshot-2.jpg',
+              imageable: nil
+            }
+
+          assert_equal(expected, serialization(@simple_picture))
+        end
+
+        def test_attributes_serialization_with_polymorphic_has_many
```
cgmckeever: @beauby 'heterogeneous has-many" tests for :json and :attribute adapters added.  If my understanding was correct, all the plumbing of the Attributes class acted as expected, tests demonstrate the nesting (which I feel is in alignment with previous versions [https://github.com/rails-api/active_model_serializers/pull/1726#issuecomment-219323927])

beauby:
Oh, right. I had just skimmed through the PR and had not understood it fully. I'm :+1: with the idea.

remear:
This adds a `.DS_Store`. Would you please remove it.

cgmckeever:
@remear LOL, I just caught that as well, pushed it ..

PR #1728, opened 2016-05-16
adds mac files to .gitignore

cgmckeever:
#### Purpose

Macs like to leave breadcrumb files which get picked up by GIT.  This excludes them via .gitignore

#### Changes

update to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,13 @@ tmp
 .ruby-gemset
 vendor/bundle
 tags
+
+# silly macs
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+Icon?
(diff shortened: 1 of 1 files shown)


remear:
I solve this by listing such files in a global .gitignore. https://robots.thoughtbot.com/global-gitignore

NullVoxPopuli:
LGTM

beauby:
IMO this should not have been merged. This issue arose a few times already, and the answer has always been that `.gitignore` is for **project specific** ignored files. OS specific ignored files should be handled by the appropriate git configuration on the user's workstation.

NullVoxPopuli:
Well, it's easier for us to accept contributions if there is less for us to worry about

PR #1729, opened 2016-05-16
adds prompt for version script and a little more documentation

cgmckeever:
#### Purpose

Wanted to pause version tester script on failures, and allow to halt execution.

#### Changes

adds a read/condition to failure case

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,10 +74,15 @@ Run a single test
 `$ rake test TEST=path/to/test.rb TESTOPTS="--name=test_something"`
 
 Run tests against different Rails versions by setting the RAILS_VERSION variable
-and bundling gems.
+and bundling gems.  (save this script somewhere executable and run from top of AMS repository)
 
 ```bash
-for version in 4.0 4.1 4.2 master; do
+#!/usr/bin/env bash
+
+rcommand='puts YAML.load_file("./.travis.yml")["env"]["matrix"].join(" ").gsub("RAILS_VERSION=", "")'
+versions=$(ruby -ryaml -e "$rcommand")
+
+for version in ${versions[@]}; do
   export RAILS_VERSION="$version"
   rm -f Gemfile.lock
   bundle check || bundle --local || bundle
@@ -88,7 +93,12 @@ for version in 4.0 4.1 4.2 master; do
   else
     # red in ANSI
     echo -e "\033[31m **** Tests failed against Rails ${RAILS_VERSION} **** \033[0m"
-  fi
+    read -p '[Enter] any key to continue, [q] to quit...' prompt
+    if [ "$prompt" = 'q' ]; then
+      unset RAILS_VERSION
+      exit 1
+    fi
+fi
   unset RAILS_VERSION
 done
 ```


CONTRIBUTING.md
```diff
@@ -74,9 +74,11 @@ Run a single test
 `$ rake test TEST=path/to/test.rb TESTOPTS="--name=test_something"`
 
 Run tests against different Rails versions by setting the RAILS_VERSION variable
-and bundling gems.
+and bundling gems.  (save this script somewhere executable and run from top of AMS repository)
```
bf4: nice. I just wrote this as something to paste into the console.  

I think a better solution to recommend something more robust would be to recommend using [wwtd](https://github.com/grosser/wwtd) and just use the current Ruby `wwtd --local`

cgmckeever: Dont know much about WWTD .... I was originally gonna run it in the console, but found it simpler just to ``../ams_tester.sh`` and bail out when something crapped out 😩

bf4: for example, this script is now out of date since we don't run on 4.0 anymore.  wwtd just using the travis file as a source of truth.  you could, of course do something like `ruby -ryaml -e 'puts YAML.load_file("./.travis.yml")["env"]["matrix"]' | xargs` instead to make the list

NullVoxPopuli:
LGTM!

CONTRIBUTING.md
```diff
@@ -74,9 +74,11 @@ Run a single test
...
 
 ```bash
+#!/bin/bash
```
remear: Perhaps `#!/usr/bin/env bash` instead.

CONTRIBUTING.md
remear: wwtd sounds like a great idea. Also, pulling the versions from .travis.yml would be very nice.

remear:
Some of my thoughts on this...

My preference would be to report failures at the end by version as I like to fire off this script and work on something else until it's finished. Maybe wwtd could do this. The prompt requires someone to sit there and acknowledge the failures before moving on to the next version. I'm not sure I'd ever bail out at such a prompt since it's been helpful in the past to know in a single run of the script that something passes on some versions and not others.

CONTRIBUTING.md
```diff
@@ -74,10 +74,15 @@ Run a single test
...
 Run tests against different Rails versions by setting the RAILS_VERSION variable
-and bundling gems.
+and bundling gems.  (save this script somewhere executable and run from top of AMS repository)
 
 ```bash
-for version in 4.0 4.1 4.2 master; do
+#!/usr/bin/env bash
+
+rcommand='puts YAML.load_file("./.travis.yml")["env"]["matrix"].join(" ").gsub("RAILS_VERSION=", "")'
+versions=$(ruby -ryaml -e "$rcommand")
```
cgmckeever: @bf4 looking up versions based on your yml suggestion

CONTRIBUTING.md
```diff
@@ -88,7 +93,12 @@ for version in 4.0 4.1 4.2 master; do
   else
     # red in ANSI
     echo -e "\033[31m **** Tests failed against Rails ${RAILS_VERSION} **** \033[0m"
-  fi
+    read -p '[Enter] any key to continue, [q] to quit...' prompt
+    if [ "$prompt" = 'q' ]; then
+      unset RAILS_VERSION
+      exit 1
+    fi
```
cgmckeever: @remear totally understand your thoughts on leaving this auto-pilot.  My only counter argument is, that this script is really nice to get a new contributor going - and jumping on errors right away (for me) was nice.  As an experienced contributor, you probably have some more advanced tools in your belt and/or can just comment out L96 to skip the prompt.

CONTRIBUTING.md
bf4: >  (save this script somewhere executable and run from top of AMS repository)

can we make this even easier?

CONTRIBUTING.md
```diff
@@ -74,10 +74,15 @@ Run a single test
...
+and bundling gems.  (save this script somewhere executable and run from top of AMS repository)
 
 ```bash
-for version in 4.0 4.1 4.2 master; do
+#!/usr/bin/env bash
+
+rcommand='puts YAML.load_file("./.travis.yml")["env"]["matrix"].join(" ").gsub("RAILS_VERSION=", "")'
+versions=$(ruby -ryaml -e "$rcommand")
+
+for version in ${versions[@]}; do
```
bf4: @cgmckeever I am envious of your bash skills.  I've never seen that and it is awesome

CONTRIBUTING.md
bf4: agreed. but if we make it too nice, then it should just go in `bin`, right?

cgmckeever: I thought about this, but wasnt sure about a script floating around a gem, so I just left it in the docs

Issue #1719, opened 2016-05-07
Dynamic serializer selection


KevinColemanInc:
Currently the serializer param on `has_x` takes an object.

It would be cool if it could accept a method name as well.

Use case:

Dynamically setting the serializer for polymorphic relationships:

    has_one :user, polymorphic: true, serializer: :serializer_picker
 
    def serializer_picker
       if user.class == 'Admin'
         SparseAdminSerializer
       end
     end

cgmckeever:
cant you just do this as an attribute then and make a method for said attribute?

```
class SomeSerializer < ApplicationSerializer
  attributes :id, :somefakerelationship

  def somefakerelationship
     ## do stuff
    ## profit?
  end

```

KevinColemanInc:
@cgmckeever true, but I want to specify the serializer of the thing that spawned. For my particular use case, I am have a model called "Notifications"  It has a `source_obj` polymorphic relationship on it such that many different models can be a `source_obj`.

The default model's serializer is often too big for what I want.  So I want to the source_obj to have use special "sparse" serializers for each object. e.g. sparse_user_serializer, sparse_profile_serializer, etc. etc. instead of the default model serializer of user_serializer or profile_serializer.

does that make sense?

I think my hack to fix this will be to do:

    class SomeSerializer < ApplicationSerializer
      attributes :id, :source_obj

      def source_obj
        object.source_obj.to_json
      end
    end

so I am at least not accidently getting any N+1s

groyoh:
@KevinColemanInc use this:

```ruby
class MySerializer < ActiveModel::Serializer
  def self.serializer_for(model, options)
    return SparseAdminSerializer if model.class == 'Admin'
    super
  end

  # the rest of the serializer
end
```

cgmckeever:
@groyoh NICE!

KevinColemanInc:
Cool! I will use that.  I was hoping to be able to something like this:

beauby:
We should document this. PR anyone?

PR #1730, opened 2016-05-16
Adds documentation for overriding default serializer based on conditions

cgmckeever:
#### Purpose

Conversation on 'dynamic serializers' spawned an interesting solution via @groyoh 
Just documenting it here as its a 💯 #protip  

#### Changes

Updated serializer docs

#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/issues/1719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Fixes:
 
 Misc:
 - [#1673](https://github.com/rails-api/active_model_serializers/pull/1673) Adds "How to" guide on using AMS with POROs (@DrSayre)
+- [#1730](https://github.com/rails-api/active_model_serializers/pull/1730) Adds documentation for overriding default serializer based on conditions (@groyoh/@cgmckeever)
 
 ### [v0.10.0.rc5 (2016-04-04)](https://github.com/rails-api/active_model_serializers/compare/v0.10.0.rc4...v0.10.0.rc5)
 
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -369,3 +369,19 @@ class PostSerializer < ActiveModel::Serializer
   end
 end
 ```
+
+## Overriding association serializer lookup
+
+If you want to define a specific serializer lookup for your associations, you can override
+the `ActiveModel::Serializer.serializer_for` method to return a serializer class based on defined conditions.
+
+```ruby
+class MySerializer < ActiveModel::Serializer
+  def self.serializer_for(model, options)
+    return SparseAdminSerializer if model.class == 'Admin'
+    super
+  end
+
+  # the rest of the serializer
+end
+```


docs/general/serializers.md
```diff
@@ -369,3 +369,19 @@ class PostSerializer < ActiveModel::Serializer
   end
 end
 ```
+
+## Overriding default serializer
+
+You can use the `serializer_for` method to return a different serializer based on
+use/scope specific conditions.
```
cgmckeever: wording welcome!

docs/general/serializers.md
```diff
@@ -369,3 +369,19 @@ class PostSerializer < ActiveModel::Serializer
   end
 end
 ```
+
+## Overriding default serializer
```
groyoh: I'd rather change it to `## Overriding association serializer lookup` as this should only works for the relationships of this specific serializer.

docs/general/serializers.md
```diff
@@ -369,3 +369,19 @@ class PostSerializer < ActiveModel::Serializer
...
+
+You can use the `serializer_for` method to return a different serializer based on
```
groyoh: Can you specify that the `serializer_for` is a class method? What do you think of something like:

```diff
+If you want to define a specific serializer lookup for your associations, you can override the `ActiveModel::Serializer.serializer_for` method to return a serializer class based on the model .
```

?

groyoh:
@cgmckeever thanks for this PR! :100:

docs/general/serializers.md
```diff
@@ -369,3 +369,19 @@ class PostSerializer < ActiveModel::Serializer
...
+
+```ruby
+class MySerializer < ActiveModel::Serializer
+  def self.serializer_for(model, options)
+    return SparseAdminSerializer if model.class == 'Admin'
+    super
+  end
+
+  # the rest of the serializer
+end
```
cgmckeever: @groyoh much better suggestion - updated

NullVoxPopuli:
This is super handy to document!

NullVoxPopuli:
LGTM

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Fixes:
 
 Misc:
 - [#1673](https://github.com/rails-api/active_model_serializers/pull/1673) Adds "How to" guide on using AMS with POROs (@DrSayre)
+- [#1730](https://github.com/rails-api/active_model_serializers/pull/1730) Adds documentation for overriding default serializer based on conditions (@groyoh/@cgmckeever)
```
cgmckeever: @NullVoxPopuli added

NullVoxPopuli: wooooo

KevinColemanInc:
@groyoh hey! so I don't think this documentation is correct.  I tried implementing this and it doesn't seem to have the desired effect.

    class MySerializer < ActiveModel::Serializer
      def self.serializer_for(model, options)

Override the serializer_for method on MySerializer doesn't do anything, because the code always uses the definition of `serializer_for ` in SerializableResource not whatever model itself is referencing.
https://github.com/rails-api/active_model_serializers/blob/aad7779a3fd918bed5c17438c4626883d6fe6c5f/lib/active_model_serializers/serializable_resource.rb#L57

So I think the way it is currently structured, it is impossible to have custom serializers for polymorphic associations in a given parent serializer.

any thoughts?

PR #1733, opened 2016-05-17
Handle situations when serializer is initialized with a nil object.

arkiver:
#### Purpose

Gracefully handle situations when serializer is accidentally/unknowingly
initialized with a nil object.

#### Changes

1. A simple IF that checks if object exists before calling a method on it.
2. Line breaks for code readability

This case may come up when using with the airblade/paper_trail gem.

For example:

employee = Employee.first.version_at(date) # version may or may not exist on that date.
EmployeeSerializer.new(employee).attributes

fails with undefined method ```read_attribute_for_serialization``` for Nil class.

### DIFF
--- a/test/serializers/read_attribute_for_serialization_test.rb
+++ b/test/serializers/read_attribute_for_serialization_test.rb
@@ -7,9 +7,11 @@ module ActiveModel
       class Parent < ActiveModelSerializers::Model
         attr_accessor :id
       end
+
       class Child < Parent
         attr_accessor :name
       end
+
       class ParentSerializer < ActiveModel::Serializer
         attributes :$id
 
@@ -17,8 +19,14 @@ module ActiveModel
           object.id
         end
       end
+
       class ChildSerializer < ParentSerializer
         attributes :name
+      end
+
+      def test_for_serializer_initialized_with_nil_object
+        child = ChildSerializer.new(nil)
+        assert_equal nil, child.read_attribute_for_serialization(:name)
       end
 
       def test_child_serializer_calls_dynamic_method_in_parent_serializer
@@ -32,6 +40,7 @@ module ActiveModel
       class ErrorResponse < ActiveModelSerializers::Model
         attr_accessor :error
       end
+
       class ApplicationSerializer < ActiveModel::Serializer
         attributes :status
 
@@ -39,9 +48,11 @@ module ActiveModel
           object.try(:errors).blank? && object.try(:error).blank?
         end
       end
+
       class ErrorResponseSerializer < ApplicationSerializer
         attributes :error
       end
+
       class ErrorResponseWithSuperSerializer < ApplicationSerializer
         attributes :error
 
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -180,7 +180,7 @@ module ActiveModel
         send(attr)
       elsif self.class._fragmented
         self.class._fragmented.read_attribute_for_serialization(attr)
-      else
+      elsif object
         object.read_attribute_for_serialization(attr)
       end
     end


test/serializers/read_attribute_for_serialization_test.rb
```diff
@@ -7,20 +7,28 @@ class ReadAttributeForSerializationTest < ActiveSupport::TestCase
...
         end
       end
+
       class ChildSerializer < ParentSerializer
         attributes :name
       end
 
+      def test_for_serializer_initialized_with_nil_object
+        child = ChildSerializer.new(nil)
+        assert_equal nil, child.read_attribute_for_serialization(:name)
```
bf4: But this is a bug in your code you should fix, not paper over, no?

arkiver: Agreed. But it gives me a trace from the gem instead of my app. So maybe in the initializer, we could raise an exception if somebody attempts to initialize with a nil because the code expects the object passed in to be not nil. I didn't want to mess with ```def initialize``` but this would be an additional check, maybe?

cgmckeever: in ``def initialize`` why not raise an error if object is an ``ActiveRecord``, as the code will still barph if object is anything other ?

arkiver: Yes that would be a better approach.

NullVoxPopuli: > why not raise an error if object isnt an ActiveRecord

Cause we support searrializing poros. We'd have to check if the object responds to the 3 or 4 methods we require.

cgmckeever: truth.  Could check for those.  

I always have internal conflicts if code should be more defensive at the lowest point (in this case returning ``nil`` if ``read_attribute_for_serialization`` cant do its thing, or if the higher up code should know not to try to serialize.

beauby: The problem with "silently failing" is that it makes misbehaving code go unnoticed and debugging very hard. When stuff fails, you know there's an issue with your code, and the sooner it fails, the easier it is to fix it.

bf4: This is also super easy to check for in userland code.  I'd consider adding a nil check in initialize that raises an exception earlier, but honestly, there's no reason that should ever be happening.  When serializing a nil association we skip it, rather than blow up. And if you pass nil to render json: it will skip using a serializer as well.
 That is, the exception orginates in the example scenario when a nil is knowingly being passed in.  Raising an exception in initialize instead of attritbutes doesn't change that.

> For example:

> employee = Employee.first.version_at(date) # version may or may not exist on that date.
EmployeeSerializer.new(employee).attributes

If you think this still merits a code change in AMS, would you mind providing a different example?

B mobile phone

arkiver: For my particular case, I was using it like so: 
```EmployeeSerializer.new(employee.version_at(date)).attributes```
I half expected that if the passed in object is nil, I would be returned a nil or {} or something similar. Calling further methods on nil or {}, etc should raise exceptions in my own code that I am very familiar with and I won't need to look at the gem's code.

Another scenario off the top of my head would be something like 
```EmployeeSerializer.new(get_employee_for_serializing).attributes```

where ```get_employee_for_serializing``` may be a method that may return a nil when you do something like a #find or Model.where(conditions).first. I can see such helper methods being used in controllers rendering json via serializers.

I realize that adding a nil check would be very easy in user's code. But IMHO I should put these in after realizing that I cannot pass in nil to a serializer after seeing a TypeError.

bf4: If you passed an account into a user serializer you would also get an error. We can't keep you from doing that

B mobile phone

PR #1741, opened 2016-05-19
Encourage pinning of pre-1.0.0 release [skip ci]

krisleech:
#### Purpose

Encourage the pinning of pre-1.0.0 release of this gem.

0.10 introduced breaking changes (notably, no root node is included in the JSON). This broke our build. Since semver allows pre-1.0.0 to introduce breaking changes this is okay. However since the gem is 5 years old but has no 1.0 release maybe we should encourage the pinning of the gem in the README.

#### Changes

Documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ It is generally safe and recommended to use the master branch.
 Add this line to your application's Gemfile:
 
 ```
-gem 'active_model_serializers'
+gem 'active_model_serializers', '~> 0.10.0'
 ```
 
 And then execute:


NullVoxPopuli:
I'm not opposed to that. It's a good habit to have if people want to be able to do `bundle update`

@bf4, thoughts?

README.md
```diff
@@ -56,7 +56,7 @@ It is generally safe and recommended to use the master branch.
 Add this line to your application's Gemfile:
 
 ```
-gem 'active_model_serializers'
+gem 'active_model_serializers', '~> 0.10.0'
```
bf4: seems like a good idea. There's been discussion of doing something similar to what Rake and Node.js have done and jump to `10.0` (I'm for it), but it hasn't been resolved yet.  pre-1.0 semver is weird.

@krisleech Thanks for sharing this PR.  Anything else you might contribute back to help out others is much appreciated!

PR #1746, opened 2016-05-23
Clarify that only 4.0 is dropped

sandstrom:


remear:
I'm not sure this really needs further clarification. It says `Drop support for Rails 4.0`. It means what it says, `4.0`, not `4.1`, not `4.2`, not `4.x`.

sandstrom:
Okay, probably just me then 😄

PR #1750, opened 2016-05-25
[WIP] fix belongs_to causing unnecessary db hits

NullVoxPopuli:
#### Purpose

we only need id and type for belongs to relationships, and we can optimize render time by not touching objects at the other end of the belongs_to
#### Changes

Just notes for now. So others can see where the problem is. We're going to have to tweak how we build the relationship reflection objects, or maybe make all calls to the reflection lazy. idk.

I feel like in order to do this properly, we are going to need to more directly tie in to ActiveRecord.
as @beauby said in #1100: 

>  What happens if we have belongs_to :author in the serializer, and belongs_to :author, class_name: "User" in the model? Then I fear AMS wouldn't know how to set the type property of the Resource Identifier.

So.. not sure if the team wants to have that much coupling with AR. I think its needed in order to properly address this issue though. 
#### Related GitHub issues
#762
#1100 (failing test by @ssendev)
#### TODO before merging
- [ ] Remove binding.pry
- [ ] Actually fix the issue

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -56,6 +56,11 @@ module ActiveModel
         :nil
       end
 
+      # This gets the value of the reflection.
+      # this needs to be avoided if we have a belongs_to relationship
+      # because we only care about the id and type, and not about the object.
+      # This causes n+1 queries unnecessarily
+      #
       # @param serializer [ActiveModel::Serializer]
       # @yield [ActiveModel::Serializer]
       # @return [:nil, associated resource or resource collection]
@@ -70,6 +75,9 @@ module ActiveModel
       #     end
       #   end
       def value(serializer)
+        require 'pry-byebug'
+        binding.pry
+
         @object = serializer.object
         @scope = serializer.scope
 
@@ -78,9 +86,15 @@ module ActiveModel
           if block_value != :nil
             block_value
           elsif @_include_data
+            # this is what needs to be avoided in the case of belongs_to.
+            # should an id and type hash just be returned?
+            # maybe how associations work in general needs to be re-looked at
             serializer.read_attribute_for_serialization(name)
           end
         else
+          # this is what needs to be avoided in the case of belongs_to.
+          # should an id and type hash just be returned?
+          # maybe how associations work in general needs to be re-looked at
           serializer.read_attribute_for_serialization(name)
         end
       end
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -126,6 +126,21 @@ module ActiveModel
         assert expected_association_keys.include? :site
       end
 
+      class BelongsToTestPostSerializer < ActiveModel::Serializer
+        belongs_to :blog
+      end
+
+      test 'belongs_to should not load record' do
+        post = Post.new
+        post.blog_id = 'blog'
+        post.blog = 'should not touch this'
+
+        actual = serializable(post, adapter: :json_api, serializer: BelongsToTestPostSerializer).as_json
+        expected = { data: { id: 'post', type: 'posts', relationships: { blog: { data: { id: 'blog', type: 'blogs' } } } } }
+
+        assert_equal expected, actual
+      end
+
       class InlineAssociationTestPostSerializer < ActiveModel::Serializer
         has_many :comments
         has_many :comments, key: :last_comments do
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,7 @@ group :bench do
 end
 
 group :test do
+  gem 'pry-byebug'
   gem 'sqlite3',                          platform: (@windows_platforms + [:ruby])
   gem 'activerecord-jdbcsqlite3-adapter', platform: :jruby
   gem 'codeclimate-test-reporter', require: false
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -87,10 +87,14 @@ module ActiveModel
         return unless object
 
         Enumerator.new do |y|
+          # _reflections is an AMS::BelongsToReflection or
+          # an AMS::HasManyReflection
           self.class._reflections.each do |reflection|
             next if reflection.excluded?(self)
             key = reflection.options.fetch(:key, reflection.name)
             next unless include_tree.key?(key)
+            # until this is called, the reflection / association
+            # is _not_ evaluated
             y.yield reflection.build_association(self, instance_options)
           end
         end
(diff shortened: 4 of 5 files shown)


test/serializers/associations_test.rb
```diff
@@ -126,6 +126,21 @@ def test_associations_custom_keys
         assert expected_association_keys.include? :site
       end
 
+      class BelongsToTestPostSerializer < ActiveModel::Serializer
+        belongs_to :blog
+      end
+
+      test 'belongs_to should not load record' do
```
NullVoxPopuli: test courtesy of @ssendev

sandstrom:
This would be a godsend. 🌟  Currently using AMS and need manual work-arounds to avoid these costly DB-lookups.

I agree with using reflection to inspect the relationships of the model. Specifying a type manually would be rare and only needed for polymorphic relationships.

I don't use ActiveRecord, so not 100% sure but I think `Author.reflect_on_association(:books).klass` works[1]. Mongoid, another popular ORM, also has good reflection support.

Would be great if the decided-upon solution would support both Mongoid and ActiveRecord.

[1] https://github.com/rails/rails/blob/master/activerecord/lib/active_record/reflection.rb

NullVoxPopuli:
I guess what we _could_ do is maybe have support for looking up custom types / polymorphic types via ActiveRecord if that technique is available, and then also provide a way to override the serialized type name per relationship in the serializer?

sandstrom:
@NullVoxPopuli If it's easy to modify I think that's alright.

Just a short thrown-together snippet to illustrate my thought. Basically I'll try reflection for AR + it's public and overrideable so that someone using Mongoid can easily adjust in their `ApplicationSerializer`.

```ruby
# public
# override to use your own method, e.g. for Mongoid
def retrieve_relation_id
  if defined?(ActiveRecord)
    # Author.reflect_on_association(:books).klass
    # ... something else using AR
  else
    # fallback to the current, naive solution
    record.relation.id
  end
end
```

lib/active_model/serializer/reflection.rb
```diff
@@ -78,9 +86,15 @@ def value(serializer)
           if block_value != :nil
```
bf4: in JSONAPI::Resources, the relationship object has a `foreign_key_on` option that may be `self`. 

[docs](https://github.com/cerebris/jsonapi-resources/blob/f4ab5683de49c2a506eea14096bd81a0ec85e94e/README.md#options)

> The relationship methods (`relationship`, `has_one`, and `has_many`) support the following options:
>  * `foreign_key` - the method on the resource used to fetch the related resource. Defaults to `<resource_name>_id` for has_one and `<resource_name>_ids` for has_many relationships.
> `to_one` relationships support the additional option:
>  * `foreign_key_on` - defaults to `:self`. To indicate that the foreign key is on the related resource specify `:related`.
> Examples:
> ```ruby
> class ExpenseEntryResource < JSONAPI::Resource
>   has_one :currency, class_name: 'Currency', foreign_key: 'currency_code'
> end
> ```


 see https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/relationship.rb for full code

```ruby
module JSONAPI
  class Relationship
    attr_reader :foreign_key

    def initialize(name, options = {})
      @foreign_key = options[:foreign_key] ? options[:foreign_key].to_sym : nil
    end

    def belongs_to?
      false
    end

    class ToOne < Relationship
      attr_reader :foreign_key_on

      def initialize(name, options = {})
        super
        @foreign_key ||= "#{name}_id".to_sym
        @foreign_key_on = options.fetch(:foreign_key_on, :self)
      end

      def belongs_to?
        foreign_key_on == :self
      end

    class ToMany < Relationship
      def initialize(name, options = {})
        super
        @foreign_key ||= "#{name.to_s.singularize}_ids".to_sym
      end
    end
  end
end
```

NullVoxPopuli: So, I'm just trying to get a handle on how all this existing reflection code works.
@bf4, can you look over this? I want to make sure I know what's happening.

It looks like the most problematic area is in reflection.rb#build_association

```ruby

        # evaluates the relationship
        # - this is only needed if we want to use 'includes', otherwise, we only need ids.
        association_value = value(subject)

        # we only need a serializer if we are using includes
        # without includes, all we need are the id and type
        serializer_class = subject.class.serializer_for(association_value, reflection_options)
        reflection_options[:include_data] = @_include_data

        # this would cover the case of includes not being used for this association
        if serializer_class
          begin
              association_value,
              serializer_options(subject, parent_serializer_options, reflection_options)
            )
            reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
          end
        elsif !association_value.nil? && !association_value.instance_of?(Object)
      
          # would this just be a hash of id / type?
          # maybe the relationship objects should have id/type accessors?
          # has many would just have an array of ids?
          reflection_options[:virtual_value] = association_value
        end

        # this is just for rendering
        Association.new(name, serializer, reflection_options, @_links, @_meta)
      end
```

if I understand things correctly, it's really just the `value` method and how the association/reflection is represented that needs to change?

richmolj: @NullVoxPopuli I know @bf4 knows this area better than I do, but wanted to address what I think one key problematic area is, and how it relates to https://github.com/rails-api/active_model_serializers/pull/1720:

```ruby
# evaluates the relationship
# - this is only needed if we want to use 'includes', otherwise, we only need ids.
association_value = value(subject)

# we only need a serializer if we are using includes
# without includes, all we need are the id and type
```

The issue I see here is sometimes you want a `link`, not `id/type` combo. In this case:

* We need to fire `#value` to A) see if we want `id/type` in the response (`include_data`) and B) get the actual links data.
* This will run the association block
* If you have an in-line association, you're now firing the unnecessary SQL again.

In code:

```ruby
belongs_to :author do
  link :related do
    href "/foo/bar"
  end

  # we only want to run this code if :author is included, but instead it
  # has to run in any case to get the link href above
  Author.first || Author.default
end
```

> I guess what we could do is maybe have support for looking up custom types / polymorphic types via ActiveRecord if that is available, and then also provide a way to override the serialized type name per relationship in the serializer?

Unfortunately I think this would cause the same inline-association issue. Just for the sake of argument, let's say the solution was a `type` line, you'd have something like this:

```ruby
belongs_to :author do
  type 'users'

  Author.first || Author.default
end
```

In order to get the `type` value, you need to run the block, which is going to fire the unnecessary SQL again.

This is the why https://github.com/rails-api/active_model_serializers/pull/1720 changes inline-associations to be lazily evaluated by wrapping the data loading in a proc:

```ruby
belongs_to :author do
  type 'users'
  
  # we can now conditionally/lazily fire this code
  load_data { Author.first || Author.default }
end
```

NullVoxPopuli:
@beauby do your serializers account for this? or would this be a feature of `jsonapi-rails`?

beauby:
@NullVoxPopuli They do as long as a link is defined for the corresponding relationship.

NullVoxPopuli:
Closing for now.

ActiveRecord performance should be addressed in a more up to date branch (this branch just adds some in-code docs)

PR #1752, opened 2016-05-26
Remove default key case change message

remear:
#### Purpose

Removes the post install message that warned the default key case for the `json_api` adapter had changed.

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -57,10 +57,4 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'grape', ['>= 0.13', '< 1.0']
   spec.add_development_dependency 'json_schema'
   spec.add_development_dependency 'rake', ['>= 10.0', '< 12.0']
-
-  spec.post_install_message = <<-EOF
-NOTE: The default key case for the JsonApi adapter has changed to dashed.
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
LGTM

active_model_serializers.gemspec
```diff
@@ -57,10 +57,4 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'grape', ['>= 0.13', '< 1.0']
   spec.add_development_dependency 'json_schema'
   spec.add_development_dependency 'rake', ['>= 10.0', '< 12.0']
-
-  spec.post_install_message = <<-EOF
-NOTE: The default key case for the JsonApi adapter has changed to dashed.
-See https://github.com/rails-api/active_model_serializers/blob/master/docs/general/key_transforms.md
-for more information on configuring this behavior.
```
bf4: might want to include that it's a performance hit, and to use `ActiveModelSerializers.config.key_transform = :unaltered` if you don't need it :)

NullVoxPopuli: does that need to be in the post install message though? or our docs?

remear: docs, IMO.

PR #1753, opened 2016-05-26
Add key transform performance note to docs

remear:
#### Purpose

Note in the docs that key transforms can be expensive and using `:unaltered` will provide a performance boost.


NullVoxPopuli:
why do we have twice as many CI checks than normal?

Issue #1442, opened 2016-01-18
[RFC] Automatic serializer lookup


beauby:
# TL;DR
What should be the lookup mechanism when inferring a serializer for a resource?
# Current situation
Work has been done to get us to a more flexible situation in this regard.
Currently, for a resource `CoolNamespace::CoolResource`, AMS will look for a `CoolResourceSerializer` in the following namespaces:
+ If it is being serialized as an association of some other resource using `OtherNamespace::OtherResourceSerializer`, first look for `OtherNamespace::OtherResourceSerializer::CoolResourceSerializer` (this is designed so that defining single-purpose serializers for associations is less cumbersome),
+ otherwise, look for `CoolNamespace::CoolResourceSerializer` (this is kind of the "normal" lookup),
+ if none is found, recursively do the same for the superclass of `CoolResource`.
Other possible places we might want to look at:
+ the calling controller's namespace (#1436), i.e. when doing `render json: @posts` inside `Api::V1::PostsController#index`, look for `Api::V1::PostSerializer`,
+ *all* of the calling controller's prefix namespaces (in the previous case it would be `::Api::V1`, `::Api`, and `::`,
+ the parent serializer's namespace, i.e. when serializing `CoolResource` as an association through `OtherNamespace::OtherResourceSerializer`, look for `OtherNamespace::CoolResourceSerializer` (so that when you do `render @post, serializer: Experimental::CoolFeature::PostSerializer`, the post's author can be inferred to be `Experimental::CoolFeature::AuthorSerializer`),
+ a user-specified list of namespaces,
+ possibly many other things.

Please chip in with your ideas, use-cases and strong opinions!

beauby:
(cc @bf4, @joaomdmoura)

bf4:
Change to pr?

bf4:
Mention this is controller by global option or adapter:nil

bf4:
Mention 0.9 behavior w serializer_nanespace option to limit the cases ams had to handle. Could ve an option to serialization context

natematykiewicz:
I believe this is related to my ticket: #1337 

[...]

Issue #1745, opened 2016-05-22
Empty collection rendering with incorrect root


mockdeep:
#### Expected behavior vs actual behavior

When I upgraded to AMS 0.10.0, a started seeing this issue. When an empty collection is rendered within a controller, the wrong root key is transmitted to the client side. I get:

```
{"":[]}
```

when I expect:

```
{"tasks":[]}
```

I have a tasks controller that looks like the following. In this case it's returning an empty array, not an ActiveRecord::Relation object:

```ruby
class TasksController < ApplicationController
  def index
    respond_to do |format|
      format.json { render json: current_user.undone_and_pending_tasks }
    end
  end
end
```

vladra:
Having very similar error.

When I render an array of integers, automatic serializer which is used is `Rendered ActiveModel::Serializer::CollectionSerializer with Array`

But when I render empty array, it is selecting wrong serializer: `Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModelSerializers::Adapter::Json`

And this makes output `{"": []}` instead of `[]`

beauby:
@mockdeep If your `undone_and_pending_tasks` method returns an empty array, then your `render json: current_user.undone_and_pending_tasks` is essentially the same as `render json: []`, for which AMS has no way of guessing the name you want for the root.
The recommended way is to have your method return an AR collection. You could also explicitly set the root by doing `render json: [], root: "posts"`.

mockdeep:
@beauby This seems to work as expected on version 0.9.5, though. I'm not sure why, but does it not infer the type from the controller?

beauby:
No currently it does not, although it would be doable via `serialization_context`. Wanna make a PR for this?

mockdeep:
@beauby is that something that is done in 0.9.5 but not in 0.10.x? Not sure if I'll get the time to work on it, but if there's a pretty straightforward path I could take a look.

beauby:
I'm not very familiar with the 0.9 codebase (0.10 was a total rewrite based off 0.8), so I can't really say. I have a PR on my list to move the `root` stuff from the serializer to the adapter (where it belongs). I'll try to get that done tonight, then I can outline to you how to do the controller-based inference based on the new code.

bf4:
Also see https://github.com/rails-api/active_model_serializers/issues/1756

groyoh:
[...]

PR #1757, opened 2016-05-30
configurable serializer_lookup option

NullVoxPopuli:
#### Purpose

people need to be able to customize how the serializer look up is performed. Especially when it comes to API versioning, as there are quite a few ways to do serializer lookup and determining what way to map api version to serializers.
#### Changes

Adds a config option for adding additional serializer lookups.
#### Caveats

When configuring this, the dev can easily break everything.
#### Related GitHub issues
- #1442
- #886

Things left to do:
- [x] Tests
   - [x] Test for changing the Proc to have a different name matching scheme
   - [x] Test to demonstrate taking the version from a request header
- [x] Benchmarks vs `master`
- [x] Documentation

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -48,34 +48,34 @@ require 'coverage'
 Coverage.start
 
 ## ADD SOME CUSTOM REPORTING AT EXIT
-SimpleCov.at_exit do
-  next if $! and not ($!.kind_of? SystemExit and $!.success?)
-
-  header = "#{'*' * 20} SimpleCov Results #{'*' * 20}"
-  results = SimpleCov.result.format!.join("\n")
-  exit_message = <<-EOF
-
-#{header}
-{{RESULTS}}
-{{FAILURE_MESSAGE}}
-
-#{'*' * header.size}
-  EOF
-  percent = Float(SimpleCov.result.covered_percent)
-  if percent < @minimum_coverage
-    failure_message = <<-EOF
-Spec coverage was not high enough: #{percent.round(2)}% is < #{@minimum_coverage}%
-    EOF
-    exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', failure_message)
-    @output.puts exit_message
-    abort(failure_message) if @generate_report
-  elsif @running_ci
-    exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', <<-EOF)
-Nice job! Spec coverage (#{percent.round(2)}%) is still at or above #{@minimum_coverage}%
-    EOF
-    @output.puts exit_message
-  end
-end
+# SimpleCov.at_exit do
+#   next if $! and not ($!.kind_of? SystemExit and $!.success?)
+#
+#   header = "#{'*' * 20} SimpleCov Results #{'*' * 20}"
+#   results = SimpleCov.result.format!.join("\n")
+#   exit_message = <<-EOF
+#
+# #{header}
+# {{RESULTS}}
+# {{FAILURE_MESSAGE}}
+#
+# #{'*' * header.size}
+#   EOF
+#   percent = Float(SimpleCov.result.covered_percent)
+#   if percent < @minimum_coverage
+#     failure_message = <<-EOF
+# Spec coverage was not high enough: #{percent.round(2)}% is < #{@minimum_coverage}%
+#     EOF
+#     exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', failure_message)
+#     @output.puts exit_message
+#     abort(failure_message) if @generate_report
+#   elsif @running_ci
+#     exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', <<-EOF)
+# Nice job! Spec coverage (#{percent.round(2)}%) is still at or above #{@minimum_coverage}%
+#     EOF
+#     @output.puts exit_message
+#   end
+# end
 
 ## CAPTURE CONFIG IN CLOSURE 'AppCoverage.start'
 ## to defer running until test/test_helper.rb is loaded.
@@ -105,6 +105,7 @@ end.new do
   else
     formatters = []
   end
-  SimpleCov.formatters = formatters
+  # Using default formatters...
+  # SimpleCov.formatters = formatters
 end
 # rubocop:enable Style/MultilineBlockChain
--- a/.simplecov
+++ b/.simplecov
@@ -48,34 +48,34 @@ require 'coverage'
 Coverage.start
 
 ## ADD SOME CUSTOM REPORTING AT EXIT
-# SimpleCov.at_exit do
-#   next if $! and not ($!.kind_of? SystemExit and $!.success?)
-#
-#   header = "#{'*' * 20} SimpleCov Results #{'*' * 20}"
-#   results = SimpleCov.result.format!.join("\n")
-#   exit_message = <<-EOF
-#
-# #{header}
-# {{RESULTS}}
-# {{FAILURE_MESSAGE}}
-#
-# #{'*' * header.size}
-#   EOF
-#   percent = Float(SimpleCov.result.covered_percent)
-#   if percent < @minimum_coverage
-#     failure_message = <<-EOF
-# Spec coverage was not high enough: #{percent.round(2)}% is < #{@minimum_coverage}%
-#     EOF
-#     exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', failure_message)
-#     @output.puts exit_message
-#     abort(failure_message) if @generate_report
-#   elsif @running_ci
-#     exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', <<-EOF)
-# Nice job! Spec coverage (#{percent.round(2)}%) is still at or above #{@minimum_coverage}%
-#     EOF
-#     @output.puts exit_message
-#   end
-# end
+SimpleCov.at_exit do
+  next if $! and not ($!.kind_of? SystemExit and $!.success?)
+
+  header = "#{'*' * 20} SimpleCov Results #{'*' * 20}"
+  results = SimpleCov.result.format!.join("\n")
+  exit_message = <<-EOF
+
+#{header}
+{{RESULTS}}
+{{FAILURE_MESSAGE}}
+
+#{'*' * header.size}
+  EOF
+  percent = Float(SimpleCov.result.covered_percent)
+  if percent < @minimum_coverage
+    failure_message = <<-EOF
+Spec coverage was not high enough: #{percent.round(2)}% is < #{@minimum_coverage}%
+    EOF
+    exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', failure_message)
+    @output.puts exit_message
+    abort(failure_message) if @generate_report
+  elsif @running_ci
+    exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', <<-EOF)
+Nice job! Spec coverage (#{percent.round(2)}%) is still at or above #{@minimum_coverage}%
+    EOF
+    @output.puts exit_message
+  end
+end
 
 ## CAPTURE CONFIG IN CLOSURE 'AppCoverage.start'
 ## to defer running until test/test_helper.rb is loaded.
@@ -105,7 +105,6 @@ end.new do
   else
     formatters = []
   end
-  # Using default formatters...
-  # SimpleCov.formatters = formatters
+  SimpleCov.formatters = formatters
 end
 # rubocop:enable Style/MultilineBlockChain
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -82,6 +82,12 @@ which is shorthand for
 
 Each of the array entries represent a proc. A serializer lookup proc will be yielded 3 arguments. `resource_class`, `serializer_class`, and `namespace`.
 
+Note that:
+ - `resource_class` is the class of the resource being rendered
+ - by default `serializer_class` is `ActiveModel::Serializer`
+   - for association lookup it's the "parent" serializer
+ - `namespace` correspond to either the controller namespace or the [optionally] specified [namespace render option](./rendering.md#namespace)
+
 An example config could be:
 
 ```ruby
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -60,6 +60,50 @@ application, setting `config.key_transform` to `:unaltered` will provide a perfo
 What relationships to serialize by default.  Default: `'*'`, which includes one level of related
 objects. See [includes](adapters.md#included) for more info.
 
+
+##### serializer_lookup_chain
+
+Configures how serializers are searched for. By default, the lookup chain is
+
+```ruby
+ActiveModelSerializers::LookupChain::DEFAULT
+```
+
+which is shorthand for
+
+```ruby
+[
+  ActiveModelSerializers::LookupChain::BY_PARENT_SERIALIZER,
+  ActiveModelSerializers::LookupChain::BY_NAMESPACE,
+  ActiveModelSerializers::LookupChain::BY_RESOURCE_NAMESPACE,
+  ActiveModelSerializers::LookupChain::BY_RESOURCE
+]
+```
+
+Each of the array entries represent a proc. A serializer lookup proc will be yielded 3 arguments. `resource_class`, `serializer_class`, and `namespace`.
+
+An example config could be:
+
+```ruby
+ActiveModelSerializers.config.serializer_lookup_chain = [
+  lambda do |resource_class, serializer_class, namespace|
+    "API::#{namespace}::#{resource_class}"
+  end
+]
+```
+
+If you simply want to add to the existing lookup_chain. Use `unshift`.
+
+```ruby
+ActiveModelSerializers.config.serializer_lookup_chain.unshift(
+  lambda do |resource_class, serializer_class, namespace|
+    # ...
+  end
+)
+```
+
+See [lookup_chain.rb](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/lookup_chain.rb) for further explanations and examples.
+
 ## JSON API
 
 ##### jsonapi_resource_type
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -60,17 +60,10 @@ module ActiveModel
 
     # @api private
     def self.serializer_lookup_chain_for(klass, namespace = nil)
-      chain = []
-
-      resource_class_name = klass.name.demodulize
-      resource_namespace = klass.name.deconstantize
-      serializer_class_name = "#{resource_class_name}Serializer"
-
-      chain.push("#{namespace}::#{serializer_class_name}") if namespace
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
-
-      chain
+      lookups = ActiveModelSerializers.config.serializer_lookup_chain
+      Array[*lookups].flat_map do |lookup|
+        lookup.call(klass, self, namespace)
+      end.compact
     end
 
     # Used to cache serializer name => serializer class
--- a/lib/active_model/serializer/concerns/configuration.rb
+++ b/lib/active_model/serializer/concerns/configuration.rb
@@ -32,6 +32,14 @@ module ActiveModel
         config.jsonapi_include_toplevel_object = false
         config.include_data_default = true
 
+        # example output
+        # => [
+        # "::Api::V1::ResNamespace::ResourceSerializer",
+        # "::Api::ResNamespace::ResourceSerializer",
+        # "::ResNamespace::ResourceSerializer" ,
+        # "::ResourceSerializer"]
+        config.serializer_lookup_chain = ActiveModelSerializers::LookupChain::DEFAULT.dup
+
         config.schema_path = 'test/support/schemas'
       end
     end
--- a/lib/active_model/serializer/concerns/configuration.rb
+++ b/lib/active_model/serializer/concerns/configuration.rb
@@ -32,12 +32,24 @@ module ActiveModel
         config.jsonapi_include_toplevel_object = false
         config.include_data_default = true
 
-        # example output
+        # For configuring how serializers are found.
+        # This should be an array of procs.
+        #
+        # The priority of the output is that the first item
+        # in the evaluated result array will take precedence
+        # over other possible serializer paths.
+        #
+        # i.e.: First match wins.
+        #
+        # @example output
         # => [
-        # "::Api::V1::ResNamespace::ResourceSerializer",
-        # "::Api::ResNamespace::ResourceSerializer",
-        # "::ResNamespace::ResourceSerializer" ,
-        # "::ResourceSerializer"]
+        #   "CustomNamespace::ResourceSerializer",
+        #   "ParentSerializer::ResourceSerializer",
+        #   "ResourceNamespace::ResourceSerializer" ,
+        #   "ResourceSerializer"]
+        #
+        # If CustomNamespace::ResourceSerializer exists, it will be used
+        # for serialization
         config.serializer_lookup_chain = ActiveModelSerializers::LookupChain::DEFAULT.dup
 
         config.schema_path = 'test/support/schemas'
--- a/lib/active_model_serializers/lookup_chain.rb
+++ b/lib/active_model_serializers/lookup_chain.rb
@@ -12,7 +12,7 @@ module ActiveModelSerializers
     #
     # Example:
     #  British::Author => British::AuthorSerializer
-    BY_NAMESPACE = lambda do |resource_class, _serializer_class, _namespace|
+    BY_RESOURCE_NAMESPACE = lambda do |resource_class, _serializer_class, _namespace|
       resource_namespace = namespace_for(resource_class)
       serializer_name = serializer_from(resource_class)
 
@@ -23,7 +23,7 @@ module ActiveModelSerializers
     #
     # Example:
     #  Api::V3::AuthorsController => Api::V3::AuthorSerializer
-    BY_CONTROLLER_NAMESPACE = lambda do |resource_class, _serializer_class, namespace|
+    BY_NAMESPACE = lambda do |resource_class, _serializer_class, namespace|
       resource_name = resource_class_name(resource_class)
       namespace ? "#{namespace}::#{resource_name}Serializer" : nil
     end
@@ -52,9 +52,9 @@ module ActiveModelSerializers
     end
 
     DEFAULT = [
-      BY_CONTROLLER_NAMESPACE,
-      BY_PARENT_SERIALIZER,
       BY_NAMESPACE,
+      BY_PARENT_SERIALIZER,
+      BY_RESOURCE_NAMESPACE,
       BY_RESOURCE
     ].freeze
 
--- a/lib/active_model_serializers/lookup_chain.rb
+++ b/lib/active_model_serializers/lookup_chain.rb
@@ -24,7 +24,8 @@ module ActiveModelSerializers
     # Example:
     #  Api::V3::AuthorsController => Api::V3::AuthorSerializer
     BY_CONTROLLER_NAMESPACE = lambda do |resource_class, _serializer_class, namespace|
-      "#{namespace}::#{resource_class}Serializer"
+      resource_name = resource_class_name(resource_class)
+      namespace ? "#{namespace}::#{resource_name}Serializer" : nil
     end
 
     # Allows for serializers to be defined in parent serializers
--- a/lib/active_model_serializers/lookup_chain.rb
+++ b/lib/active_model_serializers/lookup_chain.rb
@@ -1,0 +1,79 @@
+module ActiveModelSerializers
+  module LookupChain
+    # Standard appending of Serializer to the resource name.
+    #
+    # Example:
+    #   Author => AuthorSerializer
+    BY_RESOURCE = lambda do |resource_class, _serializer_class, _namespace|
+      serializer_from(resource_class)
+    end
+
+    # Uses the namespace of the resource to find the serializer
+    #
+    # Example:
+    #  British::Author => British::AuthorSerializer
+    BY_NAMESPACE = lambda do |resource_class, _serializer_class, _namespace|
+      resource_namespace = namespace_for(resource_class)
+      serializer_name = serializer_from(resource_class)
+
+      "#{resource_namespace}::#{serializer_name}"
+    end
+
+    # Uses the controller namespace of the resource to find the serializer
+    #
+    # Example:
+    #  Api::V3::AuthorsController => Api::V3::AuthorSerializer
+    BY_CONTROLLER_NAMESPACE = lambda do |resource_class, _serializer_class, namespace|
+      "#{namespace}::#{resource_class}Serializer"
+    end
+
+    # Allows for serializers to be defined in parent serializers
+    # - useful if a relationship only needs a different set of attributes
+    #   than if it were rendered independently.
+    #
+    # Example:
+    #   class BlogSerializer < ActiveModel::Serializer
+    #     class AuthorSerialier < ActiveModel::Serializer
+    #     ...
+    #     end
+    #
+    #     belongs_to :author
+    #     ...
+    #   end
+    #
+    #  The belongs_to relationship would be rendered with
+    #    BlogSerializer::AuthorSerialier
+    BY_PARENT_SERIALIZER = lambda do |resource_class, serializer_class, _namespace|
+      return if serializer_class == ActiveModel::Serializer
+
+      serializer_name = serializer_from(resource_class)
+      "#{serializer_class}::#{serializer_name}"
+    end
+
+    DEFAULT = [
+      BY_CONTROLLER_NAMESPACE,
+      BY_PARENT_SERIALIZER,
+      BY_NAMESPACE,
+      BY_RESOURCE
+    ].freeze
+
+    module_function
+
+    def namespace_for(klass)
+      klass.name.deconstantize
+    end
+
+    def resource_class_name(klass)
+      klass.name.demodulize
+    end
+
+    def serializer_from_resource_name(name)
+      "#{name}Serializer"
+    end
+
+    def serializer_from(klass)
+      name = resource_class_name(klass)
+      serializer_from_resource_name(name)
+    end
+  end
+end
--- a/lib/active_model_serializers/lookup_chain.rb
+++ b/lib/active_model_serializers/lookup_chain.rb
@@ -52,8 +52,8 @@ module ActiveModelSerializers
     end
 
     DEFAULT = [
-      BY_NAMESPACE,
       BY_PARENT_SERIALIZER,
+      BY_NAMESPACE,
       BY_RESOURCE_NAMESPACE,
       BY_RESOURCE
     ].freeze
--- a/test/support/serialization_testing.rb
+++ b/test/support/serialization_testing.rb
@@ -17,6 +17,14 @@ module SerializationTesting
     ActiveModelSerializers.config.jsonapi_namespace_separator = original_separator
   end
 
+  def with_prepended_lookup(lookup_proc)
+    original_lookup = ActiveModelSerializers.config.serializer_lookup_cahin
+    ActiveModelSerializers.config.serializer_lookup_chain.unshift lookup_proc
+    yield
+  ensure
+    ActiveModelSerializers.config.serializer_lookup_cahin = original_lookup
+  end
+
   # Aliased as :with_configured_adapter to clarify that
   # this method tests the configured adapter.
   # When not testing configuration, it may be preferable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Breaking changes:
 
 Fixes:
 
+- [#1973](https://github.com/rails-api/active_model_serializers/pull/1973) Fix namespace lookup for collections and has_many relationships (@groyoh)
 - [#1887](https://github.com/rails-api/active_model_serializers/pull/1887) Make the comment reflect what the function does (@johnnymo87)
 - [#1890](https://github.com/rails-api/active_model_serializers/issues/1890) Ensure generator inherits from ApplicationSerializer when available (@richmolj)
 - [#1922](https://github.com/rails-api/active_model_serializers/pull/1922) Make railtie an optional dependency in runtime (@ggpasqualino)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Fixes:
 
 Features:
 
+- [#1757](https://github.com/rails-api/active_model_serializers/pull/1757) Make serializer lookup chain configurable. (@NullVoxPopuli)
 - [#1968](https://github.com/rails-api/active_model_serializers/pull/1968) (@NullVoxPopuli)
   - Add controller namespace to default controller lookup
   - Provide a `namespace` render option
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -24,10 +24,6 @@ module ActionController
       @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
     end
 
-    def namespace_for_serializer
-      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
-    end
-
     def serialization_scope
       return unless _serialization_scope && respond_to?(_serialization_scope, true)
 
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -24,6 +24,10 @@ module ActionController
       @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
     end
 
+    def namespace_for_serializer
+      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+    end
+
     def serialization_scope
       return unless _serialization_scope && respond_to?(_serialization_scope, true)
 
--- a/lib/active_model/serializer/collection_serializer.rb
+++ b/lib/active_model/serializer/collection_serializer.rb
@@ -71,7 +71,9 @@ module ActiveModel
       end
 
       def serializer_from_resource(resource, serializer_context_class, options)
-        serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
+        serializer_class = options.fetch(:serializer) do
+          serializer_context_class.serializer_for(resource, namespace: options[:namespace])
+        end
 
         if serializer_class.nil?
           ActiveModelSerializers.logger.debug "No serializer found for resource: #{resource.inspect}"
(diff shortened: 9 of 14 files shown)


lib/active_model/serializer/configuration.rb
```diff
@@ -29,6 +29,28 @@ def config.array_serializer
...
+          namespaces = current_resource.split('::')
+          resource_serializer = namespaces.slice!(-1)
+
+          lookups << '::' + resource_serializer
+          namespaces.length.times do |i|
+            current_namespace = namespaces[i - namespaces.length..-1].join('::')
+            lookups << '::' + current_namespace + '::' + resource_serializer
+          end
+
+          lookups
```
bf4: This should be in serialization_context, maybe?.

It's also pretty easy to solve with explicit code in the controller.  then just pass in `serializer: resource_serializer` (or in index, `each_serializer: resource_serializer`) and you're good to go

```ruby
    def resource_serializer
      instance_variable_get("@#{resource_name}_serializer") || self.resource_serializer = "#{self.class.parent}::#{resource_class.name}Serializer".constantize
    end

    def resource_serializer=(new_resource_serializer)
      instance_variable_set("@#{resource_name}_serializer", new_resource_serializer)
    end

    # The resource class based on the controller
    # @return [Class]
    def resource_class
      @resource_class ||= resource_name.classify.constantize
    end

    # The singular name for the resource class based on the controller
    # @return [String]
    def resource_name
      @resource_name ||= controller_name.singularize
    end

```

NullVoxPopuli: Is that what we want to recommend to people?

I think it's pretty straightforward to do this, but maybe it could be documented somewhere? So people who ask can work with something? Idk 

The other thing I was thinking, is to Lee this option nil, so existing behavior remains the same, and provide the option? 

But yeah, I'm on the fence on if ams should even provide the capability to configure the lookup.

Maybe, if rails pushes more API centric usage, it would wake more sense?

groyoh: You guys could have a look at #1745. There is a discussion around the same idea (lookup based on controller). Maybe we can come up with a consensus on this.

NullVoxPopuli:
needs rebase.

Is there a direction we want to go with this?
I think having serializer lookup being configurable would be a good way to help out people who have weird namespacing situations.

NullVoxPopuli:
@rails-api/ams ^

lib/active_model/serializer/configuration.rb
NullVoxPopuli: oh!, @bf4 that serializer lookup defined in the controller wouldn't keep relationships under the same namespace, would it? I think that would be harder to debug than the proc here

NullVoxPopuli:
rebased

NullVoxPopuli:
@bf4 @richmolj I'd like to revisit this, and get your thoughts on it.

What is present in this PR, is the change to support using a proc for serializer lookup and an _example_ proc.
Obviously, the final version of this PR would have the proc be empty (if we wanted it that way).

I am in favor of the either of the following

 - Move serializer lookup entirely into the proc

or
 
 - have the proc be empty, keeping current / default lookup. The proc would only provide additional lookups, and the default would not be able to be changed (just prepended to the custom)


For both scenarios:

 - Provide tests that show procs and the common situations we get asked about with customizing the lookup chain. (API versioning, API fallback, etc)

richmolj:
@NullVoxPopuli after reading [this comment of yours](https://github.com/rails-api/active_model_serializers/pull/1757#discussion_r78190986), I'm sold.

> Move serializer lookup entirely into the proc

I'd favor a slight modification to this, moving the serializer lookup into a *collection* of procs. Something like:

```ruby
ActiveModelSerializers.config.serializer_lookup_chain << proc_to_check_object_instance_method
ActiveModelSerializers.config.serializer_lookup_chain << proc_for_automatic_namespaced_lookup
ActiveModelSerializers.config.serializer_lookup_chain << proc_for_automatic_lookup
```

That way our internals match the externals (simpler), and users can easily re-order or add new ones that would take precedence. So basically middleware, where the first one to return a match wins.

lib/active_model/serializer/concerns/configuration.rb
```diff
@@ -31,6 +31,28 @@ def config.array_serializer
...
 
+        # example output
+        # => [
+        # "::Api::V1::ResNamespace::ResourceSerializer",
+        # "::Api::ResNamespace::ResourceSerializer",
+        # "::ResNamespace::ResourceSerializer" ,
+        # "::ResourceSerializer"]
+        config.serializer_lookup = proc { |controller_name|
+          lookups = []
+          current_resource = controller_name.gsub(/Controller\z/, '')
```
bf4: controller_name isn't correct. It's more lookup_namespace and then you want to `lookup_namespace.remove(/Controller\z/)` because you know someone might pass in a controller?

NullVoxPopuli: that's true. I'll switch it to lookup_namespace, and mention in a comment that it *can* and often is the controller name, cause of https://github.com/rails-api/active_model_serializers/pull/1757/files#diff-5624f9cf35be4dc254412d979fd293c4R63 klass is a controller (I guess, most of the time)

NullVoxPopuli:
> moving the serializer lookup into a collection of procs.

@richmolj: I think I could get on board with that -- but we need a place to store the procs that makes sense.

I looking at the existing code, I think it only makes sense to have the following as 'default':

```ruby
  module AMS::SerializerLookup
     DEFAULT = -> (lookup_namespace) {
        chain = []

      # Example: 
      # BlogSerializer::AuthorSerializer
      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
      # Example: 
      # Api::AuthorSerializer
      chain.push("#{resource_namespace}::#{serializer_class_name}")

      chain
     }    

```
I don't think it makes sense to split this up, because both chain.push lines need serializer_class_name


then is serializer.rb:
```ruby
    def self.serializer_lookup_chain_for(klass)
      ActiveModelSerializers.config.serializer_lookup_chain.map { |p| p.call(klass) }.flatten
    end
```

NullVoxPopuli:
I just realized, that these procs are going to need the serializer class name to be passed to them as well (esp for the case of defining nested serializers)

richmolj:
My thinking was something like

```ruby
chain.push ->(klass) { klass.serializer if klass.respond_to?(:serializer) }
chain.push ->(klass) { "#{klass.name}Serializer".safe_constantize }
chain.push ->(klass) { "#{klass.name.demodulize}Serializer".safe_constantize }
```

Even better, pass an instance instead of the class and do the same.

But I'm splitting hairs here. As long as this is a customizable array of procs being passed an instance to be serialized, or the class of that instance, I'm 👍

NullVoxPopuli:
I think it the interface would have to be:

String, Class
-> (lookup_namespace, serializer_class) { ... }

Because, serializer_lookup_chain_for only gets passed the controller class (not the instance). I also thinking passing the instance would invite too much coupling.

But, with the above change to the proc API, the procs could be:

      # Example: 
      # BlogSerializer::AuthorSerializer
      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer

```ruby
# Downside to splitting this up so much, is that there is some duplicate code
# Like creating the serializer name string from since it's going to be the suffix
# of everything
# This could be a third parameter so it's not calculated all the time.
# however, this proc would then have 3 parameters, and I don't know how I feel about that.
module AMS::SerializerLookup
   DEFAULT = [
      IN_PARENT_SERIALIZER
      BY_NAMESPACE
   ] 

  IN_PARENT_SERIALIZER = -> (lookup_namespace, serializer_class) {
    # logic for giving the parent::child serializer name
  }

  BY_NAMESPACE = -> (lookup_namespace, serializer_class) {
    logic for giving a serializer of the same namespace as lookup_namespace
  }
end
```

```ruby
# serializer.rb
    def self.serializer_lookup_chain_for(klass)
      ActiveModelSerializers.config.serializer_lookup_chain.map { |p| 
        p.call(klass, self) 
      }.flatten.compact
    end
```

Config Somewhere:
```ruby
# ( already default (in configuration.rb), so this would be redundant )
AMS.config.serializer_lookup_chain = AMS::SerializerLookup::DEFAULT
or 
AMS.config.serializer_lookup_chain << whatever_proc
or
AMS.config.serializer_lookup_chain = []
AMS.config.serializer_lookup_chain << some_proc
```

richmolj:
> Because, serializer_lookup_chain_for only gets passed the controller class (not the instance).

Right, but I think we could change [this line](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L47) pretty easily. However, you note the performance impact of string instantiation which I agree with. It looks like [the result is cached](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L89) so it's not a performance issue if we pass the class, but it would be if we passed the instance.

So I can get behind the idea of passing the class, instead of the instance, for that reason. My only thought with the instance was you could pick the serializer at runtime based on a property or something.

I'm not totally sure why `lookup_namespace` is needed in addition to the class though.

richmolj:
Oh hold on I think I may have found the source of my confusion. The goal is to based the lookup on the *controller* namespace, not the *object's* namespace. That makes sense now. My preference would be to pass the controller instance itself, instead of just its namespace, but either way.

NullVoxPopuli:
I'll get some tests written for this, this weekend to demonstrate the different scenarios I can think of where this could be useful.

richmolj:
After thinking about it a little bit more, I think you were right. `lookup_namespace` and `serializer_klass` should be the arguments passed to the procs. Then use the caching in `serializer_lookup_chain_for` to avoid a performance penalty.

This is a great idea. I don't have this use case yet, but I expect to eventually.

lib/active_model/serializer.rb
```diff
@@ -60,15 +60,12 @@ class << self
...
-      chain = []
-
       resource_class_name = klass.name.demodulize
-      resource_namespace = klass.name.deconstantize
-      serializer_class_name = "#{resource_class_name}Serializer"
-
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
 
+      lookups = ActiveModelSerializers.config.serializer_lookup_chain
```
NullVoxPopuli: @richmolj this looks nice - It's already chached, see line 86

lib/active_model/serializer/concerns/configuration.rb
```diff
@@ -31,6 +31,31 @@ def config.array_serializer
...
         config.jsonapi_include_toplevel_object = false
 
+        # example output
+        # => [
+        # "::Api::V1::ResNamespace::ResourceSerializer",
+        # "::Api::ResNamespace::ResourceSerializer",
+        # "::ResNamespace::ResourceSerializer" ,
+        # "::ResourceSerializer"]
+        #
+        # NOTE: lookup_namespace is often the ControllerClass.name
```
NullVoxPopuli: this is false. the first parameter is the resource class

lib/active_model_serializers/lookup_chain.rb
```diff
@@ -0,0 +1,43 @@
+module ActiveModelSerializers
+  module LookupChain
+    BY_RESOURCE = lambda do |_resource_class, resource_class_name, _serializer_class|
+      "#{resource_class_name}Serializer"
+    end
+
+    BY_NAMESPACE = lambda  do |resource_class, resource_class_name, _serializer_class|
+      resource_namespace = resource_class.name.deconstantize
+      serializer_name = "#{resource_class_name}Serializer"
```
NullVoxPopuli: serializer_name is duplicated, and so is resource_namespace.

This is why I'm more leaning towards this NOT being an array of procs, so that we can keep the code duplication down.

richmolj: To remove the duplication you could do something like:

```ruby
class SerializerLookup
  class << self
    def by_namespace(lookup_namespace, klass)
      "#{resource_namespace(klass)}::#{serializer_name(klass)}"
    end

    def by_parent_serializer(lookup_namespace, klass)
      # re-use same logic
    end

    private

    def resource_namespace(klass)
      klass.name.deconstantize
    end

    def serializer_name(klass)
      "#{resource_class_name}Serializer"
    end
end

procs << SerializerLookup.method(:by_namespace)
procs << SerializerLookup.method(:by_parent_serializer)
```

That said - I think an array of procs is a good idea, but I'm not chained to it.

NullVoxPopuli: ha. _chained_. :-)

I like this pattern.

richmolj: hahahaha I didn't even realize...no pun intended 😂

NullVoxPopuli:
At the very least all this is finally getting documented

lib/active_model_serializers/lookup_chain.rb
bf4: @NullVoxPopuli @richmolj a class without state should just be a module using either `module_function` or `extend self` depending on if there's a chance you might want to include it (like file_utils) or just style

NullVoxPopuli:
@remear do you have opinions on this?

test/feature/configuration/serializer_lookup/using_serializable_resource_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'test_helper'
+
+describe 'feature' do
+  describe 'configuration' do
+    describe 'serializer lookup' do
+      describe 'using SerializableResource' do
+        it 'uses the child serializer' do
```
NullVoxPopuli: more tests need to exist in this file such that it can be the example of possible lookup chain scenarios. 

Including:
 - Versioned APIs (both via URL and request header)
 - Versioned APIs with fallback to earlier version
 - Mimicing the controller's namespace.

NullVoxPopuli:
Working on tests now, just came up with a need for this

groyoh:
@NullVoxPopuli is this ready for review?

NullVoxPopuli:
@groyoh yeah, just working on a namespaced controller test atm.

lib/active_model/serializer.rb
```diff
@@ -60,16 +60,10 @@ class << self
...
-      resource_class_name = klass.name.demodulize
-      resource_namespace = klass.name.deconstantize
-      serializer_class_name = "#{resource_class_name}Serializer"
-
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
-
-      chain
+      lookups = ActiveModelSerializers.config.serializer_lookup_chain
+      Array[*lookups].map do |lookup|
```
groyoh: What about `flat_map`?

NullVoxPopuli:
looks like this isn't done, as I need to pass in the controller context to the lookup procs

lib/active_model/serializer.rb
groyoh: I simply meant: 

```ruby
 Array[*lookups].flat_map do |lookup|
  lookup.call(klass, self)
end.compact
```

:)

NullVoxPopuli: oh ok. that would be a safe way to handle the accidental scenario of lookups being a nested array

NullVoxPopuli: actually, I had it backwards.

found this: 
```
a.flat_map(&b) works exactly like a.map(&b).flatten!(1).
```
is that useful?

NullVoxPopuli: oh wait. I'm flattening at the end( I didn't look at the whole method).

flat_map looks good :-)

NullVoxPopuli:
@rails-api/ams now it's ready for review. No idea why my test is poisoning the config though.

It seems that I can never set the config option `serializer_lookup_chain` back to what it was before the lookup test runs.

NullVoxPopuli:
I had to move the lookup proc test to an isolated test. :-\

NullVoxPopuli:
coverage report is waaaay off in travis

NullVoxPopuli:
this is what I get:

```
Finished in 0.915576s, 578.8708 runs/s, 969.8817 assertions/s.

530 runs, 888 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /home/lprestonsegoiii/Development/NullVoxPopuli/active_model_serializers/coverage. 1496 / 1515 LOC (98.75%) covered.
Running RuboCop...
Inspecting 165 files
.........................
```
This is what The broken travis builds get:
```
# Running:
.
Finished in 0.022875s, 43.7159 runs/s, 87.4318 assertions/s.
1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
Formatter CodeClimate::TestReporter::Formatter failed with CodeClimate::TestReporter::Formatter::InvalidSimpleCovResultError: undefined method `values' for #<SimpleCov::Result:0x00000006c46d90> (/home/travis/build/rails-api/active_model_serializers/vendor/bundle/ruby/2.3.0/gems/codeclimate-test-reporter-1.0.0/lib/code_climate/test_reporter/formatter.rb:21:in `rescue in format')
```
Looks like something broke with simple cov

NullVoxPopuli:
omg, I think I fixed it
yay
The errors for the past few builds were just simplecov / codeclimate-reporter related.

NullVoxPopuli:
hmm, trying this in an app -- something's off

NullVoxPopuli:
ok, I 'think' it works

```ruby
ActiveModelSerializers.config.serializer_lookup_chain.unshift(
  lambda do |_, _, controller_class|
    if controller_class
      ap controller_class
      ap controller_class.name.gsub(/sController$/, 'Serializer')
      controller_class.name.gsub(/sController$/, 'Serializer')
    end
  end
  # ] + ActiveModelSerializers::LookupChain::DEFAULT.dup
)
```

However, I think there may be more work to be done on passing along the namespace from a controller to the nested serializers/relationships

NullVoxPopuli:
Just tested this with my personal app with all my weird namespacing. All tests pass. :-)

NullVoxPopuli:
For those passing by, this is how you get AMS to find serializers that follow the namespace of the controller:

```ruby
ActiveModelSerializers.config.serializer_lookup_chain.unshift(
  lambda do |resource_class, _, controller_class|
    if controller_class
      serializer_name = "::#{resource_class.name}Serializer"
      controller_class.name.gsub(/::([^:])+Controller$/, serializer_name)
    end
  end
)
```

.simplecov
```diff
@@ -48,34 +48,34 @@ require 'coverage'
...
-    @output.puts exit_message
-    abort(failure_message) if @generate_report
-  elsif @running_ci
-    exit_message.sub!('{{RESULTS}}', results).sub!('{{FAILURE_MESSAGE}}', <<-EOF)
-Nice job! Spec coverage (#{percent.round(2)}%) is still at or above #{@minimum_coverage}%
-    EOF
-    @output.puts exit_message
-  end
-end
+# SimpleCov.at_exit do
```
NullVoxPopuli: do we really need all this? it was causing the travis errors

test/feature/configuration/serializer_lookup/using_serializable_resource_test.rb
NullVoxPopuli: this'll be todo, as we'd just be testing different ways to configure -- which is open ended

test/support/serialization_testing.rb
```diff
@@ -17,6 +17,14 @@ def with_namespace_separator(separator)
     ActiveModelSerializers.config.jsonapi_namespace_separator = original_separator
   end
 
+  def with_prepended_lookup(lookup_proc)
```
NullVoxPopuli: I think this only works when running isolated tests.

.travis.yml
```diff
@@ -20,7 +20,8 @@ cache:
 
 script:
   - bundle exec rake ci
-
+after_success:
+  - codeclimate-test-reporter
```
bf4: is the api key set somewhere?

lib/active_model/serializer.rb
```diff
@@ -59,17 +59,11 @@ class << self
...
-      serializer_class_name = "#{resource_class_name}Serializer"
-
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
-
-      chain
+    def self.serializer_lookup_chain_for(klass, controller_class = nil)
+      lookups = ActiveModelSerializers.config.serializer_lookup_chain
+      Array[*lookups].flat_map do |lookup|
+        lookup.call(klass, self, controller_class)
```
bf4: this doesn't belong in the Serializer.  Serializer should know nothing about controller invoking it.

bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1757/files#r86845909

lib/active_model/serializer.rb
```diff
@@ -84,11 +78,12 @@ def self.serializers_cache
     #   1. class name appended with "Serializer"
     #   2. try again with superclass, if present
     #   3. nil
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(klass, options = {})
       return nil unless config.serializer_lookup_enabled
       serializers_cache.fetch_or_store(klass) do
         # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
-        serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
+        lookup_chain = serializer_lookup_chain_for(klass, options[:controller_class])
```
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1757/files#r86845909

lib/active_model/serializer/reflection.rb
```diff
@@ -106,6 +106,8 @@ def value(serializer, include_slice)
       #
       def build_association(parent_serializer, parent_serializer_options, include_slice = {})
         reflection_options = options.dup
+        reflection_options[:controller_class] ||= parent_serializer_options[:controller_class]
```
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1757/files#r86845909

lib/active_model_serializers/lookup_chain.rb
```diff
@@ -0,0 +1,70 @@
...
+    #   end
+    #
+    #  The belongs_to relationship would be rendered with
+    #    BlogSerializer::AuthorSerialier
+    BY_PARENT_SERIALIZER = lambda do |resource_class, serializer_class, _controller_class|
+      return if serializer_class == ActiveModel::Serializer
+
+      serializer_name = serializer_from(resource_class)
+      "#{serializer_class.name}::#{serializer_name}"
+    end
```
bf4: None of these actually use `controller_class`. ref: https://github.com/rails-api/active_model_serializers/pull/1757/files#r86845909

lib/active_model_serializers/serializable_resource.rb
```diff
@@ -55,7 +55,8 @@ def serializer
       @serializer ||=
         begin
           @serializer = serializer_opts.delete(:serializer)
-          @serializer ||= ActiveModel::Serializer.serializer_for(resource)
+          options = { controller_class: serializer_opts[:controller_class] }
```
bf4: ref: https://github.com/rails-api/active_model_serializers/pull/1757/files#r86845909

lib/active_model/serializer.rb
bf4: I'd rather just move all these class methods to a `ImplicitSerializer` utility module with a `lookup_by` method and a user-customizable strategy somewhere. e.g.

```ruby
config.serializer_lookup = ImplicitSerializer
```

Then someone can just change that to whatever they want and customize.

Now, it's still missing most of the controller context, .... or is it?  That's where `SerializationContext` comes it. It could take in a `namespace` option.

.simplecov
NullVoxPopuli: why can't we get the artifacts on travis? [Code Climate Test Reporter](https://github.com/codeclimate/ruby-test-reporter#installation) recently released a breaking change where you don't run code climate test reporter in your tests -- they now provide a shell utilitiy to upload simplecov output.

> What errors?

for example: https://travis-ci.org/rails-api/active_model_serializers/jobs/173385513

lib/active_model/serializer.rb
bf4: ref [ActionController::Serialization in 0.9](https://github.com/rails-api/active_model_serializers/pull/1436#issuecomment-171823264)

```ruby

    def namespace_for_serializer
      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
    end

    def default_serializer(resource)
      options = {}.tap do |o|
        o[:namespace] = namespace_for_serializer if namespace_for_serializer
      end

      ActiveModel::Serializer.serializer_for(resource, options)
    end

    def default_serializer_options
      {}
    end

    def serialization_scope
      _serialization_scope = self.class._serialization_scope
      send(_serialization_scope) if _serialization_scope && respond_to?(_serialization_scope, true)
    end

    def build_json_serializer(resource, options = {})
      options = default_serializer_options.merge(options)
      @namespace_for_serializer = options.fetch(:namespace, nil)

      if serializer = options.fetch(:serializer, default_serializer(resource))
        options[:scope] = serialization_scope unless options.has_key?(:scope)

        if resource.respond_to?(:to_ary)
          options[:resource_name] = controller_name 
          options[:namespace] = namespace_for_serializer if namespace_for_serializer
        end

        serializer.new(resource, options)
      end
```

.travis.yml
NullVoxPopuli: I thought that'd be configured in the travis ENV?

lib/active_model_serializers/lookup_chain.rb
NullVoxPopuli: right, these are all the lookup strategies that AMS 0.10 has been using -- just re-implemented as procs. This was a suggestion by @richmolj, I think. In case someone wanted to customize the lookup order, they could.

.simplecov
bf4: > they now provide a shell utilitiy to upload simplecov output.
not new, but I guess is a new default :)

NullVoxPopuli: Surprise!

:-(

A couple of my gems' travis builds broke cause of this.

bf4: seems like 

```ruby
if ENV['CI']
  # Since version 1.0, we now require you to run `SimpleCov` in your test/spec helper, and then
  # run the provided `codeclimate-ruby` binary separately to report your results
   # to Code Climate.
   # More information here: https://github.com/codeclimate/ruby-test-reporter/blob/master/README.md
   no format
else
  remove all the fancy text printing and just use the html formatter
end
```

.travis.yml
bf4: Well, someone needs to add it...

lib/active_model/serializer.rb
```diff
@@ -84,11 +78,12 @@ def self.serializers_cache
     #   1. class name appended with "Serializer"
     #   2. try again with superclass, if present
     #   3. nil
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(klass, options = {})
```
bf4: might as well just be `namespace` instead of `options`

lib/active_model_serializers/lookup_chain.rb
```diff
@@ -0,0 +1,70 @@
+module ActiveModelSerializers
+  module LookupChain
+    # Standard appending of Serializer to the resource name.
+    #
+    # Example:
+    #   Author => AuthorSerializer
+    BY_RESOURCE = lambda do |resource_class, _serializer_class, _namespace|
+      serializer_from(resource_class)
+    end
```
bf4: @NullVoxPopuli don't kill me, but can we put the new lookup chain in a separate PR and just add the `namespace` code in this PR?

lib/active_model/serializer.rb
```diff
@@ -59,17 +59,11 @@ class << self
...
-      resource_class_name = klass.name.demodulize
-      resource_namespace = klass.name.deconstantize
-      serializer_class_name = "#{resource_class_name}Serializer"
-
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
-
-      chain
+    def self.serializer_lookup_chain_for(klass, namespace = nil)
+      lookups = ActiveModelSerializers.config.serializer_lookup_chain
```
bf4: I think it would be better for the lookup_chain to be a class_attribute or some such on the serializer

```ruby
class_attribute(:lookup_chain) {  ActiveModelSerializers.config.serializer_lookup_chain }
```

lib/active_model_serializers/lookup_chain.rb
NullVoxPopuli: lol. It'll have to be tonight, so I can dedicate more mental capacity to splitting this apart. :-)

lib/active_model/serializer.rb
bf4: makes it easier to configure in place, and no need for changing global state

NullVoxPopuli: that'd probably allow the lookup tests to not need to be isolated.

.travis.yml
NullVoxPopuli: oh. O.O -- I guess we haven't had coverage being pushed this whole time...

NullVoxPopuli:
https://github.com/rails-api/active_model_serializers/pull/1968 should be merged before any further changes happen to this PR.

After that PR is merged, this PR is just going to provide a way for users to configure the lookup chain without monkey patching.

NullVoxPopuli:
TODO:

- [ ]  Benchmarks vs `master`
- [ ]  Documentation

.simplecov
```diff
@@ -105,7 +105,6 @@ end.new do
   else
     formatters = []
   end
-  # Using default formatters...
-  # SimpleCov.formatters = formatters
+  SimpleCov.formatters = formatters
```
beauby: This does not seem to belong in this PR.

NullVoxPopuli: It was reverted

lib/active_model_serializers/lookup_chain.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    # Uses the namespace of the resource to find the serializer
+    #
+    # Example:
+    #  British::Author => British::AuthorSerializer
+    BY_NAMESPACE = lambda do |resource_class, _serializer_class, _namespace|
```
groyoh: I think it would be more comprehensible and accurate to call this `BY_RESOURCE_NAMESPACE` and the one below `BY_NAMESPACE`. `BY_RESOURCE_NAMESPACE` makes it clear that we use the namespace from the resource. Otherwise it may be ambiguous. Also the lookup below may not be the "controller" namespace but any kind of namespace. For example, we used to have all the serializer in a `Serializer` module e.g. `Serializer::V1::AuthorSerializer`, `Serializer::V2::AuthorSerializer`.

NullVoxPopuli: Totally agree with this. I'll update on my next iteration

bf4:
needs rebase off of master, no?

NullVoxPopuli:
It is currently rebased off master.

NullVoxPopuli:
@groyoh I've addressed both your recent comments

NullVoxPopuli:
Side note: (@bf4, @groyoh), htis is the current priority for namespace lookup:

```ruby
    DEFAULT = [
      BY_NAMESPACE,
      BY_PARENT_SERIALIZER,
      BY_RESOURCE_NAMESPACE,
      BY_RESOURCE
    ].freeze
```

should custom namespace and parent serializer be switched? 
Or is this where we draw the line and just tell people to customize it?

So: 

```ruby
ActiveModelSerializers.config.serializer_lookup_chain = [
  ActiveModelSerializers::LookupChain::BY_PARENT_SERIALIZER,
  ActiveModelSerializers::LookupChain::BY_NAMESPACE,
  ActiveModelSerializers::LookupChain::BY_RESOURCE_NAMESPACE,
  ActiveModelSerializers::LookupChain::BY_RESOURCE
]
``` 
?

groyoh:
I think the second one makes a bit more sense. For example, if I define a `Serializers::User`, `Serializer::Post` and `Serializer::Post::User`, I'd clearly want the child serializer to be prioritized. Otherwise defining child serializer and another serializer with the same name may cause issues.

NullVoxPopuli:
That's what I was thinkin.

NullVoxPopuli:
I think this is re-ready for review for those that are interested.

docs/general/configuration_options.md
```diff
@@ -60,6 +60,50 @@ application, setting `config.key_transform` to `:unaltered` will provide a perfo
...
+```ruby
+[
+  ActiveModelSerializers::LookupChain::BY_PARENT_SERIALIZER,
+  ActiveModelSerializers::LookupChain::BY_NAMESPACE,
+  ActiveModelSerializers::LookupChain::BY_RESOURCE_NAMESPACE,
+  ActiveModelSerializers::LookupChain::BY_RESOURCE
+]
+```
+
+Each of the array entries represent a proc. A serializer lookup proc will be yielded 3 arguments. `resource_class`, `serializer_class`, and `namespace`.
```
groyoh: Shouldn't we explain that:
* by default `serializer_class` is `ActiveModel::Serializer`
* for association lookup it's the "parent" serializer
* `namespace` correspond to either the controller namespace or the one specified with `namespace` option

?

docs/general/configuration_options.md
```diff
@@ -60,6 +60,50 @@ application, setting `config.key_transform` to `:unaltered` will provide a perfo
...
+  lambda do |resource_class, serializer_class, namespace|
+    "API::#{namespace}::#{resource_class}"
+  end
+]
+```
+
+If you simply want to add to the existing lookup_chain. Use `unshift`.
+
+```ruby
+ActiveModelSerializers.config.serializer_lookup_chain.unshift(
```
groyoh: That's an opportunity to add a `ActiveModelSerializers.add_lookup(&block)` I guess.

docs/general/configuration_options.md
NullVoxPopuli: lib/active_model/serializer/field.rb.

Attribute, and Reflection inherit from it.

NullVoxPopuli: couldn't hurt. I'll add that.

docs/general/configuration_options.md
NullVoxPopuli: maybe in another pr?

groyoh:
Looks good to me.  @NullVoxPopuli  is there anything missing for merging? Are you still planning to benchmark this?

docs/general/configuration_options.md
groyoh: Thanks :100:

groyoh:
@NullVoxPopuli maybe this PR could benefit from having an entry in the CHANGELOG? I mean it's an addition to AMS API so would be good to have it in the CHANGELOG.

NullVoxPopuli:
changelog added. :-)

NullVoxPopuli:
> Looks good to me. @NullVoxPopuli is there anything missing for merging? Are you still planning to benchmark this?

Working on it now, but other than that, I don't think anything is missing -- think this is a good set of changes for people to learn about and modify the lookup chain.

NullVoxPopuli:
Here are the benchmark results:

```
Configurable Lookup Chain 246237.80395070254/ips; 21 objects
Old Lookup Chain (v0.10) 791285.5954198999/ips; 9 objects
Benchmark results:
{
  "commit_hash": "999c858",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "benchmark_run[environment]": "2.3.1p112",
  "runs": [
    {
      "benchmark_type[category]": "Configurable Lookup Chain",
      "benchmark_run[result][iterations_per_second]": 246237.804,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 21
    },
    {
      "benchmark_type[category]": "Old Lookup Chain (v0.10)",
      "benchmark_run[result][iterations_per_second]": 791285.595,
      "benchmark_run[result][total_allocated_objects_per_iteration]": 9
    }
  ]
}
```

NullVoxPopuli:
That performance hit is kinda concerning? Do you think it's still worth merging this, since in real-life, the serializer lookups are cached?

groyoh:
If you change the benchmark to test 

```ruby
ActiveModelSerializers::SerializableResource.new(
  PrimaryResource.new(id: 1, title: "title", body: "body")
).as_json
```

The results get around:

```json
{
  "label": "Configurable Lookup Chain",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 6755.365764011151,
  "iterations_per_second_standard_deviation": 2.664548542418537,
  "total_allocated_objects_per_iteration": 175
}
{
  "label": "Old Lookup Chain (v0.10)",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 6779.273728491176,
  "iterations_per_second_standard_deviation": 6.165852224601526,
  "total_allocated_objects_per_iteration": 175
}
```

Which is not that big of a deal IMO and we could certainly optimize it at a later point.

I tried it out quickly and we should probably spend a bit more time benchmarking it (especially with namespace lookup, child serializers, etc). But if the lookup process only takes a tiny of time compared to the whole serialization process, then we're pretty safe.

NullVoxPopuli:
Cool. Well then, I think we're good to go. :-)

Changelog has been added.

groyoh:
@NullVoxPopuli I'll do some deeper benchmarks later and merge it if everything looks fine. Does that work for you?

NullVoxPopuli:
It does! I'm in no rush. I also have this:
2016-11-14-112349_303x79_scrot

So feel free to push benchmark changes.

groyoh:
@NullVoxPopuli have a look at https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/collection_serializer.rb#L74. We do not pass the `namespace` options when initializing the serializers meaning that if we have:

```ruby
module Api
  module V1
    class PrimaryResourceSerializer < ActiveModel::Serializer
      attributes :title, :body
      has_many :has_many_relationships
    end
    class HasManyRelationshipSerializer < ActiveModel::Serializer
      attribute :body
    end
  end
end
```

The `has_many` lookup chain will be will be

```ruby
[
  "AmsBench::Api::V1::PrimaryResourceSerializer::HasManyRelationshipSerializer",   
  "::HasManyRelationshipSerializer",
  "HasManyRelationshipSerializer"
]
```

instead of

```ruby
[
  "AmsBench::Api::V1::PrimaryResourceSerializer::HasManyRelationshipSerializer",   
  "AmsBench::Api::V1::HasManyRelationshipSerializer",
  "HasManyRelationshipSerializer"
]
```

Pretty sure that's a bug but don't have time right now to look further into it. Will check again later. Also not sure if it belongs to this PR either or a different bug PR.

NullVoxPopuli:
Oh, collections. Kk. Good find. I can make a PR to fix soon

groyoh:
Once #1973 is merged, I'll add my benchmark changes/results. Otherwise the benchmark won't be accurate.

NullVoxPopuli:
cool. I'm just waiting for / watching travis

NullVoxPopuli:
@groyoh #1973 is merged

groyoh:
This is the final results: 

```json
{
  "label": "Configurable Lookup Chain",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 2104.3974604253326,
  "iterations_per_second_standard_deviation": 2.280937936044573,
  "total_allocated_objects_per_iteration": 593
}
{
  "label": "Old Lookup Chain (v0.10)",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 2097.130687654142,
  "iterations_per_second_standard_deviation": 6.055893452308528,
  "total_allocated_objects_per_iteration": 593
}
```

groyoh:
I just figured out there is something strange with the benchmark. I need to test it further.

NullVoxPopuli:
bro. rubocop

NullVoxPopuli:
I wish I could :+1: a commit

NullVoxPopuli:
wait, so the old lookup chain is slower than the new configurable one? with the same objects? wat

NullVoxPopuli:
anywho, that benchmark looks good to me :+1:

lib/active_model/serializer/concerns/configuration.rb
```diff
@@ -32,6 +32,26 @@ def config.array_serializer
...
+        # @example output
+        # => [
+        #   "CustomNamespace::ResourceSerializer",
+        #   "ParentSerializer::ResourceSerializer",
+        #   "ResourceNamespace::ResourceSerializer" ,
+        #   "ResourceSerializer"]
+        #
+        # If CustomNamespace::ResourceSerializer exists, it will be used
+        # for serialization
+        config.serializer_lookup_chain = ActiveModelSerializers::LookupChain::DEFAULT.dup
```
beauby: Why the `.dup`?

NullVoxPopuli: DEFAULT is a frozen array, and we want mutatability?

beauby: `DEFAULT` should be frozen, but are there real use cases for injecting stuff in the lookup chain rather than replacing it altogether?

NullVoxPopuli: I'm not entirely sure on that one. 

Do we get a performance boost if serializer_lookup_chain is frozen? I wouldn't think it matters, since config is a singletonish thing

groyoh:
@NullVoxPopuli turns out the whole serializer lookup is cached (https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L97) so we cannot see any difference once it's cached...in the end the benchmark I added does not actually says much.

NullVoxPopuli:
Excellent

NullVoxPopuli:
any objections to merging?

groyoh:
Yup, sorry gotta push a last commit. Will do it in 5min (and I will make sure I run rubocop :grin: ).

groyoh:
So the end result is:

* if I have this code

```ruby
configurable.call
old.call
```

```json
{
  "label": "Configurable Lookup Chain",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 2097.0843979297792,
  "iterations_per_second_standard_deviation": 2.6226889129638966,
  "total_allocated_objects_per_iteration": 593
}
{
  "label": "Old Lookup Chain (v0.10)",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 2140.142899144878,
  "iterations_per_second_standard_deviation": 3.831519850041984,
  "total_allocated_objects_per_iteration": 593
}
```

* if I have this code

```ruby
old.call
configurable.call
```

```json
{
  "label": "Old Lookup Chain (v0.10)",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 2103.7895845497637,
  "iterations_per_second_standard_deviation": 2.8044629763972666,
  "total_allocated_objects_per_iteration": 593
}
{
  "label": "Configurable Lookup Chain",
  "version": "0.10.2",
  "rails_version": "4.2.7.1",
  "iterations_per_second": 2177.0852661503927,
  "iterations_per_second_standard_deviation": 1.3779892072416242,
  "total_allocated_objects_per_iteration": 593
}
```

So it looks that the one running last is running faster (?). But looking at both result I'd say the difference is not significant.

NullVoxPopuli:
cool. Running the old one actually would invalidate the results of the configurable lookup chain, because of the in-bench monkey patch.

But excellent! once travis passes, lets merge.

groyoh:
So I'll merge once Travis give the go and no one objects!

groyoh:
@NullVoxPopuli in my last commit I make sure to clear the cache before each benchmark. That way we ensure that the lookup chain method is actually called and that both benchmark are similar.

PR #1758, opened 2016-05-30
Update README to write how to configure adapter

tricknotes:
For the new user, how to customize adapter is an important concern.
Additionally, README says:
> Check how to change the adapter in the sections below.

However how to change adapter was not documented anywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,12 @@ Thanks!
 
 ## High-level behavior
 
+Choose an adapter from [adapters](lib/active_model_serializers/adapter):
+
+``` ruby
+ActiveModelSerializers.config.adapter = :json_api # Default: `:attributes`
+```
+
 Given a [serializable model](lib/active_model/serializer/lint.rb):
 
 ```ruby


README.md
```diff
@@ -96,6 +96,12 @@ Thanks!
 
 ## High-level behavior
 
+Choise an adapter from [adapters](lib/active_model/serializer/adapter):
```
NullVoxPopuli: `Choise` should be choose.

NullVoxPopuli:
I'm surprised that wasn't already on that document. nice catch!

README.md
tricknotes: Good catch @NullVoxPopuli !
I fixed it 🔧 https://github.com/rails-api/active_model_serializers/pull/1758/commits/90cfa92b5102dd11f7755b427b21318602279061

README.md
```diff
@@ -96,6 +96,12 @@ Thanks!
 
 ## High-level behavior
 
+Choose an adapter from [adapters](lib/active_model/serializer/adapter):
```
beauby: Adapters have been migrated to `lib/active_model_serializers/adapter` IIRC.

tricknotes: Ouch! That's right.

NullVoxPopuli:
LGTM, @tricknotes, thanks for the update!

Issue #1756, opened 2016-05-30
Exclude namespace in root key in 0.10?


ehannes:
Using the `:json` adapter, is there some way to exclude the namespace part from the root key?

I used version 0.9 before and my JSON representation of a `User` at api/v1/users/1 was something like:

    { "user": { "id": 1, "first_name": "Foo", "last_name": "Bar" } }

I recently upgraded to version 0.10 and now I get the namespace included in the root key:

    { "api/v1/user": { "id": 1,"first_name": "Foo", "last_name": "Bar" } }

I want the root key to be like before, that is "user" instead of "api/v1/user".

NullVoxPopuli:
According to https://github.com/rails-api/active_model_serializers/blob/master/docs/general/adapters.md#json

The root key can't be overridden :-(

I think @beauby was doing some work on this.

bf4:
You should be able to change it by controlling `Serializer#json_key`.  See `Adapter::Base#root`

```ruby
      def root
        serializer.json_key.to_sym if serializer.json_key
      end
````

ehannes:
@NullVoxPopuli I read that, but I also read this at https://github.com/rails-api/active_model_serializers/blob/master/docs/general/rendering.md#overriding-the-root-key:

> Overriding the resource root only applies when using the JSON adapter.

Is that documentation obsolete?

ehannes:
> So, looks like you can make a pr to update the docs :)

@bf4 What docs are you suggesting to be updated? The one that @NullVoxPopuli linked to?

ehannes:
I have now two working solutions:

1. Set `type` in the serializer as @bf4 suggested.

2. Set the root key explicitly when rendering, something like:
`render json: @user, root: 'user'`

The second alternative is obviously only suitable when changing the root temporary.

I think I'll go the first alternative and programmatically determine the root key through a module or something. That has the advantage to work with the `:json_api` adapter as well.

bf4:
Whatever would have helped you before you made the issue :) probably need to update those docs 

B mobile phone

remear:
> The root key can't be overridden, and will be derived from the resource being serialized.

needs to be corrected in https://github.com/rails-api/active_model_serializers/blob/master/docs/general/adapters.md#json. It should be updated to make known, in brief, it *is* possible to override the root key in the ways mentioned in this issue and link to the other guide sections:

[...]

PR #1761, opened 2016-05-31
Fix root key documentation

### LINKED ISSUE #1486: Override root when using JSON adapter is possible
The [documentation](https://github.com/rails-api/active_model_serializers/blob/master/docs/general/adapters.md#json) for the JSON adapter say:

> The root key can't be overridden, and will be derived from the resource being serialized.

However, within a controller, when I do
```ruby
render json: @my_resource_collection, root: 'foobar'
```

I am able to override the root to `foobar`. What is the documentation referring to?

groyoh:
#### Purpose

Fix wrong documentation stating that the root key cannot be overridden for the JSON adapter.

#### Related GitHub issues

Closes #1486 and #1756 .

#### Additional information

@ehannes @AlexVPopov does this sounds better to you?

### DIFF
--- a/docs/general/adapters.md
+++ b/docs/general/adapters.md
@@ -67,9 +67,11 @@ Doesn't follow any specific convention.
 
 ### JSON
 
-The response document always with a root key.
+The json response is always rendered with a root key.
 
-The root key **can't be overridden**, and will be derived from the resource being serialized.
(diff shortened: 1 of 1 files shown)


AlexVPopov:
:+1:

PR #1762, opened 2016-06-01
respect `config.default_includes` in #serializable_hash

y-yagi:
#### Purpose 

`default_includes` config was added in #1426.
However, it has not been used in `#serializable_hash`, want to use it even `#serializable_hash`. 

#### Changes 

Use `config.default_includes` to default value of `#serializable_hash`.


lib/active_model/serializer.rb
```diff
@@ -151,7 +151,7 @@ def success?
     #     serializer.as_json(include: { posts: { include: { comments: { only: :body } }, only: :title } })
     def serializable_hash(adapter_opts = nil)
       adapter_opts ||= {}
-      adapter_opts = { include: '*', adapter: :attributes }.merge!(adapter_opts)
+      adapter_opts = { include: config.default_includes, adapter: :attributes }.merge!(adapter_opts)
```
bf4: good call. would you mind adding a Changelog entry?

bf4: but should be `ActiveModelSerializer.default_includes`, not `config.default_includes`

y-yagi: You mean `ActiveModelSerializers.config.default_includes`?  `ActiveModelSerializers` module does not have `default_includes` method, it is an error to `ActiveModelSerializers.default_includes`.

bf4:
I merged some big refactoring in this area recently. Take a look and let me know how I can help with this PR.

y-yagi:
Thank you for pointing out!
After confirming master, it seems already to use `config.default_includes` in #serializable_hash. 
(I think that it's capable of influence of #1766.)
Therefore, I will close this PR. Thanks!

PR #1768, opened 2016-06-01
Fix docs for deserialization relationships

soberstadt:
#### Purpose

I found in the tests that deserialization expects relationships to be under a `relationships` key.
https://github.com/rails-api/active_model_serializers/blob/master/test/action_controller/json_api/deserialization_test.rb#L68

#### Changes

One key in the example document to be deserialized, from `associations` to `relationships`.


NullVoxPopuli:
Good catch!

PR #1777, opened 2016-06-06
Added ability for type to take a proc or a block

NullVoxPopuli:
#### Purpose

I want to be able to dynamically set the type, especially for polymorphic associations.

``` ruby
class BlockTypeSerializer < ActiveModel::Serializer
   type proc { |object|
     "block-#{object.class.model_name}"
  }
end
```
#### Caveats

So for, only a proc -- we should unify all our apis with block options, though, yeah?

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -188,9 +188,7 @@ module ActiveModel
 
     # Used by adapter as resource root.
     def json_key
-      root ||
-        (_type && _type.is_a?(Proc) ? _type.call(object) : _type) ||
-        object.class.model_name.to_s.underscore
+      root || _type || object.class.model_name.to_s.underscore
     end
 
     def read_attribute_for_serialization(attr)
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -188,7 +188,9 @@ module ActiveModel
 
     # Used by adapter as resource root.
     def json_key
-      root || _type || object.class.model_name.to_s.underscore
+      root ||
+        (_type && _type.is_a?(Proc) ? _type.call(object) : _type) ||
+        object.class.model_name.to_s.underscore
     end
 
     def read_attribute_for_serialization(attr)
--- a/lib/active_model/serializer/type.rb
+++ b/lib/active_model/serializer/type.rb
@@ -26,14 +26,7 @@ module ActiveModel
         #     type proc { |object| object.class.name }
         #
         def type(type = nil, &block)
-          type = block if block_given?
-          return unless type
-
-          self._type = if type.respond_to?(:call) && type.arity == 1
-                         type
-                       else
-                         type.to_s
-                       end
+          self._type = block || type
         end
       end
     end
--- a/lib/active_model/serializer/type.rb
+++ b/lib/active_model/serializer/type.rb
@@ -16,8 +16,13 @@ module ActiveModel
         # @example
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     type 'authors'
+        #
+        # @example
+        # TODO: actually do block, not Proc
+        #   class AdminAuthorSerializer < ActiveModel::Serializer
+        #     type { |object| object.class.name }
         def type(type)
-          self._type = type && type.to_s
+          self._type = type.is_a?(Proc) ? type : type.to_s if type
         end
       end
     end
--- a/lib/active_model/serializer/type.rb
+++ b/lib/active_model/serializer/type.rb
@@ -21,8 +21,15 @@ module ActiveModel
         # TODO: actually do block, not Proc
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     type { |object| object.class.name }
-        def type(type)
-          self._type = type.is_a?(Proc) ? type : type.to_s if type
+        def type(type = nil, &block)
+          type = block if block_given?
+          return unless type
+
+          self._type = if type.respond_to?(:call) && type.arity == 1
+                         type
+                       else
+                         type.to_s
+                       end
         end
       end
     end
--- a/lib/active_model/serializer/type.rb
+++ b/lib/active_model/serializer/type.rb
@@ -18,9 +18,13 @@ module ActiveModel
         #     type 'authors'
         #
         # @example
-        # TODO: actually do block, not Proc
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     type { |object| object.class.name }
+        #
+        # @example
+        #   class AdminAuthorSerializer < ActiveModel::Serializer
+        #     type proc { |object| object.class.name }
+        #
         def type(type = nil, &block)
           type = block if block_given?
           return unless type
--- a/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
+++ b/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
@@ -20,8 +20,8 @@ module ActiveModelSerializers
         private
 
         def type_for(serializer)
-          return serializer._type.call(serializer.object) if serializer._type.is_a?(Proc)
-          return serializer._type if serializer._type
+          return serializer._type.call(serializer.object) if serializer._type.respond_to?(:call)
+          return serializer._type.to_s if serializer._type
           if ActiveModelSerializers.config.jsonapi_resource_type == :singular
             serializer.object.class.model_name.singular
           else
--- a/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
+++ b/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
@@ -20,6 +20,7 @@ module ActiveModelSerializers
         private
 
         def type_for(serializer)
+          return serializer._type.call(serializer.object) if serializer._type.is_a?(Proc)
           return serializer._type if serializer._type
           if ActiveModelSerializers.config.jsonapi_resource_type == :singular
             serializer.object.class.model_name.singular
--- a/test/adapter/json_api/type_test.rb
+++ b/test/adapter/json_api/type_test.rb
@@ -28,9 +28,7 @@ module ActiveModel
           end
 
           class LambdaTypeSerializer < ActiveModel::Serializer
-            type do |object|
-              "lambda-#{object.class.model_name}"
-            end
+            type ->(object) { "lambda-#{object.class.model_name}" }
           end
 
           setup do
--- a/test/adapter/json_api/type_test.rb
+++ b/test/adapter/json_api/type_test.rb
@@ -15,8 +15,18 @@ module ActiveModel
             type :profile
           end
 
+          class BlockTypeSerializer < ActiveModel::Serializer
+            type proc { |object|
+              "block-#{object.class.model_name}"
+            }
+          end
+
           setup do
             @author = Author.new(id: 1, name: 'Steve K.')
+          end
+
+          test 'the type can take a block' do
+            assert_type(@author, 'block-author', serializer: BlockTypeSerializer)
           end
 
           def test_config_plural
--- a/test/adapter/json_api/type_test.rb
+++ b/test/adapter/json_api/type_test.rb
@@ -15,14 +15,24 @@ module ActiveModel
             type :profile
           end
 
-          class BlockTypeSerializer < ActiveModel::Serializer
+          class ProcTypeSerializer < ActiveModel::Serializer
             type proc { |object|
-              "block-#{object.class.model_name}"
+              "proc-#{object.class.model_name}"
             }
+          end
+
+          class BlockTypeSerializer < ActiveModel::Serializer
+            type do |object|
+              "block-#{object.class.model_name}"
+            end
           end
 
           setup do
             @author = Author.new(id: 1, name: 'Steve K.')
+          end
+
+          test 'the type can take a proc' do
+            assert_type(@author, 'proc-author', serializer: ProcTypeSerializer)
           end
 
           test 'the type can take a block' do
--- a/test/adapter/json_api/type_test.rb
+++ b/test/adapter/json_api/type_test.rb
@@ -27,6 +27,12 @@ module ActiveModel
             end
           end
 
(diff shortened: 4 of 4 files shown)


lib/active_model/serializer.rb
```diff
@@ -188,7 +188,9 @@ def as_json(adapter_opts = nil)
 
     # Used by adapter as resource root.
     def json_key
-      root || _type || object.class.model_name.to_s.underscore
+      root ||
+        (_type && _type.is_a?(Proc) ? _type.call(object) : _type) ||
+        object.class.model_name.to_s.underscore
     end
```
bf4: let's not complicate the code.  You can already pass in `root` or override `json_key` or `_type`. Better just to document this (esp until we do the clean up code for this... i.e. make it simpler!)

NullVoxPopuli: this particular change actually isn't needed. yay!!! :-)

lib/active_model/serializer/type.rb
```diff
@@ -16,8 +16,13 @@ module ClassMethods
...
         #   class AdminAuthorSerializer < ActiveModel::Serializer
         #     type 'authors'
+        #
+        # @example
+        # TODO: actually do block, not Proc
+        #   class AdminAuthorSerializer < ActiveModel::Serializer
+        #     type { |object| object.class.name }
         def type(type)
-          self._type = type && type.to_s
+          self._type = type.is_a?(Proc) ? type : type.to_s if type
```
bf4: might as well check if it responds to `:call` (and arity is 0)

bf4: man, all these blocks have got to be causing memory leaks

bf4: see [CollectionSerializer#json_key](https://github.com/rails-api/active_model_serializers/pull/1618/files#diff-78e515848bf3a42cc6aa1e892f6798caR30) for some more discussion about what I think would be better to do first

groyoh: @bf4 why do you think it causes memory leak? I assume you would have a limited amount of serializer and therefore a limited amount of proc and contexts too.

bf4: could be.  I've never really investigated it

NullVoxPopuli:
@bf4, I changed to a duck-type-y check for the non string/symbol type.

@bf4 & @groyoh : added some additional stuff to support actual blocks (rather than just procs)

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
```diff
@@ -20,6 +20,7 @@ def as_json
         private
 
         def type_for(serializer)
+          return serializer._type.call(serializer.object) if serializer._type.is_a?(Proc)
```
NullVoxPopuli: is there a scenario in which a lambda / Proc / block wouldn't respond to call or be of type Proc?

lib/active_model/serializer/type.rb
```diff
@@ -16,8 +16,24 @@ module ClassMethods
...
+        #     type proc { |object| object.class.name }
+        #
+        def type(type = nil, &block)
+          type = block if block_given?
+          return unless type
+
+          self._type = if type.respond_to?(:call) && type.arity == 1
+                         type
+                       else
+                         type.to_s
```
beauby: The `.to_s` part doesn't really belong here I think (which would allow to simplify this method).

beauby: You could put it [there](https://github.com/rails-api/active_model_serializers/pull/1777/files#diff-65abd6a4fd7fd55fe487ff138e0841dbR24). Thus, this method becomes:
```ruby
def type(type = nil, &block)
  self._type = block || type
end

File: lib/active_model_serializers/adapter/json_api/resource_identifier.rb (hunk 3 above)
Comment (beauby) [inline lib/active_model_serializers/adapter/json_api/resource_identifier.rb]: I don't think so, but since we fundamentally do not care whether it is a `Proc` or not as long as it is a callable, we might as well adopt duck typing here.

File: test/adapter/json_api/type_test.rb [hunk 5]
```diff
@@ -15,10 +15,30 @@ class SymbolTypeSerializer < ActiveModel::Serializer
...
+              "proc-#{object.class.model_name}"
+            }
+          end
+
+          class BlockTypeSerializer < ActiveModel::Serializer
+            type do |object|
+              "block-#{object.class.model_name}"
+            end
+          end
+
```
beauby: Could you add a test with a lambda?
```ruby
type ->(object) { "lambda-#{object.class.model_name}" }
```

lib/active_model/serializer/type.rb
NullVoxPopuli: brilliant!

test/adapter/json_api/type_test.rb
NullVoxPopuli: sure thing

NullVoxPopuli: added!

NullVoxPopuli: errr, copy paste error. one sec

NullVoxPopuli: now there is a lambda test ^_~

bf4:
@NullVoxPopuli I'd really appreciate if this PR simplified how we set root, rather than add another way... I mean, I see how it would make it consistent with other parts of the code base, but it's been taken up your time and without cleanup, doesn't add much value, imo.

As I wrote in the CollectionSerializer

```ruby
     # TODO: unify naming of root, json_key, and _type.  Right now, a serializer's
      # json_key comes from the root option or the object's model name, by default.
      # But, if a dev defines a custom `json_key` method with an explicit value,
      # we have no simple way to know that it is safe to call that instance method.
      # (which is really a class property at this point, anyhow).
      # rubocop:disable Metrics/CyclomaticComplexity
      # Disabling cop since it's good to highlight the complexity of this method by
      # including all the logic right here.
      def json_key
        return root if root
        # 1. get from options[:serializer] for empty resource collection
        key = object.empty? &&
          (explicit_serializer_class = options[:serializer]) &&
          explicit_serializer_class._type
        # 2. get from first serializer instance in collection
        key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
        # 4. key may be nil for empty collection and no serializer option
        key && key.pluralize
      end
      # rubocop:enable Metrics/CyclomaticComplexity
```

and then in the single resource Serializer

```ruby
    with_options instance_writer: false, instance_reader: true do |serializer|
      serializer.class_attribute :_type # @api private    
   end

        # Set the JSON API type of a serializer.
        # @example
        #     type 'authors'
        def self.type(type)
          self._type = type && type.to_s
        end

    # Used by adapter as resource root.
    def json_key
      root || _type || object.class.model_name.to_s.underscore
    end
```

and in Adapter::Base

```ruby
def root
  serializer.json_key.to_sym if serializer.json_key
end
```

And docs:

- [json_key](https://github.com/rails-api/active_model_serializers/blob/835aad3c7493f7cc33fe624434cf4c38082f3bdd/docs/general/serializers.md#json_key)
- [root](https://github.com/rails-api/active_model_serializers/blob/835aad3c7493f7cc33fe624434cf4c38082f3bdd/docs/general/serializers.md#root)
- [overriding the root key](https://github.com/rails-api/active_model_serializers/blob/835aad3c7493f7cc33fe624434cf4c38082f3bdd/docs/general/rendering.md#overriding-the-root-key)
- [how to add a root key](https://github.com/rails-api/active_model_serializers/blob/db87f8d85e7a46ef195a555cb7e2509f1affc622/docs/howto/add_root_key.md)

NullVoxPopuli:
@bf4 thanks to @beauby, ActiveModel::Serializer::Type#type is now
```ruby
        def type(type = nil, &block)
          self._type = block || type
        end
```

and then the only other change is in AMS::Adapter::JsonApi::ResourceIdentifier#type_for:
```ruby
          return serializer._type.call(serializer.object) if serializer._type.respond_to?(:call)
          return serializer._type.to_s if serializer._type
```

which is pretty stinkin simple.

But, I can see that adding this would make the collection_serializer#json_key even more complicated. (one additional line to evaluate `key` as a proc if it responds to `call`. (So currently, if someone uses a block type and tries to render a collection using that serializer, it's probably going to error (so we're missing a test)).

@bf4 based on your comment on `json_key`, is the concern more the naming? or the complexity? (or both?).

for naming, is there anything preventing us from just always using `_type` for what the serializer thinks its type is, and maybe `_evaluated_type` to do all the logic (current json_key in collection_serializer). Like, is the issue just that we're _using_ `json_key` as a method? I know we want as little possibility for method collisions as possible. 

But idk, the collection_serializer complexity needs to be the way it is doesn't it? maybe it could just change to:

```ruby
    def _evaluated_type
        return root if root

        # 1. get from options[:serializer] for empty resource collection
        key = object.empty? &&
          (explicit_serializer_class = options[:serializer]) &&
          explicit_serializer_class._evaluate_type(object)

        # 2. get from first serializer instance in collection
        key ||= (serializer = serializers.first) && serializer._evaluated_type

        key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil

        # 4. key may be nil for empty collection and no serializer option
        key && key.pluralize
      end
```

This would mean we'd need to add a class method, `_evaluate_type` so that we can use blocks with passed serializers, as well as make use of it in serializer instances, and memoize or something.

```ruby
# serializer
def self._evaluate_type(object)
  return _type.call(object) if _type.respond_to?(:call)
  _type.to_s
end

def _evaluated_type
  @_evaluated_type ||= self.class._evaluate_type(object)
end
```

? 

I have no idea. This was pretty much stream of consciousness.

lib/active_model/serializer/type.rb
```diff
@@ -16,8 +16,17 @@ module ClassMethods
...
+        # @example
+        #   class AdminAuthorSerializer < ActiveModel::Serializer
+        #     type { |object| object.class.name }
+        #
+        # @example
+        #   class AdminAuthorSerializer < ActiveModel::Serializer
+        #     type proc { |object| object.class.name }
+        #
+        def type(type = nil, &block)
+          self._type = block || type
```
bf4: lost the `type && type.to_s`

bf4: I mean, it's in `return serializer._type.to_s if serializer._type` below, but why not just do the conditional once :)

bf4: also, how does this not blow up the serializer's json_key?

bf4:
@NullVoxPopuli 

> based on your comment on json_key, is the concern more the naming? or the complexity? (or both?).

complexity.  there shouldn't be so many code paths to do the same thing. 

but the complexity is due to all the different names for the same thing.  we need to pick a name, root, type, json_key, whatever, and just condense all of it, and just make the distinction between a class attribute and an instance attribute

stuffing the logic in `_evaluated_type` doesn't solve anything. just moving the chairs on the 🚢  :)

lib/active_model/serializer/type.rb
NullVoxPopuli: > also, how does this not blow up the serializer's json_key?
 
maybe we having a missing test?

> I mean, it's in return serializer._type.to_s if serializer._type below, but why not just do the conditional once :)

yeah, idk .I could do `block || type.try(:to_s)` to compactly achieve the same, I think

bf4: you put a `try` in there and I'll tell you to change it, and I don't mean to `.try!`

NullVoxPopuli:
can we just decide on _type? since that's the convention that JSON API uses? (`type`) (and because we all <3 JSON API)

lib/active_model/serializer/type.rb
NullVoxPopuli: lol. maybe we should figure out the bigger problem first of json_key, root, _type? :-)
I'll update / redo this PR when we figure all that out.

bf4:
> can we just decide on _type? since that's the convention that JSON API uses? (type) (

I'm generally interested in pushing the JSON API naming/objects/fields/options into the other adapters, etc. e.g. no reason  JSON adapter can't use fields or include

lib/active_model/serializer/type.rb
bf4: oops, missed smiley, sorry.  

This should probably just be a `Type` object that descends from our Field value object (iirc)

NullVoxPopuli: I looked at that when I was initially implemented this, but to convey the idea / new api, I wanted as few code changes as possible.

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
```diff
@@ -20,7 +20,8 @@ def as_json
         private
 
         def type_for(serializer)
-          return serializer._type if serializer._type
+          return serializer._type.call(serializer.object) if serializer._type.respond_to?(:call)
```
groyoh: I think using `serializer.instance_eval(serializer._type)` would be more appropriate for instance if you want to use a function defined on the serializer or something similar and is closer to the other block usage we have.

tommiller1:
Is this getting merged any time soon? I need this in my base controller to get rid of ugly 'namespace_v2_modelname' root names.

bf4:
@tommiller1 the root is already configurable in I think 4 ways, which is why I commented https://github.com/rails-api/active_model_serializers/pull/1777#issuecomment-224308628

tommiller1:
Nice thanks. Fixed it with:
def root
      object.class.model_name.param_key
end

tommiller1:
How do you specify root name for an empty array?

NullVoxPopuli:
TODO: I need to convert the type part of the serializer API to a similar object to Field, so that it can have similar functionality without all this extra logic.

groyoh:
@NullVoxPopuli do you mind if I take over this PR and update it?

NullVoxPopuli:
Go for it, do you need access?

groyoh:
No it should be fine.

NullVoxPopuli:
Cleaning up my PRs. If this work is still desired, feel free to cherry-pick

PR #1624, opened 2016-03-27
Silence @_routes warnings

bf4:

PR #1778, opened 2016-06-06
initial support for hypertext application language

maletor:
Adds HAL as an adapter. It is not completely finished but it should be considered a start. If this is the right direction let me know and I can finish it off with some tests.

### DIFF
--- a/lib/active_model_serializers/adapter/hypertext_application_language.rb
+++ b/lib/active_model_serializers/adapter/hypertext_application_language.rb
@@ -1,0 +1,74 @@
+module ActiveModelSerializers
+  module Adapter
+    class HypertextApplicationLanguage < Base
+      extend ActiveSupport::Autoload
+      autoload :Link
+      autoload :PaginationLinks
+
+      def initialize(serializer, options = {})
+        super
+        @include_directive = JSONAPI::IncludeDirective.new(
+          options[:include], allow_wildcard: true)
+      end
+
+      def serializable_hash(options = nil)
+        options = serialization_options(options)
+
+        if serializer.respond_to?(:each)
+          serializable_hash_for_collection(options)
+        else
+          serializable_hash_for_single_resource(options)
+        end
+      end
+
+      private
+
+      def serializable_hash_for_collection(options)
+        hash = {}
+        hash[:_links] = { self: { href: instance_options.fetch(:serialization_context).request_url } }
+        hash[:_links].update(pagination_links_for(serializer))
+        hash[:_embedded] = {}
+        hash[:_embedded][:items] =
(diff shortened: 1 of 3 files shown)


lib/active_model_serializers/adapter/hypertext_application_language/link.rb
```diff
@@ -0,0 +1,77 @@
+module ActiveModelSerializers
+  module Adapter
+    class HypertextApplicationLanguage
+      class Link
+        include SerializationContext::UrlHelpers
+
+        def initialize(serializer, value)
+          # handles warning actionpack-4.0.13/lib/action_dispatch/routing/route_set.rb
+          # warning: instance variable @_routes not initialized
+          @_routes ||= nil
```
bf4: I thought I fixed that at some point... owait, maybe you based this on that? https://github.com/rails-api/active_model_serializers/pull/1624/files

bf4:
I'm by default not interested in adding anything to AMS that increases its maintenance burden right now, but this looks pretty clean.  👍 If you're willing to help maintain it, I'd be happy to work with you to finish the PR. It's interesting to see the common code in here and the JsonApi links/paginationlinks that could be extracted.

lib/active_model_serializers/adapter/hypertext_application_language.rb
```diff
@@ -0,0 +1,74 @@
...
+        end
+      end
+
+      def resource_object_for(options)
+        if serializer.class.cache_enabled?
+          @cached_attributes.fetch(serializer.cache_key(self)) do
+            serializer.cached_fields(options[:fields], self)
+          end
+        else
+          serializer.cached_fields(options[:fields], self)
```
bf4: just a note that anything related to caching is in flux right now, and the direction I'm think is that caching should decorate the serialization process so that we can pipeline it.  So, maybe it's best just to leave the caching stuff out for now, or implement it outside of the adapter?

jgrau:
From reading this it looks like the missing pieces are: 

1. Remove the caching behaviour
2. Add tests

Anything else?

bf4:
@jgrau Someone to maintain it... I'd like to schedule this for 0.99 or 1.0.0

jgrau:
@maletor would you be interested in helping maintaining this?

remear:
I'm thinking I'd like to see this as a `active_model_serializers_adapter_hal` gem so it exists for others' use, but can be maintained separately from AMS. When 0.99/1.0.0 are on the horizon we can talk about pulling the HAL adapter into the core if it still makes sense.

maletor:
I'll see about creating that gem. Probably in no short order, though.

PR #1797, opened 2016-06-11
Add include_data :via_include_parameter

richmolj:
@NullVoxPopuli @bf4, taking a final stab at simplifying the code from https://github.com/rails-api/active_model_serializers/pull/1720. I created a separate PR since the original solves more issues and I didn't want to blow away the tests if you wanted to use them. Feel free to just close this PR if you still want to go in another direction, don't want this to seem pushy 😄 I believe this is the simplest possible solution, so maybe helpful to understand the problem if nothing else.

This now adds only one new variable name, `include_slice`. This is the slice of the include directive hash being  processed by the current serializer. All the changes to `lib/active_model/serializer/associations.rb` and `lib/active_model_serializers/adapter/json_api.rb` are just passing that slice around so we can access it when we finally decide whether or not to have the `data` key in the response. The changes to `lib/active_model/serializer/reflection.rb` access that variable to make the decision.

Why do we need the current slice instead of the whole directive? Because multiple entities can have the same relationship, e.g. `/blogs/?include=posts.tags,tags` should include both blog tags and post tags, but `/blogs/?include=posts.tags` should only include post tags

Users now opt-in to this pattern **using the current DSL**. The current DSL conditionally includes the `data` key in the response via `include_data true/false`:

```ruby
  has_many :comments do
    include_data false
  end
end
```

We now add another option `:via_include_parameter`. This means we only include the `data` key when the relationship is being sideloaded:

```ruby
  has_many :comments do
    include_data :via_include_parameter
  end
end
```

### DIFF
--- a/lib/active_model/serializer/concerns/associations.rb
+++ b/lib/active_model/serializer/concerns/associations.rb
@@ -83,7 +83,8 @@ module ActiveModel
       #   +default_include_directive+ config value when not provided)
       # @return [Enumerator<Association>]
       #
-      def associations(include_directive = ActiveModelSerializers.default_include_directive)
+      def associations(include_directive = ActiveModelSerializers.default_include_directive, include_slice = nil)
+        include_slice ||= include_directive
         return unless object
 
         Enumerator.new do |y|
@@ -91,7 +92,8 @@ module ActiveModel
             next if reflection.excluded?(self)
             key = reflection.options.fetch(:key, reflection.name)
             next unless include_directive.key?(key)
-            y.yield reflection.build_association(self, instance_options)
+
+            y.yield reflection.build_association(self, instance_options, include_slice)
           end
         end
       end
--- a/lib/active_model/serializer/concerns/configuration.rb
+++ b/lib/active_model/serializer/concerns/configuration.rb
@@ -22,6 +22,7 @@ module ActiveModel
         config.default_includes = '*'
         config.adapter = :attributes
         config.key_transform = nil
+        config.jsonapi_pagination_links_enabled = true
         config.jsonapi_resource_type = :plural
         config.jsonapi_namespace_separator = '-'.freeze
         config.jsonapi_version = '1.0'
--- a/lib/active_model/serializer/concerns/configuration.rb
+++ b/lib/active_model/serializer/concerns/configuration.rb
@@ -29,6 +29,7 @@ module ActiveModel
         # Make JSON API top-level jsonapi member opt-in
         # ref: http://jsonapi.org/format/#document-top-level
         config.jsonapi_include_toplevel_object = false
+        config.include_data_default = true
 
         config.schema_path = 'test/support/schemas'
       end
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -37,7 +37,7 @@ module ActiveModel
       def initialize(*)
         super
         @_links = {}
-        @_include_data = true
+        @_include_data = Serializer.config.include_data_default
         @_meta = nil
       end
 
@@ -69,17 +69,15 @@ module ActiveModel
       #       Blog.find(object.blog_id)
       #     end
       #   end
-      def value(serializer)
+      def value(serializer, include_slice)
         @object = serializer.object
         @scope = serializer.scope
 
-        if block
-          block_value = instance_exec(serializer, &block)
-          if block_value != :nil
-            block_value
-          elsif @_include_data
-            serializer.read_attribute_for_serialization(name)
-          end
+        block_value = instance_exec(serializer, &block) if block
+        return unless include_data?(include_slice)
+
+        if block && block_value != :nil
+          block_value
         else
           serializer.read_attribute_for_serialization(name)
         end
@@ -106,11 +104,11 @@ module ActiveModel
       #
       # @api private
       #
-      def build_association(parent_serializer, parent_serializer_options)
-        association_value = value(parent_serializer)
+      def build_association(parent_serializer, parent_serializer_options, include_slice = {})
+        association_value = value(parent_serializer, include_slice)
         reflection_options = options.dup
         serializer_class = parent_serializer.class.serializer_for(association_value, reflection_options)
-        reflection_options[:include_data] = @_include_data
+        reflection_options[:include_data] = include_data?(include_slice)
         reflection_options[:links] = @_links
         reflection_options[:meta] = @_meta
 
@@ -136,6 +134,14 @@ module ActiveModel
       attr_accessor :object, :scope
 
       private
+
+      def include_data?(include_slice)
+        if @_include_data == :if_sideloaded
+          include_slice.key?(name)
+        else
+          @_include_data
+        end
+      end
 
       def serializer_options(parent_serializer, parent_serializer_options, reflection_options)
         serializer = reflection_options.fetch(:serializer, nil)
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -235,17 +235,17 @@ module ActiveModelSerializers
         @primary = []
         @included = []
         @resource_identifiers = Set.new
-        serializers.each { |serializer| process_resource(serializer, true) }
+        serializers.each { |serializer| process_resource(serializer, true, @include_directive) }
         serializers.each { |serializer| process_relationships(serializer, @include_directive) }
 
         [@primary, @included]
       end
 
-      def process_resource(serializer, primary)
+      def process_resource(serializer, primary, include_slice = {})
         resource_identifier = ResourceIdentifier.new(serializer, instance_options).as_json
         return false unless @resource_identifiers.add?(resource_identifier)
 
-        resource_object = resource_object_for(serializer)
+        resource_object = resource_object_for(serializer, include_slice)
         if primary
           @primary << resource_object
         else
@@ -255,21 +255,21 @@ module ActiveModelSerializers
         true
       end
 
-      def process_relationships(serializer, include_directive)
-        serializer.associations(include_directive).each do |association|
-          process_relationship(association.serializer, include_directive[association.key])
+      def process_relationships(serializer, include_slice)
+        serializer.associations(include_slice).each do |association|
+          process_relationship(association.serializer, include_slice[association.key])
         end
       end
 
-      def process_relationship(serializer, include_directive)
+      def process_relationship(serializer, include_slice)
         if serializer.respond_to?(:each)
-          serializer.each { |s| process_relationship(s, include_directive) }
+          serializer.each { |s| process_relationship(s, include_slice) }
           return
         end
         return unless serializer && serializer.object
-        return unless process_resource(serializer, false)
+        return unless process_resource(serializer, false, include_slice)
 
-        process_relationships(serializer, include_directive)
+        process_relationships(serializer, include_slice)
       end
 
       # {http://jsonapi.org/format/#document-resource-object-attributes Document Resource Object Attributes}
@@ -293,7 +293,7 @@ module ActiveModelSerializers
       end
 
       # {http://jsonapi.org/format/#document-resource-objects Document Resource Objects}
-      def resource_object_for(serializer)
+      def resource_object_for(serializer, include_slice = {})
         resource_object = serializer.fetch(self) do
           resource_object = ResourceIdentifier.new(serializer, instance_options).as_json
 
@@ -304,7 +304,7 @@ module ActiveModelSerializers
         end
 
         requested_associations = fieldset.fields_for(resource_object[:type]) || '*'
-        relationships = relationships_for(serializer, requested_associations)
+        relationships = relationships_for(serializer, requested_associations, include_slice)
         resource_object[:relationships] = relationships if relationships.any?
 
         links = links_for(serializer)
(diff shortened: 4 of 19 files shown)


lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -235,17 +235,17 @@ def resource_objects_for(serializers)
...
         @resource_identifiers = Set.new
-        serializers.each { |serializer| process_resource(serializer, true) }
+        serializers.each { |serializer| process_resource(serializer, true, @include_directive) }
         serializers.each { |serializer| process_relationships(serializer, @include_directive) }
 
         [@primary, @included]
       end
 
-      def process_resource(serializer, primary)
+      def process_resource(serializer, primary, include_slice = {})
```
bf4: directive now aka slice

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -432,12 +432,12 @@ def resource_object_for(serializer)
       #     id: 'required-id',
       #     meta: meta
       #   }.reject! {|_,v| v.nil? }
-      def relationships_for(serializer, requested_associations)
+      def relationships_for(serializer, requested_associations, include_slice)
         include_directive = JSONAPI::IncludeDirective.new(
           requested_associations,
```
bf4: now have slice and directive

bf4:
I feel like we shouldn't need to introduce the include_slice directive (i.e have two include directives) for this to work.  Otherwise, the public interface change is good.

test/adapter/json_api/include_param_test.rb
```diff
@@ -0,0 +1,166 @@
...
+          def setup
+            IncludeParamAuthorSerializer.comment_loader = Class.new(CustomCommentLoader).new
+            @tag = Tag.new(id: 1337, name: 'mytag')
+            @author = IncludeParamAuthor.new(
+              id: 1337,
+              tags: [@tag]
+            )
+          end
+
+          def test_relationship_not_loaded_when_not_included
```
bf4: I think this test is the central feature in this PR, at least for the JSONAPI adapter

richmolj:
> I feel like we shouldn't need to introduce the include_slice directive (i.e have two include directives) for this to work. Otherwise, the public interface change is good.

I feel you, but not sure how to avoid it. If you have an `include_directive` like `{ posts: { comments: :author }, reviews: { comments: {} }`, and you only ever know about the entire directive and not the slice currently being processed, how would `CommentSerializer` know if `author` data should be populated? It's conditional depending on the nesting level.

richmolj:
> It should be modifying the include and fields params as each serializer calls the next, no?

@bf4 This may be a matter of phrasing, but to be clear - since `fields` and `include` live on the **adapter**, the adapter recursively iterates through associations (one serializer doesn't call the next - I wish they did, the issue would likely be avoided). During iteration, since we're only dealing with a slice of the overall directive, I renamed the parameter named `include_slice` to make this more obvious.

I feel like the current PR is already doing what you are suggesting, so let me try to rephrase - I think the confusion is with the current codebase, not this PR. `include_directive` means **3 different things** in the current codebase:

* The `@include_directive` - the overall directive passed to `render`
* The `include_directive` in parameters that I now refer to as `include_slice`. This is the slice of the `@include_directive` we're currently iterating over.
* The `include_directive` in `#relationships_for` is a totally separate concept, derived from the `fields` param and used to iterate over associations.

IOW, you could actually refer to it as `include_slice` or `include_directive` **everywhere**...except the [current implementation of #relationships_for](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L435-L450) screws us up. This introduces [another variable named 'include_directive'](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L436-L439) that is **not the include directive we've been dealing with this whole time**. It's a separate directive [derived from the 'fields' parameter](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L306-L307). It uses this new directive [to iterate over the associations](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L440) - and we can't pass only the current slice (you'll get a ton of test failures as we often need to process relationships even when `include` is not present). But we need that current slice down the line in `reflection.rb`, which is why we end up referring to both `include_directive` and `include_slice` in the same method and confusion ensues.

Hope this helps - as always, your time and effort here are much appreciated.

NullVoxPopuli:
Some thoughts:

I don't think `fields` should be recursive, unless we added the same notation that we do for include (`'field1,field2.subfield1`).

What if the API was this:

```ruby
# current behavior
has_many :comments


has_many :comments do
  # must explicitly be included
  # or maybe something related to the resource identifier object?
  # http://jsonapi.org/format/#document-resource-identifier-objects
  skip_unless_included
end
```
I feel like `include_data` is a little ambiguous, as what we want to avoid is the Resource Identifier object isn't it?



I don't like the additional `include_` variable.

bf4:
Hmm, fields_directive...

B mobile phone

NullVoxPopuli:
> Hmm, fields_directive...

at least initially, we could just have fields span attributes and relationships, flatly. No need for a fields directive, I don't think. at least.. not yet.

richmolj:
@NullVoxPopuli correct me if I'm wrong, but `fields` is not currently recursive. I'm actually not sure how/why it would ever be - JSONAPI [fields](http://jsonapi.org/format/#fetching-sparse-fieldsets) is flat and namespaced, e.g. ` /articles?include=author&fields[articles]=title,body&fields[people]=name`. Something like `field1.subfield1` would deviate from the spec AFAIK. [#process_relationships](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L258-L261) is the source of recursion here.

> skip_unless_included

:+1: to `include_data` being ambiguous, but keep in mind the current behavior of this gem uses `include_data false` independent of the `?include` parameter. So replacing with `skip_unless_included` probably doesn't match well for the current use cases you guys are supporting.

> I don't like the additional include_ variable.

I think that is something everyone, including me, agrees on :joy: My thing is: what's the alternative? I can't see how you avoid it without fundamentally changing how `adapter/json_api.rb` works (avoiding the recursion and switching to a serializers-call-serializers pattern).

NullVoxPopuli:
> JSONAPI fields is flat and namespaced

I think we're on the same page here.

> So replacing with skip_unless_included probably doesn't match well for the current use cases you guys are supporting.

how do you mean?

> avoiding the recursion and switching to a serializers-call-serializers pattern

ah, I see. that is obnoxious then.

bf4:
`skip_unless_included` should be default for what jsonapi passes into serializer 'attributes'

bf4:
i.e.

```ruby
# https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb
      resource = fetch_attributes(options[:fields], cached_attributes, adapter_instance)
# https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/attributes.rb
 next if attr.excluded?(self)
            next unless requested_attrs.nil? || requested_attrs.include?(key)
```

richmolj:
@NullVoxPopuli regarding `skip_unless_included` and the current behavior of the gem: Assuming people are using `include_data false` right now with the current gem, they are excluding the `data` key **independent of the ?include param**. So renaming this option something specific to the `?include` param (`skip_unless_included`) seems off to me. That said, I agree with your general thought and for my personal use of AMS I'd be :+1:

NullVoxPopuli:
> skip_unless_included should be default for what jsonapi passes into serializer 'attributes'

@bf4, do you mean we _shouldn't_ include the id/type for relationships (by default) (no options passed)

> Assuming people are using include_data false right now with the current gem, they are excluding the data key

@richmolj  I actually didn't know that was a thing. ha

richmolj:
@bf4 while I agree `skip_unless_included` should be a default, I am not sure 'attributes' is the place for it, for two reasons:

One - I *think* the payload output would be something like:

```
{
  relationships: {
    author: {
      data: {
        id: 1, type: 'people', attributes: {}
      }
    }
  }
}
```

Which means the resource identifiers would be included, a DB hit is still happening, and we haven't solved the issue.

That said, even if we got the payload to look like this:

```
{
  relationships: {
    author: {
      data: {}
    }
  }
}
```

It's still a no-go. This is telling clients "this object does not have an author", versus "this object may or may not have an author, we're just not including author information in the response".

bradens:
I need this so bad, is there anything I can do to help get something to solve this problem released?

richmolj:
@bf4 - per our discussion yesterday:

* Renamed `via_include_parameter` to `if_sideloaded`, so it now reads `include_data :if_sideloaded`.
* Added `ActiveModel::Serializer.config.include_data_default` configuration option.

Though we both agree this should be the default, I didn't change the current default out of backwards compatibility concerns. If you're OK with it I'm OK with it, though.

lib/active_model/serializer/concerns/configuration.rb
```diff
@@ -29,6 +29,7 @@ def config.array_serializer
         # Make JSON API top-level jsonapi member opt-in
         # ref: http://jsonapi.org/format/#document-top-level
         config.jsonapi_include_toplevel_object = false
+        config.include_data_default = true
```
bf4: jsonapi adapter should default to false, right?

lib/active_model/serializer/reflection.rb
```diff
@@ -69,19 +69,17 @@ def include_data(value = true)
...
-          if block_value != :nil
-            block_value
-          elsif @_include_data
-            serializer.read_attribute_for_serialization(name)
-          end
-        else
+        block_value = instance_exec(serializer, &block) if block
+        return unless include_data?(include_slice)
+
+        if [nil, :nil].include?(block_value)
```
bf4: nil is valid iirc

lib/active_model/serializer/reflection.rb
```diff
@@ -137,6 +135,14 @@ def build_association(parent_serializer, parent_serializer_options)
 
       private
 
+      def include_data?(include_slice)
+        if @_include_data == :if_sideloaded
+          include_slice.key?(name)
+        else
+          @_include_data
+        end
```
bf4: 👍

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -304,7 +304,7 @@ def resource_object_for(serializer)
         end
 
         requested_associations = fieldset.fields_for(resource_object[:type]) || '*'
```
bf4: urg, `*`

lib/active_model/serializer/concerns/configuration.rb
richmolj: current default is true. per the PR description, I vote we change that default but up to you guys

lib/active_model/serializer/concerns/associations.rb
```diff
@@ -83,15 +83,17 @@ def associate(reflection)
...
-      def associations(include_directive = ActiveModelSerializers.default_include_directive)
+      def associations(include_directive = ActiveModelSerializers.default_include_directive, include_slice = nil)
+        include_slice ||= include_directive
         return unless object
 
         Enumerator.new do |y|
           self.class._reflections.values.each do |reflection|
             next if reflection.excluded?(self)
             key = reflection.options.fetch(:key, reflection.name)
             next unless include_directive.key?(key)
```
bf4: in this context, when would `include_directive.key?(key)` not be equal to `include_slice.key?(key)`?

lib/active_model/serializer/reflection.rb
richmolj: great catch 👍 , made the change

lib/active_model/serializer/concerns/associations.rb
richmolj: This is one of the gross aspects of the current adapter but

* we get the [fields](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L306), defaulted to '*'.

* they end up in [relationships_for](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L435)

* that calls [#associations](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api.rb#L440) leading us to this code.

That's opposed to `include_slice`, which is always the slice of the *real* include directive passed to `render`.

In other words, the current code has two 'include directives'. One real one that is what you think it is, and a separate one that is derived from a relationship's `fields`

richmolj: So in this case, for instance, let's say your include directive was:

```ruby
{ posts: { comments: {} } }
```

When you are iterating the relations and you hit the `comments` portion, the `include_directive` (the current code) is actually '*', assuming no sparse fieldsets requested. But the `include_slice` would be {}.

kevintyll:
Is there any concern for ensuring that the relationship object contains at least of of `data`, `links`, or `meta` elements?
Per the spec, one is required, but if the association is set up with no link, or meta defined, and `include_data :if_sideloaded`, then the result will not be compliant?

Looking forward to this feature BTW.  I just implemented my own monkey patch to do exactly this.  This solution is better.

richmolj:
@kevintyll - definitely agree. If the relation is not influded you should see an empty meta key, eg 'comments: { meta: {} }'. I'm on mobile right now but I should add a test asserting that is the case.

lib/active_model_serializers/adapter/json_api.rb
NullVoxPopuli: :-(

bf4:
@richmolj issue title /description should also be updated, I think, no?

richmolj:
@bf4 agreed, will close this and submit a new one tonight.

richmolj:
Moved to https://github.com/rails-api/active_model_serializers/pull/1931

PR #1802, opened 2016-06-14
update integration documentation for ember and JSONAPI to include…

kyleshevlin:
…key_transform unaltered to match the attribute and relationship underscore transformations.

#### Purpose

Add missing info in integration documentation. If Ember serializers use underscores, key_transform needs to be set to unaltered.

#### Changes

Documentation is updated.

#### Caveats

None.

### DIFF
--- a/docs/integrations/ember-and-json-api.md
+++ b/docs/integrations/ember-and-json-api.md
@@ -20,36 +20,29 @@ To solve this, in Ember, both the adapter and the serializer will need some modi
 First, set the adapter type in an initializer file:
 
 ```ruby
-# config/initializers/ams_config.rb
+# config/initializers/active_model_serializers.rb
 ActiveModelSerializers.config.adapter = :json_api
 ```
 
 or:
 
 ```ruby
-# config/initializers/ams_config.rb
+# config/initializers/active_model_serializers.rb
 ActiveModelSerializers.config.adapter = ActiveModelSerializers::Adapter::JsonApi
 ```
 
-You will also want to set the `key_transform` to `:underscore` since you will adjust the attributes in your Ember serializer to use underscores instead of dashes later.
+You will also want to set the `key_transform` to `:unaltered` since you will adjust the attributes in your Ember serializer to use underscores instead of dashes later. You could also use `:underscore`, but `:unaltered` is better for performance.
 
 ```ruby
-# config/initializers/ams_config.rb
-ActiveModelSerializers.config.key_transform = :underscore
+# config/initializers/active_model_serializers.rb
+ActiveModelSerializers.config.key_transform = :unaltered
 ```
 
-There are multiple mimetypes for json that should all be parsed similarly, so
-in `config/initializers/mime_types.rb`:
+Lastly, in order to properly handle JSON API responses, we need to register a JSON API renderer, like so:
 
 ```ruby
-api_mime_types = %W(
-  application/vnd.api+json
-  text/x-json
-  application/json
-)
-
-Mime::Type.unregister :json
-Mime::Type.register 'application/json', :json, api_mime_types
+# config/initializers/active_model_serializers.rb
+require 'active_model_serializers/register_jsonapi_renderer'
 ```
 
 ### Adapter Changes
--- a/docs/integrations/ember-and-json-api.md
+++ b/docs/integrations/ember-and-json-api.md
@@ -17,8 +17,30 @@ To solve this, in Ember, both the adapter and the serializer will need some modi
 
 ### Server-Side Changes
 
-there are multiple mimetypes for json that should all be parsed similarly, so
+First, set the adapter type in an initializer file:
+
+```ruby
+# config/initializers/ams_config.rb
+ActiveModelSerializers.config.adapter = :json_api
+```
+
+or:
+
+```ruby
+# config/initializers/ams_config.rb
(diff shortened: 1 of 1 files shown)


docs/integrations/ember-and-json-api.md
```diff
@@ -17,8 +17,30 @@ To solve this, in Ember, both the adapter and the serializer will need some modi
...
+```
+
+or:
+
+```ruby
+# config/initializers/ams_config.rb
+ActiveModelSerializers.config.adapter = ActiveModelSerializers::Adapter::JsonApi
+```
+
+You will also want to set the `key_transform` to `:underscore` since you will adjust the attributes in your Ember serializer to use underscores instead of dashes later.
```
NullVoxPopuli: There actually isn't a reason to do this anymore, since ember accepts our default key transform without modification. It's still worth mentioning to make sure both server and client side apps have manching key transforms though.

docs/integrations/ember-and-json-api.md
```diff
@@ -17,8 +17,30 @@ To solve this, in Ember, both the adapter and the serializer will need some modi
...
+
+```ruby
+# config/initializers/ams_config.rb
+ActiveModelSerializers.config.adapter = :json_api
+```
+
+or:
+
+```ruby
+# config/initializers/ams_config.rb
```
bf4: nitpick: convention here would be `config/initializers/active_model_serializers.rb`

docs/integrations/ember-and-json-api.md
bf4: I'd say the bigger point here is that if you'd like to handle key transform on the client (Ember) side, to use `key_transform: :unaltered` since it's much faster.

docs/integrations/ember-and-json-api.md
```diff
@@ -17,8 +17,30 @@ To solve this, in Ember, both the adapter and the serializer will need some modi
...
+ActiveModelSerializers.config.key_transform = :underscore
+```
+
+There are multiple mimetypes for json that should all be parsed similarly, so
 in `config/initializers/mime_types.rb`:
+
 ```ruby
 api_mime_types = %W(
   application/vnd.api+json
```
bf4: @NullVoxPopuli we should update this mime types docs as it is not advisable.  i.e. it's wrong, but if you want to do it, fine, we don't recommend it

```ruby
api_mime_types = %W(
  application/vnd.api+json
  text/x-json
  application/json
)

Mime::Type.unregister :json
Mime::Type.register 'application/json', :json, api_mime_types
```

Better to [`require 'active_model_serializers/register_jsonapi_renderer'`](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/register_jsonapi_renderer.rb)

NullVoxPopuli: Agreed, this needs no be updated as well.

NullVoxPopuli: *To

I guess mobile github doesn't allow editing comments .:-(

docs/integrations/ember-and-json-api.md
kyleshevlin: I am not very knowledgeable on this gem, and perhaps there is an actual issue here and I'm happy to raise it if someone can help me discover it, but you are correct that Ember accepts payloads from AMS correctly, as setting the adapter correctly dashes names on outgoing payloads; however, it fails on incoming payloads with dashes. I could not get a PATCH to update attributes that were dasherized instead of underscored.

Thus, when someone referred me to this integration and it works once I adjust the key_transform to underscore, it seemed like submitting a PR to update the docs was the right thing to do.

As is, in theory, you shouldn't need to update keyForAttribute or keyForRelationship in Ember since they are dasherized, and you shouldn't need a key_transform change in AMS since it's default for `:json_api` is `:dash`;

If you plan to keep this integration here with the recommendation to underscore Ember's keys, than the keys also need to be transformed in AMS to the best of my understanding.

docs/integrations/ember-and-json-api.md
kyleshevlin: Happy to change.

docs/integrations/ember-and-json-api.md
NullVoxPopuli: If you look at my aeonvera repository, you'll see how to handle dashed
attributes on patch. I think about every controller does the same technique.

But yes, the overall guide should be updated to show how to handle dashed
everything.

That said, having the underscore option is handy of key transform
performance is a concern, so I'd still like to keep notes on how to use
underscored attributes everywhere. 😉

kyleshevlin: I have made an update using `:unaltered` and requiring the JSON API renderer file

NullVoxPopuli:
LGTM

docs/integrations/ember-and-json-api.md
```diff
@@ -17,17 +17,32 @@ To solve this, in Ember, both the adapter and the serializer will need some modi
...
+
 ```ruby
-api_mime_types = %W(
-  application/vnd.api+json
-  text/x-json
-  application/json
-)
-
-Mime::Type.unregister :json
-Mime::Type.register 'application/json', :json, api_mime_types
```
bf4: 💯https://github.com/rails/rails/pull/23712

PR #1824, opened 2016-06-30
Add doc for setting conditional serialization_scope

yogeshjain999:
#### Purpose
I want to set serialization_scope as different object rather than current_user, for only specific actions in given controller.

#### Changes
Adding documentation for how to conditionally set `serialization_scope` in controllers.

#### Related GitHub issues
#1711

### DIFF
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -314,6 +314,38 @@ So that when we render the `#edit` action, we'll get
 
 Where `can_edit` is `view_context.current_user.admin?` (true).
 
+You can also tell what to set as `serialization_scope` for specific actions.
+
+For example, use `admin_user` only for `Admin::PostSerializer` and `current_user` for rest.
+
+```ruby
+class PostsController < ActionController::Base
+
+  before_action only: :edit do
+    self.class.serialization_scope :admin_user
+  end
+
+  def show
+    render json: @post, serializer: PostSerializer
+  end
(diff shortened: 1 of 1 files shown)


docs/general/serializers.md
```diff
@@ -314,6 +314,38 @@ So that when we render the `#edit` action, we'll get
...
+
+  def admin_user
+    User.new(id: 2, name: 'Bob', admin: true)
+  end
+
+  def current_user
+    User.new(id: 2, name: 'Bob', admin: false)
+  end
+end
+```
```
cgmckeever: I feel the documentation could be better served by showing how to change scopes for a single serializer versus a different scope for two serializers.  I get the concept is the same, but it is not as straight forward IMO

bf4: do you mean like this: https://github.com/rails-api/active_model_serializers/blob/master/docs/general/serializers.md#scope ?

bf4:
@yogeshjain999 look good to me. Is this basically the code you're using?

docs/general/serializers.md
cgmckeever: wow -- that is some deep HOWTO ;)

I read it a couple times, and loosely am going to answer you question with a 'yes' ....

yogeshjain999:
yes, @bf4. I'm using the same approach.

PR #1839, opened 2016-07-11
Add tests for fields option demonstrating usage on both attributes and relationships

NullVoxPopuli:
#### Purpose

Adds tests demonstrating proper syntax / format for using `fields` on both attributes and relationships.

#### Wierdness

The root key needs to be specified in order to whitelist relationships, which feels weird.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ Breaking changes:
 Features:
 
 Fixes:
- - [#1839](https://github.com/rails-api/active_model_serializers/pull/1839) `fields` option whitelists both attributes and relationships. (@NullVoxPopuli)
 
 Misc:
+- [#1839](https://github.com/rails-api/active_model_serializers/pull/1839) `fields` tests demonstrating usage for both attributes and relationships. (@NullVoxPopuli)
 
 ### [v0.10.2 (2016-07-05)](https://github.com/rails-api/active_model_serializers/compare/v0.10.1...v0.10.2)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Breaking changes:
 Features:
 
 Fixes:
+ - [#1839](https://github.com/rails-api/active_model_serializers/pull/1839) `fields` option whitelists both attributes and relationships. (@NullVoxPopuli)
 
 Misc:
 
--- a/lib/active_model/serializer/fieldset.rb
+++ b/lib/active_model/serializer/fieldset.rb
@@ -20,21 +20,11 @@ module ActiveModel
       private
 
       def parsed_fields
-        # TODO: this is not very flexible.
-        # - we should probably support symbol, array, hash, etc
-        #
-        # JSONAPI::IncludeDirective supports these, but also changes the resulting structure.
-        # is it worth it to duplicate some of that functionality?
-        # or should IncludeDirective be modified to only allow one layer of options?
         if raw_fields.is_a?(Hash)
-          symbolize_hash(raw_fields)
+          raw_fields.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
         else
           {}
         end
-      end
-
-      def symbolize_hash(hash)
-        hash.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
       end
     end
   end
--- a/lib/active_model/serializer/fieldset.rb
+++ b/lib/active_model/serializer/fieldset.rb
@@ -20,11 +20,7 @@ module ActiveModel
       private
 
       def parsed_fields
-        if raw_fields.is_a?(Hash)
-          raw_fields.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
-        else
-          {}
-        end
+        JSONAPI::IncludeDirective.new(raw_fields).to_hash
       end
     end
   end
--- a/lib/active_model/serializer/fieldset.rb
+++ b/lib/active_model/serializer/fieldset.rb
@@ -18,6 +18,7 @@ module ActiveModel
       attr_reader :raw_fields
 
       private
+
       def parsed_fields
         # TODO: this is not very flexible.
         # - we should probably support symbol, array, hash, etc
--- a/lib/active_model/serializer/fieldset.rb
+++ b/lib/active_model/serializer/fieldset.rb
@@ -18,9 +18,22 @@ module ActiveModel
       attr_reader :raw_fields
 
       private
-
       def parsed_fields
-        JSONAPI::IncludeDirective.new(raw_fields).to_hash
+        # TODO: this is not very flexible.
+        # - we should probably support symbol, array, hash, etc
+        #
+        # JSONAPI::IncludeDirective supports these, but also changes the resulting structure.
+        # is it worth it to duplicate some of that functionality?
+        # or should IncludeDirective be modified to only allow one layer of options?
+        if raw_fields.is_a?(Hash)
+          symbolize_hash(raw_fields)
+        else
+          {}
+        end
+      end
+
+      def symbolize_hash(hash)
+        hash.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
       end
     end
   end
--- a/test/action_controller/json_api/fields_test.rb
+++ b/test/action_controller/json_api/fields_test.rb
@@ -36,14 +36,14 @@ module ActionController
           get :render_fields_works_on_relationships
           response = JSON.parse(@response.body)
           expected = {
-            "data" => {
-              "id" => "1337",
-              "type" => "posts",
-              "relationships" => {
-                "author" => {
-                  "data" => {
-                  "id" => "1",
-                  "type" => "authors"
+            'data' => {
+              'id' => '1337',
+              'type' => 'posts',
+              'relationships' => {
+                'author' => {
+                  'data' => {
+                    'id' => '1',
+                    'type' => 'authors'
                   }
                 }
               }
--- a/test/action_controller/json_api/fields_test.rb
+++ b/test/action_controller/json_api/fields_test.rb
@@ -1,0 +1,57 @@
+require 'test_helper'
+
+module ActionController
+  module Serialization
+    class JsonApi
+      class FieldsTest < ActionController::TestCase
+        class FieldsTestController < ActionController::Base
+          class PostSerializer < ActiveModel::Serializer
+            type 'posts'
+            attributes :title, :body, :publish_at
+            belongs_to :author
+            has_many :comments
+          end
+
+          def setup_post
+            ActionController::Base.cache_store.clear
+            @author = Author.new(id: 1, first_name: 'Bob', last_name: 'Jones')
+            @comment1 = Comment.new(id: 7, body: 'cool', author: @author)
+            @comment2 = Comment.new(id: 12, body: 'awesome', author: @author)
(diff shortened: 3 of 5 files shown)


NullVoxPopuli:
> @NullVoxPopuli, thanks for your PR! By analyzing the annotation information on this pull request, we identified @arenoir, @beauby and @dubadub to be potential reviewers

waaaaaaaaaaaaaa O.O

neat

lib/active_model/serializer/fieldset.rb
```diff
@@ -20,11 +20,7 @@ def fields_for(type)
       private
 
       def parsed_fields
-        if raw_fields.is_a?(Hash)
-          raw_fields.each_with_object({}) { |(k, v), h| h[k.to_sym] = v.map(&:to_sym) }
-        else
-          {}
-        end
+        JSONAPI::IncludeDirective.new(raw_fields).to_hash
```
NullVoxPopuli: Not sure if this would open us up to too many issues, but it works with all the tests.

test/serializers/fieldset_test.rb
```diff
@@ -5,7 +5,7 @@ class Serializer
     class FieldsetTest < ActiveSupport::TestCase
       def test_fieldset_with_hash
         fieldset = ActiveModel::Serializer::Fieldset.new('post' => %w(id title), 'comment' => ['body'])
-        expected = { post: [:id, :title], comment: [:body] }
+        expected = { post: { id: {}, title: {} }, comment: { body: {} } }
```
NullVoxPopuli: this is the downside of using the include directive, as its meant for nested relationships.

NullVoxPopuli: I just didn't want to re-write that code (slightly modified) here.

bf4: ooh, this is bad ui change...

lib/active_model/serializer/fieldset.rb
bf4: I was thinking the change should be in the adapter that passes in `fields: {}` instead of `fields: nil` when `options[:fields].nil?`

see https://github.com/rails-api/active_model_serializers/blob/91b37cee2c0815234d94214219497c0a26d33ffd/lib/active_model_serializers/adapter/json_api.rb#L300-L306

bf4: or just that `fields_for` should return `{}`, n/m  https://github.com/rails-api/active_model_serializers/blob/91b37cee2c0815234d94214219497c0a26d33ffd/lib/active_model/serializer/fieldset.rb#L12-L14


```diff
      def fields_for(type)
-        fields[type.singularize.to_sym] || fields[type.pluralize.to_sym]
+        fields[type.singularize.to_sym] || fields[type.pluralize.to_sym] || {}
      end
```

NullVoxPopuli: well, the issue IS with parsed_fields because it removes the way you have to specify relationships.

```
root_type: [:relationship_type]
```

test/serializers/fieldset_test.rb
NullVoxPopuli: It be worth it to actually write logic/tests to allow specification of both attributes and relationships, rather than treat everything like a relationship

lib/active_model/serializer/fieldset.rb
NullVoxPopuli: alternatively, I could maybe change how the relationship's call to fields_for is handled. maybe it would be a bit more intuitive that way.

test/serializers/fieldset_test.rb
bf4: but it's also a breaking change...

bf4: cc @beauby

NullVoxPopuli: fwiw, all the tests pass.
however, I'm working on something a little more specific to the exact problem, rather than just blanketly using IncludeDirective

NullVoxPopuli: I may have actually been confused about the intended API, investigating

NullVoxPopuli: I need to write a test for full render call, rather than the adapter

lib/active_model/serializer/fieldset.rb
```diff
@@ -18,14 +18,23 @@ def fields_for(type)
...
 
       private
-
       def parsed_fields
+        # TODO: this is not very flexible.
+        # - we should probably support symbol, array, hash, etc
+        #
+        # JSONAPI::IncludeDirective supports these, but also changes the resulting structure.
+        # is it worth it to duplicate some of that functionality?
+        # or should IncludeDirective be modified to only allow one layer of options?
```
NullVoxPopuli: example:

```ruby
# the root type *should* be assumed (it's currently not with JSON API
fields: :author
fields :author, :comments
fields :access_token # from the docs
```
docs: https://github.com/rails-api/active_model_serializers/blob/master/docs/general/rendering.md#fields
and: https://github.com/rails-api/active_model_serializers/blob/master/docs/general/fields.md 

There is an inconsistency between APIs in the different adapters.

JSON API requires the root key in the fields specification, whereas the other adapters do not require the root. 

So, maybe this PR should focus on unifying this between all adapters.
 - @bf4, @beauby

test/action_controller/json_api/fields_test.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          def render_fields_works_on_relationships
+            setup_post
+            render json: @post, serializer: PostSerializer, adapter: :json_api, fields: { posts: [:author] }
+          end
+        end
+
+        tests FieldsTestController
+
+        test 'fields works on relationships' do
```
NullVoxPopuli: On a side note, I think we need some organization of the tests. 
maybe split them in to Unit, and Integration, and the further folders in Integration by topic of what's being tested.
Unit would follow the structure of `lib` - If this sounds like a good idea, lemme know -- this could cause conflicts with all the PRs, though. :-\

I just have a hard time finding things in `tests`

NullVoxPopuli:
In other news, I learned that multi-word types MUST use the same key-casing as how the key transform is configured, which I didn't think was very apparent, as there is no logging or anything. I'll add that to the docs.

lib/active_model/serializer/fieldset.rb
bf4: btw, I've been thinking we should just name everything `key`. attribute key, relationship key, object key, collection key, and forget about json_key and have key make use of type and root options when available.

NullVoxPopuli: I think that would be a great topic for a PR this week. Hmm.

NullVoxPopuli:
any objections to this?

lib/active_model/serializer/fieldset.rb
bf4: Related https://github.com/rails-api/active_model_serializers/pull/1857

While working on that I've discovered an architectural problem that will require some careful unraveling

In brief

serializer.attributes(fields) builds a hash from calling value (aka read_attribute_for_serialization ) on the Attribute objects with matching names

serializer.associations(include_directive) iterates over all the Reflection objects and rejects excluded and yields an Association object built jit by the Reflection

The Association is used by calling association.serializer.serializable_hash

As currently implemented, the reflection is evaluated (method is called) when the Association is built.  This prevents lazy evaluation

I have a hack in that pr to delay evaluation until the serializer asks for it, which works when the serializer calls serializable_hash

However, the JSONAPI adapter iterates the attributes and associations directly and passes the association into a new Relationship which promptly evaluates the association before determining whether to include it.  Complicating this is that there is an assumption in the code that we mist evaluate the association so that we can infer the serializer class so that we can instantiate it, but it is already too late.

The fundamental way that a serializer is told which fields and associations to evaluate is inconsistent and far to eager.  The already-evaluated-association object permeates the code.... Should be much simpler

Also would be good to make Fields and similar apis consistent

NullVoxPopuli: I have a hack that brings fields functionality to the Attributes adapter (I use this for CSV generation)
```ruby
module ActiveModelSerializers
  module Adapter
    class Attributes < Base
        options[:fields] ||= instance_options[:fields]
        hash = serializer.serializable_hash(instance_options, options, self)
        fields = JSONAPI::IncludeDirective.new(options[:fields]).to_hash
        apply_fields_whitelist(hash, fields)
      end

      # this is a little backwards, but it's needed until AMS has a more unified interface
      def apply_fields_whitelist(hash, fields)
        return hash.map{ |e| apply_fields_whitelist(e, fields) } if hash.is_a?(Array)

        result = {}
        fields_keys = fields.keys
        hash.each do |k, v|
          next unless fields_keys.include?(k)
          if v.is_a?(Hash)
            result[k] = apply_fields_whitelist(v, fields[k])
          else
            result[k] = v
          end
        end

        result
      end
    end
  end
end
```

this sort of thing def needs to be unified. 
As for unification timeline, I think I'll start this weekend, if not Monday, but I'd like to see #1857 merged, as well as this one, and the test refactor merged. :-)

NullVoxPopuli: > While working on that I've discovered an architectural problem that will require some careful unraveling

As I've been looking over everything the past week planning out my big refactor / unification, I've been realizing the whole of AMS is a giant bowl of spaghetti.

bf4: Well, 0.10 is. 0.8 and 0.9 are still pretty clean.  So sad.

cgmckeever: Is it at least glutten free?  🍝  🍷

NullVoxPopuli: 0.10 will become clean soon enough. :-)

The important thing is we have enough tests to accurately keep functionality as we refactor

bf4: If we can return to a clean adapterless cache unaware serializer that can take in options to serialize its attributes and association... and I'd love to use just functions here, like in knuckles, we could rock this

NullVoxPopuli: > and I'd love to use just functions here, like in knuckles

what do you mean by these two statements? 
for first, do you mean more FP pure functions?

bf4:
@NullVoxPopuli The PR description make it sound like it's adding a feature, but the actual is that it is adding tests.  I've added a commit undoing the change in the FieldSet since is not a functional change, as far as I can tell.

While making that change, some of the discussion has gotten lost in the diff.

- [where should the `fields` default value be set](https://github.com/rails-api/active_model_serializers/pull/1839#discussion_r70325716)
- [discussion of differences in fields usage by attributes/associations and general ams architecture](https://github.com/rails-api/active_model_serializers/pull/1839#discussion_r70355684)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Breaking changes:
 Features:
 
 Fixes:
+ - [#1839](https://github.com/rails-api/active_model_serializers/pull/1839) `fields` option whitelists both attributes and relationships. (@NullVoxPopuli)
```
bf4: not a fix, I don't think just tests? https://github.com/rails-api/active_model_serializers/pull/1839#issuecomment-238433456

NullVoxPopuli: yeah, just tests

NullVoxPopuli:
PR description updated

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Features:
 Fixes:
 
 Misc:
+- [#1839](https://github.com/rails-api/active_model_serializers/pull/1839) `fields` tests demonstrating usage for both attributes and relationships. (@NullVoxPopuli)
```
NullVoxPopuli: @bf4 changelog entry moved, and re-worded.

Issue #1843, opened 2016-07-13
RFC - Refactor (Part 1) to get all the adapters using the same API


NullVoxPopuli:
Current Issues with adapters:

 - repeated code:
   - include logic
   - building relationships
   - building the attributes hash (from fields whitelisting, for example) 
- all the logic for retrieving fields from the cache is different per adapter
- `JSON` and `Attributes` are missing features that `JSON API` offers 
   - If adapters are what we intend them to be, they'll all be just _formatters_... sort of...

So, hopefully this can spark some new design ideas / ways to solve the above, and make AMS clean again.

Adapters should have an API modeled after JSON API 
 - `data`
  - primary serializer / resource(s)
 - `included`
   - flat list -- these objects would have the AR integration that doesn't make DB queries to belongs_to records... but in here is where the include logic would be handled, so we don't have to deal with it in the individual adapters.
 - `meta`
 - `links`
 - `errors`
 
A lot of these are already accessible to the serializer, but we've been relying on the adapter to traverse all of it.  

> All of the relationships and everything should already be traversed before getting to the adapter, so the adapter can just focus on how to format the output

An adapter would be responsible for converting each of these structures into whatever format it wishes. 

Thoughts before I start working on this? ;-)

bf4:
? https://github.com/rails-api/active_model_serializers/blob/master/docs/rfcs/template.md

NullVoxPopuli:
I didn't know that existed, thanks!

bf4:
> repeated code

- serializers have an `attributes` methods. It executes `read_attribute_for_serialization` on each attribute in `self.class._attributes`.
  - the list of serialized attributes can be limited by passing in an argument `options[:fields]`
- serializers have a list of defined `associations`
  - for each associated, a serializer is found, and we call `attributes` on it.
    - we pass in to attributes `options[:includes][association_name]`, see serializer for what this does, like fields

> If adapters are what we intend them to be, they'll all be just formatters... sort of...

yup, basically

- Attributes Adapter
  - given a serializer, passes options into `attributes` and then into `associations`. Associations use a key. This is basically the behavior of an active model /active record object's `serializable_hash` method it gets from `ActiveModel::Serializers::JSON` and `ActiveModel::Serialization`.
- JSON adapter
[...]

PR #1845, opened 2016-07-14
Add ActiveModel::Serializer#default_include

iMacTia:
### Purpose
Let's assume we have the current serializers:

```ruby
  attributes :id
  attribute :name, key: :title

  has_many :posts
end

  attributes :id, :title

  has_one :author
  has_one :category
end

  attributes :id, :name

  has_one :address 
end

class CategorySerializer < ActiveModel::Serializer
  attributes :id, :name
end
```

The only way to specify nested associations inclusion, at the moment, is through the controller's render `include` option, as shown in the current examples:

```ruby
# blogs_controller.rb
render json: @blog, include: 'posts.category, posts.author.address'

# posts_controller.rb
render json: @posts, include: 'category, author.address'
```

The issue is that in this case (I admit is probably a stupid example, but stay with me), when rendering a post, we always want to render also its category, its author and its author address.

### Changes
Add two new class methods to `ActiveModel::Serializer`:

**default_include**: associations that you want to include by default (if you use json/attributes adapter, this defaults to all first-level associations). This value will be overriden by the controller's render include.
**always_include**: associations that should always be serialized even if the object is deeply nested. This CAN'T BE OVERRIDDEN from the controller.

Both methods will accept a parameter which will be parsed by JSONAPI::IncludeDirective.
Optionally, a block can be provided that will return this value, in this case the block will be evaluated on the **serialiser instance**.

### Related GitHub issues
#1333, #1849 

### Additional helpful information
Tests are currently failing since we're waiting for [this PR](https://github.com/beauby/jsonapi/pull/27) to be merged on `jsonapi` gem. Hopefully @beauby will have a look at it soon :)

### Discussion
@NullVoxPopuli, @bf4: what should happen if the default_include method is called many times in the same serialiser? Current implementation will override the previous value, does it make sense for you as well?


lib/active_model/serializer/associations.rb
```diff
@@ -65,6 +65,16 @@ def has_one(name, options = {}, &block) # rubocop:disable Style/PredicateName
...
         end
 
+        # Set _default_include to the parsed value of +include+.
+        # @param includes value to be parsed by JSONAPI::IncludeDirective::Parser
+        # @param options options for JSONAPI::IncludeDirective::Parser, default { allow_wildcard: true }
+        # @return [void]
+        #
+        def default_include(include, options = {})
+          default_options = { allow_wildcard: true }
+          self._default_include = JSONAPI::IncludeDirective.new(include, default_options.merge(options))
```
NullVoxPopuli: I'm surprised ruby lets you name a variable `include`

test/serializers/default_include_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
```
NullVoxPopuli: can you add a test for if a post has default_includes, but the object you're serializing does not?

I don't think we want default_includes to be 'respected' the whole way down the render chain

NullVoxPopuli:
>  what should happen if the default_include method is called many times in the same serialiser?

I think overriding the previous value is perfect!



Nice work overall, only had a couple comments :-)

NullVoxPopuli:
also, I wonder if we should move @beauby 's jsonapi to rails-api or rails?

lib/active_model/serializer/associations.rb
iMacTia: I agree this is probably not the smartest name, will refactor it to `args` ;)

test/serializers/default_include_test.rb
iMacTia: Sorry not sure I got this one. I think this is what I did.

> Can you add a test for if a post has default_includes, but the object you're serializing does not?

`default_include` is specified in the PostSerializer, but in the test I'm serialising a Blog object.

lib/active_model/serializer/associations.rb
iMacTia: Actually, changed to `include_args` to follow JSONAPI::IncludeDirective#initialize convention 😄

iMacTia:
@NullVoxPopuli great! And thanks a lot for your comments 😄!
It's always hard to put your hands on someone else's project, and sometime you miss conventions/logics. So please feel free to ask for changes if you think they could improve the PR ;)

test/serializers/default_include_test.rb
NullVoxPopuli: oh, sorry. I guess I wrote too quickly. Here is there scenario I'd like tested:
```
ParentSerializer
  has_many Child

ChildSerializer
  default_include [:grand_child]
```

when serializing the parent with `include: 'children'` (this was probably a bad example, thanks English pluralization), only the child should show up, and not the grand child.

iMacTia: Ah in this case I think my current test is not gonna work.
I'm assuming in this case that serialising the Blog we'll see the authors in the result, which are in the `default_include` of PostSerializer.

This is because I'm using `adapter: :attributes` in the test, which by default include first level associations (the equivalent of having `include: 'posts'` in JSON-API adapter).

Also, my implementation actually "merges" the controller include with the serialiser one. Are you saying the controller one should actually override the serialiser's?

Maybe a quick "expectations" diagram would help in this case 😅.
Let's take my example of Blog -> Posts (default_include :author) -> Author

1. What's the expected result for a simple `render json: Blog.first`?
2. What's the expected result for `render json: Blog.first, include: 'posts'`?
3. Let's assume posts have also another association "category", what's the expected result for `render json: Blog.first, include: 'posts.category'?

beauby:
@nullvoxpopuli: The move to rails-api is planned, I just want to stabilize some stuff first.

test/serializers/default_include_test.rb
NullVoxPopuli: good clarification!

1. 
  data:
    - blog
2. 
  data: 
    - blog 
  included: 
    - posts
 - I guess this could be up for debate, but I think for deeply nested includes, it *should* rely on the the controller include option, to reduce accidental / unavoidable data inclusion. i.e.: I think default_includes on only the top level serializable object allows  for the most flexibility

3. 
  data: 
    - blog
  include: 
    - posts
    - category

iMacTia: Ok this make things clearer, but in all these examples the `default_option` in PostSerializer was always ignored/overridden.
Could you please list the same three expected results, but this time assuming that in BlogSerializer there's a `default_include: 'posts'`?

NullVoxPopuli: Assume:
 - `BlogSerializer` has `default_include 'posts'`
 - `PostSerializer` has `default_includes 'author'`
 - Blog has_many Posts, Post has many authors, and belongs_to a category

One.) `render json: Blog.first`
```
data: 
  blog
included:
  posts
```

Two.) `render json: Blog.first, include: 'posts'`
```
data: 
  blog
included:
  posts
```

Three.) `render json: Blog.first, include: 'posts.category'`
```
data: 
  blog
included:
  posts
  categories
```

iMacTia: Ok great, last one (hopefully, sorry for bothering you with all these examples 😅)

Assume:

* `BlogSerializer` has `default_include 'posts.author'`
* `PostSerializer` has `default_include 'author'`
* Blog `has_many :posts`, Post `has_one :author` and `belongs_to :category`

What happens to cases 2 and 3? I'm particularly interested in the "conflict" between

    render json: Blog.first, include: 'posts.category'

and 

    # blog_serializer.rb
    default_include 'posts.author'

NullVoxPopuli: no problem :-)

Two.) `render json: Blog.first, include: 'posts'`
```
data: 
  blog
included:
  posts
  authors
```

Three.) `render json: Blog.first, include: 'posts.category'`
```
data: 
  blog
included:
  posts
  authors
  categories
```

So, the include option is only merged with the default includes in the top-most serializer, in my perfect world, at least :-)

iMacTia: Ok, thanks @NullVoxPopuli.
My initial idea was to have a cascade-include behaviour, but I can see why you prefer going down this route. I'll probably need to move my code then because AFAIK `associations` method is called for each serialiser, even nested ones.
It would make more sense to have the merge happening when the `ActiveModelSerializers::SerializableResource` is created for rendering

NullVoxPopuli: Yeah, though, the week this is released, I bet someone wants the cascading functionality you have currently.

So... idk. I'm not a huge fan of all the configuration options, but as long as there are thorough tests for them, I think It could be ok.

iMacTia: It should be quite easy for me to implement both ways to do it.
The only question is: should I add an option on default_include (`root_only`, or `always`), or should I make another class attribute to support both of them 😃 ?

NullVoxPopuli: I like `always_include` and `default_include` :-)

iMacTia:
Just pushed another commit to add `always_include` as well.
Now both implementations are available!   🎉 🎉 🎉 🎉
Tests are still failing, will check them carefully after I'm done with the PR on @beauby project 😄 
In the meantime, @NullVoxPopuli, if you have any idea for additional tests, just let me know 😉

test/serializers/default_include_test.rb
```diff
@@ -0,0 +1,79 @@
...
+        @posts = [
+            Post.new(id: 1, title: 'The first post', author: @authors.first, category: @categories.first),
+            Post.new(id: 2, title: 'The second post', author: @authors.first, category: @categories.first)]
+        @blog = Blog.new(id: 2, name: 'The Blog', posts: @posts)
+      end
+
+      test '#default_include option populate "included" for json_api adapter' do
+        serialized = ActiveModelSerializers::SerializableResource.new(@blog, serializer: BlogSerializer, adapter: :json_api).as_json
+
+        assert_equal serialized[:included].size, 3
```
NullVoxPopuli: for both these tests and the always tests, I'd actually make sure that the correct types are rendered, just so whomever is looking at these tests in the future can very easily see _what_ is being rendered.

test/serializers/default_include_test.rb
```diff
@@ -0,0 +1,52 @@
...
+        attributes :id
+        attribute :name, key: :title
+
+        has_many :posts
+      end
+      class PostSerializer < ActiveModel::Serializer
+        default_include [:author]
+        attributes :id, :title
+
+        has_one :author
```
bf4: I think I'd prefer 'has_one :author, include: :all' as a more declarative and easier to implement... I wrote more on this somewhere else... Advantage would include specifying included fields and smaller public api

NullVoxPopuli: but what does include all mean? include all dependencies? include all authors?

if anything, `include: :always` would make more sense.

Though, implementation is going to be slightly more complicated than the current.

> Advantage would include specifying included fields

can you explain thoughts on this?

test/serializers/always_include_test.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      test '#always_include merges with the render include for json_api adapter' do
+        serialized = ActiveModelSerializers::SerializableResource.new(@blog, serializer: BlogSerializer, adapter: :json_api, include: 'posts.category').as_json
+
+        assert_equal serialized[:included].size, 4
+        assert_equal 'The first post', serialized[:included].first[:attributes][:title]
+      end
+
+      test '#always_include option include associations for attributes adapter' do
+        serialized = ActiveModelSerializers::SerializableResource.new(@blog, serializer: BlogSerializer, adapter: :attributes).as_json
```
bf4: There's a test helper 'serializable'... On phone sorry small details

test/serializers/default_include_test.rb
bf4: Quoting from my comment in https://github.com/rails-api/active_model_serializers/issues/1843#issuecomment-232387846

> - serializers have an `attributes` methods. It executes `read_attribute_for_serialization` on each attribute in `self.class._attributes`.
  - the list of serialized attributes can be limited by passing in an argument `options[:fields]`
- serializers have a list of defined `associations`
  - for each associated, a serializer is found, and we call `attributes` on it.
    - we pass in to attributes `options[:includes][association_name]`, see serializer for what this does, like fields

So, that 

| | render options | serializer attributes under JSON adapter | serializer attributes under JSONAPI adapter |
|---|------------- |--------------- | ------------------|
| `has_one :author` | none | `AuthorSerializer.new(object.author).attributes` | not serialized
| `has_one :author` | `include: [:author]` | `AuthorSerializer.new(object.author).attributes` | `AuthorSerializer.new(object.author).attributes`
| `has_one :author` | `include: [:author], fields: { authors: [:name]}` | `AuthorSerializer.new(object.author).attributes([:name])` | `AuthorSerializer.new(object.author).attributes([:name])`
| `has_one :author, include: :all` | none | `AuthorSerializer.new(object.author).attributes` | `AuthorSerializer.new(object.author).attributes`
| `has_one :author, include: [:name]` | none | `AuthorSerializer.new(object.author).attributes([:name])` | `AuthorSerializer.new(object.author).attributes([:name])`
| `has_one :author, include: :all` | `include: [:author], fields: {authors: [:name]}` | `AuthorSerializer.new(object.author).attributes([:name])` | `AuthorSerializer.new(object.author).attributes([:name])`

bf4:
Looks like my comment on api syntax got lost in a diff change https://github.com/rails-api/active_model_serializers/pull/1845#discussion_r70874638

test/serializers/default_include_test.rb
bf4: @NullVoxPopuli 

> but what does include all mean? include all dependencies? include all authors?
> if anything, include: :always would make more sense.

`:all`, meaning 'all fields', was really just an attempt at something better than `true` and something like for all attributes `include: true` or limit them with `include: true, fields: [:name]`.  I think there's definitely a better way to phrase this, but main point is location of the new api and the kinds of words to use.  

So that we could exchange a line in the controller

```ruby
    render json: post, include: [:metrics, :media, :publishes, :author],
        fields: { media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio) }
```

with

```ruby
class PostSerializer
  has_many :metrics, include: :all
  has_one :media, include: [:title]
  has_many :publishers, exclude: [:publisher_bio]
end
render json: post
```

iMacTia: I can see the advantage of the `fields` option, but wouldn't be easy to confuse `include` and `exclude` with `only` and `except`? Actually, aren't they doing the same thing?

iMacTia: Just as a background, `include` as an association option was already discussed on #1113, but finally dismissed in favour of a controller-level include.
There were many reasons for these (I strongly advise a quick read on the above issue), but it was mainly due to the confusion this could produce.

I think that `default_include` and `always_include` would provide a useful additional feature, without getting too much out of context and without creating confusion.

It's really just a simple way to keep code DRY and avoid copy-pasting includes in your controllers when the same serialiser is used.

This doesn't mean that we can't implement also an association-level `include`, I see it as a complementary feature, and not a substitution for this one 😄

test/serializers/always_include_test.rb
iMacTia: Thanks for the tip!
Found it and issued another commit using it 😄!

test/serializers/default_include_test.rb
iMacTia: Added a check for them in my latest commit 😄

iMacTia:
FYI just added a temporary monkey-patch to add the IncludeDirective#merge method and all tests are passing 🎉 🎉 🎉 🎉!
Maybe we need tougher tests 😄 Any idea?

bf4:
@iMacTia would you mind updating the PR description with a higher level outline of the problem this PR is addressing?

> Provide a class method (ActiveModel::Serializer#default_include) to specify which serialiser associations should ALWAYS be included when that serialiser is used.
Parameters are the same as the include option available on controller's render.

Is very implementation-specific.  

The problem, if I understand it, is that when using the JSON API adapter, and we have an API that has default `include` behavior, it can be tiresome to write ```render json: post, include: [:metrics, :media, :publishes, :author], fields: { media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio) }``` in multiple places.

The goal, if I understand it, is to add a way, when using JSON API, for the serializer to specify default included relationships and their fields, right?

and the proposed implementation of that is (?)

```ruby
  always_include 'metrics,media,publishers,author'
  default_include 'metrics,media,publishers,author'
end
render json: post, fields: { media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio) }
```

where `always_include` resolves the need to specify `include: [:metrics, :media, :publishes, :author]`

(without carefully reading all the code, it's not clear to me the difference between always include and default include. )

It should be noted that I proposed something like

```
class PostSerializer
  has_many :metrics, include: :all
  has_one :media, include: [:title]
  has_many :publishers, exclude: [:publisher_bio]
end
render json: post
```

where `include/exclude` map to the JSON API `include` and `field`. They bear no relation to the AMS `if/unless` conditional exclusion nor `only/except` json serialization options in ActiveModel::Serialization

tommiller1:
Awesome work getting this working! 

My couple of cents;

The always_include method should be specified on the association itself ex. "has_one :author, always_include: true". Basically always_include should make a serializer treat the association as an attribute. Attributes are always included even for deeply nested associations. Ex: including 'posts.comments.images.uploader' in a render call will return all attributes in this entire chain but no associations. With always_include we want to change this behaviour so that all associations along this chain with always_include set to true are included. 

The default_include method should be just that; default includes if there are no other includes. It should be completely ignored if we specify includes on the render call (not merged). Also currently default_includes doesnt really offer anything new, we are just moving the includes from the controller to the serializer. You should consider making default_includes an instance method so we can specify parameters in the render call that we use in the default_include method to bunch together common includes. Ex:

	render json: @post, show: true

	PostSerializer
		def default_includes
			if instance_options[:show]
				['comments.images', 'images', 'dragons']
			elsif instance_options[:no_dragons]
				['comments.images', 'images']
			end 
		end
	
Anyway, I am happy this is getting included in some form. Thanks for taking the time to work on this, highly appreciated!

bolshakov:
I have the similar feature implemented in our project. 

```ruby
    # Method may be used when you have to define an attribute which 
    # should be serialized using AMS. So you can define always included relations.
    #
    # @param [String, Symbol]
    # @return [void]
    # @example
    #   class InvoiceSerializer < ApplicationSerializer
    #     attribute :id
    #     serialized_attribute :balance
    #   end
    #
    # It will be rendered the following way without using expands:
    #
    #   {
    #     "object": "invoice",
    #     "id": "3f8b90da-2a74-4f0a-8f5e-e63835348b72",
    #     "balance": {
    #       "object": "money",
    #       "currency": "BTC",
    #       "value": 664
    #     }
    #   }
    # 
    def serialized_attribute(name)
      attribute(name)

      define_method name do
        ActiveModel::SerializableResource
          .new(object.send(name))
          .serializable_hash[:data]
      end
    end
```

NullVoxPopuli:
So, @iMacTia , it looks like I was wrong about the top level `default_includes` API, and that people want what @bf4 suggested. Do you think you have time to investigate that route?

iMacTia:
@NullVoxPopuli @bf4 
Understood, but I think this PR is too messy now, I prefer to work a new branch and open another PR/Issue to discuss about it

iMacTia:
As discussed in #1849, I'll now finish to work on this one and hopefully we'll have `default/always _include` feature soon 🎉 

Improvements to work on:

1. Make `default_include` and `always_include` optionally accepting a block to be evaluated at instance level.
2. Update tests

Point 1 was discussed in #1849 after the input from @tommiller1 and @NullVoxPopuli proposed a nice solution:

```ruby
class PostSerializer < AM::S
  default_include do |object, serializer_instance|
    # whatever logic in here, must return a value to be parsed as JSONAPI::IncludeDirective
  end
end
```

We also discussed about a possible relationship-level include, but that will be implemented in a separate PR.

iMacTia:
Updated PR description

tommiller1:
"This value will be merged/overriden by the controller's render include."

Will it be merged or overidden? Default includes should be overridden by the includes on the render call imo. Merging them is not intuitive behaviour.

iMacTia:
@tommiller1 Yes, agreed. I think that's just a copy-and-paste error 😄 fixing now

bf4:
@iMacTia can you update the description with a nice summary of the behavior, something like in https://github.com/rails-api/active_model_serializers/pull/1845#discussion_r70874638 ?

I've also been thinking that this is currently possible by overriding `serializable_hash` in userland without any changes to AMS

I thought I wrote it down somewhere, but in any case, here's an untested example:

we can replace

```ruby
    render json: post, include: [:metrics, :media, :publishes, :author],
        fields: { media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio) }
```

with

```ruby
render json: post
```

if we modify serializable_hash

```ruby
class PostSerializer < ApplicationSerializer
  has_many :metrics
  has_one :media
  has_many :publishers
    adapter_options ||= {}
    adapter_options[:include] ||= [:metrics, :media, :publishes, :author]
    options[:fields] ||= {}
    options[:fields].reverse_merge!(media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio))
    super
  end
end
```

bf4:
@iMacTia can you update the description with a nice summary of the behavior, something like in https://github.com/rails-api/active_model_serializers/pull/1845#discussion_r70874638 ?

I've also been thinking that this is currently possible by overriding `serializable_hash` in userland without any changes to AMS

I thought I wrote it down somewhere, but in any case, here's an untested example:

we can replace

```ruby
    render json: post, include: [:metrics, :media, :publishes, :author],
        fields: { media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio) }
```

with

```ruby
render json: post
```

if we modify serializable_hash

```ruby
class PostSerializer < ApplicationSerializer
  has_many :metrics
  has_one :media
  has_many :publishers
    adapter_options ||= {}
    adapter_options[:include] ||= [:metrics, :media, :publishes, :author]
    options[:fields] ||= {}
    options[:fields].reverse_merge!(media: [:title], publishers: (PublisherSerializer._attributes - :publisher_bio))
    super
  end
end
```

oyeanuj:
@iMacTia Really looking forward to your PR getting merged! :) Any updates by any chance?

bf4:
@oyeanuj This PR is really about moving some behavior from userland to AMS, which is why we're not in a rush to add more complexity to AMS.  

As stated in the PR description, this behavior is already provided by passing in (query) options.  

And as stated elsewhere, you can probably get the desired default behavior by taking advantage of serializer inheritance https://github.com/rails-api/active_model_serializers/pull/1845#issuecomment-234319407

oyeanuj:
@bf4 Thanks for the quick response! While I understand the complexity part of your message above, I am not sure if this behavior is already provided. 

In the PR above, the behavior is done through controller options and in the comment linked through serializer inheritance. 

**Regd. the controller option** 

That doesn't make for a easy transition from 0.8.x and 0.9.x to 0.10.x as one would have to move a lot of the behavior from serializer to the controller and specify it every time the object is rendered (as mentioned in more detail in #1333).

**Regd the `serializable_hash` option**

That is an interesting option but I had two concerns with it. 

1. Arguably and subjectively, that doesn't look like the clean Rails way of doing it :) I could see that `reverse_merge` statement get pretty nasty in a slightly larger serializer.

2. With the above code, could you do something like? On a quick test, that didn't work out but I can spend more time on it.
````ruby
render json: post, except: [:publishers]
````

Also, my aim with the comment isn't to suggest which approach is better (older or newer), but to point out that there was a usage established in the earlier versions (so not necessarily moving complexity from userland per se), and that same behavior isn't possible yet (especially, if you are not using the JSONAPI adapter). Since those version did provide this in the AMS, my assumption was that it would still be considered as fair territory for AMS 0.10.x and 1.x in the future..

Totally understand if the `default_includes` or `always_includes` is not a direction you choose, but hope this perspective helps in making the decision.

If I am missing something basic in my understanding of the `serializable_hash` option, that would still allow for something like this, please let me know and I'll stop bothering you all on this thread :)

Appreciate all of your effort on this project, and PR (and hoping that didn't come off as an entitled comment)!

Thank you!

**Comment by bf4** [comment]
Great comment Anuj!  

B mobile phone

**Comment by remear** [comment]
I'm not really convinced that default_includes/always_includes or any include logic should be in the serializers. IMO, the serializers should be concerned with how to serialize the model to which they pertain. My opinion is that `includes` is an option that pertains to document assembly, and should, therefore, be done via the controller. If you'd like a default include set, perhaps even different defaults for various actions, configure them and supply them in the appropriate render calls.

**Comment by NullVoxPopuli** [comment]
This is actually the consensus, thanks for touching on this again, @remear. 

There are just too many differing and conflicting scenarios that everyone wants all at once.

**Comment by oyeanuj** [comment]
@remear @NullVoxPopuli Thats a bummer :( Maybe we can document it for those upgrading from 0.9 to 0.10 since it was supported in the earlier version. Speaking of, is there a guide for to help with that? (Maybe I just wasn't able to find one.)

Also, is #1865 still fair game?

@iMacTia What has been your workaround/interim solution for this? Or have you been maintaining a different fork for this?

**Comment by NullVoxPopuli** [comment]
I believe #1865 is still fair game as it's a render option, rather than a serializer option

@bf4 mentioned a work around. https://github.com/rails-api/active_model_serializers/issues/1865#issuecomment-236937372

**Comment by iMacTia** [comment]
@oyeanuj I was in a rush and had to refactor controllers to use the `include` option there. The only workaround I used to avoid too much duplication was defining constants at Serializer level to be used in the controller.
Example:

```ruby
class PostSerializer
  INCLUDES = [:author, :comments]
  
  has_one :author
  has_many :comments
end

# and then in the controller
render @post, include: PostSerializer::INCLUDES
```

I DON'T LIKE this solution, but again I was in a rush to deploy the application with the updated AMS gem, so I had to...
I'm now on holiday till early next week, as soon as I'm back I'll look for a better solution. probably defining relations as attributes as most people do.

Really a shame we didn't find an agreement on this. Even though I understand @remear and @NullVoxPopuli motivation, I still think this leaves a good portion of developer in the ugly situation of writing un-DRY unmaintainable code. It is riskious if you start with 0.10 from day one, a pain if you have to refactor your app from older version as you can easily miss a controller render (especially implicit ones) and we don't all live in the ideal world where every response is checked by automated tests...

NullVoxPopuli:
for what it's worth, I think includes should be in the controller, which managing includes there is pretty dry and maintainable. 

But, skirting around a problem (lack of tests) by adding a large amount of automated  functionality is suuuuuuuuuper risky and smelly. 

You're more than welcome to monkey patch, and @bf4 and I have been wanting to have a showcase page of the different forks / extensions / monkey patches that people use for different situations so that if someone with teh same issue comes across the showcase page, you'll still end up helping a brotha out. :-)

iMacTia:
@NullVoxPopuli I agree with all the above but the first sentence.
I can't remember if there'a an example that show this already, but in case we have a resource that always needs to be rendered with some subresources and that gets rendered in different controller options, than you definitely end up copy-pasting your include directive in all those controller actions.

I'm not complaining and as I said I understand your decision, this issue probably doesn't affect the JSONAPI adapter but is more related to the Attributes one and I know your focus is on the former.
Moreover, there are workarounds to do it without this PR so I think it's OK. At least we now have an issue where this thing was discussed, decided and closed, so other devs in the future knows what they have to do.

NullVoxPopuli:
>  than you definitely end up copy-pasting your include directive in all those controller actions.

You could move that include string to a constant in your controller.

:-\

iMacTia:
In fact my temporary solution was to create a constant (in the serializer, not the controller, but it's actually the same).

However, this is still prone to errors (like adding a new action in the future and forgetting the include, or missing an implicit render of that Serializer, which actually happened during my refactoring).
So my preferred solution at this point will be to have the association defined as an attribute to mimic the `default_include` functionality.

Again, I'm not blaiming anyone, I understand your decision. I'm just glad this was discussed :)

oyeanuj:
Agree with @iMacTia, glad this was discussed. 

Implicit renders of the associations, and that too, across controllers is the most problematic area for me as well, and doesn't feel like a clean solution. But given that this is a breaking change, I feel like this would be a great addition as a 'recipe', 'showcase' or a 'monkey-patch' and will help out others who might upgrade in the future.

robinsingh-bw:
```
# usage: always_include :image
    # the included object must be an association (not a method/attribute)!
    def self.always_include(name, options = {})
      attribute(name, options)
      define_method name do
        if obj = object.send(name)
          resource = ActiveModelSerializers::SerializableResource.new(obj)
          resource.serialization_scope = instance_options[:scope]
          resource.serialization_scope_name = instance_options[:scope_name]
          resource.serializable_hash
        end
      end
    end
```

iMacTia:
That's a great monkey-patch example, thanks @robinsingh-bw!
It will be my starting point to get this done properly :)!
Thanks for sharing!

oyeanuj:
@iMacTia Did you end up using @robinsingh-bw 's monkeypatch or another solution? Is there a gist of before -> after of your code that you could share, from 0.9 to 0.10? Or a gist of your usage?

@robinsingh-bw stupid question, but where are you including that monkeypatch? In the ApplicationSerializer?

robinsingh-bw:
In my BaseSerializer that all serializers extend from, its not really a monkeypatch as we are not overwriting any core methods. I have updated it to support custom methods defined on the serializers (in addition to object relations):
`    # usage: always_include :image
    def self.always_include(name, options = {})
      options[:key] ||= name
      method_name = name.to_s + '_serialized'
      define_method method_name do
        data = respond_to?(name) ? send(name) : object.send(name)
        if data
          options = instance_options.dup
          options.delete(:serializer)
          options.delete(:each_serializer)

          # use Attributes adapter here instead of Json as we dont want a root object
          options[:adapter] = ActiveModelSerializers::Adapter::Attributes

          resource = ActiveModelSerializers::SerializableResource.new(data, options)
          resource.serializable_hash
        end
      end

      attribute(method_name, options)
    end`

Issue #1831, opened 2016-07-06
request to follow SemVer/mention it in README


pravi:
Do you follow Semantic Versioning http://semver.org? If yes, can this be mentioned in README? If not, can you consider it?

bf4:
We do use semantic versioning. I'd welcome a PR to add it to the readme.

pravi:
@bf4 are you planning for a 1.0 release soon?

bf4:
When we have a 1.0 we'll release it.

PR #1846, opened 2016-07-14
add SemVer compliance text

pravi:
#### Purpose

To give confidence to users of this library about API stability.

#### Related GitHub issues

Closes: #1831

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,16 @@ And then execute:
 $ bundle
 ```
 
+## Semantic Versioning
+
+This project conforms to [semver](http://semver.org/). As a result of this policy, you can (and should) specify a dependency on this gem using the [Pessimistic Version Constraint](http://guides.rubygems.org/patterns/) with two digits of precision. For example:
+
+spec.add_dependency 'active_model_serializers', '~> 1.0'
+
+This means your project is compatible with licensee 1.0 up until 2.0. You can also set a higher minimum version:
+
+spec.add_dependency 'active_model_serializers', '~> 1.1'
+
 ## Getting Started
 
 See [Getting Started](docs/general/getting_started.md) for the nuts and bolts.


README.md
```diff
@@ -65,6 +65,16 @@ And then execute:
 $ bundle
 ```
 
+## Semantic Versioning
+
+This project conforms to [semver](http://semver.org/). As a result of this policy, you can (and should) specify a dependency on this gem using the [Pessimistic Version Constraint](http://guides.rubygems.org/patterns/) with two digits of precision. For example:
+
+spec.add_dependency 'active_model_serializers', '~> 1.0'
```
bf4: correction `~> 0.10.0`

README.md
```diff
@@ -65,6 +65,16 @@ And then execute:
...
+
+This means your project is compatible with licensee 1.0 up until 2.0. You can also set a higher minimum version:
```
bf4: typo: `license/version`.  But more generally, I don't think all this text is necessary as it's included in the link documentation.

bf4: Are there any prominent projects that use semver that devote this much text?

NullVoxPopuli: I think most just say, `This project conforms to [semver]`. And that's it.

README.md
NullVoxPopuli: I was thinking OP was planning for the future.

README.md
bf4: with version.  and maybe something like


```plain
[![SemVer 2.0](https://img.shields.io/badge/SemVer-2.0-lightgrey.svg)](http://semver.org/)
```

[SemVer 2.0](http://semver.org/)

NullVoxPopuli: Hey, that's super concise. I like that

pravi: Many projects do this, full list here https://wiki.debian.org/Teams/Ruby/UpstreamPledge

README.md
pravi: Well, its really meaningfull only for a stable release, as even under SemVer, you are allowed to break compatibility in any 0.x version.

jgrau: I would be confused if the example in the readme listed a version that does not exist. I would suggest using the current most recent version. Once we get to `1.0` I am sure a `grep` would catch the outdated version and update it.

Issue #1696, opened 2016-04-20
Allow for associations redefinition on inherited serializers


beauby:
Fixes #1688 
Closes #1690

c.f. #1686

PR #1848, opened 2016-07-15
Bugfix/redefine associations on inherited serializers

EhsanYousefi:
#### Purpose
Allow for associations redefinition on inherited serializers
#### Related GitHub issues
[1696](https://github.com/rails-api/active_model_serializers/issues/1696)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Misc:
 
 Fixes:
 - [#1814](https://github.com/rails-api/active_model_serializers/pull/1814) Ensuring read_multi works with fragment cache
-- [#1848](https://github.com/rails-api/active_model_serializers/pull/1848) Redefine associations on inherited serializers
+- [#1848](https://github.com/rails-api/active_model_serializers/pull/1848) Redefine associations on inherited serializers. (@EhsanYousefi)
 
 Misc:
 - [#1808](https://github.com/rails-api/active_model_serializers/pull/1808) Adds documentation for `fields` option. (@luizkowalski)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Misc:
 
 Fixes:
 - [#1814](https://github.com/rails-api/active_model_serializers/pull/1814) Ensuring read_multi works with fragment cache
+- [#1848](https://github.com/rails-api/active_model_serializers/pull/1848) Redefine associations on inherited serializers
 
 Misc:
 - [#1808](https://github.com/rails-api/active_model_serializers/pull/1808) Adds documentation for `fields` option. (@luizkowalski)
--- a/lib/active_model/serializer/associations.rb
+++ b/lib/active_model/serializer/associations.rb
@@ -13,7 +13,7 @@ module ActiveModel
       included do
         with_options instance_writer: false, instance_reader: true do |serializer|
           serializer.class_attribute :_reflections
-          self._reflections ||= []
+          self._reflections ||= {}
         end
 
         extend ActiveSupport::Autoload
@@ -74,7 +74,8 @@ module ActiveModel
         # @api private
         #
         def associate(reflection)
-          self._reflections << reflection
+          key = reflection.options[:key]
+          key ? self._reflections[key] = reflection : self._reflections[reflection.name] = reflection
         end
       end
 
@@ -86,7 +87,7 @@ module ActiveModel
         return unless object
 
         Enumerator.new do |y|
-          self.class._reflections.each do |reflection|
+          self.class._reflections.values.each do |reflection|
             next if reflection.excluded?(self)
             key = reflection.options.fetch(:key, reflection.name)
             next unless include_directive.key?(key)
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -1,5 +1,4 @@
 require 'test_helper'
-
 module ActiveModel
   class Serializer
     class AssociationsTest < ActiveSupport::TestCase
@@ -104,13 +103,13 @@ module ActiveModel
         end
 
         assert(
-          PostSerializer._reflections.all? do |reflection|
-            inherited_klass._reflections.include?(reflection)
+          PostSerializer._reflections.values.all? do |reflection|
+            inherited_klass._reflections.values.include?(reflection)
           end
         )
 
         assert(
-          inherited_klass._reflections.any? do |reflection|
+          inherited_klass._reflections.values.any? do |reflection|
             reflection.name == :top_comments
           end
         )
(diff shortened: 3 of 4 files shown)


NullVoxPopuli:
can you address the rubocop issues?
thanks!

EhsanYousefi:
Done!

lib/active_model/serializer/associations.rb
```diff
@@ -74,7 +74,8 @@ def has_one(name, options = {}, &block) # rubocop:disable Style/PredicateName
         # @api private
         #
         def associate(reflection)
-          self._reflections << reflection
+          key = reflection.options[:key]
+          key ? self._reflections[key] = reflection : self._reflections[reflection.name] = reflection
```
NullVoxPopuli: why the difference between key and name here?

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,111 @@ def test_illegal_conditional_associations
           assert_match(/:if should be a Symbol, String or Proc/, exception.message)
         end
       end
+
+      class InheritedSerializerTest < ActiveSupport::TestCase
+        InheritedPostSerializer = Class.new(PostSerializer) do
+          belongs_to :author, custom_option: true
```
NullVoxPopuli: what's `custom_option`?

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,111 @@ def test_illegal_conditional_associations
...
+              posts: nil,
+              roles: nil,
+              bio: nil
+            }
+
+            assert_equal(associations, expected_associations)
+            assert_equal(@author_serializer.associations.count, 3)
+          end
+
+          without_redefine.call
```
NullVoxPopuli: why are you using lambdas like this?

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,111 @@ def test_illegal_conditional_associations
...
+
+          with_redefine = lambda do
+            associations = {}
+            @inherited_post_serializer.associations.each do |ass|
+              associations[ass.name] = ass.options[:custom_option]
```
NullVoxPopuli: ah, so you're using custom option to test the presence of an association. 
This feels kinda hacky, and adds additional logic into tests that isn't part of the public API.

can you re-work these tests to use public API things? thanks!

lib/active_model/serializer/associations.rb
EhsanYousefi: Because you can't handle some situations solely with name i.e:
`has_many :comments
  has_many :comments, key: :last_comments do
     object.comments.last(1)
   end`

test/serializers/associations_test.rb
EhsanYousefi: Because of lexical scoping.

test/serializers/associations_test.rb
EhsanYousefi: Sure, why not. :D

test/serializers/associations_test.rb
NullVoxPopuli: do the tests fail without the lambdas?

lib/active_model/serializer/associations.rb
NullVoxPopuli: this may be something that AMS needs to change in the future then. Cause this is silly. haha

key should be :comments for the first
and key should be last_comments for the second.

test/serializers/associations_test.rb
EhsanYousefi: Not they don't, using lambdas make code more functional.

lib/active_model/serializer/associations.rb
EhsanYousefi: Yes, it is silly indeed, but, but there's nothing we can do, at least for now.

test/serializers/associations_test.rb
NullVoxPopuli: well, why not split apart the tests? having so much setup and assert in a single test is just asking to be split up.

EhsanYousefi: There's not too much setup, but OK, I'll split them apart.

EhsanYousefi:
I've done some improvements on tests, take a look at them please.

test/serializers/associations_test.rb
```diff
@@ -1,5 +1,5 @@
 require 'test_helper'
-
+require 'pry'
```
NullVoxPopuli: this needs to be removed

test/serializers/associations_test.rb
```diff
@@ -290,6 +290,48 @@ def test_illegal_conditional_associations
...
+          @inherited_post_serializer = InheritedPostSerializer.new(@post, custom_options: true)
+          @inherited_author_serializer = InheritedAuthorSerializer.new(@author)
+        end
+
+        def test_redefined_has_many_and_has_one
+          author_associations = @author_serializer.associations.to_a
+          inherited_author_associations = @inherited_author_serializer.associations.to_a
+          assert_equal(@author_serializer.associations.count, 3)
+          assert_equal(@inherited_author_serializer.associations.count, 4)
+          assert_equal((author_associations - inherited_author_associations).count, 2)
```
NullVoxPopuli: can you add assertions to this that show what associations are being produced? 
counting associations doesn't actually prove that this is working, just that there are different associations.

test/serializers/associations_test.rb
EhsanYousefi: Ah, how careless I'm. Sorry for that.

test/serializers/associations_test.rb
EhsanYousefi: Done.

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,82 @@ def test_illegal_conditional_associations
...
+          has_many :roles, polymorphic: true
+          has_one :bio, key: :biography
+        end
+
+        def setup
+          @author = Author.new(name: 'Steve K.')
+          @post = Post.new(title: 'New Post', body: 'Body')
+          @post_serializer = PostSerializer.new(@post, custom_options: true)
+          @author_serializer = AuthorSerializer.new(@author)
+          @inherited_post_serializer = InheritedPostSerializer.new(@post, custom_options: true)
```
NullVoxPopuli: the `custom_options` aren't needed anymore :-)

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,82 @@ def test_illegal_conditional_associations
...
+            @author_serializer.associations.map(&:name).sort,
+            [:posts, :roles, :bio].sort
+          )
+          assert_equal(
+            @inherited_author_serializer.associations.map(&:name).sort,
+            [:posts, :roles, :bio, :bio].sort
+          )
+          assert_equal(
+            map_with_key(@inherited_author_serializer.associations).sort,
+            [:posts, :roles, :bio, :biography].sort
```
NullVoxPopuli: These testes look MUCH better. :-)

Just a couple more changes, and I think I'll be ready to merge.
1. I find myself having to look at your test setup, and back at the different assertions to figure out what is being tested. Do you think you could break up some of these assertions into different tests?

Something like:

```ruby
test 'a serializer inheriting from the author serializer can have an additional bio field named biography' do
  expected = [:bio, :biography, :posts, :roles]
  result = map_with_key(@inherited_author_serializer.associations).sort
end
```

that `test 'name of test' do` syntax is also what we're trying to get people to do going forward, but yeah.


So, if you could change the tests to be like that, that'd be great! I know it's more computing of the same thing for each test, but readability for others looking at the project is what the goal is here. :-)


the second thing, is just to add a changelog entry :-)

test/serializers/associations_test.rb
EhsanYousefi: :D I've just removed them.

NullVoxPopuli: yay!

test/serializers/associations_test.rb
EhsanYousefi: I took your advise and made some changes, take a look at them please.

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,61 @@ def test_illegal_conditional_associations
...
+          @inherited_post_associations = @inherited_post_serializer.associations.to_a
+        end
+
+        test 'an author serializer must have [posts,roles,bio] associations' do
+          expected = [:posts, :roles, :bio].sort
+          result = @author_serializer.associations.map(&:name).sort
+          assert_equal(result, expected)
+        end
+
+        test 'a post serializer must have [author,comments,blog] associations' do
```
NullVoxPopuli: these tests look great!

EhsanYousefi: Yes, it is indeed, thanx to you.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Misc:
 
 Fixes:
 - [#1814](https://github.com/rails-api/active_model_serializers/pull/1814) Ensuring read_multi works with fragment cache
+- [#1848](https://github.com/rails-api/active_model_serializers/pull/1848) Redefine associations on inherited serializers
```
NullVoxPopuli: can you add `(@EhsanYousefi )` to the end of this? :-) gotta make sure you get credit!

NullVoxPopuli:
Just gotta add your name, and then we'll be set!

CHANGELOG.md
EhsanYousefi: :D done.

lib/active_model/serializer/associations.rb
```diff
@@ -13,7 +13,7 @@ module Associations
       included do
         with_options instance_writer: false, instance_reader: true do |serializer|
           serializer.class_attribute :_reflections
-          self._reflections ||= []
+          self._reflections ||= {}
```
bf4: Is this a breaking change? Need to confirm before release

lib/active_model/serializer/associations.rb
bf4: This isn't proper use of ternary, let's look at cleaning up

test/serializers/associations_test.rb
```diff
@@ -290,6 +289,61 @@ def test_illegal_conditional_associations
           assert_match(/:if should be a Symbol, String or Proc/, exception.message)
         end
       end
+
+      class InheritedSerializerTest < ActiveSupport::TestCase
+        InheritedPostSerializer = Class.new(PostSerializer) do
```
bf4: @nullvoxpopuli let's not introduce more dynamics classes right now unless we need to

NullVoxPopuli: I'll release a PR later today that gets rid of them, so people stoop copying the patterns they see.

lib/active_model/serializer/associations.rb
NullVoxPopuli: Do we need to modify rubocop to catch this?

lib/active_model/serializer/associations.rb
NullVoxPopuli: I can run my aeonvera project's tests with this branch/master, if you think that'll be reassuring.

lib/active_model/serializer/associations.rb
EhsanYousefi: I don't get it, why it is improper use?

NullVoxPopuli: because you have assignments in the true/false areas of ternary.

proper form has has the ternary only produce the value, and not do any mutations.

So, the only way to re-write the ternary without breaking that rule would be to convert it to an if block, unless you did:

```ruby
key_name = key ? key : reflection.name
self._reflections[key_name] = reflection
```

EhsanYousefi: `key_name = key ? key : reflection.name
self._reflections[key_name] = reflection`
Is really better solution than mine.
Assignments produce value in ruby(unlike Scala), and thats why I did that.
Considering your notes I'll refactor the code.

NullVoxPopuli: feel free to open a new PR, as this one is already closed. (It'll also keep the review process cleaner, too)

But yeah, basically:
 - short lines
 - each line should do only one thing

PR #1857, opened 2016-07-26
Smarter association id lookup-- no db hit on belongs_to for id-only

### LINKED ISSUE #1552: Serializer pulls relations when belongs_to declared but not included

bf4:
If the serializer has an association, 
- If adapter is not JSON API, serialize it. 
- else if it is JSON API and the related resource 
  - is not included, serialize just the ids
  - is included, serialize the all fields, or the requested fields

The thing where it doesn't actually call the association is a 
special case where AMS needs to know it can get the relationship id 
from the resource itself, so that userland code doesn't need to do any fancy association preloading.

The way this has been solved in [JSONAPI-Resources](https://github.com/cerebris/jsonapi-resources/blob/f4ab5683de49c2a506eea14096bd81a0ec85e94e/README.md#options)
is to have special rules for looking up association ids.
And see [JSONAPI::Relationship](https://github.com/cerebris/jsonapi-resources/blob/f4ab5683de49c2a506eea14096bd81a0ec85e94e/lib/jsonapi/relationship.rb)

In brief:

| association | foreign_key method | loads relation (i.e. when foreign key is on relation ) |
| --- | --- | --- |
| `has_many   :authors` | `:author_ids` | true |
| `has_one    :author` | `:author_id` | true |
| `belongs_to :author` | `:author_id` | false |

Caveat:

This will only work on active record objects

Includes:
- #1100

Follows:
- Closes #1552
- #1750
- #1797

Related:
- #1843
- #1555

### DIFF
--- a/lib/active_model/serializer/association.rb
+++ b/lib/active_model/serializer/association.rb
@@ -38,6 +38,10 @@ module ActiveModel
         reflection.options[:meta]
       end
 
+      def belongs_to?
+        reflection.foreign_key_on == :self
+      end
+
       def polymorphic?
         true == reflection_options[:polymorphic]
       end
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -47,11 +47,23 @@ module ActiveModel
     #
     # So you can inspect reflections in your Adapters.
     class Reflection < Field
+      attr_reader :foreign_key, :type
+
       def initialize(*)
         super
         options[:links] = {}
         options[:include_data_setting] = Serializer.config.include_data_default
         options[:meta] = nil
+        @type = options.fetch(:type) do
+          class_name = options.fetch(:class_name, name.to_s.camelize.singularize)
+          class_name.underscore.pluralize.to_sym
+        end
+        @foreign_key =
+          if collection?
+            "#{name.to_s.singularize}_ids".to_sym
+          else
+            "#{name}_id".to_sym
+          end
       end
 
       # @api public
@@ -148,6 +160,11 @@ module ActiveModel
         else
           serializer.read_attribute_for_serialization(name)
         end
+      end
+
+      # @api private
+      def foreign_key_on
+        :related
       end
 
       # Build association. This method is used internally to
--- a/lib/active_model_serializers/adapter/json_api/relationship.rb
+++ b/lib/active_model_serializers/adapter/json_api/relationship.rb
@@ -43,14 +43,21 @@ module ActiveModelSerializers
         end
 
         def data_for_one(association)
-          # TODO(BF): Process relationship without evaluating lazy_association
-          serializer = association.lazy_association.serializer
-          if (virtual_value = association.virtual_value)
-            virtual_value
-          elsif serializer && association.object
-            ResourceIdentifier.new(serializer, serializable_resource_options).as_json
+          if association.belongs_to? &&
+              parent_serializer.object.respond_to?(association.reflection.foreign_key)
+            id = parent_serializer.object.send(association.reflection.foreign_key)
+            type = association.reflection.type.to_s
+            ResourceIdentifier.for_type_with_id(type, id, serializable_resource_options)
           else
-            nil
+            # TODO(BF): Process relationship without evaluating lazy_association
+            serializer = association.lazy_association.serializer
+            if (virtual_value = association.virtual_value)
+              virtual_value
+            elsif serializer && association.object
+              ResourceIdentifier.new(serializer, serializable_resource_options).as_json
+            else
+              nil
+            end
           end
         end
 
--- a/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
+++ b/lib/active_model_serializers/adapter/json_api/resource_identifier.rb
@@ -22,6 +22,13 @@ module ActiveModelSerializers
           JsonApi.send(:transform_key_casing!, raw_type, transform_options)
         end
 
+        def self.for_type_with_id(type, id, options)
+          {
+            id: id.to_s,
+            type: type_for(:no_class_needed, type, options)
+          }
+        end
+
         # {http://jsonapi.org/format/#document-resource-identifier-objects Resource Identifier Objects}
         def initialize(serializer, options)
           @id   = id_for(serializer)
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -137,6 +137,34 @@ module ActiveModel
         assert expected_association_keys.include? :site
       end
 
+      class BelongsToBlogModel < ::Model
+        attributes :id, :title
+        associations :blog
+      end
+      class BelongsToBlogModelSerializer < ActiveModel::Serializer
+        type :posts
+        belongs_to :blog
+      end
+
+      def test_belongs_to_doesnt_load_record
+        attributes = { id: 1, title: 'Belongs to Blog', blog: Blog.new(id: 5) }
+        post = BelongsToBlogModel.new(attributes)
+        class << post
+          def blog
+            fail 'should use blog_id'
+          end
+
+          def blog_id
+            5
+          end
+        end
+
+        actual = serializable(post, adapter: :json_api, serializer: BelongsToBlogModelSerializer).as_json
+        expected = { data: { id: '1', type: 'posts', relationships: { blog: { data: { id: '5', type: 'blogs' } } } } }
+
+        assert_equal expected, actual
+      end
+
       class InlineAssociationTestPostSerializer < ActiveModel::Serializer
         has_many :comments
         has_many :comments, key: :last_comments do
--- a/lib/active_model/serializer/belongs_to_reflection.rb
+++ b/lib/active_model/serializer/belongs_to_reflection.rb
@@ -2,6 +2,10 @@ module ActiveModel
   class Serializer
     # @api private
     class BelongsToReflection < Reflection
+      # @api private
+      def foreign_key_on
+        :self
+      end
     end
   end
 end


NullVoxPopuli:
Do you plan to fix the bug, or is this just to demonstrate the problem? (either is cool)

lib/active_model/serializer/reflection.rb
```diff
@@ -106,42 +116,44 @@ def value(serializer)
...
+              reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
+            end
+          elsif !association_value.nil? && !association_value.instance_of?(Object)
+            reflection_options[:virtual_value] = association_value
+          end
+        }
         reflection_options = options.dup
-        serializer_class = subject.class.serializer_for(association_value, reflection_options)
         reflection_options[:include_data] = @_include_data
+        reflection_options[:new_options] = {
```
bf4: hack to allow for refactoring...

lib/active_model/serializer/association.rb
```diff
@@ -9,10 +9,46 @@ class Serializer
...
+
+      def to_many?
+        reflection.name == :has_many
+      end
+
+      def belongs_to?
+        reflection.foreign_key_on == :self
+      end
+
+      def serializer
```
bf4: the serializer really shouldn't be evaluated until needed. Unfortunately, a lot of our code very much depends on expecting it to be evaluated. This implementation makes it possible to make it lazy.  It's not ideal, but is just a step to getting it right.

Unfortunately, associations are evaluated in two places: in `Serializer#relationship_value_for` and in `JsonApi#relationships_for`. Would be easier in one spot.  

The other problem, is that associations are built using the include directive, but it's not clear when the attributes are requested, especially since the include directive relies on the `*` behavior

You'd think

```ruby
        requested_associations = fieldset.fields_for(resource_object[:type]) || '*'
        relationships = relationships_for(serializer, requested_associations)
```

would imply which fields to get....

NullVoxPopuli: this is good to know.

test/serializers/associations_test.rb
```diff
@@ -125,6 +125,26 @@ def test_associations_custom_keys
...
+
+      def test_belongs_to_doesnt_load_not_included_record_for_jsonapi
+        post = Post.new(blog_id: 'the_blog_id')
+        def post.blog
+          fail 'should use blog_id'
+        end
+        def post.read_attribute_for_serialization(name)
+          # Bug in how ActiveModelSerializers::Model handles attributes and methods
+          send(name)
+        end
```
bf4: The current impl is

```ruby
      if key == :id || key == 'id'
        attributes.fetch(key) { id }
      else
        attributes[key]
      end
    end
```

Which doesn't send and relies overly much on the attributes method, which is only set at initialize... This is definitely a bug that needs to be fixed.....

Until I wrote this, the test was't raising the error

bf4:
I think I got it working and all tests passing! Now time for review.

test/adapter/json_api/relationship_test.rb
```diff
@@ -4,157 +4,157 @@ module ActiveModelSerializers
...
+        #   test_relationship(expected, meta: meta, options: { include_data: true }, links: links)
+        # end
+        #
+        # private
+        #
+        # def test_relationship(expected, params = {})
+        #   parent_serializer = AuthorSerializer.new(@author)
+        #   relationship = Relationship.new(parent_serializer, @serializer, nil, params)
+        #   assert_equal(expected, relationship.as_json)
+        # end
```
bf4: These tests, like much of the JSONAPI code, are tightly coupled to the implementation.   

They are asserting a relationship between two serializers are that are not necessarily actually related.

The impl in this PR is to pass in the parent serializer and the association.  

In these tests, the association would need to be derived from the `@serializer`. I couldn't get this test suite to pass, to am relying on the remainder of the test suite.

bf4: (DO NOT merge with this commented out... tbd what to do about it)

lib/active_model_serializers/adapter/json_api/relationship.rb
```diff
@@ -38,15 +38,24 @@ def as_json
 
         private
 
-        def data_for(serializer)
-          if serializer.respond_to?(:each)
+        def data_for(association)
+          if belongs_to_with_foreign_key?(association)
```
bf4: This is the change desired by the PR

lib/active_model/serializer/reflection.rb
```diff
@@ -106,42 +116,44 @@ def value(serializer)
       #
       # @api private
       #
-      def build_association(subject, parent_serializer_options)
-        association_value = value(subject)
+      def build_association(parent_serializer, parent_serializer_options)
+        association_value_proc = proc do
+          value(parent_serializer)
+        end
+        serializer_proc = ->(association_value, reflection_options) {
```
bf4: hack to allow minimal code changes while lazy-evaluating the association value and its derived serializer

lib/active_model/serializer/association.rb
```diff
@@ -9,10 +9,58 @@ class Serializer
...
+        "#{reflection.type.to_s.singularize}_type"
+      end
+
+      def value
+        @value ||= association_value = reflection_options.fetch(:association_value_proc).call
+      end
+
+      def serializer
+        defined?(@reified_serializer) ? @reified_serializer : reify!
+      end
```
bf4: Hack to work around this being an attribute of the Struct, and current public interface method, and me not wanting to make too many unnecessary changes

lib/active_model/serializer/association.rb
```diff
@@ -9,10 +9,58 @@ class Serializer
...
+        reflection.name == :has_many
+      end
+
+      def belongs_to?
+        reflection.foreign_key_on == :self
+      end
+
+      def polymorphic_type
+        "#{reflection.type.to_s.singularize}_type"
+      end
```
bf4: not used at present, but should be

lib/active_model/serializer/collection_reflection.rb
```diff
@@ -2,6 +2,12 @@ module ActiveModel
   class Serializer
     # @api private
     class CollectionReflection < Reflection
+      def initialize(*)
+        super
+        @class_name = options.fetch(:class_name, name.to_s.camelize.singularize)
+        @type = class_name.underscore.pluralize.to_sym
+        @foreign_key ||= "#{name.to_s.singularize}_ids".to_sym
```
bf4: `class_name` and `type` should be used in more cases for inferring a serializer, rather that requiring an instantiated associated object.  This can move us in a direction where a serializer on its own an only serialize itself and its associations and only an adapter can get it to serialize deeper. And each association on the serializer can have the serializer set else is inferred.

lib/active_model/serializer/association.rb
```diff
@@ -9,10 +9,58 @@ class Serializer
...
+        @reflection_options = options.delete(:new_options)
+      end
+
       # @return [Symbol]
       def key
-        options.fetch(:key, name)
+        @key ||= options.fetch(:key) { name }
+      end
+
+      def links
```
NullVoxPopuli: So, something that's confusing for me, is that this association here is a struct, where as the generic reflection defined below is a class, and they are both under the `active_model/serializer` namespace.  *I* know they have different purposes, but I think the organization of the classes should reflect the different purposes as well. Maybe a namespace for definitions, and a namespace for representation? or maybe just a namespace for representation, since in a serializer, we're most often working with the definitions of things.

thoughts?

lib/active_model/serializer/reflection.rb
NullVoxPopuli: I was just about to ask about this, thanks

lib/active_model/serializer/association.rb
bf4: yup, it got missed in the convergent evolution of the other classes.

bf4:
cc @youroff @ssendev @sandstrom @richmolj @joaomdmoura  @vasilakisfil @beauby

bf4:
(Test failure right now are regularly occurring but intermittent... tests pass though, and pr needs refactoring, so this is a good chance to look)

lib/active_model/serializer/association.rb
bf4: In my wip sketch of how associations work, you see everything else comes from Field

```
ActiveModel::Serializer # serializer_instance
|>  #as_json(adapter_opts = nil)
    |>  #serializable_hash(adapter_options, options, adapter_instance)
        adapter_options ||= {}
        include_directive = options.fetch(:include_directive) { ActiveModel::Serializer.include_directive_from_options(adapter_options) }
        fields = options[:fields]
        |
        |>  #attributes(fields)
        |   self.class._attributes_data.each_with_object({}) do |(key, attr), hash|
        |     |>(attr) ActiveModel::Serializer::Attribute (is a ActiveModel::Serializer::Field)
        |       next if attr.excluded?(self) # serializer_instance
        |       next unless fields.nil? || fields.include?(key)
        |       hash[key] = attr.value(self) # read_attribute_for_serialization # serializer_instance
        |
        |>  #associations(include_directive)
            relationships = {}
            self.class._reflections.values.each do |reflection|
            |>(reflection) is a ActiveModel::Serializer::Reflection subclass (is a ActiveModel::Serializer::Field subclass)
              next if reflection.excluded?(self) # serializer_instance
              next unless include_directive.key?(key) # serializer_instance
              #reflection.build_association(self, instance_options) # serializer_instance
                association_serializer.new(
                  read_attribute_for_serialization(name) # association name
                )
                |>(association) is ActiveModel::Serializer::Association (is a struct)
                  yield |association|
                    adapter_opts = adapter_options.merge(include_directive: include_directive[association.key])
                    relationships[association.key] ||= association.serializer.serializable_hash(adapter_opts)

JSON API Adapter
  initialize
    @include_directive = JSONAPI::IncludeDirective.new(options[:include], allow_wildcard: true)
    @fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(options.delete(:fields))
  serializable_hash
    is_collection = serializer.respond_to?(:each)
    serializers = is_collection ? serializer : [serializer]
    @primary = []
    @included = []
    @resource_identifiers = Set.new
    serializers.each { |serializer| process_resource(serializer, true) }
    serializers.each { |serializer| process_relationships(serializer, @include_directive) }
      |>(process_relationships)
        serializer.associations(include_directive).each do |association|
          |>(process_relationship)
            association_serializer = association.serializer
            association_include_directive = include_directive[association.key]
              serializer.each { |s| process_relationship(s, include_directive) }
              return
            end
            return unless serializer && serializer.object
            return unless process_resource(serializer, false)
            |>(process_resource)
              resource_identifier = ResourceIdentifier.new(serializer, instance_options).as_json
              return false unless @resource_identifiers.add?(resource_identifier)

              resource_object = resource_object_for(serializer)
              |>(resource_object_for)
                resource_object = ResourceIdentifier.new(serializer, instance_options).as_json
                requested_fields = fieldset && fieldset.fields_for(resource_object[:type])
                attributes = attributes_for(serializer, requested_fields)
                |>(attributes_for)
                |
                |>(relationships_for)
                  include_directive = JSONAPI::IncludeDirective.new(
                    fieldset.fields_for(resource_object[:type]) || '*'
                    requested_associations,
                    allow_wildcard: true
                  )
                  serializer.associations(include_directive).each_with_object({}) do |association, hash|
                    hash[association.key] = Relationship.new(args).as_json
                    |>
                      association_options = args.fetch(:options, {})
                      data =
                        if association.to_many?
                          serializer = association.serializer
                          serializer.map { |s| ResourceIdentifier.new(s, serializable_resource_options).as_json }
                        elsif association_options[:virtual_value]
                          association_options[:virtual_value]
                        elsif association.belongs_to?
                          foreign_key = association.reflection.foreign_key
                          if @object.respond_to?(foreign_key)
                            {
                              id: @object.send(foreign_key),
                              type: association.reflection.type.to_s
                            }
                          else
                            serializer && serializer.object &&
                              ResourceIdentifier.new(serializer, serializable_resource_options).as_json
                          end

                      json[:data] = data if association_options[:include_data]
```

sandstrom:
@bf4 this is a great feature!

Regarding "This will only work on active record objects", would it be possible to define an interface (+ default handling for ActiveRecord) such that other ORMs can use it too (or use some other mechanism, basically anything that'll make this useful for other ORMs that are similar to ActiveRecord). For example Mongoid.

bf4:
Well, I wrote it using respond_to, so if it quacks that way it will work.  I'm not familiar with the interface needed by mongoid

B mobile phone

lib/active_model/serializer/collection_reflection.rb
```diff
@@ -2,6 +2,12 @@ module ActiveModel
   class Serializer
     # @api private
     class CollectionReflection < Reflection
+      def initialize(*)
+        super
+        @class_name = options.fetch(:class_name, name.to_s.camelize.singularize)
```
NullVoxPopuli: should these maybe be methods?

lib/active_model/serializer/reflection.rb
```diff
@@ -106,42 +116,44 @@ def value(serializer)
...
-
-        if serializer_class
-          begin
-            serializer = serializer_class.new(
-              association_value,
-              serializer_options(subject, parent_serializer_options, reflection_options)
-            )
-          rescue ActiveModel::Serializer::CollectionSerializer::NoSerializerError
-            reflection_options[:virtual_value] = association_value.try(:as_json) || association_value
+      def build_association(parent_serializer, parent_serializer_options)
```
NullVoxPopuli: this needs to be broken down

lib/active_model/serializer/singular_reflection.rb
```diff
@@ -2,6 +2,12 @@ module ActiveModel
   class Serializer
     # @api private
     class SingularReflection < Reflection
+      def initialize(*)
+        super
+        @class_name = options.fetch(:class_name, name.to_s.camelize)
```
NullVoxPopuli: should these maybe be methods on `Reflection`?

lib/active_model_serializers/adapter/json_api/relationship.rb
```diff
@@ -38,15 +38,24 @@ def as_json
...
+            id = @object.send(association.reflection.foreign_key)
+            type = association.reflection.type.to_s
```
NullVoxPopuli: i think this block should be extracted out to a method. So, if we wanted, we could test the construction of the ResourceIdentifier.
Though, This may be an opportunity for polymorphic build_resource_identifier later down the road

richmolj:
Looks like great work on a good feature 👍 However, I'm curious about the PR description:

> The way this has been solved in JSONAPI-Resources is to have special rules for looking up association ids.

I'd like to better understand this. JSONAPI-resources will not serialize the resource identifiers (`id/type`) if the relationship is not specified in the `?include` parameter (at least, this is the default, there may be an option I don't know about). So there would never be a scenario where they would have to avoid fetching the relationship while still providing the `id/type`, AFAIK.

IOW:

> else if it is JSON API and the related resource
    *is not included, serialize just the ids
    *is included, serialize the all fields, or the requested fields

This logic differs from JSONAPI-resources, which causes issues for all relationship types. This PR fixes the `belongs_to` type. Is there any plan to eventually change this logic so that all relationship types are fixed?

NullVoxPopuli:
>  Is there any plan to eventually change this logic so that all relationship types are fixed?

yeah, it will likely just be a call to `"#{relationship_name}_ids"` and then `each_with_object({})` to turn it in to resource identifiers. a has_many would make one request still (if not already loaded)

> So there would never be a scenario where they would have to avoid fetching the relationship while still providing the id/type

I think the issue with this is coming up with a clear, easy to stand API to determine what is happening.
e.g.: how do we specify a relationship that always includes the resource identifier? and how do we only include it when `includes` is used.

this is tricky, because frameworks like Ember utilize these relationship identifiers for front-end lazy-loading. (`async: true` (default) in ember-data)

richmolj:
@NullVoxPopuli I'm an Ember user. I agree including the id/types is a valid use case to support a client like that. So this PR is great to see.

FWIW, though, I get around this by providing `links`. An ember `async: true` relationship will follow the `related` link to lazy-load data (I believe it will do this even if resource identifiers are present but could be wrong). It's particularly important to do this when a `has_many` relationship contains tens of thousands of records (ie a post with ten thousand comments).

So my question is, are you guys still planning to support this use case (only including resource identifiers when included, as jsonapi-resources does) as well?

NullVoxPopuli:
> So my question is, are you guys still planning to support this use case (only including resource identifiers when included, as jsonapi-resources does) as well?

We need to, we just don't know the best api for it in the serializers. How does jsonapi-resources differentiate between the two situations?

richmolj:
@NullVoxPopuli Cool, thanks, just wanted to confirm my understanding.

> How does jsonapi-resources differentiate between the two situations?

AFAIK jsonapi-resources does not support resource identifiers when the relationship is not in the `?include` parameter. If it does, it's some customization I don't know about (does not seem to be in the README).

>  we just don't know the best api for it in the serializers

Two thoughts: If you want to use the current API, you could enhance `include_data` - `include_data :always` vs. `include_data :if_included`. I personally think a better API would be a class attribute on the serializer so you don't have to customize every relationship, but either way.

(my apologies if this is derailing the thread)

richmolj:
FWIW, that's the bit actually relevant to this PR and jsonapi-resources - it does have a separate belongs_to logic, but for a different purpose than this PR is trying to support. I believe it is used to generate the `link` to the resource (e.g. `/authors/1`). From the README:

```ruby
# Resource Linkage
# Controls the serialization of resource linkage for non compound documents
# NOTE: always_include_to_many_linkage_data is not currently implemented
config.always_include_to_one_linkage_data = false
```

lib/active_model/serializer/collection_reflection.rb
bf4: as oppose to in the initializer?

lib/active_model/serializer/reflection.rb
bf4: yup

lib/active_model/serializer/singular_reflection.rb
bf4: Make sense

I think I modeled it on JSONAPI::Resources... (which has a simpler structure as well)

lib/active_model_serializers/adapter/json_api/relationship.rb
bf4: You mean like 

```ruby
if association.belongs_to_with_foreign_key?
  id = association.reflect_id(@object)
  type = association.reflect_type
```

lib/active_model/serializer/collection_reflection.rb
NullVoxPopuli: I'm just wary of too much logic in an initializer / any method

lib/active_model_serializers/adapter/json_api/relationship.rb
NullVoxPopuli: that would be a good change too.

what I was kind of thinking is abstracting out more methods for the conditions to read more like what the the functionality is:

like, on line 46, maybe: `association.is_collection?`

and then the building of the ResourceIdentifier(s, opt).as_json could be its own method, since serializable_resource_options are outside the scope of _this_ method. maybe:

```ruby
def jsonized_resource_for(serializer)
  ResourceIdentifier.new(serializer, serializable_resource_options).as_json
end
```

idk

that also allows us to test the logic of `data_for` without worrying about if the implementation of `ResourceIdentifier` is correct.

test/adapter/json_api/relationship_test.rb
```diff
@@ -152,18 +152,20 @@ def test_relationship_with_everything
...
 
-          meta = test_options.delete(:meta)
-          options = test_options.delete(:options)
-          links = test_options.delete(:links)
-          association_serializer = @serializer
+          options = test_options.delete(:options) || {}
+          options[:new_options] = {
+            meta: test_options.delete(:meta),
+            links: test_options.delete(:links),
+            reflection: parent_serializer._reflections.to_a.first
```
bf4: I know this is broken (in the rebase off of master), but just wanted to push the work for the moment

test/adapter/json_api/relationship_test.rb
```diff
@@ -4,132 +4,204 @@ module ActiveModelSerializers
...
+
         setup do
-          @blog = Blog.new(id: 1)
-          @author = Author.new(id: 1, name: 'Steve K.', blog: @blog)
-          @serializer = BlogSerializer.new(@blog)
-          ActionController::Base.cache_store.clear
+          @serializable_resource_options = {} # adapter.instance_options
+        end
+
+        def build_author_relationship(author_attributes, association_name)
```
bf4: I completely rewrote the test harness and assertion in this file.. could use a second pair of eyes, but at least I think it is straightforward enough to read the assertions and understand, and to read the harness and understand what it's doing (even if not so clear why it's so complicated, and how could it be simpler).

test/serializers/associations_test.rb
```diff
@@ -125,6 +125,27 @@ def test_associations_custom_keys
...
+
+      def test_belongs_to_doesnt_load_not_included_record_for_jsonapi
+        post = Post.new(blog_id: 'the_blog_id')
+        def post.blog
+          fail 'should use blog_id'
+        end
+
+        def post.read_attribute_for_serialization(name)
+          # Bug in how ActiveModelSerializers::Model handles attributes and methods
+          send(name)
```
bf4: ❗

lib/active_model/serializer/associations.rb
```diff
@@ -84,17 +84,22 @@ def associate(reflection)
...
-            y.yield reflection.build_association(self, instance_options)
+          self.class._reflections.keys.each do |association_key|
+            assoc = association(association_key, include_directive)
+            next if assoc.nil?
+            y.yield assoc
           end
         end
       end
+
+      def association(association_key, include_directive = ActiveModelSerializers.default_include_directive)
```
bf4: not required in this PR, and only tested implicitly, but seems to make using associations a little easier

test/adapter/json_api/relationship_test.rb
NullVoxPopuli: I think it's pretty straight forward, usage wise.

though, why do you need to reset the reflection stuff?
is that for isolation or something else?
my concern is that if something is happening to those normally that is being avoided here, it could reveal a hole in testing.

test/adapter/json_api/relationship_test.rb
```diff
@@ -4,132 +4,204 @@ module ActiveModelSerializers
...
 
         def test_relationship_with_data
-          expected = {
-            data: {
-              id: '1',
-              type: 'blogs'
-            }
-          }
-          test_relationship(expected, options: { include_data: true })
+          author_attributes = { id: 1, name: 'Steve K.', blog: Blog.new(id: 1) }
```
NullVoxPopuli: much better than test_relationship :-)

test/serializers/associations_test.rb
NullVoxPopuli: !!!

NullVoxPopuli: does that comment saying you're implementing the bug, or avoiding it? I'm not clear on that one.

lib/active_model_serializers/adapter/json_api/resource_identifier.rb
```diff
@@ -22,6 +22,13 @@ def self.type_for(class_name, serializer_type = nil, transform_options = {})
           JsonApi.send(:transform_key_casing!, raw_type, transform_options)
         end
 
+        def self.for_type_with_id(type, id, options)
```
NullVoxPopuli: nice method name :-)

lib/active_model/serializer/associations.rb
```diff
@@ -84,17 +84,22 @@ def associate(reflection)
...
+        return nil if object.nil?
+        reflection = _reflections[association_key]
+        return nil if reflection.excluded?(self)
+        return nil unless include_directive.key?(association_key)
```
NullVoxPopuli: does an association include the resource identifier?

is the include directive the one that determines `included`ness? 
otherwise, how does the include_directive get populated?

Like, if I have an author, and the blogs relationship is rendered as just resource identifiers. How does this code play in to that?

test/adapter/json_api/relationship_test.rb
bf4: Just that normally these are created by the serializer using these as defaults in the reflection initializer, but here I am mutating them, which changes them for all the tests, and dup and deep_dup didn't work.  So it's a hack to make the serializer behave in each test as if it were being defined for the first time.  

B mobile phone

test/serializers/associations_test.rb
bf4: It's a bug that needs to be fixed, that changes in attributes via access don't affect the attribute hash, and that the read is something something

lib/active_model/serializer/reflection.rb
```diff
@@ -106,42 +116,44 @@ def value(serializer)
...
+          meta: @_meta,
+          reflection: self,
+          association_value_proc: association_value_proc,
+          serializer_proc: serializer_proc,
+          parent_serializer: parent_serializer,
+          parent_serializer_options: parent_serializer_options
+        }
 
-        Association.new(name, serializer, reflection_options, @_links, @_meta)
+        Association.new(name, nil, reflection_options)
```
bf4: TODO: remove this option from the Association struct since it's not using it

lib/active_model/serializer/association.rb
```diff
@@ -9,10 +9,58 @@ class Serializer
     # @example
     #  Association.new(:comments, CommentSummarySerializer)
     #
-    Association = Struct.new(:name, :serializer, :options, :links, :meta) do
+    Association = Struct.new(:name, :serializer, :options) do
```
bf4: this should probably just subclass  `ActiveModel::Serializer::Field` like everything else. see https://github.com/rails-api/active_model_serializers/pull/1857/files#r73813576

NullVoxPopuli: yes

NullVoxPopuli:
An issue I noticed when playing around with this briefly:


pasted image at 2016_08_30 10_15 pm

sponsor should be of type `organizations`, and sponsored should be of type `events`

belongs_to associations now use the association_name for the model name, rather than the type of the corresponding serializer

bf4:
I just rebased and broke a lot of stuff... I'll need to review the diff for what changes we actually want at this point.  So, yes, lots of failures now that need to be addressed, it is known.

furkanayhan:
@bf4 Hey, thank you for this improvement.
But there is something wrong with polymorphic-belongs_to relationships.

Examle:

serializer:
```ruby
class CommentSerializer < ActiveModel::Serializer
  # ....
  belongs_to :commentable
end
```

request:
`/comments?include=commentable`

result:
```js
//...
"relationships": {
  "commentable": {
    "data": {
      "id": "1",
      "type": "commentables"
    }
  }
}
//...
```

Can you check it out?

lib/active_model/serializer/reflection.rb
```diff
@@ -47,11 +47,23 @@ class Serializer
...
     class Reflection < Field
+      attr_reader :foreign_key, :type
+
       def initialize(*)
         super
         options[:links] = {}
         options[:include_data_setting] = Serializer.config.include_data_default
         options[:meta] = nil
+        @type = options.fetch(:type) do
+          class_name = options.fetch(:class_name, name.to_s.camelize.singularize)
```
andyhite: This line is causing the bug reported here: https://github.com/rails-api/active_model_serializers/issues/2160 In particular, the use of the field's `name` if the `class_name` and `type` aren't specified breaks belongs_to associations.

PR #1861, opened 2016-07-27
Specify how to get the instance of the serializer

Tonkpils:
#### Changes

The docs mentioned `#serializer` returned the instance but it actually returns the class of the serializer that will be used. Added an extra about using `#serializer_instance`


bf4:
@Tonkpils Thanks for this. This has been an occasional source of confusion for me, as well.

PR #1863, opened 2016-08-01
Tests Refactor

NullVoxPopuli:
#### Purpose

Light re-organization to hopefully be a little more clear where certain types of tests are / go.

#### Changes

Test files move around


test/test_helper.rb
```diff
@@ -47,9 +47,8 @@ def serialization_options(options)
 
 require 'support/rails5_shims'
 
-require 'fixtures/active_record'
-
-require 'fixtures/poro'
+require 'support/fixtures/active_record'
+require 'support/fixtures/poro'
```
bf4: I think the Rails convention is for fixtures to go in the test directory http://guides.rubyonrails.org/testing.html#the-low-down-on-fixtures though admittedly these are not the same datatype as referenced in the link and I have seen both file locations...  thoughts?

bf4:
@NullVoxPopuli the diff is kind of hard to read. Can you summarize the renamings?

NullVoxPopuli:
I moved most of the tests to `unit/` and, grape and action_controller to `acceptance/` I also moved some general files to `support/`, like fixtures, but the error I'm getting is because Travis can't find the file for some reason. It's like there is a different current path when Travis runs the tests. I might need to do a require relative or something to get those tests passing.

Some of the weirdness with our unit tests is the am/s vs ams pathing.  I'd like to get that straightened out later in my giant refactor / unification.

test/test_helper.rb
NullVoxPopuli: I like to try to keep the number of folders per folder reasonably lowish. I guess the argument for moving to fixtures also could be "the tests don't *require* fixtures, it just makes reusing objects a little more DRY"

bf4: ok

bf4: (but DRY is the wrong word, here)

test/unit/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -38,7 +38,7 @@ class WithoutRenderer < JsonApiRendererTest
       require 'active_record'
       require 'support/rails5_shims'
       require 'active_model_serializers'
-      require 'fixtures/poro'
+      require_relative '../../support/fixtures/poro'
```
bf4: when tests run they include the `test` directory in the search path, so `require 'support/fixtures/poro'` is what you want

bf4:
Is `unit/acceptance` the current convention?

test/test_helper.rb
NullVoxPopuli: yeah. 

I'm kinda thinking tests should utilize less of the poros, and just define what they need in the test class. But... for the future, I guess.

NullVoxPopuli:
in rails, the tests are organized by top-level app folder (mostly)  - the exception is `feature` and `requests`

so, rails kinda has two things going on, where feature, and requests are both types of acceptance testing, and everything else is unit/integration.

I'd like to see rails eventually have less folders under ./tests or ./specs in the future, and be more explicit about the types of tests. (and especially getting away from controller tests altogether, and just doing request tests/specs, as there is less hacking with request specs in the test env.)

test/unit/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
NullVoxPopuli: cool, hopefully that fixes it. I don't know why I want straigh to ../support. :-\ (I had to check the commit history to make sure I didn't already try `require 'support/....`

NullVoxPopuli:
hey that change worked! Thanks, @bf4

bf4:
I'd like to get another maintainer to 👍 since it's a structural and philosophical change to the tests and we want to minimize that

Rails is getting rid of controller tests.  As far as I can tell, Rails still has `models` and `integration` tests.  The unit/acceptance thing went away at some point IIRC

NullVoxPopuli:
Fair enough.

One thing I'd like to point out for whoever reviews next is that rails divides up test folders by layer in the over all (MVC(xyz)) layers.

For gems, I think it's good to differentiate between the types of tests, as in the example of acceptance tests -- it's easier for people to find where / how to use the gem in intended environments. I think the acceptance folder could be divided in to topics, and would eventually be our runnable/testable documentation.

The unit tests should still follow the directory structure of lib, and try to be as true to the idea of a 'unit test' as possible - where if something is broken, a test should be able to tell you exactly how to fix it.

NullVoxPopuli:
Closing in favor of another testing strategy

PR #1767, opened 2016-06-01
Replace fail/rescue CollectionSerializer::NoSerializerError with throw/catch :no_serializer

bf4:
Closes https://github.com/rails-api/active_model_serializers/pull/1693

Fixes https://github.com/rails-api/active_model_serializers/issues/1075

PR #1867, opened 2016-08-02
Infer serializer from collection name for empty results root

mchitten:
#### Purpose

In the event that a collection object is empty and has a name, it's possible to determine the serializer from the name of the object. This may not always be the case -- this could raise `NameError` for a class that does not exist, or `NoSerializerError` if its matching serializer does not exist, in which case this falls back to the original `name.underscore` implementation.
#### Changes

Updates `CollectionSerializer` to try to determine serializer from object name.
#### Related GitHub issues

Possibly(?) a fix for #1745.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Features:
 Fixes:
 
 - [#1833](https://github.com/rails-api/active_model_serializers/pull/1833) Remove relationship links if they are null (@groyoh)
+- [#1867](https://github.com/rails-api/active_model_serializers/pull/1867) Infer collection root preferably from serializer matching collection name, if a named collection. (@mchitten)
 
 Misc:
 
--- a/lib/active_model/serializer/collection_serializer.rb
+++ b/lib/active_model/serializer/collection_serializer.rb
@@ -34,6 +34,7 @@ module ActiveModel
       # we have no simple way to know that it is safe to call that instance method.
       # (which is really a class property at this point, anyhow).
       # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/PerceivedComplexity
       # Disabling cop since it's good to highlight the complexity of this method by
       # including all the logic right here.
       def json_key
@@ -45,11 +46,18 @@ module ActiveModel
         # 2. get from first serializer instance in collection
         key ||= (serializer = serializers.first) && serializer.json_key
         # 3. get from collection name, if a named collection
-        key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
+        key ||= if object.respond_to?(:name) && (object_name = object.name)
+                  if (object_class = object_name.safe_constantize)
+                    serializer_class = serializer_from_resource(object_class.new, ActiveModel::Serializer, options) { nil }
+                  end
+                  (serializer_class && serializer_class.json_key) || object_name.underscore
+                end
+
         # 4. key may be nil for empty collection and no serializer option
         key && key.pluralize
       end
       # rubocop:enable Metrics/CyclomaticComplexity
+      # rubocop:enable Metrics/PerceivedComplexity
 
       def paginated?
         object.respond_to?(:current_page) &&
@@ -73,8 +81,9 @@ module ActiveModel
       def serializer_from_resource(resource, serializer_context_class, options)
         serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
 
-        if serializer_class.nil? # rubocop:disable Style/GuardClause
-          fail NoSerializerError, "No serializer found for resource: #{resource.inspect}"
+        if serializer_class.nil?
+          block_given? && yield ||
+            fail(NoSerializerError, "No serializer found for resource: #{resource.inspect}")
         else
           serializer_class.new(resource, options.except(:serializer))
         end
--- a/test/collection_serializer_test.rb
+++ b/test/collection_serializer_test.rb
@@ -23,6 +23,21 @@ module ActiveModel
         resource
       end
 
+      def build_named_collection_with_matching_serializer(*resource)
+        serialized_resource = Class.new(::Model)
+        serializer = Class.new(ActiveModel::Serializer) do
+          def json_key
+            'resources'
+          end
+        end
+
+        Object.const_set(:SerializedResource, serialized_resource)
+        Object.const_set(:SerializedResourceSerializer, serializer)
+
+        resource.define_singleton_method(:name) { 'SerializedResource' }
+        resource
+      end
+
       def test_has_object_reader_serializer_interface
         assert_equal @serializer.object, @resource
       end
@@ -74,6 +89,11 @@ module ActiveModel
       def test_json_key_with_resource_with_name_and_no_serializers
         serializer = collection_serializer.new(build_named_collection)
         assert_equal 'me_resources', serializer.json_key
+      end
+
+      def test_json_key_with_resource_with_name_and_serializer
+        serializer = collection_serializer.new(build_named_collection_with_matching_serializer)
+        assert_equal 'resources', serializer.json_key
       end
 
       def test_json_key_with_resource_with_nil_name_and_no_serializers


lib/active_model/serializer/collection_serializer.rb
```diff
@@ -44,8 +44,16 @@ def json_key
           explicit_serializer_class._type
         # 2. get from first serializer instance in collection
         key ||= (serializer = serializers.first) && serializer.json_key
-        # 3. get from collection name, if a named collection
-        key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
+        if object.respond_to?(:name)
+          begin
+            # 3. use json_key from serializer from collection name
+            key ||= serializer_from_resource(object.name.safe_constantize.new, ActiveModel::Serializer, options).json_key
+          rescue NoSerializerError, NameError
```
bf4: Sort of wishing I had merged https://github.com/rails-api/active_model_serializers/pull/1767 at this point...

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -44,8 +44,16 @@ def json_key
           explicit_serializer_class._type
         # 2. get from first serializer instance in collection
         key ||= (serializer = serializers.first) && serializer.json_key
-        # 3. get from collection name, if a named collection
-        key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
+        if object.respond_to?(:name)
```
bf4:
```ruby
        key ||= begin
          object_name = object.respond_to?(:name) ? object.name : nil
          if (object_class = object_name && object_name.safe_constantize)
            serializer_class = serializer_from_resource(object_class.new, ActiveModel::Serializer, options) { nil }
          end
          serializer_class.nil? : object_name.underscore : serializer_class.json_key
        end
```

just thinking aloud...

```diff
     def serializer_from_resource(resource, serializer_context_class, options)
        serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }

        if serializer_class.nil? # rubocop:disable Style/GuardClause
-          fail NoSerializerError, "No serializer found for resource: #{resource.inspect}"
+          block_given? ? yield : 
+            fail NoSerializerError, "No serializer found for resource: #{resource.inspect}"
        else
          serializer_class.new(resource, options.except(:serializer))
        end
      end
```

mchitten: This is great, and definitely a better idea than the `begin...rescue`. I had a couple of tweaks, but updated with more or less the same.

bf4: Thanks, was a bit messy as I wrote it... what do you think about changing to throw/catch?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Features:
 Fixes:
 
 - [#1833](https://github.com/rails-api/active_model_serializers/pull/1833) Remove relationship links if they are null (@groyoh)
+- [#1867](https://github.com/rails-api/active_model_serializers/pull/1867) Infer collection root from collection name if possible (@mchitten)
```
bf4: this is really something like 

> Infer collection root preferably from the serializer for the collection name then the collection name, if a named collection

?

test/collection_serializer_test.rb
```diff
@@ -1,5 +1,15 @@
...
 
+class SerializedResource < ActiveModelSerializers::Model
+  # See test line 94
+end
+
+class SerializedResourceSerializer < ActiveModel::Serializer
+  def json_key
+    'resource'
+  end
+end
```
bf4: Seems you are wanting these in the global scope. Can we perhaps dynamically create and remove these classes or alias exiting ones? (.e.g SerializedResource = Post...)

test/collection_serializer_test.rb
```diff
@@ -23,6 +33,11 @@ def build_named_collection(*resource)
         resource
       end
 
+      def build_named_collection_with_matching_serializer(*resource)
+        resource.define_singleton_method(:name) { 'SerializedResource' }
```
bf4: do you have a real-world example of this?  Previously there was a test for a monkey-patched array instance and it made me unsure if this was a real feature

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -73,7 +80,8 @@ def serializers_from_resources
       def serializer_from_resource(resource, serializer_context_class, options)
         serializer_class = options.fetch(:serializer) { serializer_context_class.serializer_for(resource) }
 
-        if serializer_class.nil? # rubocop:disable Style/GuardClause
+        if serializer_class.nil?
+          yield if block_given?
```
bf4: For this pattern to work, it must either yield or raise.  As written it will yield then raise

```ruby
block_given? and yield or
           fail NoSerializerError, "No serializer found for resource: #{resource.inspect}"
```

would work, I think (not sure about the `and`)

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -45,7 +45,14 @@ def json_key
...
         # 3. get from collection name, if a named collection
-        key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
+        key ||= begin
+                  object_name = object.respond_to?(:name) ? object.name : nil
+                  if (object_name && object_class = object_name.safe_constantize)
+                    serializer_class = serializer_from_resource(object_class.new, ActiveModel::Serializer, options) { nil }
+                  end
+                  serializer_class.nil? ? (object_name && object_name.underscore) : serializer_class.json_key
+                end
+
```
bf4: ping me when you're done futzing with this. back to work :)

test/collection_serializer_test.rb
mchitten: When I first came across this weirdness, it was actually while testing, and on our API I wasn't seeing the issue. I just did some more poking around and found that it might actually have something to do with the `#none` method.

In this case, `tags` is the expected json_key as defined in the `ResourceTagSerializer`.

```
[1] pry(main)> ActiveModel::Serializer::CollectionSerializer.new(ResourceTag.none, {}).json_key
=> "resource_tags"
[2] pry(main)> ActiveModel::Serializer::CollectionSerializer.new(ResourceTag.last(3), {}).json_key
=> "tags"
```

lib/active_model/serializer/collection_serializer.rb
mchitten: Sorry -- was fighting an uphill battle with rubocop's `Metrics/PerceivedComplexity`. Based on your comment line 38 I figured it might be okay to disable that for now? Let me know if that's misguided and you'd prefer me to break out into other methods.

lib/active_model/serializer/collection_serializer.rb
mchitten: I think the `throw` / `catch` idea makes a lot of sense. Does it make sense to throw that in here too? With the exception of the `yield` as you mentioned, I feel like it doesn't actually affect this code anymore sans-`rescue`.

mchitten:
@bf4 Ready for another look when you get the chance

mchitten:
Hey @bf4, any additional thoughts on this? Thanks!

test/collection_serializer_test.rb
```diff
@@ -23,6 +23,21 @@ def build_named_collection(*resource)
...
+        serializer = Class.new(ActiveModel::Serializer) do
+          def json_key
+            'resources'
+          end
+        end
+
+        Object.const_set(:SerializedResource, serialized_resource)
+        Object.const_set(:SerializedResourceSerializer, serializer)
+
+        resource.define_singleton_method(:name) { 'SerializedResource' }
```
bf4: 1. I'd rather this be in a block so that we can also teardown the classes after defining them
1. I think we can consolidate the code here a little

```ruby
def with_serializable_named_model(name)
  model = Class.new(::Model) do
    def self.name
      'SerializableResource'
    end
  end
  Object.const_set(:SerializableResource, model)
  serializer = Class.new(ActiveModel::Serializer) do
    def json_key
      name
    end
  end
  Object.const_set(:SerializableResourceSerializer, serializer)
  yield model
ensure
  Object.send(:remove_const, :SerializableResource)
  Object.send(:remove_const, :SerializableResourceSerializer)
end
```

then

```ruby
def test_json_key_with_resource_with_name_and_serializer
  with_serializable_named_collection('resources') do |model|
     serializer_instance = collection_serializer.new([model])
     assert_equal 'resources', serializer_instance.json_key
  end
end
```

bf4: (would also be good to note for the reviewer how the test would fail before the code changes)

bf4: Not sure how important to your use case, but note that in the test I changed 

```diff
- collection_serializer.new(model)
+ collection_serializer.new([model])
```

since the collection serializer should take in an array of models, or at least something enumerable.  Was that a bug in the test or in your implementation?

NullVoxPopuli:
needs rebase.

@bf4 can you summarize all your comments so @mchitten can finish this? :-)

dyoganand:
Will this be merged??

bf4:
@dyoganand This needed some revisions. Would you be interested in taking it on?

bf4:
same issue: https://github.com/rails-api/active_model_serializers/issues/2087#issuecomment-290205455

IzumiSy:
When is this PR going to be merged? I am being bothered the bug which this PR potentially fixes up exactly at the moment.

samuelpismel:
2 years... Two years! And this isn't merged...

NullVoxPopuli:
@samuelpismel at this point, if you want newish support / a library that easier to maintain, I'd use http://jsonapi-rb.org/
if you don't have a jsonapi style api, jbuilder is probably good enough.

This project has gone through so many rewrites...
The master branch has also been cleared out "in prep" for the next rewrite -- so this PR would need to target the 0.10.x branch.

:-\

Apologies for the circumstances. :(

samuelpismel:
Thanks man, you are correct, sorry about that.

Issue #1587, opened 2016-03-13
Add warning in documentation to benchmark testing before use


bf4:
Per https://github.com/rails-api/active_model_serializers/issues/1586 caching may actually slow down your rendering.

Until it's fixed, it would be good to add a warning to the documentation.

PR #1872, opened 2016-08-04
Add warning in caching documentation

aldreth:
References https://github.com/rails-api/active_model_serializers/issues/1587

### DIFF
--- a/docs/general/caching.md
+++ b/docs/general/caching.md
@@ -2,6 +2,12 @@
 
 # Caching
 
+## Warning
+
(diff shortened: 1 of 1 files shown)


bf4:
Awesome, thanks!

bf4:
(The caching issues are actually not so bad at present. I've done a lot of work on them and they should generally be positive at this point, but not as good as they should be)

aldreth:
Yeah, they hit an app I'm working on pretty hard, so I've been following progress with interest.

bf4:
@aldreth Thanks.  I've been mostly thinking about simplifying AMS internals for the push to 1.0 and that will likely include having caching as a separate layer, so that it will be easier to manage.  In the meantime, I haven't been working on that area. If you're interested I can give you a bit of a tour.

PR #1878, opened 2016-08-12
Use ActiveSupport::Cache.expand_cache_key for cache key expansions

markiz:
#### Purpose

Allows using non-primitive objects as object cache key. 
For example, we might want to return an array of objects that define `#cache_key` as our serialized object's `#cache_key`

#### Changes

`join("/")` changed to `ActiveSupport::Cache.expand_cache_key`

#### Caveats

Dependency on ActiveSupport (which we already have)

The test is a bit contrived, sorry about that.
`#expand_cache_key` can be redefined to change the actual implementation.

Previous version of AMS did use `ActiveSupport::Cache.expand_cache_key`

### DIFF
--- a/test/cache_test.rb
+++ b/test/cache_test.rb
@@ -4,6 +4,26 @@ require 'tempfile'
 
 module ActiveModelSerializers
   class CacheTest < ActiveSupport::TestCase
+    class AuthorWithCustomCacheKey < Author
+      class DerivedCacheKey
+        def initialize(object, method)
+          @object = object
+          @method = method
+        end
+
+        def cache_key
+          @object.__send__(@method)
+        end
+      end
+
+      def cache_key
+        [
+          DerivedCacheKey.new(self, :name),
+          DerivedCacheKey.new(self, :id)
+        ]
+      end
+    end
+
     class UncachedAuthor < Author
       # To confirm cache_key is set using updated_at and cache_key option passed to cache
       undef_method :cache_key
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
what are the benefits of this? / what's the use case?

markiz:
@NullVoxPopuli 

We use non-primitive object as cache keys, another contrived example:

```
class User
  def cache_key
    [super, latest_post, latest_newsfeed_item, latest_friend_request] # where latest_post, latest_newsfeed_item, latest_friend_request are ActiveRecord models, possibly nil
  end
end
```

There wouldn't be any problem if AMS didn't try to `join` those.
In 0.8 it did the right thing with `expand_cache_key`
We can probably live with overridable `expand_cache_key` too (and override it to do what we want). But using `join` here breaks compat for us.

bf4:
@markiz lgtm, 

1. would you mind adding a change log?
1. would you mind looking at the object_cache_key methods in the Caching mixin?

markiz:
@bf4: 

1. Sure (4ce7dc0)
2. I'm not sure I follow. `object_cache_key` duplicates some of the functionality in the `expand_cache_key`, but it also has some differing behavior, so I don't see any immediate need to change it.

NullVoxPopuli:
@markiz thanks for that. I didn't know this was an old feature, and I had no idea people even did this. Certainly looks looks better than building your own cache key string. :-)

bf4:
@NullVoxPopuli yeah.. I'm not sure why it was removed.  We use ActiveSupport::Cache which uses expand_cache_key but ¯\_(ツ)_/¯

test/cache_test.rb
```diff
@@ -4,6 +4,26 @@
...
+        def cache_key
+          @object.__send__(@method)
+        end
+      end
+
+      def cache_key
+        [
+          DerivedCacheKey.new(self, :name),
+          DerivedCacheKey.new(self, :id)
+        ]
```
bf4: @markiz my only real concern is that this class is really weird-looking. Would you mind adding some comments why you designed it this way? Be nice to future devs :)

test/cache_test.rb
```diff
@@ -4,6 +4,26 @@
...
+      class DerivedCacheKey
+        def initialize(object, method)
+          @object = object
+          @method = method
+        end
+
+        def cache_key
+          @object.__send__(@method)
+        end
+      end
```
bf4: Maybe

```ruby
class AuthorWithCustomCacheKey < Author
  HasCacheKey = Struct.new(:cache_key)
  
  # Rather than `[name, id]` as the cache key, demonstrate that `:cache_key` is called on
  # any element that responds to it.
  def cache_key
    [HasCacheKey.new(name), HasCacheKey.new(id)]
  end
end
```

test/cache_test.rb
bf4: see https://github.com/rails-api/active_model_serializers/pull/1878/files#r74627820

test/cache_test.rb
```diff
@@ -4,6 +4,26 @@
 
 module ActiveModelSerializers
   class CacheTest < ActiveSupport::TestCase
+    class AuthorWithCustomCacheKey < Author
```
bf4: maybe `AuthorWithCacheKeyObjects` ?

bf4: or `AuthorWithExpandableCacheKeyElements`?

markiz: I like that last one, I'll change pr to use that

markiz: 7b6302a

test/cache_test.rb
markiz: d0e54da

markiz: d0e54da

Issue #1854, opened 2016-07-23
Serializing a Plain-Old Ruby Object (PORO) via Inherited ActiveModelSerializers::Model fails


browrp:
### Expected
In Rails, when using a Plain Old Ruby Object (PORO) with Active Model Serializers and inheriting from the **_ActiveModelSerializers::Model_** class as defined in the guides on page https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/serialize_poro.md, I expected the PORO to serialize to JSON.  

```
class Project < ActiveModelSerializers::Model

    def initialize
       project_sponsors = Array.new
    end

  attr_accessor :uuid_key, :project_name
  attr_accessor :project_sponsors

end
```

```
class ProjectSponsor < ActiveModelSerializers::Model
  attr_accessor :uuid_key, :first_name, :last_name
end
```

```
    render :status => 200,
           :json => { :success => true,
                      :info => "Project Sponsor Overview",
                      :data => { :project_list => @x}}

```

I expect to see:

```
{
	"success": true,
	"info": "Project Sponsor Overview",
	"data": {
		"project_list": [
			{
				"uuid_key": "afde-90-zxy",
				"project_name": "The X Files",
				"project_sponsors": [
					{
						"uuid_key": "a1a-x24",
						"first_name": "Mary",
						"last_name": "Smith"
					},
					{
						"uuid_key": "b1b-z39",
						"first_name": "Tom",
						"last_name": "Jones"
					}
				]
			},
			{
				"uuid_key": "bkij-11-bb",
				"project_name": "Sleepy Hollow",
				"project_sponsors": [
					{
						"uuid_key": "hhi1-h23",
						"first_name": "Henry",
						"last_name": "McNeal"
					},
					{
						"uuid_key": "iixf-i09",
						"first_name": "Icabob",
						"last_name": "Crane"
					}
				]
			}
		]
	}
}

```

### Actual
What I get is the following error:
```
NoMethodError in ProjectOverviewController#index
undefined method `keys' for nil:NilClass

```
I found that the Project's class' initializer is causing this error.  When I remove it the error stops but the JSON is returned is empty; none of the attributes are serialized:

```
{
	"success": true,
	"info": "Project Sponsor Overview",
	"data": {
		"project_list": [
			{
			},
			{
			}
		]
	}
}
```

However, when I follow alternate approach as described in the guide...
```
# my_model.rb
class MyModel
  alias :read_attribute_for_serialization :send
[...]

PR #1880, opened 2016-08-15
Add additional doc for using AMS with POROs

tommi-lew:
#### Purpose

For additional clarity.
#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/issues/1854

### DIFF
--- a/docs/howto/serialize_poro.md
+++ b/docs/howto/serialize_poro.md
@@ -29,4 +29,10 @@ class MyModel < ActiveModelSerializers::Model
 end
 ```
 
-The default serializer would be `MyModelSerializer`.
+The default serializer would be `MyModelSerializer`:
+```ruby
+# my_model_serializer.rb
+class MyModelSerializer < ActiveModel::Serializer
+  attributes :id, :name, :level
+end
+```


docs/howto/serialize_poro.md
```diff
@@ -29,4 +29,10 @@ class MyModel < ActiveModelSerializers::Model
...
 ```
 
-The default serializer would be `MyModelSerializer`.
\ No newline at end of file
+The default serializer would be `MyModelSerializer`:
+```ruby
+# my_model_serializer.rb
+class MyModelSerializer < ActiveModel::Serializer
+  attributes :id, :name, :level
+end
```
bf4: Would you mind commenting here on how this change improves the documentation? There's definitely room for improvement, in particular, I don't think the word `default` is helpful.  It's more like the automatic lookup would resolve to a serializer with that name.  Besides that, how do you see adding an example serializer helpful?  I'd rather have a link to the serializers docs for more info.

DrSayre: @tommi-lew @bf4 I'm fine with any changes to this. I put that sentence there to point out the serializer is the same name as one created with ActiveRecord.

bf4: Would you have thought otherwise?

NullVoxPopuli:
@DrSayre do you think you can finish this up?

richmolj:
I'm kinda down on this and the current documentation. We're encouraging people to use `ActiveModelSerializers::Model`, which most serializers do not need. Most serializers only require: 

```ruby
alias :read_attribute_for_serialization :send
```

And if we made an internal change they wouldn't even need that.

The exceptions are `ErrorSerializer` and caching. I suggest the documentation change be this - otherwise people (like me) are needlessly adding stuff to their POROs that is not needed.

bf4:
Closing as stale

PR #1881, opened 2016-08-15
ActiveModelSerializers::Model successor initialized with string keys fix

yevhene:
#### Purpose
Fixing strange behavior of ActiveModelSerializers::Model successor when initialized with string keys hash

#### Changes
ActiveModelSerializers::Model is stringify keys on initialization

Example:
```ruby
class User < ActiveModelSerializers::Model
  attr_accessor :name
end

  attribute :name
end

resource = User.new('key' => value)
serializer = UserSerializer.new(resource)
serializer.to_json #=> '{"name": null}'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Features:
 Fixes:
 
 - [#1833](https://github.com/rails-api/active_model_serializers/pull/1833) Remove relationship links if they are null (@groyoh)
+- [#1881](https://github.com/rails-api/active_model_serializers/pull/1881) ActiveModelSerializers::Model correctly works with string keys (@yevhene)
 
 Misc:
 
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
this looks good. Simple, to the point. I actually didn't know why someone would want to do this, but I checked AR, and it supports this. So.. neat. 

Could you add a changelog entry?

bf4:
@yevhene I'm pretty sure that behavior comes from ActiveModel::Model.  Also, `serializer.to_json` is not the recommended way to serialize a model.

yevhene:
@bf4 Yes, this behavior is comes with ActiveModel::Model, but ActiveModelSerializers::Model overrides it for some reason. Problem not in `to_json`.

yevhene:
@NullVoxPopuli I was faced with this when I take raw data from database and wrap it with ActiveModelSerializers::Model descendant.

yevhene:
@NullVoxPopuli I updated the changelog.

NullVoxPopuli:
cool!, thanks!

PR #1883, opened 2016-08-16
(WIP) Fix Serializer Lookup Chain for Namespaces

thewatts:
#### Purpose

This fixes dynamic lookup of namespaced serializers through relationships.

#### Changes

This adjusts the internal API of the serializer method used for dynamic lookup by adding an additional constant string to the lookup chain used to find the related serializer. 

#### Caveats

No tests. I'm a testing guy - so, if someone can point me to where I should be testing this in the project, I will 100% write them.

I'm guessing I'll need to start with `test/serializers/associations_test.rb`

Also - this is currently only working for a single nested resource, I'll adjust in short order to provide a dynamic amount of namespaced items.

#### Related GitHub issues
Not sure

-- The Problem:

Currently, when using a relationship from within a namespace, ex:

```ruby
module Api
  class LessonSerializer < ApplicationSerializer
    belongs_to :trait
  end
end
```

The method to fetch a serializer builds a lookup chain of possible
constant names that gets looped over with `safe_constantize`. When using
relationships from within a namespace (like above), the lookup chain
would return this:

```ruby
[
  "Api::LessonSerializer::TraitSerializer",
  "::TraitSerializer",
]
```

This works fine the first time, as calling `.safe_constantize` on
`"Api::LessonSerializer::TraitSerializer"` returns the correct constant
in this case: `Api::TraitSerializer`.

However, the _second_ time that the lookup has to happen (from a
different relationship), ex:

```ruby
module Api
  class CourseSerializer < ApplicationSerializer
    belongs_to :trait
  end
end
```

The lookup chain returns a similar response:

```ruby
[
  "Api::CourseSerializer::TraitSerializer",
  "::TraitSerializer",
]
```

But at this point, calling `.safe_constantize` on
`"Api::CourseSerializer::TraitSerializer"` returns `nil`.

Here's an example from the Rails console:
safe_constantize example

It looks like the main problem [lies in this issue here](https://github.com/rails/rails/issues/9286)
(many thanks to @willcosgrove for finding it!)

So this means that calling `.safe_constantize` on the same name string
returns the constant the first time (based on
`ActiveSupport::Dependencies`), but the _second_ time it gets called on
the same string it returns `nil`.

safe_constantize of the same string

-- The Fix:

I get the namespace of the `name` of the parent serializer, and use that
to also build and add a string to the lookup chain array.

So in our first example above, it adds: `"Api::TraitSerializer"` to the
lookup chain, allowing the serializer to be found every time.


NullVoxPopuli:
Hey @thewatts thanks for your PR!

There have actually been many approaches to this before:

My favorite: https://github.com/rails-api/active_model_serializers/pull/1757 (bias though)
https://github.com/rails-api/active_model_serializers/pull/1436
https://github.com/rails-api/active_model_serializers/pull/1338
https://github.com/rails-api/active_model_serializers/pull/1295
https://github.com/rails-api/active_model_serializers/pull/1196
https://github.com/rails-api/active_model_serializers/pull/1157

I think that rails issue you linked to comes down to how ruby does constant lookup in an auto-loading environment (rails) http://cirw.in/blog/constant-lookup.html

That said, for tests, I'd would add to the main AMS::Serializer tests and add tests for the look up chain method

NullVoxPopuli:
I suppose you'd also wasn't to add integration tests to action controller tests somewhere. Maybe a new file for lookup?

thewatts:
@NullVoxPopuli thanks for the reply!

And sorry for the late reply on my end - that's quite the chain of issues/PRs to follow.
(I'll admit - I skimmed to get the gist).

Honestly, dynamically defining namespace by controller (or through configuration as in your PR) hadn't crossed my mind.

I had just noticed that the dynamic lookup of some `belongs_to` relationships were failing when two different serializers shared the same relationship resource.

I've looked at making things more dynamic, so that deeper nested items could be found correctly (ie: `Api::V1::SuperDuper::TraitSerializer`). I've added the commit to this PR.

I can see the desire for configuration though, especially for relationships where you have a `V2` serializer that needs to pull a serializer relationship from a `V1` type.

Ex:

```ruby
module Api
  module V1
    class TraitSerializer < ApplicationSerializer
    end
  end
end

module Api
  module V2
    class LessonSerializer < ApplicationSerializer
      belongs_to :trait
    end
  end
end
```

In that case - `Api::V1::TraitSerializer` wouldn't be found.

---

Though my additions create a quick fix (at least to my own problem) - it looks like there's already some solid direction going with those other PRs.

Has there been a consensus on the final direction? If so - I won't be offended if this PR is closed ;)

thewatts:
@NullVoxPopuli if this PR is helpful, I absolutely don't mind adding tests (and of course - fixing the one I just broke 😅 )

lib/active_model/serializer.rb
```diff
@@ -62,13 +62,27 @@ class << self
...
+      resource_class_name   = klass.name.demodulize
+      resource_namespace    = klass.name.deconstantize
       serializer_class_name = "#{resource_class_name}Serializer"
 
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
+      tmp_namespace   = name.to_s
+      name_namespaces = [tmp_namespace]
+
+      until tmp_namespace.empty?
```
NullVoxPopuli: do you need to have two loops here? I imagine you could add to the chain array as you deconstantized the namespaces.

Could you also add some documentation (comments) in here explaining how the chain array looks after different steps of this process?

NullVoxPopuli:
@thewatts I think this is def worth while (it's why I've worked on it a bit).
I can't remember what the hesitation is in changing this is though. Maybe @bf4 can speak more to that.

lib/active_model/serializer.rb
thewatts: @NullVoxPopuli Good call! Thanks for the feedback!

Adjustments added in most recent commit. Let me know your thoughts!

thewatts:
@NullVoxPopuli for the lookup chain and versioning... would we try to specifically look for versioned namespaces?

ex: if `V2` exists, assume `V1` does as well? That could be tricky, since technically relationships could work both ways. `V1` of a Serializer starts using a `V2` of another.

lib/active_model/serializer.rb
```diff
@@ -60,16 +60,41 @@ class << self
...
+      # [
+      #   "Api::V1::Admin::PostSerializer",
+      #   "Api::V1::PostSerializer",
+      #   "Api::PostSerializer",
+      # ]
+      parent_namespace = name.to_s
 
-      chain
+      if self != ActiveModel::Serializer
+        until parent_namespace.empty?
```
NullVoxPopuli: This is getting there! Why not use parent namespace.each?

lib/active_model/serializer.rb
```diff
@@ -60,16 +60,41 @@ class << self
...
+          lookup_chain.push("#{parent_namespace}::#{serializer_class_name}")
+          parent_namespace = parent_namespace.deconstantize
+        end
+      end
+
+      # And finally - this adds the last namespaced candidate based on the
+      # provided klass, returning the full chain.
+      #
+      # Examples:
+      # - Post -> "::PostSerializer"
```
NullVoxPopuli: Are these backwards? the top level one should be at the bottom, yeah?

NullVoxPopuli:
Could you add a change log entry as well?
Thanks!

lib/active_model/serializer.rb
NullVoxPopuli: Actually, I think using .each would turn this in to too much work. Haha

lib/active_model/serializer.rb
```diff
@@ -60,16 +60,41 @@ class << self
...
-      resource_namespace = klass.name.deconstantize
+      resource_class_name   = klass.name.demodulize
+      resource_namespace    = klass.name.deconstantize
       serializer_class_name = "#{resource_class_name}Serializer"
 
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
+      # Loops over a namespaced Serializer's name to add multiple
+      # namespaced serializer candidates for lookup in the `lookup_chain` from
+      # the "parent" serializer.
```
NullVoxPopuli: This is the kind of documentation I'd like to see all over AMS :-)

lib/active_model/serializer.rb
thewatts: We could def use `.each` if we split the string on `::` - but we'd have to build the namespaces a little differently.

I'm game for either.

NullVoxPopuli: I think the way you have it may be the cleanest. 
Just trying to think of a way it could possibly throw an exception.. (I haven't thought of one so far)

lib/active_model/serializer.rb
thewatts: We want th ordered by longest namespace first, right? 

Otherwise if a non-namespaced item exists, ex: `CommentSerializer`, it would be found _before_ the namespaced version, `Api::CommentSerializer` (which definitely isn't what we want if the parent is itself namespaced)

lib/active_model/serializer.rb
thewatts: Yeah it's a tough balance - solid documentation can be a bonus - but it's the first thing to become stale unless tightly monitored.

Typically I try to have the code be self documenting (so to speak), but at this point we're deep into the belly of the beast :)

NullVoxPopuli: Yeah, everyone's goal is self-documenting code. But as long as churn is low, documentation is great to have. I don't see serializer lookup changing very often :-)

lib/active_model/serializer.rb
NullVoxPopuli: If we represent these in array form (with more than two, so we can be sure we're talking about the same direction)
Do you mean:
```
Api::V2::PostSerializer
Api::PostSerializer
PostSerializer
```

?

Or are you saying this more has to do with the namespace on the model? (I actually think namespacing in controllers vs models is why this gets punted :( )

So, for an  Namespaced::Post object:

```
Api::V2::Namespaced::PostSerializer
Api::Namespaced::PostSerializer
Namespaced::PostSerializer
PostSerializer? # should this one even be here?
```

thewatts: I was thinking the top example - but for me, this specifically broke when referencing a relationship from inside a namespaced serializer.

Ex: `belongs_to :trait`.

In that case, the `klass` being sent through the chain building method is never namespaced.

Is there another instance top of mind where this method gets used? Is it when being used from inside a controller?

NullVoxPopuli: currently, controller namespace is a big one.

Like api/v2/namespace/posts_controller.

can you add tests?

thewatts: Yeah no prob on those tests - but as a heads up - it'll be a little while before I can submit them.

NullVoxPopuli: That's fine. The pace of AMS isn't anything speedy, and as far as i know, this feature isn't pressing.

bf4:
Just wanna let you all know I'm aware of this but haven't yet looked at it :)

lib/active_model/serializer.rb
```diff
@@ -60,16 +60,41 @@ class << self
...
 
-      resource_class_name = klass.name.demodulize
-      resource_namespace = klass.name.deconstantize
+      resource_class_name   = klass.name.demodulize
+      resource_namespace    = klass.name.deconstantize
       serializer_class_name = "#{resource_class_name}Serializer"
 
-      chain.push("#{name}::#{serializer_class_name}") if self != ActiveModel::Serializer
-      chain.push("#{resource_namespace}::#{serializer_class_name}")
+      # Loops over a namespaced Serializer's name to add multiple
```
bf4: This kind of comment should be above the method definition, no?

NullVoxPopuli:
is this still an issue on master?

thewatts:
@NullVoxPopuli - I see that you just pushed an update here, #1968. 

This specific PR was for trying to auto-build the namespace lookup chain used during declarations in the `Serializer` classes.

ex:

```ruby
module Api
  module V1
    class AuthorSerializer < ApplicationSerializer
      has_many :posts # this would fail to fine the correct serializer
```

My apologies for not staying on top of it - shortly after the PR was submitted - I found out I was losing my job @ my employer (cutbacks). I've since been in the transition to other employment.

It looks like your PR adds the ability to specify the namespace directly in a `before_action` callback which may render the needs for this PR obsolete.

If you still want it in master - I can rebase off of master and make the necessary adjustments, add tests, and update the Changelog (will take me a week or so to get the available time).

If not - no worries 😄

NullVoxPopuli:
sorry for your loss -- company cut backs are never fun -- even for those that stay. 

Anywho, for this PR, I'd like to see some tests so that we know *exactly* what situation your PR addresses.

but being able to specify a namespace hopefully will help a ton of people.

PR #1887, opened 2016-08-19
Correct comment

johnnymo87:
#### Purpose
Make the comment reflect what the function does

#### Changes
Correct comment

### DIFF
--- a/lib/active_model_serializers/serializable_resource.rb
+++ b/lib/active_model_serializers/serializable_resource.rb
@@ -49,7 +49,7 @@ module ActiveModelSerializers
 
     # Get serializer either explicitly :serializer or implicitly from resource
     # Remove :serializer key from serializer_opts
-    # Replace :serializer key with :each_serializer if present
+    # Remove :each_serializer if present and set as :serializer key
     def serializer
       @serializer ||=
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Nice find! Would you like to add a changelog to take credit for the correction?

lib/active_model_serializers/serializable_resource.rb
```diff
@@ -49,7 +49,7 @@ def serializer_instance
 
     # Get serializer either explicitly :serializer or implicitly from resource
     # Remove :serializer key from serializer_opts
-    # Replace :serializer key with :each_serializer if present
+    # Replace :each_serializer key with :serializer if present
```
bf4: Hmm, I don't think it's exactly clear either way, now that I look at it.  Should be

> Remove `:each_serializer` if present and set as `:serializer` key

bf4: (or something like that)

johnnymo87:
@NullVoxPopuli @bf4 updated

bf4:
@johnnymo87 Merged. Sorry I missed this was ready.

PR #1888, opened 2016-08-22
Updated ember integration guide to use the proper renderer

Patsy-issa:
#### Purpose

Update the guide to use the proper renderer
#### Changes

Uses the `jsonapi` rendered that is registered in [register_jsonapi_renderer](https://github.com/rails-api/active_model_serializers/blob/49ee823a5311799b3e96b0384ca4fd7a9cf39a34/lib/active_model_serializers/register_jsonapi_renderer.rb)

### DIFF
--- a/docs/integrations/ember-and-json-api.md
+++ b/docs/integrations/ember-and-json-api.md
@@ -117,7 +117,7 @@ will generate the path `/posts/{postId}?include='comments'`
 
 So then in your controller, you'll want to be sure to have something like:
 ```ruby
-render json: @post, include: params[:include]
+render jsonapi: @post, include: params[:include]
 ```
 
 If you want to use `include` on a collection, you'd write something like this:
--- a/docs/integrations/ember-and-json-api.md
+++ b/docs/integrations/ember-and-json-api.md
@@ -42,7 +42,9 @@ Lastly, in order to properly handle JSON API responses, we need to register a JS
 
 ```ruby
 # config/initializers/active_model_serializers.rb
-require 'active_model_serializers/register_jsonapi_renderer'
+ActiveSupport.on_load(:action_controller) do
+  require 'active_model_serializers/register_jsonapi_renderer'
+end
 ```
 
 ### Adapter Changes


bf4:
@Patsy-issa Awesome!  Would you be interested in adding a commit to put the `require` statement in a lazy loading block (i.e. run when action controller is loaded)

I recommend using lazy loading both for performance and load-order issues ;)

```ruby
  require 'active_model_serializers/register_jsonapi_renderer'
end
```

Patsy-issa:
Updated to use lazy loading.

docs/integrations/ember-and-json-api.md
```diff
@@ -117,7 +119,7 @@ will generate the path `/posts/{postId}?include='comments'`
 
 So then in your controller, you'll want to be sure to have something like:
 ```ruby
-render json: @post, include: params[:include]
+render jsonapi: @post, include: params[:include]
```
bf4: @Patsy-issa My only other comment, and maybe we should discuss this in a PR, is that it's generally not a good idea to blindly pass in query params like `params[:include]`.  You should instead use a whitelist of valid params, like `params.require(:include).permit(:author, :comments)`

bf4: Since include is an array, I don' think `permit` is the correct method call.

bf4:
Would you like to add yourself to changelog under Unreleased, Misc, or just for me to merge this? @Patsy-issa

Patsy-issa:
I m thinking let's not merge this in and instead I ll open a pr addressing the previous changes + the permitted params as well as check on the ember side to see if the adapter and serializer code is still valid.

Patsy-issa:
The js is a bit inconsistent, for example the adapter uses destructuring to assign `underscore` and `pluralize` yet the serializer uses var assignment `var underscore = Ember.String.underscore;`

bf4:
the AMS adapter or Ember-data adapter?

Patsy-issa:
The ember data one

docs/integrations/ember-and-json-api.md
NullVoxPopuli: for whitelisting include, you'd have to do:

```ruby
render jsonapi: @post, include: include_string
# ....

def include_string
  [
    'relationship1',
    'relationship2'
  ].select { |r| params[:include].include?(r) }
end
```

PR #1889, opened 2016-08-24
Support key transformation for Attributes adapter

iancanderson:
#### Purpose
The `:attributes` adapter is the default one, but it did not support
key transformation. This was very surprising behavior, since the
"Configuration Options" page in the guides didn't mention that this
behavior was not supported by the attributes adapter.

#### Changes
This PR adds key transform support to the attributes adapter, and
adds documentation about the default transform for the attributes
adapter (which is `:unaltered`).

#### Caveats
This could potentially break existing apps that have a `key_transform` set that also use the `Attributes` adapter. Before this change, the `key_transform` setting would be a no-op. After this change, the `key_transform` would take effect and therefore change the behavior of their current apps.

However, the current behavior seems like a bug to me, and I think it is worth fixing.

### DIFF
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -4,7 +4,9 @@ module ActiveModelSerializers
       def serializable_hash(options = nil)
         options = serialization_options(options)
         options[:fields] ||= instance_options[:fields]
-        serializer.serializable_hash(instance_options, options, self)
+        serialized_hash = serializer.serializable_hash(instance_options, options, self)
+
+        self.class.transform_key_casing!(serialized_hash, instance_options)
       end
     end
   end
--- a/lib/active_model_serializers/key_transform.rb
+++ b/lib/active_model_serializers/key_transform.rb
@@ -11,6 +11,7 @@ module ActiveModelSerializers
     # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
     def camel(value)
       case value
+      when Array then value.map { |item| camel(item) }
       when Hash then value.deep_transform_keys! { |key| camel(key) }
       when Symbol then camel(value.to_s).to_sym
       when String then value.underscore.camelize
@@ -25,6 +26,7 @@ module ActiveModelSerializers
     # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
     def camel_lower(value)
       case value
+      when Array then value.map { |item| camel_lower(item) }
       when Hash then value.deep_transform_keys! { |key| camel_lower(key) }
       when Symbol then camel_lower(value.to_s).to_sym
       when String then value.underscore.camelize(:lower)
@@ -40,6 +42,7 @@ module ActiveModelSerializers
     # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L185-L187 ActiveSupport::Inflector.dasherize}
     def dash(value)
       case value
+      when Array then value.map { |item| dash(item) }
       when Hash then value.deep_transform_keys! { |key| dash(key) }
       when Symbol then dash(value.to_s).to_sym
       when String then value.underscore.dasherize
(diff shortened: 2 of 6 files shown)


test/fixtures/poro.rb
```diff
@@ -46,6 +46,13 @@ def arguments_passed_in?
   end
 end
 
+class Person < Model
```
NullVoxPopuli: can you move these to your test file?

We really don't like the poros file, as it makes deugging tests a little more convoluted than it should be

NullVoxPopuli:
Looks good, can you add a changelog entry?

test/fixtures/poro.rb
iancanderson: Agreed! I'll move them

iancanderson:
@NullVoxPopuli thanks for the quick reply! I added a changelog entry, moved the model definition into the test, and squashed 👍

iancanderson:
@NullVoxPopuli I'm having trouble actually using this in practice - hold off on merging for now

bf4:
@iancanderson This is a feature you want?

bf4:
The 'Attributes' adapter is probably going to go away. I've already moved most of it back into the Serializer itself.  Then the JSON adapter will simply support a `:root` option just like [ActiveModel::Serializers::JSON](https://github.com/rails/rails/blob/5-0-stable/activemodel/lib/active_model/serializers/json.rb) in Rails does (and probably therefore also support `ActiveRecord::Base.include_root_in_json = true`, etc.)

iancanderson:
@bf4 What we need is an adapter that can serialize without a JSON root that can also transform keys.

See the most recent commit that handles the issue we had with the first approach. I'll comment in-line as well to point out where the problem was.

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -4,7 +4,9 @@ class Attributes < Base
       def serializable_hash(options = nil)
         options = serialization_options(options)
         options[:fields] ||= instance_options[:fields]
-        serializer.serializable_hash(instance_options, options, self)
+        serialized_hash = serializer.serializable_hash(instance_options, options, self)
```
iancanderson: The problem was that since there is no root key for the Attributes adapter, this `serialized_hash` could actually be an array, which was not handled by the `KeyTransform` functions.

iancanderson: (This happens when a `CollectionSerializer` is used, which maps the resources into an array)

iancanderson:
@bf4 I understand that the attributes adapter is "going away", but it's currently the default serializer. Until it's gone, (and until the json adapter gets an option to not include the root), I think this patch is worthwhile.

The changes made in `KeyTransform` to support an array would also be necessary for the json adapter to support payloads without a root key, so I think this PR is actually a stepping stone to what you eventually want. What do you think?

lib/active_model_serializers/key_transform.rb
```diff
@@ -11,6 +11,7 @@ module KeyTransform
     # @see {https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/methods.rb#L66-L76 ActiveSupport::Inflector.camelize}
     def camel(value)
       case value
+      when Array then value.map { |item| camel(item) }
```
NullVoxPopuli: This really should have been here the whole time anyway. Welcome change!

@bf4 do you know if there is much of performance improvement that could be made be re-ordering these? or would it not be worth it?

test/adapter/attributes_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+module ActiveModelSerializers
+  module Adapter
+    class AttributesTest < ActiveSupport::TestCase
+      class Person
```
NullVoxPopuli: yeah, lets hope other contributors pick up on this pattern.

iancanderson:
@NullVoxPopuli @bf4 anything I can do to move this forward?

lib/active_model_serializers/key_transform.rb
bf4: This is a separate feature that should be noted in the changelog, I think. Otherwise, LGTM

NullVoxPopuli:
@iancanderson can you move your changelog entry to be under the features heading?

iancanderson:
@NullVoxPopuli changelog moved, and squashed the commits in https://github.com/rails-api/active_model_serializers/pull/1889/commits/20e91fa21f9778f2a50a7f2f4c26e010bf820128

NullVoxPopuli:
woo!

PR #1894, opened 2016-08-29
Update ember-and-json-api.md

Ikariusrb:
#### Purpose
Update the ember integration documentation

#### Changes
Update documentation for how to get ember to use include directives
Add documentation on how to update controllers to tell Rails we are consuming and generating jsonapi data.

Removed ember-data adapter change to support include directives, as it's now built-in.  Updated the documentation for how to add include directives to ember store queries, and added documentation covering how to tell Rails that we are consuming and generating jsonapi data

### DIFF
--- a/docs/integrations/ember-and-json-api.md
+++ b/docs/integrations/ember-and-json-api.md
@@ -38,12 +38,40 @@ You will also want to set the `key_transform` to `:unaltered` since you will adj
 ActiveModelSerializers.config.key_transform = :unaltered
 ```
 
-Lastly, in order to properly handle JSON API responses, we need to register a JSON API renderer, like so:
+In order to properly handle JSON API responses, we need to register a JSON API renderer, like so:
 
 ```ruby
 # config/initializers/active_model_serializers.rb
 require 'active_model_serializers/register_jsonapi_renderer'
 ```
+Rails also requires your controller to tell it that you accept and generate JSONAPI data.  To do that, you use `respond_to` in your controller handlers.  You can add `ActionController::MimeResponds` to your application controller to enable this:
+
+```ruby
+class ApplicationController < ActionController::API
+  include ActionController::MimeResponds
+end
+```
+Then, in your controller you can tell rails you're accepting and rendering the jsonapi format:
+```ruby
+ # POST /post
+  def create
+    @post = Post.new(post_params)
+    respond_to do |format|
+      if @post.save
+        format.jsonapi { render jsonapi: @post, status: :created, location: @post }
+      else
+        format.jsonapi { render jsonapi: @post.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+  
+    # Only allow a trusted parameter "white list" through.
+    def post_params
+      ActiveModelSerializers::Deserialization.jsonapi_parse!(params).permit(:title, :body)
+    end
+end
+```
+
 
 ### Adapter Changes
 
@@ -67,18 +95,6 @@ export default  DS.JSONAPIAdapter.extend({
     return pluralize(underscored);
   },
 
-  // allows queries to be sent along with a findRecord
-  // hopefully Ember / EmberData will soon have this built in
-  // ember-data issue tracked here:
-  // https://github.com/emberjs/data/issues/3596
-  urlForFindRecord(id, modelName, snapshot) {
-    let url = this._super(...arguments);
-    let query = Ember.get(snapshot, 'adapterOptions.query');
-    if(query) {
-      url += '?' + Ember.$.param(query);
-    }
-    return url;
-  }
 });
 ```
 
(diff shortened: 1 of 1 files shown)


docs/integrations/ember-and-json-api.md
```diff
@@ -38,12 +38,40 @@ You will also want to set the `key_transform` to `:unaltered` since you will adj
...
 ```
 
-Lastly, in order to properly handle JSON API responses, we need to register a JSON API renderer, like so:
+In order to properly handle JSON API responses, we need to register a JSON API renderer, like so:
 
 ```ruby
 # config/initializers/active_model_serializers.rb
 require 'active_model_serializers/register_jsonapi_renderer'
 ```
+Rails also requires your controller to tell it that you accept and generate JSONAPI data.  To do that, you use `respond_to` in your controller handlers.  You can add `ActionController::MimeResponds` to your application controller to enable this:
```
NullVoxPopuli: why is this? was this mentioned in the slack convo you had with @bf4? might be a good idea to state *why* you *need* respond_to.

docs/integrations/ember-and-json-api.md
```diff
@@ -38,12 +38,40 @@ You will also want to set the `key_transform` to `:unaltered` since you will adj
...
+        format.jsonapi { render jsonapi: @post, status: :created, location: @post }
+      else
+        format.jsonapi { render jsonapi: @post.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+  
+    # Only allow a trusted parameter "white list" through.
+    def post_params
+      ActiveModelSerializers::Deserialization.jsonapi_parse!(params, only: [:title, :body] )
```
NullVoxPopuli: is the exclamation point required? I'm using this without the exclamation point. Exclamation point implies mutation.

docs/integrations/ember-and-json-api.md
```diff
@@ -67,18 +95,6 @@ export default  DS.JSONAPIAdapter.extend({
     return pluralize(underscored);
   },
 
-  // allows queries to be sent along with a findRecord
```
NullVoxPopuli: no longer needed since Ember 2.5!!!! (I think it was 2.5)

docs/integrations/ember-and-json-api.md
```diff
@@ -102,18 +118,16 @@ export default DS.JSONAPISerializer.extend({
 
 ```
 
+
 ## Including Nested Resources
 
-Previously, `store.find` and `store.findRecord` did not allow specification of any query params.
-The ActiveModelSerializers default for the `include` parameter is to be `nil` meaning that if any associations are defined in your serializer, only the `id` and `type` will be in the `relationships` structure of the JSON API response.
+Ember-data can request related records by using `include`.  To tell it to make that request, use the form below when requesting records.
```
NullVoxPopuli: Ember Data is referred to as a "Ember Data" rather than "Ember-data", see: https://github.com/emberjs/data

docs/integrations/ember-and-json-api.md
```diff
@@ -102,18 +118,16 @@ export default DS.JSONAPISerializer.extend({
...
 For more on `include` usage, see: [The JSON API include examples](./../general/adapters.md#JSON-API)
 
-With the above modifications, you can execute code as below in order to include nested resources while doing a find query.
-
 ```javascript
-store.findRecord('post', postId, { adapterOptions: { query: { include: 'comments' } } });
+store.findRecord('post', postId, { include: 'comments' } );
 ```
-will generate the path `/posts/{postId}?include='comments'`
+which will generate the path /posts/{postId}?include='comments'
```
NullVoxPopuli: line 134 should be changed to `render jsonapi: ...`

docs/integrations/ember-and-json-api.md
NullVoxPopuli: I would reword "use the form below" it sounds awkward, imo.

docs/integrations/ember-and-json-api.md
bf4: I prefer it rather than silent failure and invalid data

docs/integrations/ember-and-json-api.md
bf4: This is more like the following story:

`render jsonapi: foo`

Client: Hello. I am sending you JSONAPI data.  I'll accept yours answers only as JSON API data
Server: Here's some JSON API data

`respond_to do |format| format.jsonapi { render jsonapi: foo } end`

Client: Hello. I am sending you JSONAPI data.  I'll accept yours answers only as JSON API data
Server: You want JSON API data? Here's some JSON API data

docs/integrations/ember-and-json-api.md
NullVoxPopuli: ah, that makes sense. 
In my app, I disabled document validating, cause I'm using embedded documents. I think that was pre-`!` though. I don't remember much of that code and what is called when.

docs/integrations/ember-and-json-api.md
NullVoxPopuli: I like this story.

bf4: Which means

> Rails also requires your controller to tell it that you accept and generate JSONAPI data

Isn't quite true. Rails even has an [ignore_accept_header](https://github.com/rails/rails/search?utf8=%E2%9C%93&q=ignore_accept_header&type=Code)

bf4:
You may want to rebase this off of master.

docs/integrations/ember-and-json-api.md
bf4: should this be included as an comment somewhere? People are still on 1.13

docs/integrations/ember-and-json-api.md
```diff
@@ -102,18 +118,16 @@ export default DS.JSONAPISerializer.extend({
 
 ```
 
+
 ## Including Nested Resources
```
bf4: Would be great to mention the whole 'embedded resources' controversy here and how it's not JSON API so we don't do it, but ....

docs/integrations/ember-and-json-api.md
NullVoxPopuli: I'm not sure if 1.13 is officially supported other than some bug fixes.

docs/integrations/ember-and-json-api.md
NullVoxPopuli: yeah, maybe mention the difference between including, and embedding, and that there are ways to embed, such as: https://github.com/NullVoxPopuli/aeonvera/blob/master/config/initializers/active_model_serializers.rb#L18

NullVoxPopuli:
@Ikariusrb do you think you can finish this up?

PR #1897, opened 2016-08-31
Refactor Association into Field like everything else

bf4:
Intended to simplify work in https://github.com/rails-api/active_model_serializers/pull/1857

PR #1898, opened 2016-08-31
Organize Serializer concerns into concerns/

bf4:
I keep losing track of which things in the serializer/ folder
are part of the serializer that have been put in their own 'concern' file
and which files represent their own object.

Related to #1857 and #1897


NullVoxPopuli:
imafan.jpg

NullVoxPopuli:
(I've also had the same issue)

NullVoxPopuli:
I think one small further / next step would be to differentiate between the relationship hierarchy, and everything else.

NullVoxPopuli:
but that requires class/module namespace changing... and ... blegh

Issue #1877, opened 2016-08-09
Inheriting from ActiveModelSerializers::Model to support serialization of a PORO is a bad practice


vasilakisfil:
#### Expected behavior vs actual behavior
Expected:
```ruby
class SimpleObject
  include ActiveModelSerializers::Model #includes neccesary methods

end
```

Actual
```ruby
class SimpleObject < ActiveModelSerializers::Model

end
```

What would happen if my SimpleObject was inheriting from another class?

#### Additonal helpful information
#1272

NullVoxPopuli:
that's a good point.

if you want to make a quick lil PR that extracts the body of AMS::Model into a module and update the docs for using it, that would be hugely appreciated.

Maybe call it `ModelMixin` or something? idk. that way AMS::Model could include it, and people's existing code wouldn't break.

vasilakisfil:
alright!

bf4:
Why AMS::Model is an executable documentation of the API.  Why is inheritance a problem?

bf4:
Oh, you'd rather just be able to include a mixin. So, here's the issue: you might not want the precise implementation of AMS::Model.  In fact, there's a bug in it.

bf4:
Also related

- https://github.com/rails-api/active_model_serializers/issues/1283
- https://github.com/rails-api/active_model_serializers/pull/1873#issuecomment-237859399

bf4:
@vasilakisfil Is there anything else you'd like to discuss here?  FWIW, I use AMS::Model, but there is a known bug in it around the attributes hash and read attribute for serialization not returning the same thing, which is due to different implementation decisions in AM::Model and AM::Serializers::JSON.. I haven't resolved it yet

vasilakisfil:
No I think I am fine although I haven't seen the bug you are talking about.

richmolj:
I rather agree with @vasilakisfil a mixin would be preferable. My understanding from [this comment on #1873](https://github.com/aldreth/active_model_serializers/blob/aa3e0f4a7d0082c357c745984cefd4aac7f49179/lib/active_model_serializers/model.rb#L32-L38) is that the bug is:

```ruby
class Post < ActiveModelSerializers::Model
 attr_accessor :title
end

# works
Post.new(title: 'my title').as_json # { title: 'my title' }

# does not work
post = Post.new
post.title = 'my title'
post.as_json # {}
```

[...]

PR #1903, opened 2016-09-01
Fix AMS::Model accessor vs. attributes mutation

bf4:
So far when I fix it, other tests break, but I think that's because the subclass 'Model' has that crazy method_missing

### DIFF
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -30,6 +30,8 @@ module ActiveModelSerializers
     end
 
     def read_attribute_for_serialization(key)
+      return send(key) if respond_to?(key)
+
       if key == :id || key == 'id'
         attributes.fetch(key) { id }
       else
--- a/test/active_model_serializers/model_test.rb
+++ b/test/active_model_serializers/model_test.rb
@@ -35,8 +35,6 @@ module ActiveModelSerializers
       # Change via accessor
       instance.one = :not_one
 
-      expected_attributes = { one: :not_one, two: 2, three: 3 }
-      assert_equal expected_attributes, instance.attributes
       assert_equal :not_one, instance.one
       assert_equal :not_one, instance.read_attribute_for_serialization(:one)
     end
@@ -50,21 +48,16 @@ module ActiveModelSerializers
       instance = klass.new(original_attributes)
 
       # Initial value
-      expected_attributes = { id: :ego, one: 1, two: 2, three: 3 }
-      assert_equal expected_attributes, instance.attributes
       assert_equal 1, instance.one
       assert_equal 1, instance.read_attribute_for_serialization(:one)
 
       # Change via accessor
       instance.id = :superego
 
-      expected_attributes = { id: :superego, one: 1, two: 2, three: 3 }
-      assert_equal expected_attributes, instance.attributes
       assert_equal :superego, instance.id
       assert_equal :superego, instance.read_attribute_for_serialization(:id)
     ensure
       self.class.send(:remove_const, :SomeTestModel)
     end
-
   end
 end
--- a/test/active_model_serializers/model_test.rb
+++ b/test/active_model_serializers/model_test.rb
@@ -4,7 +4,7 @@ module ActiveModelSerializers
   class ModelTest < ActiveSupport::TestCase
     include ActiveModel::Serializer::Lint::Tests
 
-    def setup
+    setup do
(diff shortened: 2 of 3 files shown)


richmolj:
[This comment](https://github.com/rails-api/active_model_serializers/issues/1877#issuecomment-244746036) notes how to fix. Though I prefer we [get rid of this thing](https://github.com/rails-api/active_model_serializers/pull/1911)

NullVoxPopuli:
:+1: for fixing by just renaming AMS::Model to AMS::Serializable

richmolj:
I am not sure how that would fix it 😉

NullVoxPopuli:
@bf4, @richmolj I fixed the tests.

Had to add send if respond_to?

and removed the attributes comparison, because it doesn't make sense when using attr_accessor (which is what the send covers)

NullVoxPopuli:
hopefully the tests pass and it'll be ready to merge

NullVoxPopuli:
darn testing gems not working on jruby

test/active_model_serializers/model_test.rb
```diff
@@ -18,5 +18,46 @@ def test_initialization_with_string_keys
 
       assert_equal model_instance.read_attribute_for_serialization(:key), value
     end
+
+    def test_attributes_can_be_read_for_serialization
```
richmolj: It looks like this test is about a change via accessor instead of constructor. Maybe the test name could be a little more descriptive in that regard?

test/active_model_serializers/model_test.rb
```diff
@@ -18,5 +18,46 @@ def test_initialization_with_string_keys
...
+
+      assert_equal :not_one, instance.one
+      assert_equal :not_one, instance.read_attribute_for_serialization(:one)
+    end
+
+    def test_id_attribute_can_be_read_for_serialization
+      klass = Class.new(ActiveModelSerializers::Model) do
+        attr_accessor :id, :one, :two, :three
+      end
+      self.class.const_set(:SomeTestModel, klass)
```
richmolj: Why is this needed?

lib/active_model_serializers/model.rb
```diff
@@ -30,6 +30,8 @@ def updated_at
     end
 
     def read_attribute_for_serialization(key)
+      return send(key) if respond_to?(key)
```
richmolj: Hm, I'm not totally sure this would work. You'd get into this scenario:

```ruby
class Post < ActiveModelSerializers::Model
  attr_accessor :title
end

post = Post.new('title' => 'my title')
post.read_attribute_for_serialization(:title) # => nil
```

NullVoxPopuli: that's true. I think that's just another point for the current AMS::Model being broken. :-(

Do you have a PR for your module idea?

richmolj: Nope, I was waiting to get approval of the RFC. Because what would that module be? Would it contain the constructor logic (in which case we are in the business of overriding the superclass's constructor which is gross)? And if not, what *would* it include?

As far as I can tell it would just be `alias read_attribute_for_serialization send`, and in that case I'm not sure what the point of it is.

So, TBH AMS::Model is nonsensical to me, which means I don't think I can touch the code or enhance the documentation in a meaningful way. I created the RFC to try and improve that situation, but doesn't look like it is getting any traction.

richmolj: FWIW you *can* fix the current class with something like [this](https://github.com/richmolj/active_model_serializers/blob/45043f29ba743ae9390aefc8965edebc45e08960/lib/active_model_serializers/model.rb#L13-L17) and [this](https://github.com/richmolj/active_model_serializers/blob/45043f29ba743ae9390aefc8965edebc45e08960/lib/active_model_serializers/model_mixin.rb#L17). That got the tests passing, including the scenario above, though it's just POC code.

bf4:
consider:

```
    class_attribute :attribute_names
    self.attribute_names = []

    def self.attributes(*names)
      attr_accessor(*names)
      define_attribute_methods(names)
      self.attribute_names += names
    end

    def initialize(attributes = {})
      attributes.each do |attr, value|
        self.public_send("#{attr}=", value)
      end if attributes
    end

    def attributes
      attribute_names.each_with_object({}) {|field, result|
        result[field] = public_send(field)
      }
    end

```

bf4:
Closed by #1984

PR #1910, opened 2016-09-03
Enhance PORO documentation

vasilakisfil:
#### Purpose
I found it a bit frustrating sometimes when I just wanted to serialize an object using an AMS serializer, but I first had to define a class for it and either inherit or duck type the needed AMS methods. I showcase how this can be done quickly and dirty but I explicitly say that it is not recommended.

#### Changes
Enhance PORO documentation

Sometimes I felt that POROs (its methods actually that have the same name with the serializer defined attributes) and Hashes (the values which belong to keys which have the same name with the serializer defined attributes) should be serialized automatically without doing anything when specifying a serializer.


docs/howto/serialize_poro.md
```diff
@@ -21,12 +24,56 @@ class MyModel
 end
 ```
 
-Fortunately, ActiveModelSerializers provides a [`ActiveModelSerializers::Model`](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/model.rb) which you can use in production code that will make your PORO a lot cleaner.  The above code now becomes:
+### Inheriting ActiveModelSerializers::Model
+Fortunately, ActiveModelSerializers provides a [`ActiveModelSerializers::Model`](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/model.rb) which you can use in production code that will make your PORO a lot cleaner.
```
bf4: What's missing from these docs is that AMS::Model also serves as executable documentation of the serializable interface, which can be copied or referenced without actually using the class.

docs/howto/serialize_poro.md
```diff
@@ -21,12 +24,56 @@ class MyModel
...
+    an_instance = OpenStruct.new(id: 1, name: 'Just an example')
+
+    an_instance.singleton_class.send(:alias_method, :read_attribute_for_serialization, :send)
+
+    render jsonapi: an_instance, serializer: Api::V1::UserSerializer
+  end
+end
+```
+
+If your object is just a hash you can do the following trick:
```
bf4: I would like to strongly recommend not trying to serialize primitives like Hashes.  AMS is intended to models.  There's a lot of parts of the AMS contract that would be weird to have on a hash.  https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/lint.rb

vasilakisfil: What do you mean by saying "models" ?

vasilakisfil: ping @bf4

bf4: @vasilakisfil by model I mean a class that represents some domain concern.  A Hash, being a primitive, doesn't do that. I think there's is a layer of AMS should be doing hash transformations, but that's not it's primary input, and that layer isn't yet exposed.  

The imaginary pipeline is something like:

```plain
Given a model
Define attributes and associations to serializer (via ActiveModel::Serializer)
Serialize model to 'serializable_hash' per serializer and fields/include options (coordinated by adapter)

Given a serializable hash
Transform in adapter
```

and then we would have a caching layer available in the pipeline

docs/howto/serialize_poro.md
```diff
@@ -21,12 +24,56 @@ class MyModel
...
+```ruby
+class Api::V1::UsersController
+  def show
+    #this is just an example
+    an_instance = {id: 1, name: 'Just an example'}
+    an_instance.singleton_class.send(:alias_method, :read_attribute_for_serialization, :[])
```
bf4: I think maybe what we want here is something like:

Given a `user` instance, the serializer should act the `user` to get the defined attributes and associations.  However, since the transformation for a `user` with `id: 1` and `name: 'Mortimer'` `could be compared to something like  `user_attributes = user.attributes; user_attributes.slice(:id, :name)`, wouldn't it be nice if we could start from `user_attributes` as `{ id: 1, name: 'Mortimer' }` and have the serializer treat that hash as user_attributes?

Issue #1890, opened 2016-08-25
ApplicationSerializer not picked up by generator


ganlron:
#### Expected behavior vs actual behaviour
I've created an ApplicationSerializer in app/serializers/application_serializer.rb and was expecting rails g serializer foo to inherit from this, but it still defaults to ActiveModel::Serializer.

I can get around this by passing ApplicationSerializer with --parent and it works, but wasn't expected behaviour

#### Environment

Rails 5.0.0.1
ruby 2.3.1p112

bf4:
@ganlron which AMS version? I'm unfamiliar with `--parent`.  You ran `rails g serializer User --parent ApplicationSerializer`?

bf4:
Oh, ha, it's our own option https://github.com/rails-api/active_model_serializers/blob/master/lib/generators/rails/serializer_generator.rb#L9

Looks like it's because the generator checks if the ApplicationSerializer is defined, and it may not have been loaded yet?  So, logic there could be better?

ganlron:
LOL ya, was your option :)

bf4:
I'm pretty sure you could fix this by changing the `defined?` check


```ruby

      def parent_class_name
        options[:parent] or  ::ApplicationSerializer.name
      rescue NameError
          'ActiveModel::Serializer'
      end
```

PR #1912, opened 2016-09-05
Ensure generator picks up ApplicationSerializer

richmolj:
#### Purpose

Fixes https://github.com/rails-api/active_model_serializers/issues/1890

#### Changes

Look at filesystem instead of using `defined?`

#### Caveats

If the file was there but the class was not named correctly, I think you'd get an unfriendly error.

#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/issues/1890

ApplicationSerializer could exist, but not be loaded. So, we check
existence by looking at the filesystem instead of defined?.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Breaking changes:
 Fixes:
 
 - [#1887](https://github.com/rails-api/active_model_serializers/pull/1887) Make the comment reflect what the function does (@johnnymo87)
+- [#1890](https://github.com/rails-api/active_model_serializers/issues/1890) Ensure generator inherits from ApplicationSerializer when available (@richmolj)
 
 Features:
 
--- a/lib/generators/rails/serializer_generator.rb
+++ b/lib/generators/rails/serializer_generator.rb
@@ -25,7 +25,7 @@ module Rails
       def parent_class_name
         if options[:parent]
           options[:parent]
-        elsif defined?(::ApplicationSerializer)
+        elsif 'ApplicationSerializer'.safe_constantize
           'ApplicationSerializer'
         else
           'ActiveModel::Serializer'
--- a/test/generators/serializer_generator_test.rb
+++ b/test/generators/serializer_generator_test.rb
@@ -20,11 +20,10 @@ class SerializerGeneratorTest < Rails::Generators::TestCase
   end
 
   def test_uses_application_serializer_if_one_exists
-    Object.const_set(:ApplicationSerializer, Class.new)
-    run_generator
-    assert_file 'app/serializers/account_serializer.rb', /class AccountSerializer < ApplicationSerializer/
-  ensure
-    Object.send :remove_const, :ApplicationSerializer
+    stub_safe_constantize(expected: 'ApplicationSerializer') do
+      run_generator
+      assert_file 'app/serializers/account_serializer.rb', /class AccountSerializer < ApplicationSerializer/
+    end
   end
 
   def test_uses_given_parent
@@ -52,6 +51,24 @@ class SerializerGeneratorTest < Rails::Generators::TestCase
       else
         assert_no_match(/\n\nend/, content)
       end
+    end
+  end
+
+  private
+
+  def stub_safe_constantize(expected:)
+    String.class_eval do
+      alias_method :old, :safe_constantize
+    end
+    String.send(:define_method, :safe_constantize) do
+      Class if self == expected
+    end
+
+    yield
+  ensure
+    String.class_eval do
+      alias_method :safe_constantize, :old
+      undef_method :old
     end
   end
 end


richmolj:
@bf4 attempted the fix you suggest in [this comment] but didn't work. Not sure if there is a better filesystem stub for this, I'm used to rspec.

richmolj:
Hm it actually just dawned on me if you *don't* have a namespace/application_serializer but do have a non-namespaced one, we won't fall back to the non-namespaced one. Do we care about this use case / what's the correct behavior?

lib/generators/rails/serializer_generator.rb
```diff
@@ -23,12 +23,14 @@ def association_names
...
-        elsif defined?(::ApplicationSerializer)
-          'ApplicationSerializer'
+        if parent = options[:parent]
+          parent
         else
-          'ActiveModel::Serializer'
+          namespace = class_path.map(&:camelize).join('::')
+          application_serializer = [namespace, 'ApplicationSerializer']
+            .reject(&:blank?).join('::')
+          path = File.join('app/serializers', class_path, "application_serializer.rb")
```
bf4: since this is only relevant in a Rails app, I'd rather rely on Rails autoloading for a more robust support and less code on us.

i.e.

```ruby
  return parent if (parent = options[:parent])
  ApplicationSerializer.name
rescue NameError
  'ActiveModel::Serializer'
end
```

bf4: (or use `safe_constantize` or different logic, but same idea)

test/generators/serializer_generator_test.rb
```diff
@@ -20,11 +20,10 @@ def test_generates_a_namespaced_serializer
   end
 
   def test_uses_application_serializer_if_one_exists
-    Object.const_set(:ApplicationSerializer, Class.new)
-    run_generator
+    File.stub(:exists?, true) do
+      run_generator
+    end
```
bf4: we might want to make this an isolated test...

lib/generators/rails/serializer_generator.rb
richmolj: Yep, I gave that a shot based on your original comment. Does not seem to work, I believe because these files aren't `required` at this time.

richmolj:
@bf4 actually you were right, [this](https://github.com/rails-api/active_model_serializers/issues/1890#issuecomment-242818408) won't work but `safe_constantize` will. Removed the namespace stuff, added changelog.

New test had to get a little wonky though.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Breaking changes:
 Fixes:
 
 - [#1887](https://github.com/rails-api/active_model_serializers/pull/1887) Make the comment reflect what the function does (@johnnymo87)
+- [#1890](https://github.com/rails-api/active_model_serializers/issues/1890) Ensure generator Inherits from ApplicationSerializer when available (@richmolj)
```
bf4: `s/Inherits/inherits`

test/generators/serializer_generator_test.rb
```diff
@@ -54,4 +53,21 @@ def test_with_no_attributes_does_not_add_extra_space
...
+    String.send(:define_method, :safe_constantize) do
+      Class if self == expected
+    end
+
+    yield
+  ensure
+    String.class_eval do
+      alias_method :safe_constantize, :old
+    end
+  end
```
bf4: Or just

```ruby
class << test_string'
  define_method(:safe_constantize) do 
    Class if self == expected
  end
end
```

bf4: Or you could `undef` in the ensure

richmolj: I don't think the `class << "ApplicationSerializer"` will work - gave it a shot and failed. Wouldn't that only patch the singleton of that particular *instance* of String?

test/generators/serializer_generator_test.rb
```diff
@@ -20,11 +20,10 @@ def test_generates_a_namespaced_serializer
   end
 
   def test_uses_application_serializer_if_one_exists
-    Object.const_set(:ApplicationSerializer, Class.new)
-    run_generator
-    assert_file 'app/serializers/account_serializer.rb', /class AccountSerializer < ApplicationSerializer/
-  ensure
-    Object.send :remove_const, :ApplicationSerializer
+    stub_safe_constantize(expected: 'ApplicationSerializer') do
```
NullVoxPopuli: what's a failure look like when `self != expected` (inverse of line 64)?

richmolj: `safe_constantize` would return `nil`, meaning the generator would (incorrectly) inherit from `ActiveModel::Serializer` instead of `ApplicationSerializer`:

```
  1) Failure:
SerializerGeneratorTest#test_uses_application_serializer_if_one_exists [/Users/lrichmond1/code/active_model_serializers/test/generators/serializer_generator_test.rb:25]:
Expected /class AccountSerializer < ApplicationSerializer/ to match "class AccountSerializer < ActiveModel::Serializer\n  attributes :id, :name, :description\n  has_one :business\nend\n".
```

bf4:
@richmolj LGTM, you wanna merge?

PR #1913, opened 2016-09-05
Add helpful testing gems

richmolj:
Adds gems [m](https://github.com/qrush/m) (run minitest by line numbers like rspec) and [pry-byebug](https://github.com/deivid-rodriguez/pry-byebug) (for binding.pry support).

Addresses https://github.com/rails-api/active_model_serializers/pull/1911#discussion_r77548416

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,6 +45,9 @@ group :test do
   gem 'sqlite3',                          platform: (@windows_platforms + [:ruby])
   gem 'activerecord-jdbcsqlite3-adapter', platform: :jruby
   gem 'codeclimate-test-reporter', require: false
+  gem 'm', '~> 1.5'
+  gem 'pry', '~> 0.10'
+  gem 'pry-byebug', '~> 3.4', platform: :ruby
 end
 
 group :development, :test do
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -9,6 +9,7 @@ rescue LoadError
   STDERR.puts 'Running without SimpleCov'
 end
 
+require 'pry'
 require 'timecop'
 require 'rails'
 require 'action_controller'


test/test_helper.rb
```diff
@@ -9,6 +9,12 @@
   STDERR.puts 'Running without SimpleCov'
 end
 
+begin
```
bf4: If you just require `pry`, it will load all pry plugins.  This will lead to less churn if/when we add more pry configurations.  I won't block if this doesn't seem interesting to you.  It would require adding `pry` to the gemfile, then we could remove the `rescue` here since pry is all platforms

richmolj: sure ill give that a shot

Gemfile
```diff
@@ -45,6 +45,8 @@ group :test do
   gem 'sqlite3',                          platform: (@windows_platforms + [:ruby])
   gem 'activerecord-jdbcsqlite3-adapter', platform: :jruby
   gem 'codeclimate-test-reporter', require: false
+  gem 'm', '~> 1.5'
+  gem 'pry', '~> 0.10'
```
bf4: I mean, put in `pry-byebug` as well, just limit it to valid platforms

richmolj: oh, duh, yes of course. updated, thanks

NullVoxPopuli: what advantage does adding `pry` provide? I thought pry-byebug included it.

test/test_helper.rb
```diff
@@ -9,6 +9,7 @@
   STDERR.puts 'Running without SimpleCov'
 end
 
+require 'pry'
```
NullVoxPopuli: shouldn't this be `require 'pry-byebug'`?

richmolj: This was actually a great suggestion by @bf4 - if you `require 'pry'` it will load any pry plugins for you. So if you are non-MRI you will only get pry (works across rubies), if you are MRI you get `pry-byebug` as well (both specified in Gemfile).

NullVoxPopuli: neato

Issue #1921, opened 2016-09-08
Railties dependency


ggpasqualino:
#### Expected behavior vs actual behavior
One of our applications is a non rails application and we use standalone activerecord 4.2.7.1, active_model_serializers 0.8.3 and ruby 2.3.1.

We are looking forward to update active_model_serializers to 0.10.x, but in this version railties is not an optional dependency anymore, and because of that it would mess our initialization.

Is there a reason behind this mandatory dependency? Is there a plan to make it work in a standalone manner as before?

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 0.10.2

Output of `ruby -e "puts RUBY_DESCRIPTION"`: ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]

OS Type & Version: OSX 10.11.6

Integrated application and version *(e.g., Rails, Grape, etc)*: 
Grape 0.17.0
Activerecord 4.2.7.1

PR #1922, opened 2016-09-11
Make railties an optional dependency

ggpasqualino:
#### Purpose
Make railties an optional dependency in order to use active_model_serializers in non rails projects again.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/issues/1921

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
   # 'rack'
   # 'rack-test', '~> 0.6.2'
 
-  spec.add_runtime_dependency 'railties', rails_versions
+  spec.add_development_dependency 'railties', rails_versions
   # 'activesupport', rails_versions
   # 'actionpack', rails_versions
   # 'rake', '>= 0.8.7'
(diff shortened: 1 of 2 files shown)


richmolj:
I'm in favor of this change. It looks like [this PR](https://github.com/rails-api/active_model_serializers/pull/1090) by @bf4 moved us from rails as a development dependency to a runtime dependency, but reading the commit/PR this seems like it might have been unintended.

@bf4 is that correct?

(side note: if we go in the direction of [deprecating ActiveModelSerializers::Model](https://github.com/rails-api/active_model_serializers/pull/1911) activemodel could move to a development dependency as well, I think)

richmolj:
Also, just a heads up we are [already only requiring the railtie if Rails is defined](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers.rb#L43) so no worries there.

@ggpasqualino even though this is a one-line change, I could see an entry in the changelog being helpful.

ggpasqualino:
@richmolj thank you. I will check the changelog and update the PR.

Sorry about this mess. I didn't realize that when I created https://github.com/rails-api/active_model_serializers/issues/1921 because I was looking at the wrong version of the file, then I realised that in master [you are already only requiring the railtie if Rails is defined](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers.rb#L43) and created this Pull Request.

active_model_serializers.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
   # 'rack'
   # 'rack-test', '~> 0.6.2'
 
-  spec.add_runtime_dependency 'railties', rails_versions
+  spec.add_development_dependency 'railties', rails_versions
```
bf4: 👍

PR #1930, opened 2016-09-16
Ensure valid jsonapi when blank relationship

richmolj:
#### Purpose

If you specify include_data false, and do not have any links for this
relationship, we would output something like:

`{ relationships: { comments: {} } }`

This is not valid jsonapi. 

#### Changes

We will now render

`{ relationships: { comments: { meta: {} } } }`

#### Caveats

N/A

#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/pull/1797

Relevant jsonapi spec: http://jsonapi.org/format/#document-resource-object-relationships

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/relationship.rb
+++ b/lib/active_model_serializers/adapter/json_api/relationship.rb
@@ -24,6 +24,7 @@ module ActiveModelSerializers
 
           meta = meta_for(association)
           hash[:meta] = meta if meta
+          hash[:meta] = {} if hash.empty?
 
           hash
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Fixes:
 - [#1887](https://github.com/rails-api/active_model_serializers/pull/1887) Make the comment reflect what the function does (@johnnymo87)
 - [#1890](https://github.com/rails-api/active_model_serializers/issues/1890) Ensure generator inherits from ApplicationSerializer when available (@richmolj)
 - [#1922](https://github.com/rails-api/active_model_serializers/pull/1922) Make railtie an optional dependency in runtime (@ggpasqualino)
+- [#1930](https://github.com/rails-api/active_model_serializers/pull/1930) Ensure valid jsonapi when relationship has no links or data (@richmolj)
 
 Features:
 
--- a/test/adapter/json_api/relationship_test.rb
+++ b/test/adapter/json_api/relationship_test.rb
@@ -54,7 +54,7 @@ module ActiveModelSerializers
         end
 
         def test_relationship_data_not_included
-          test_relationship({}, options: { include_data: false })
+          test_relationship({ meta: {} }, options: { include_data: false })
         end
 
         def test_relationship_simple_link


lib/active_model_serializers/adapter/json_api/relationship.rb
```diff
@@ -34,6 +34,11 @@ def as_json
 
         private
 
+        def ensure_valid!(json)
```
NullVoxPopuli: is this the JSON API recommended way to deal with this problem? seems funky

richmolj:
@kevintyll [brought this up](https://github.com/rails-api/active_model_serializers/pull/1797#issuecomment-247645659). I'm making this change independent of #1797 since the situation already exists in master.

lib/active_model_serializers/adapter/json_api/relationship.rb
richmolj: We definitely need one of `data`, `links`, or `meta` according to the spec. We definitely don't have the first (remember empty data is incorrect here). Empty `links` also seems invalid, so that's a no go. [The section on meta](http://jsonapi.org/format/#document-meta) seems to indicate empty meta is valid. So If a better alternative exists, I don't know it.

NullVoxPopuli: fair enough

remear: > remember empty data is incorrect here

I don't remember. Why can't it be `data: null`, or `data: []`?

richmolj: @remear it's the difference between "we're not saying anything about this relationship one way or the other" versus "this relationship is blank".

For instance, let's say you use ember-data and `post` belongs to `author`, both objects already loaded into memory. A subsequent GET to the API that returns `author: { data: null }` will blow away `post.author`. Omitting the data key does not.

remear: I'm not following. If the GET returns null in the example you gave, there would be no actual relationship persisted in the system, right? Wouldn't you expect it to update your objects according to the result of the GET you'd just performed?

NullVoxPopuli: Well, we should be testing against this: http://www.jsonschemavalidator.net/

NullVoxPopuli: with this schema: http://jsonapi.org/schema

NullVoxPopuli: So, with this payload, JSON API is valid

```
{
  "data": {
    "id": "1",
    "type": "my-type",
    "attributes": {
      "name": "my name"
    },
    "relationships": {
      "posts": {
      }
    }
  }
}
```

So this looks like it may be an issue with ember?

NullVoxPopuli: also valid:
```

Comment (remear) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: That's sortof where I was heading. As far as I can see, the actual spec file does not list any required keys for `relationships`. So I'm curious to know where the requirement for one of `data`, `links`, or `meta` stems from.

Comment (NullVoxPopuli) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: I think that's just for the root node

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: @remear from the link in this PR description

> A “relationship object” MUST contain at least one of the following:

And then specifies `data`, `links`, and `meta`.

> I'm not following. If the GET returns null in the example you gave, there would be no actual relationship persisted in the system, right?

The relationship is persisted, we just aren't returning it in this response. If you return data `null`, you are saying "this relationship doesn't exist" versus "this API response does not contain info about this relationship"

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: @NullVoxPopuli from jsonapi.org:

> The value of the relationships key MUST be an object (a “relationships object”)

That tells me, can't be null. If it can be null, this seems reasonable.

**data** being null or [] is the only no go.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Oh, by the way! This actually came up for me real-world style a few weeks ago, but I forgot until @kevintyll brought it up.

I was writing a jsonapi client in ruby, using the [jsonapi gem](https://github.com/beauby/jsonapi) AMS also uses. That gem was blowing up on parsing due to exactly this issue.

Not to say that lib is a golden source of truth, but FYI. We gotta choose to follow the docs or the schema, that's an example of other libraries we need to play nicely with favoring the docs.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Created an issue for jsonapi: https://github.com/json-api/json-api/issues/1092, we'll see if it gets any traction...

Comment (NullVoxPopuli) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: @richmolj but then the schema for json api is wrong if that's true :-\

Comment (remear) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]:
```
"relationships": {
  "description": "Members of the relationships object (\"relationships\") represent references from the resource object in which it's defined to other resource objects.",
  "type": "object",
  "patternProperties": {
    "^\\w[-\\w_]*$": {
      "properties": {
        "links": {
          "$ref": "#/definitions/links"
        },
        "data": {
          "description": "Member, whose value represents \"resource linkage\".",
          "oneOf": [
            {
              "$ref": "#/definitions/relationshipToOne"
            },
            {
              "$ref": "#/definitions/relationshipToMany"
            }
          ]
        },
        "meta": {
          "$ref": "#/definitions/meta"
        }
      },
      "additionalProperties": false
    }
  },
  "additionalProperties": false
}
```

None of those are set as required. `relationships` itself isn't a required key either.

¯\_(ツ)_/¯

> we just aren't returning it in this response

This is what I'm trying to understand. The behavior you're describing seems wrong to me.

I agree that `data: null` and `data: []` convey _empty_ relationships. If you don't want to return a relationship in the response, wouldn't you just omit the entire relationship? So if you don't want to return the relationships for posts, then it seems logical that this would be conveyed by not having `posts` in `relationships` at all. And continuing onward, if there are no relationships, why have a `relationships` key? If you want to let consumers know that a relationship does exist but isn't available from the current available data, you'd probably provide a `link` for it, right? Conveying this "unknown relationships" information with empty meta seems weird to me. What if the relationship actually had meta if you were to fetch it?

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Yup :( For sure one of 'em is wrong.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: @remear yep - I am not saying the schema requires this stuff. I'm saying the docs require this stuff, and it looks like the docs and the schema are in conflict. So I opened up the issue on jsonapi to figure out which side has the bug.

> If you want to let consumers know that a relationship does exist but isn't available from the current available data, you'd probably provide a link for it, right?

This is the use case I would like to support. I would like to let consumers know the relationship does exist but isn't available from the current data. And I may not have a separate endpoint that I can link to - there's no reason for me to write a separate endpoint if my only use case for fetching this data is when I sideload.

> What if the relationship actually had meta if you were to fetch it?

Then we would render that `meta` as normal. We only render empty meta if no keys - including `meta` itself - are present.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: @remear @NullVoxPopuli we got a [response from jsonapi](https://github.com/json-api/json-api/issues/1092) looks like the schema is incorrect in this case and we do need one of data/meta/links. Can't be null or {}.

Comment (NullVoxPopuli) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: ok, cool. so we should also submit a PR to @beauby 's jsonapi gem for that.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: @NullVoxPopuli I'd have to crack it open again for specifics, but my point earlier was that the jsonapi gem was already behaving correctly - blowing up when this value was `{}`. Might have been an error post-validation, though.

Comment (NullVoxPopuli) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: I need to submit a PR for just nil input. My issue for that was that the _wrong_ error was being thrown.

Comment (remear) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Since the schema is the only thing we can _actually_ use for validation, I consider the schema to be the authoritative source. If the schema is wrong, they need to fix it.

I still think empty meta is the wrong way to convey the concept of unknown relationships.

Comment (NullVoxPopuli) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: https://github.com/beauby/jsonapi/pull/37 :-) slightly related

Comment (remear) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Presently, I'm not convinced this isn't an application concern until formally defined by JSONAPI. I prefer explicit over implicit behaviors. I think the empty meta approach requires clients to be responsible for knowing that empty meta implies no relationship data was available **_without telling them how to go get more_**. That's why I probably would solve this by not providing `data` in relationships (omitting the data key) and providing a `link`. That would convey a relationship is possible but you'd need to go talk to the link, or a resource-level link for the "full picture" to know more.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: We're in complete agreement when a link is possible. What if there is no separate endpoint to link to?

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Also just to be clear - it is completely possible to do what you propose with this PR. It would only be relevant to the scenario no link is present.

Comment (NullVoxPopuli) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: I think the root of the issue is that we can't actually validate our code without a change to the schema. to merge this PR would just be guessing what the schema is going to be in the meantime.  It's a pretty good guess, but I'm sure everyone using JSON API in AMS would feel more comfortable if we were inline with the schema.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: I would say it is not a guess when they are directly telling us this is the case. Are you proposing we wait for them to make this change, then merge this PR? I'm fine if that is the order of operations, though it seems unnecessary imo.

In other words, I want to verify no other problems with this PR other than waiting for them to make the schema update they have told us they are going to make.

Comment (remear) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Why does it need to be a separate endpoint? What if it's an existing one with a parameter? Why not just make an an endpoint for fetching that data? What about a relationships endpoint for that resource? The answers to those questions are better determined by the application and API design.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: > Why does it need to be a separate endpoint? What if it's an existing one with a parameter?

Then you'd be linking to a collection of posts, instead of a collection of comments, which is invalid (even if the comments are sideloaded via an ?include parameter).

> Why not just make an an endpoint for fetching that data?

Because I'd rather not code and test a completely unused endpoint just because `relationships: { comments: {} }` is not valid jsonapi.

> What about a relationships endpoint for that resource?

Not needed, this relationship is only fetched by my web application when sideloading.

> The answers to those questions are better determined by the application and API design.

Agree. With the above answers, I am saying it is a reasonable API design to not have a separate endpoint, which means no `links` can be present.

Comment (remear) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: The schema is the only thing we have against which to actually validate documents. The schema must be fixed before changes like this can be merged.

Comment (richmolj) [inline lib/active_model_serializers/adapter/json_api/relationship.rb]: Merging this PR would still validate against the existing schema 😄 

But I am fine with waiting, as long as we verify there are no other concerns.

**Comment by NullVoxPopuli** [comment]
In all this discussion, I've forgotten the _root_ of the issue. @richmolj can you summarize the problem trying to be solved here?

**Comment by richmolj** [comment]
@NullVoxPopuli is there something missing from the PR description I could add? The problem is that AMS is outputting invalid jsonapi. No actual errors thrown or anything like that within AMS, but errors *might* be thrown by clients consuming this json and expecting a compliant structure.

Props for adding the word `root` to as many issues as possible 😆

**Comment by NullVoxPopuli** [comment]
Ok, now that that is merged, @richmolj what does the difference between meta: {} and links: {} mean to you?

**Comment by richmolj** [comment]
@NullVoxPopuli I have a hunch `links: {}` is invalid. I've [opened an issue](https://github.com/json-api/json-api/issues/1095) to get clarification.

To answer your question, empty links is meaningless to me, which is why I think it is invalid. Empty `meta` tells me "I'm not saying anything about this relationship one way or the other", which is the intent of this PR.

If `links: {}` is valid, I'd be happy to accept that as a solution, though I find `meta: {}` more sensible, personally.

**Comment by NullVoxPopuli** [comment]
I guess one way to think about it that if links is empty that says: there is no way to get any information about this relationship. Meaning that the only way to even get that relationship is through the endpoint you just hit.

meta: {} just feels ambiguous to me.

idk.

@remear thoughts?

**Comment by richmolj** [comment]
Honestly, I think this is a jsonapi issue, not an AMS issue. [the jsonapi maintainers have told us empty meta is preferable](https://github.com/json-api/json-api/issues/1092#issuecomment-248012089).

I hdon't care too much one way or the other as long as it's valid, but suggest we try to convince them if we want to go in a different direction. Seems like an area AMS and jsonapi should be on the same page.

**Comment by NullVoxPopuli** [comment]
well, if JSON API people say that it's preferable, then I guess that's what we go with.

File: lib/active_model_serializers/adapter/json_api/relationship.rb [hunk 2]
```diff
@@ -25,7 +25,7 @@ def as_json
           meta = meta_for(association)
           hash[:meta] = meta if meta
```
NullVoxPopuli: Now for code review. :-)

I don't like that ensure_valid! just adds a blank meta. The method name is misleading.
Since the logic is so simple, can you add that logic to this method?

richmolj: My thought was this is a placeholder for other things we might want to do, but sure no worries.

NullVoxPopuli: I don't want to have too many placeholders right now. Overall, we're trying to simplify AMS. :-)

richmolj:
@NullVoxPopuli turns out `links: {}` **is** valid FWIW.

NullVoxPopuli:
I saw that. :-)

lib/active_model_serializers/adapter/json_api/relationship.rb
richmolj: I've force pushed this change. I still find the updated code less descriptive, but my salience is very low.

PR #1935, opened 2016-09-22
Update docs for JSONAPI relationship links

ksamc:
#### Purpose
Fixes small typos and updated the docs on relationship links for JSONAPI.

#### Changes
Removed the 'data' key in the serialized relationship. Including it according to the current docs leads to unexpected behavior when using it with libraries like Ember Data.

This [method](https://github.com/rails-api/active_model_serializers/blame/master/docs/howto/add_relationship_links.md#L111) returns an empty array if the relationship is not loaded even though they might exist. Passing an empty array as data tells clients that there are no existing relationships.

Instead data should not be sent at all so that the client uses the given nested resource link to fetch them.

### DIFF
--- a/docs/howto/add_relationship_links.md
+++ b/docs/howto/add_relationship_links.md
@@ -37,7 +37,7 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 end
 ```
 
-This will resilt in (example is in jsonapi adapter):
+This will result in (example is in jsonapi adapter):
 ```json
 {
   "data": {
@@ -69,7 +69,7 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 end
 ```
 
-This will resilt in (example is in jsonapi adapter):
+This will result in (example is in jsonapi adapter):
 ```json
 {
   "data": {
@@ -104,12 +104,7 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 
   has_many :microposts, serializer: Api::V1::MicropostSerializer do
     link(:related) { api_v1_microposts_path(user_id: object.id) }
-  end
-
-  #this is needed to avoid n+1, gem core devs are working to remove this necessity
-  #more on: https://github.com/rails-api/active_model_serializers/issues/1325
-  def microposts
-    object.microposts.loaded ? object.microposts : object.microposts.none
+    include_data false
   end
 end
 ```
@@ -126,7 +121,6 @@ This will result in:
     },
     "relationships": {
       "microposts": {
-        "data": [],
         "links": {
           "related": "/api/v1/microposts?user_id=1"
         }


vasilakisfil:
Interesting! I didn't know this method (https://github.com/rails-api/active_model_serializers/pull/1710) at all and I was fixing this issue from the ember's side in a serializer.

Looks good to me!

docs/howto/add_relationship_links.md
```diff
@@ -104,12 +104,7 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 
   has_many :microposts, serializer: Api::V1::MicropostSerializer do
     link(:related) { api_v1_microposts_path(user_id: object.id) }
-  end
-
-  #this is needed to avoid n+1, gem core devs are working to remove this necessity
-  #more on: https://github.com/rails-api/active_model_serializers/issues/1325
-  def microposts
-    object.microposts.loaded ? object.microposts : object.microposts.none
+    include_data false
```
bf4: @ksamc I think this comment is still desired cc @richmolj  ?

docs/howto/add_relationship_links.md
```diff
@@ -37,7 +37,7 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 end
 ```
 
-This will resilt in (example is in jsonapi adapter):
+This will result in (example is in jsonapi adapter):
```
bf4: ok

docs/howto/add_relationship_links.md
```diff
@@ -69,7 +69,7 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
...
```
bf4: ok

docs/howto/add_relationship_links.md
bf4: @ksamc I think this comment is still desired cc @richmolj  ?

richmolj: For this specific comment, I think it is no longer needed. `include_data false` is already documented elsewhere as avoiding the N+1/loading the data.

That said, we should probably rework this section. It's totally valid to have both `data` and `links`, so there's really no inherent reason to have `include_data false` in the documentation example in the first place. We're just talking about link generation here.

Instead, I'd add a section below this (or elsewhere) on conditionally including `data`, how and why one would do this. @ksamc any interest in adding this documentation, based on the explanation in https://github.com/rails-api/active_model_serializers/pull/1931? Otherwise I will submit a separate PR.

vasilakisfil: @richmolj I would say the opposite is more often: if you add links you probably don't include the relationship.

I would suggest to merge this request and you can start a new one about conditionally adding data (probably in another section of the wiki)

bf4:
@ksamc @groyoh Needs rebase and revision following #1981 

I think it makes sense to document `include_data false` as the usual case for including relationship links.

johnunclesam:
Please commit these changes because I almost losed my mind.

PR #1936, opened 2016-09-23
Fix ActiveRecord::Relation serializer when model has explicit serializer

palkan:
#### Purpose

Due to ActiveRecord _magic_ AMS fails to correctly determine serializer for AR::Relation.
Example:

```ruby
class Post < ActiveRecord::Base
   def self.serializer_class
      PostSpecialSerializer
   end
end

  def index
    # AMS mistakenly uses PostSpecialSerializer to render relation here instead of using CollectionSerializer
    # 'cause Relation responds to all underlying model's class methods
    render json: Post.all
  end
end 
```

#### Changes

Explicitly set `serializer_class` for `ActiveRecord::Relation`.

### DIFF
--- a/lib/active_model_serializers/railtie.rb
+++ b/lib/active_model_serializers/railtie.rb
@@ -2,6 +2,7 @@ require 'rails/railtie'
 require 'action_controller'
 require 'action_controller/railtie'
 require 'action_controller/serialization'
+require 'active_record/relation_serialization'
 
 module ActiveModelSerializers
   class Railtie < Rails::Railtie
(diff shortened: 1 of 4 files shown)


richmolj:
@palkan thanks for the clear explanation. I believe the issue here is 'serializer_class' should be an instance method, but you are defining it as a class method. If you omitted this patch but defined 'serializer_class' as an instance method, would everything work as expected?

palkan:
@richmolj Oh, what a shame 😶 🌵 
Of course, it works)

Issue #1853, opened 2016-07-21
AMS should error when no serializer can be found


cantino:
I think the default behavior of AMS should be to error if a serializer cannot be found.  Otherwise, it's easy to stumble into a major security hole.

Imagine you have a controller, maybe something like:

```ruby
class RecentUsersController < ApplicationController
  def index
    render json: User.recent.limit(5)
  end
end
```

and a serializer:

```ruby
class UserSerializer < ActiveModel::Serializer
  attribute :first_name
  attribute :user_since
end
```

Now, imagine you either put that serializer at the wrong path by accident, or you named it `RecentUserSerializer`, but forgot to specify it by name in the render call.  Either way, you're now calling `as_json` on the models and handing hashed passwords out over your API.

I'm using something like this to fix the issue, but I think it should be default behavior:

```ruby
module ErrorOnMissingSerializer
  module AsJSONOverride
    def as_json(*_)
      serializable_resource = ActiveModelSerializers::SerializableResource.new(self)
      if serializable_resource.serializer
        serializable_resource.as_json
      else
        raise NotImplementedError, "No serializer was found for #{self.class.name}"
      end
    end
  end

  class Railtie < Rails::Railtie
    ActiveSupport.on_load :active_record do
      ActiveRecord::Base.include ErrorOnMissingSerializer::AsJSONOverride
    end
  end
end
```

bf4:
@cantino I think there is room for a 'strict mode', that raises an `MissingSerializer = Class.new(ArgumentError)`  instead of returning `nil` when no serializer is found.  The complication is that we'd need to ensure that the object respond to `read_attribute_for_serialization` or something so that it's not a string, hash, array, etc.

Relevant code https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer.rb#L81-L87

```diff
    # @api private
    # Find a serializer from a class and caches the lookup.
    # Preferentially returns:
    #   1. class name appended with "Serializer"
    #   2. try again with superclass, if present
-    #   3. nil
+    #   3. config.models_require_serializer? ? fail(MissingSerializer) : nil
    def self.get_serializer_for(klass)
      return nil unless config.serializer_lookup_enabled
      serializers_cache.fetch_or_store(klass) do
        # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
        serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }

[...]

Issue #1813, opened 2016-06-17
Possible regression in 0.10.1 - Object instance variable in parent serializer set to serializer in 10.1


theotherzach:
#### Expected behavior vs actual behavior

This bug came up after I upgraded AMS from 0.10.0 to 0.10.1. Note that I cannot reproduce this behavior outside of the controller.

**Given**

* A serializer defines a custom method which makes a call to the `object` method
```ruby
def total_cents
  p object
  object.quantity * object.price_cents
end
```
* A second serializer inherits from the above serializer

**When**

* The second serializer is explicitly instantiated inside of a controller. 

**Result**

* The `object` call in `total_cents` defined on the parent serializer references a serializer

**Expected**

* The `object` call in `total_cents` should reference the object being serialized

#### Steps to reproduce

I believe that these are all of the relevant files and the Rspec test that exposes the error. Please let me know if this isn't detailed enough.

```ruby
# app/serializers/api/v1/order_line_item_serializer.rb
module Api::V1

  class OrderLineItemSerializer < ActiveModel::Serializer

    attributes :id,
               :order_id,
               :variant_id,
               :quantity,
               :price_cents,
               :total_cents,
               :created_at,
               :updated_at,

               def total_cents
                  p object
                 object.quantity * object.price_cents
               end

  end

end
```

```ruby
# app/serializers/api/v1/cart/order_line_item_serializer.rb
module Api::V1

  class Cart::OrderLineItemSerializer < Api::V1::OrderLineItemSerializer

    attributes :product_name, :variant_name

    def product_name
      object.variant.product_name
    end

    def variant_name
      object.variant.name
    end

  end

end
```

```ruby
module Api::V1
# app/controllers/api/v1/order_line_items_controller.rb

  class OrderLineItemsController < BaseController

    before_action :assign_and_authorize_order_line_item, only: [:update, :destroy]

    # PATCH /api/v1/order_line_items/:id
    def update
      if @order_line_item.update_attributes(order_line_item_params)
        render json: ::Api::V1::Cart::OrderLineItemSerializer.new(@order_line_item), status: :ok
      else
        render json: @order_line_item.errors.to_json, status: :unprocessable_entity
      end
    end

    # DELETE /api/v1/orders_line_items/:id
    def destroy
      if @order_line_item.destroy
        render json: ::Api::V1::Cart::OrderLineItemSerializer.new(@order_line_item), status: :ok
      else
[...]

PR #1949, opened 2016-10-20
Add explicit serializer not found policy

GregPK:
#### Purpose

Allow users to setup a policy to fail explicity when an `ActiveRecord::Base` inherited model is being used without an explicit `Serializer` defined.

Currently, when rendering a `ActiveRecord::Base` model it falls back on the default JSON serializer, exposing all attributes.
#### Changes
- `ActiveModel::Serializer#get_serializer_for`
#### Related GitHub issues
- https://github.com/rails-api/active_model_serializers/issues/1853

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -80,22 +80,35 @@ module ActiveModel
 
     # @api private
     # Find a serializer from a class and caches the lookup.
-    # Preferentially returns:
-    #   1. class name appended with "Serializer"
-    #   2. try again with superclass, if present
-    #   3. nil
     def self.get_serializer_for(klass)
       return nil unless config.serializer_lookup_enabled
       serializers_cache.fetch_or_store(klass) do
-        # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
-        serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
-
-        if serializer_class
-          serializer_class
-        elsif klass.superclass
-          get_serializer_for(klass.superclass)
-        end
+        get_serializer_class(klass)
       end
+    end
+
+    # @api private
+    # Preferentially returns:
+    #   1. class name appended with "Serializer"
+    #   2. try again with superclass, if present
+    #   3. (optional) runs `on_serializer_not_found` policy if serializer is not found
+    #   4. nil
+    def self.get_serializer_class(klass)
+      # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
+      serializer_class = find_serializer_class(klass)
+
+      require 'pry';binding.pry
+      if serializer_class
+        serializer_class
+      elsif klass.superclass && find_serializer_class(klass.superclass)
+        get_serializer_class(klass.superclass)
+      elsif config.on_serializer_not_found.respond_to?(:call) && klass.respond_to?(:serialized_attributes)
+        config.on_serializer_not_found.call(klass)
+      end
+    end
+
+    def self.find_serializer_class(klass)
+      serializer_lookup_chain_for(klass).lazy.map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
     end
 
     # @api private
--- a/lib/active_model/serializer/concerns/configuration.rb
+++ b/lib/active_model/serializer/concerns/configuration.rb
@@ -10,6 +10,7 @@ module ActiveModel
         config = base.config
         config.collection_serializer = ActiveModel::Serializer::CollectionSerializer
         config.serializer_lookup_enabled = true
+        config.on_serializer_not_found = -> {}
 
         def config.array_serializer=(collection_serializer)
           self.collection_serializer = collection_serializer
--- a/test/serializers/serializer_for_test.rb
+++ b/test/serializers/serializer_for_test.rb
@@ -26,6 +26,50 @@ module ActiveModel
         end
       end
 
+      class SerializerNotFoundTest < ActiveSupport::TestCase
+        class Poro; end
+        class Post < ActiveRecord::Base; end
+        class PostSerializer < ActiveModel::Serializer; end
+        class Comment < ActiveRecord::Base; end
+
+        class C  < ::Model; end
+        class CSerializer < ActiveModel::Serializer; end
+        class B < C; end
+        class A < B; end
+
+        def setup
+          @previous_serializer_not_found_policy = ActiveModelSerializers.config.on_serializer_not_found
+          ActiveModelSerializers.config.on_serializer_not_found = -> (klass) { fail NameError, "Serializer for [#{klass}] not found" }
+        end
+
+        def teardown
+          ActiveModelSerializers.config.on_serializer_not_found = @previous_serializer_not_found_policy
+        end
+
+        test 'serializer not found doesnt affect POROs' do
+          assert_nothing_raised do
+            ActiveModel::Serializer.serializer_for(Poro.new)
+          end
+        end
+
+        test 'serializer not found triggers configured policy' do
+          err = assert_raises(NameError) { ActiveModel::Serializer.serializer_for(Comment.new) }
+          assert_equal(err.message, 'Serializer for [ActiveModel::Serializer::SerializerForTest::SerializerNotFoundTest::Comment] not found')
+        end
+
+        test 'serializer not found doesnt trigger on valid model' do
+          assert_nothing_raised NameError do
+            ActiveModel::Serializer.serializer_for(Post.new)
+          end
+        end
+
+        test 'serializer not found goes to the bottom of the class hierarchy' do
+          assert_equal(ActiveModel::Serializer.serializer_for(C.new), CSerializer)
+          assert_equal(ActiveModel::Serializer.serializer_for(B.new), CSerializer)
+          assert_equal(ActiveModel::Serializer.serializer_for(A.new), CSerializer)
+        end
+      end
+
       class SerializerTest < ActiveSupport::TestCase
         module ResourceNamespace
           class Post    < ::Model; end


lib/active_model/serializer.rb
```diff
@@ -80,21 +80,29 @@ def self.serializers_cache
...
+    #   2. try again with superclass, if present
+    #   3. (optional) runs `on_serializer_not_found` policy if serializer is not found
+    #   4. nil
+    def self.get_serializer_class(klass)
+      # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
+      serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
+
+      if serializer_class
+        serializer_class
+      elsif klass.superclass && ![BasicObject, Object].include?(klass.superclass)
```
NullVoxPopuli: why'd you add to the condition here?

lib/active_model/serializer.rb
```diff
@@ -80,21 +80,29 @@ def self.serializers_cache
...
+        get_serializer_for(klass.superclass)
+      elsif config.on_serializer_not_found.respond_to?(:call) && klass.respond_to?(:serialized_attributes)
+        config.on_serializer_not_found.call(klass)
```
NullVoxPopuli: I'm not sure how I feel about ending with situation 4 (as described in the method comments) as the result of an empty block.

Can you explain why you're checking for serialized_attributes in the elsif here, and what you would do if on_serializer_not_found is nil?

lib/active_model/serializer/concerns/configuration.rb
```diff
@@ -10,6 +10,7 @@ module Configuration
         config = base.config
         config.collection_serializer = ActiveModel::Serializer::CollectionSerializer
         config.serializer_lookup_enabled = true
+        config.on_serializer_not_found = -> {}
```
NullVoxPopuli: what if this is nil?
should it start out as nil?  does defaulting to an empty block feel weird?

test/serializers/serializer_for_test.rb
```diff
@@ -26,6 +26,31 @@ def test_overwritten_serializer_for_array
...
+        def setup
+          @previous_serializer_not_found_policy = ActiveModelSerializers.config.on_serializer_not_found
+          ActiveModelSerializers.config.on_serializer_not_found = -> (klass) { fail NameError, "Serializer for [#{klass}] not found" }
+        end
+
+        def teardown
+          ActiveModelSerializers.config.on_serializer_not_found = @previous_serializer_not_found_policy
+        end
+
+        def test_serializer_not_found_triggers_configured_policy
```
NullVoxPopuli: can you change these to the `test 'serializer not found triggers configured policy'` test format.

Thanks!

lib/active_model/serializer.rb
GregPK: I wanted to get a meaningful error message. Otherwise, we always get `klass == BasicObject` and hence ("Serializer for [BasicObject] not found"), since it's at the bottom of the object hierarchy. 
I agree this is a bit wonky, but I couldn't think of anything better.

NullVoxPopuli: perhaps,

But since we don't *just* deal with ActiveRecord objects, I think this scenario would behave unexpectedly with your code.

2016-10-20-115654_326x111_scrot

lib/active_model/serializer/concerns/configuration.rb
GregPK: See comments above about the effects (TLDR: none). When I set the default here, I thought that maybe someone will be looking at the code here to consider the values for `on_serializer_not_found`. I thought that an empty proc will communicate better that this was created to be (as opposed to `nil` which could be anything).

lib/active_model/serializer.rb
NullVoxPopuli: right, but what is `klass` right now? (this is a huge chance in improve the in-code docs).
 - does it represent a serializable object? (PORO, AR, etc -- none of these need serialized_hash)

So, with your test, what happens if `Comment` were defined as:

```ruby
class Comment; end
```

GregPK: Not sure what you mean - I've added a [test](https://github.com/rails-api/active_model_serializers/pull/1949/commits/d6a287bc0f396084254aac17b3951b321553a11f) to see if this doesn't mess up POROs - please take a look if this is what you're concerned about.

This is of course under the assumption that if an object is not serializable then it should not trigger the policy.

lib/active_model/serializer.rb
```diff
@@ -80,24 +80,36 @@ def self.serializers_cache
...
+      if serializer_class
+        serializer_class
+      elsif klass.superclass && !excluded_base_class?(klass.superclass)
+        get_serializer_for(klass.superclass)
+      elsif config.on_serializer_not_found.respond_to?(:call) && klass.respond_to?(:serialized_attributes)
+        config.on_serializer_not_found.call(klass)
       end
     end
 
+    def self.excluded_base_class?(klass)
```
NullVoxPopuli: this method name doesn't make sense, it has a negative word, yet the condition is testing for the positive

lib/active_model/serializer.rb
```diff
@@ -80,24 +80,36 @@ def self.serializers_cache
...
+    #   2. try again with superclass, if present
+    #   3. (optional) runs `on_serializer_not_found` policy if serializer is not found
+    #   4. nil
+    def self.get_serializer_class(klass)
+      # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
+      serializer_class = serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
+
+      if serializer_class
+        serializer_class
+      elsif klass.superclass && !excluded_base_class?(klass.superclass)
```
NullVoxPopuli: see, where I'm getting caught up on this, is that I'm reading this as:

"if the objec i'm trying to serialize has a superclass, and that superclass is neither ActiveRecord::Base, BasicObject, nor Object, then get the serializer for that ....."

And here is the problem (in my looking things over too quickly -- you're getting the serializer for the superclass, not klass -- this makes sense now -- sorry about the confusion).

So, I guess for clarity, the condition should be extracted to a method -- maybe

`is_superclass_allowed_to_be_serializable?(klass)

but then, why _don't_ we want those superclasses serializable? I know it's weird otherwise, but what if someone wanted to try to write a general serializer for `Object`? would that be super weird? idk.

GregPK: The problem here is that I only need this additional if condition to actually have a side-effect on the next if, which is really why we're coming up against this kind of reasoning issues. I'll try to come up with a better solution.

lib/active_model/serializer.rb
```diff
@@ -80,24 +80,36 @@ def self.serializers_cache
...
+    #   2. try again with superclass, if present
+    #   3. (optional) runs `on_serializer_not_found` policy if serializer is not found
+    #   4. nil
+    def self.get_serializer_class(klass)
+      # NOTE(beauby): When we drop 1.9.3 support we can lazify the map for perfs.
+      serializer_class = find_serializer_class(klass)
+
+      if serializer_class
+        serializer_class
+      elsif klass.superclass && find_serializer_class(klass.superclass)
```
NullVoxPopuli: also, I was jus about to suggest extracting the serializer_lookup to a method. :-)

lib/active_model/serializer.rb
```diff
@@ -80,24 +80,36 @@ def self.serializers_cache
...
+        get_serializer_class(klass.superclass)
+      elsif config.on_serializer_not_found.respond_to?(:call) && klass.respond_to?(:serialized_attributes)
+        config.on_serializer_not_found.call(klass)
       end
```
NullVoxPopuli: what happens if all the conditions are false? what do we do then?

NullVoxPopuli: ;-)

NullVoxPopuli:
I feel MUCH better about this now. so, you get the :+1: from me. 

I'd like one more maintainer to review though -- @bf4, @beauby?

lib/active_model/serializer.rb
```diff
@@ -80,24 +80,36 @@ def self.serializers_cache
...
     end
 
+    def self.find_serializer_class(klass)
+      serializer_lookup_chain_for(klass).map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
```
beauby: Can be lazified, i.e.:
```ruby
serializer_lookup_chain_for(klass).lazy.map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }
```

GregPK: Lazified.

lib/active_model/serializer.rb
NullVoxPopuli: Sounds like we need to check ancestors?

GregPK:
@NullVoxPopuli @beauby I cleaned up the code and took care of the ancestor issues. Please take a look.

NullVoxPopuli:
@GregPK can you fix the rubocop issue real quick?

lib/active_model/serializer.rb
```diff
@@ -80,24 +80,42 @@ def self.serializers_cache
...
+    #   3. (optional) runs `on_serializer_not_found` policy if serializer is not found
+    #   4. nil
+    def self.get_serializer_class(klass)
+      serializer_class = find_serializer_class(klass)
+
+      if serializer_class
+        serializer_class
+      elsif (ancestor_serializer_classes = ancestor_serializer_classes(klass)).present?
+        ancestor_serializer_classes.first
+      elsif config.on_serializer_not_found.respond_to?(:call) && klass.respond_to?(:serialized_attributes)
```
bf4: and this is change 2: a new condition

Since we own this config, we shouldn't need to check if it responds to call. We can either make a default strategy or just check if the config is set.

What is `serialized_attributes` that you are checking for?

GregPK: 1. The config setup you mentioned is probably the best way to go. However, if we setup a default strategy that returns a symbol, we will brake backwards compatibility for anyone relying on the historical behaviour of returning `nil`. 

2. You're right about checking for callability. I think we can actually do `elsif config.on_serializer_not_found` - if it's an empty proc it will run, and when someone messes up the config value (ie setting it to `:no_serializer_found` instead of `-> { :no_serializer_found }` he will get an explicit error about what's wrong with the code. WDYT?

2. For [`:serialized_attributes`](http://apidock.com/rails/ActiveRecord/Base/serialized_attributes/class), I'm checking whether this is actually something that represents a serializable AR resource. I'm aware that this excludes non AR-based objects. Maybe change this to `klass.respond_to?(:serialized_attributes) || klass.respond_to?(:serialized_hash)`? This will handle both cases. I would much rather not couple this to AR via `serialized_attributes` but I could not find any other way of detecting is the object is both AR-based and serializable.

NullVoxPopuli:
needs rebase

mrnugget:
Any updates on this? I'd love to see this get merged instead of having to add the monkey-patch (which is great btw.!) to our codebase.

bf4:
I suppose what I might like is a `strict` mode. i.e., if no serializer is passed explicitly, then raise an error.  This means that it would error if you tried to pass in a hash.  The problem with having a 'default serializer' is that if someone tries to serialize a hash, we have no way of knowing what 'type' it is. This is particularly problematic for JSON API.

so, it would be something like:

```diff
- return nil unless config.serializer_lookup_enabled
+ case config.serializer_lookup_enabled
+ when :strict_mode then fail StrictMode, "Won't lookup serializer for '#{klass}' In Strict mode serializers must be explicit"
+ when false then return nil
+ else nil # continue
+ end
serializers_cache.fetch_or_store(klass) do
```

bf4:
related 
- https://github.com/rails-api/active_model_serializers/issues/1813#issuecomment-227010656

mrnugget:
I thought about changing the approach to this problem, too. I tried by adding a last element to the lookup chain, that raises. The idea being: "if you reached this end of the lookup chain, there's no serializer for you". The only problem is that every element in the chain will be called for every lookup in one `flat_map` call, so it raises every time, even though the lookup may stop earlier.

In order to get the change in this PR working for our Rails 5 codebase, I needed to make a few changes:

```diff
-      elsif klass.superclass && ![BasicObject, Object].include?(klass.superclass)
+      elsif klass.superclass && ![BasicObject, Object, ActiveRecord::Base, ApplicationRecord].include?(klass.superclass)
```

The change was necessary, because `serialized_attributes` was removed in Rails 5. Now the code says: "Okay, by looking for a serializer and walking up the superclasses you've now reached ApplicationRecord/ActiveRecord::Base and so far we found no serializer, that means.. raise an exception".

beauby:
Truth is, relying on serializer inference incurs a massive perf drop, for a few saved keystrokes.

mrnugget:
What I have now running is a slightly modified version of the code in this pull request. Slightly modified in order to make it work with Rails 5.

```diff
-      elsif config.on_serializer_not_found.respond_to?(:call) && klass.respond_to?(:serialized_attributes)
+      elsif config.on_serializer_not_found.respond_to?(:call) && [ActiveRecord::Base, ApplicationRecord].include?(klass.superclass)
      config.on_serializer_not_found.call(klass)

```

The change was necessary, because `serialized_attributes` was removed in Rails 5. Now the code says: "Okay, by looking for a serializer and walking up the superclasses you've now reached ApplicationRecord/ActiveRecord::Base and so far we found no serializer, that means.. raise an exception". It works, without breaking the serialization of `Hash` and `String` for example.

NullVoxPopuli:
> Truth is, relying on serializer inference incurs a massive perf drop, for a few saved keystrokes.

@beauby do we have benchmarks on this?
Also, do we have a solution for explicit -- yet polymorphic -- serializers?

Issue #1954, opened 2016-10-24
has_one association not included in rendered json


prrrnd:
#### Expected behavior vs actual behavior

Parking has_many :cars
Car has_one :driver

```ruby
class ParkingSerializer < ActiveModel::Serializer
  attributes :id, :location
  has_many :cars
end
```

```ruby
class CarSerializer < ActiveModel::Serializer
  attributes :id, :brand
  has_one :driver
end
```

```ruby
class DriverSerializer < ActiveModel::Serializer
  attributes :id, :name
end
```

**Expected** 
A `driver` nested under each `car` in the json when rendering a `parking`

**Result**
json with parking's attributes, cars, cars attributes except `driver`

#### Steps to reproduce

See serializers above

Controller:

```ruby
# ...
def show
  @parking = Parking.find(params[:id])
  render json: @parking
end
# ...
```

#### Environment

ActiveModelSerializers Version 0.10.2
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin15]
macOS Sierra
Rails 4.2.7.1

beauby:
`JsonApi` adapter?

prrrnd:
@beauby I'm using the default one, `:attributes`

groyoh:
@prrrnd try:

```ruby
render json: @parking, include: "*.*"
```

With 0.10, associations are included by default only for the first level. If you want to include deeper associations, you need to specify it with the `include` directive. In this case, `include: "*.*"` means include any association up to 2 levels. You may also use `include: **` to include every level (not recommended) or be more specific `include: "cars.driver"`.

prrrnd:
@groyoh Thanks, I'll try that and report back.
Seems like your explanation could be part of the [documentation](https://github.com/rails-api/active_model_serializers/blob/master/docs/general/rendering.md#include), it's waiting for a PR.

groyoh:
@prrrnd yeah the documentation is definitely lacking some love. If this was helpful for you, maybe you would kind enough to help us out with the doc :smile:

prrrnd:
@groyoh The `include` directive definitely helped. I'll try to update the documentation in a PR.

PR #1958, opened 2016-10-25
Add documentation for the include directive

prrrnd:
#### Purpose

Based on #1954, adds documentation about the `include` directive when rendering.
#### Changes

Add documentation for the `include` directive
#### Caveats

Limited documentation, but still better than nothing - I guess :-)
#### Related GitHub issues
#1954

### DIFF
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -218,7 +218,14 @@ link(:link_name) { url_for(controller: 'controller_name', action: 'index', only_
 
 #### include
 
-PR please :)
+By default, associations are only included for the first level of nesting. If you want to include associations at a deeper level, you need to specify it via the `include` directive.
+
+```ruby
+# Example with a Parking having many Cars, each Car having one Driver
+render json: @parking, include: "cars.driver"
+```
+
+`include: '*.*'` means include any association up to 2 levels. You may also use `include: '**'` to include every level (not recommended) or be more specific and include named associations such as `include: 'cars.driver'`.
 
 #### Overriding the root key
 


docs/general/rendering.md
```diff
@@ -218,7 +218,14 @@ link(:link_name) { url_for(controller: 'controller_name', action: 'index', only_
...
 
-PR please :)
+By default, associations are only included for the first level of nesting. If you want to include associations at a deeper level, you need to specify it via the `include` directive.
+
+```ruby
+# Example with a Parking having many Cars, each Car having one Driver
+render json: @parking, include: "cars.driver"
+```
+
+`include: '*.*'` means include any association up to 2 levels. You may also use `include: '**'` to include every level (not recommended) or be more specific and include named associations such as `include: 'cars.driver'`.
```
bf4: I think there is cross-referenced with docs in adapters.md?

Super important: the `*` is an AMS-specific idiosyncrasy that works now, but may go away in the future.  It's a better practice if you can use explicit includes.

bf4: Should also ref which adapters this is true for. i.e. also doc `render jsonapi: @parking`  Where `@parking = Parking.new(cars: [Car.new(driver)]`. 

Should also ref that this includes all fields by default, but you may want to limit e.g. `fields: { parking: [:car], cars: [:driver]`

dkniffin:
It looks like this documentation is already available [here](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/adapters.md#include-option), and linked to from [here](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/rendering.md#include), so I wonder if this PR is necessary anymore.

PR #1960, opened 2016-10-27
Makes serializer available to resource meta blocks

### LINKED ISSUE #1940: Make instance_options/serializer accessible on "meta" method
Hi!

I have the following serializer(part of the code):

```ruby
class TreatmentSerializer < ActiveModel::Serializer
  def city_filter?
    instance_options[:request_city].present? && instance_options[:request_city]
  end

  has_many :testimonials do |serializer|
    # some code that uses serializer.city_filter?
  end

  meta do |serializer|
    # serializer in this case returns the Meta class instance instead of the serializer itself
    {
      question_count: serializer.city_filter? ? object.treatment_questions.where(city_id: instance_options[:request_city]).size : object.treatment_questions.size
    }
  end
end
```

I need to show the **question_count** based on the city of the users request, but I can't since neither the *serializer* or the *instance_options* are available on **meta** method.

Is there any sort of workaround for this now? And maybe this feature might be included in a future release?

Oh, and I'm using **json_api** adapter format.

Thanks!

groyoh:
#### Purpose

Resolves #1940.

It is now possible to access the serializer from the
resource meta block e.g.

``` ruby
class FooSerializer
  meta do |serializer|
    { foo: serializer.foo! }
  end

  def foo!
    "foo!"
  end
end
```
#### Changes

Pass serializer to the meta block.
#### Caveats

This does not provide `instance_options` to the `meta` block as `instance_options` are protected attributes.
#### Related GitHub issues
#1940

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/meta.rb
+++ b/lib/active_model_serializers/adapter/json_api/meta.rb
@@ -18,7 +18,7 @@ module ActiveModelSerializers
 
           # Use the return value of the block unless it is nil.
           if serializer._meta.respond_to?(:call)
-            @value = instance_eval(&serializer._meta)
+            @value = instance_exec(serializer, &serializer._meta)
           else
             @value = serializer._meta
           end
--- a/test/adapter/json_api/resource_meta_test.rb
+++ b/test/adapter/json_api/resource_meta_test.rb
@@ -4,16 +4,29 @@ module ActiveModel
   class Serializer
     module Adapter
       class JsonApi
-        class ResourceMetaTest < Minitest::Test
+        class ResourceMetaTest < ActiveSupport::TestCase
           class MetaHashPostSerializer < ActiveModel::Serializer
             attributes :id
-            meta stuff: 'value'
+            meta cache: { expires: 3600 }
           end
 
           class MetaBlockPostSerializer < ActiveModel::Serializer
             attributes :id
             meta do
               { comments_count: object.comments.count }
+            end
+          end
+
+          class MetaBlockParamPostSerializer < ActiveModel::Serializer
+            attributes :id
+            meta do |serializer|
+              {
+                actions: serializer.actions
+              }
+            end
+
+            def actions
+              instance_options[:actions] || []
             end
           end
 
@@ -31,23 +44,23 @@ module ActiveModel
             end
           end
 
-          def setup
+          setup do
             @post = Post.new(id: 1337, comments: [], author: nil)
           end
 
-          def test_meta_hash_object_resource
+          test 'resource meta as hash' do
             hash = ActiveModelSerializers::SerializableResource.new(
               @post,
               serializer: MetaHashPostSerializer,
               adapter: :json_api
             ).serializable_hash
             expected = {
-              stuff: 'value'
+              cache: { expires: 3600 }
             }
             assert_equal(expected, hash[:data][:meta])
           end
 
-          def test_meta_block_object_resource
+          test 'resource meta as block' do
             hash = ActiveModelSerializers::SerializableResource.new(
               @post,
               serializer: MetaBlockPostSerializer,
@@ -59,7 +72,7 @@ module ActiveModel
             assert_equal(expected, hash[:data][:meta])
           end
 
-          def test_meta_object_resource_in_array
+          test 'resource meta within collections' do
             post2 = Post.new(id: 1339, comments: [Comment.new])
             posts = [@post, post2]
             hash = ActiveModelSerializers::SerializableResource.new(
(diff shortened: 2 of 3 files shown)


bf4:
@groyoh is it not already? (without using the block argument?)

groyoh:
@bf4 unfortunately no. Have a look at https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/adapter/json_api/meta.rb#L33. We have access to `scope` and `object` but nothing else. There are a lot of inconsistencies like this throughout the API though.

lib/active_model_serializers/adapter/json_api/meta.rb
```diff
@@ -18,7 +18,7 @@ def initialize(serializer)
 
           # Use the return value of the block unless it is nil.
           if serializer._meta.respond_to?(:call)
-            @value = instance_eval(&serializer._meta)
```
NullVoxPopuli: as a side note, should meta inherit from the same struct we're using elsewhere? would that clean up some of this code?

groyoh:
As a side note, I actually think that we should update all the `instance_eval` to `instance_exec(serializer)` and align what methods (`object`, `scope`, etc.) are available.

lib/active_model_serializers/adapter/json_api/meta.rb
NullVoxPopuli: lib/active_model/serializer/field.rb.

Attribute, and Reflection inherit from it.

groyoh: Not sure how it would benefit from inheriting from `Field`. But I can definitely see a common behavior that could be reuse with:
* the block `instance_exec`
* the `:if`, `:unless` conditionals

I guess I could try something out in a different PR and see if I can come up with something clean.

NullVoxPopuli: Yeah, a different PR would probably be better, as *this* pr solves the immediate issue, and what I'm suggesting is more structural.

test/adapter/json_api/resource_meta_test.rb
```diff
@@ -93,6 +104,24 @@ def test_meta_object_empty_string_omitted
             ).serializable_hash
             refute hash[:data].key? :meta
           end
+
+          def test_meta_block_with_parameter
```
NullVoxPopuli: can you use `test 'meta block with parameter' do` instead of the `def test_` syntax? Trying to move towards the more commonly preferred syntax. :-)

groyoh:
@NullVoxPopuli I reworked the tests according to your comment :wink:

NullVoxPopuli:
I approve. As this is a bit of an API change, @bf4?

bf4:
In Grand Cayman now. Can review in a week

B mobile phone

groyoh:
@bf4 then please stop reading your mails and enjoy your holidays (?) :wink:

test/adapter/json_api/resource_meta_test.rb
```diff
@@ -31,11 +31,22 @@ class MetaBlockPostEmptyStringSerializer < ActiveModel::Serializer
             end
           end
 
-          def setup
+          class MetaBlockWithParameterSerializer < ActiveModel::Serializer
+            attributes :id
+            meta do |serializer|
+              { present: serializer.present? }
```
groyoh: I'm not sure what you are talking about. As far as I can see, in the reflection we `instance_exec` it the same way without any serializer method (https://github.com/rails-api/active_model_serializers/blob/ec045a5388a983f45a53726b5f9dd4b3ddbf0abd/lib/active_model/serializer/reflection.rb#L76). Maybe there is something I'm missing?

test/adapter/json_api/resource_meta_test.rb
```diff
@@ -31,11 +31,22 @@ class MetaBlockPostEmptyStringSerializer < ActiveModel::Serializer
...
+            end
+
+            def present?
+              true
```
groyoh: Will do.

bf4:
Looks like I should have 'iirc', was thinking of the private attr readers, but there was discussion there in that instance_exec, might be worth digging up

B mobile phone

groyoh:
I think the idea was to use `instance_exec` and yield the `serializer` so that it makes the call to the serializer explicit rather than using "some hidden methods that no one knows about". Doing it in such way could also allow us to align the usage of the DSL:

1. Yielding the serializer so that it is explicit
2. Calling helper methods only via the serializer's public interface

```ruby
class UserSerializer
  meta do |serializer|
    { actions: serializer.actions } # everything goes through the serializer's public API
  end

  link :self do |serializer|
    href serializer.url_for(:self)  # everything goes through the serializer's public API
  end

  has_many :posts do |serializer, association_serializer|
    # you get the idea
  end

  def actions # Public method
    # Some actions
  end

  def url_for(type) # Public method
    # A url
  end
end
```

To me this seems less error-prone and easier to understand than having a few methods available within the block.

PR #1962, opened 2016-10-28
Update key_transforms.md

yagudaev:
fix reference error in docs

### DIFF
--- a/docs/general/key_transforms.md
+++ b/docs/general/key_transforms.md
@@ -26,9 +26,9 @@ Key translation precedence is as follows:
 
 ##### Configuration option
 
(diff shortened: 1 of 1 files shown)


beauby:
Not an error (https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers.rb#L21).

yagudaev:
@beauby thanks. `ActiveModel::Serializer` seemed to be the way to go with earlier versions of `0.10.0.x`. Is that something that changed? Or just something more rails specific to go through `ActiveModel`?

beauby:
@yagudaev Yeah it has changes. The idea was to unburden the `ActiveModel::Serializer` class from library concerns.

PR #1968, opened 2016-11-08
This adds namespace lookup to serializer_for

NullVoxPopuli:
#### Purpose

To Provide the ability to namespace serializers in the same fashion that controllers are namespaced -- as it is very common to add `Api::V#` to both controllers and serializers.

#### Changes

Pass the controllers namespace through serializer_options and serializer_for

#### Caveats

Namespaced lookup is first, given that `namespace` is passed in to the serializable resource options.

#### Related GitHub issues

Subset of the work in #1757

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ cache:
 
 script:
   - bundle exec rake ci
-
+after_success:
+  - codeclimate-test-reporter
 env:
   global:
     - "JRUBY_OPTS='--dev -J-Xmx1024M --debug'"
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -250,7 +250,7 @@ The namespace for serializer lookup is based on the controller.
 To configure the implicit namespace, in your controller, create a before filter
 
 ```ruby
-before_filter do
+before_action do
   self.namespace_for_serializer = Api::V2
 end
 ```
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -266,11 +266,12 @@ render json: @post, namespace: Api::V2
 
 This tells the serializer lookup to check for the existence of `Api::V2::PostSerializer`, and if any relations are rendered with `@post`, they will also utilize the `Api::V2` namespace.  
 
-The namespace can _only_ be in one of the following formats:
+The `namespace` can be any object whose namespace can be represented by string interpolation (i.e. by calling to_s)
+- Module `Api::V2`
+- String `'Api::V2'`
+- Symbol `:'Api::V2'`
 
-- `Api::V2`
-- `'Api::V2'`
-- `:'Api::V2'`
+Note that by using a string and symbol, Ruby will assume the namespace is defined at the top level.
 
 
 #### serializer
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -250,8 +250,7 @@ The namespace for serializer lookup is based on the controller.
 To configure the implicit namespace, in your controller, create a before filter
 
 ```ruby
-before_filter :use_my_namespace
-def use_my_namespace
+before_filter do
   self.namespace_for_serializer = Api::V2
 end
 ```
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -245,13 +245,14 @@ Note: the `Attributes` adapter (default) does not include a resource root. You a
 
 #### namespace
 
-The namespace for serializer lookup is based on the controller, via `ActionController::Serialization#namespace_for_serializer`.
+The namespace for serializer lookup is based on the controller.
 
-To configure the implicit namespace, in your controller, define
+To configure the implicit namespace, in your controller, create a before filter
 
 ```ruby
-def namespace_for_serializer
-  @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+before_filter :use_my_namespace
+def use_my_namespace
+  self.namespace_for_serializer = Api::V2
 end
 ```
 
@@ -259,10 +260,17 @@ end
 
 
 ```ruby
+@post = Post.first
 render json: @post, namespace: Api::V2
 ```
 
-This tells the serializer lookup to check for the existence of `Api::V2::PostSerializer`, and if any relations are rendered with `@post`, they will also utilize the `Api::V2` namespace.
+This tells the serializer lookup to check for the existence of `Api::V2::PostSerializer`, and if any relations are rendered with `@post`, they will also utilize the `Api::V2` namespace.  
+
+The namespace can _only_ be in one of the following formats:
+
+- `Api::V2`
+- `'Api::V2'`
+- `:'Api::V2'`
 
 
 #### serializer
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -243,6 +243,28 @@ This will be rendered as:
 ```
 Note: the `Attributes` adapter (default) does not include a resource root. You also will not be able to create a single top-level root if you are using the :json_api adapter.
 
+#### namespace
+
+The namespace for serializer lookup is based on the controller, via `ActionController::Serialization#namespace_for_serializer`.
+
+To configure the implicit namespace, in your controller, define
+
+```ruby
+def namespace_for_serializer
+  @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+end
+```
+
+`namespace` can also be passed in as a render option:
+
+
+```ruby
+render json: @post, namespace: Api::V2
+```
+
+This tells the serializer lookup to check for the existence of `Api::V2::PostSerializer`, and if any relations are rendered with `@post`, they will also utilize the `Api::V2` namespace.
+
+
 #### serializer
 
 PR please :)
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -18,6 +18,10 @@ module ActionController
       self._serialization_scope = :current_user
     end
 
+    def namespace_for_serializer
+      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+    end
+
     def serialization_scope
       return unless _serialization_scope && respond_to?(_serialization_scope, true)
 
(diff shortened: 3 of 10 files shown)


lib/active_model_serializers/serializable_resource.rb
```diff
@@ -55,7 +55,7 @@ def serializer
       @serializer ||=
         begin
           @serializer = serializer_opts.delete(:serializer)
-          @serializer ||= ActiveModel::Serializer.serializer_for(resource)
+          @serializer ||= ActiveModel::Serializer.serializer_for(resource, serializer_opts)
```
bf4: namespace?

lib/active_model/serializer.rb
```diff
@@ -59,13 +59,14 @@ class << self
...
     # @api private
-    def self.serializer_lookup_chain_for(klass)
+    def self.serializer_lookup_chain_for(klass, namespace = nil)
       chain = []
 
       resource_class_name = klass.name.demodulize
       resource_namespace = klass.name.deconstantize
       serializer_class_name = "#{resource_class_name}Serializer"
 
+      chain.push("#{namespace.name}::#{serializer_class_name}") if namespace
```
bf4: Is `namespace.name` necessary? Seems it implies namespace is a class, whereas, if it is, when interpolating, Ruby will do the right thing whether it's a string, symbol, or class... i.e just `namespace`

bf4:
@NullVoxPopuli what is the `test/feature/configuration/serializer_lookup/in_controllers_test.rb` in the diff?

bf4:
Where do you think docs should go? Any sense of where people would look for this?

NullVoxPopuli:
> @NullVoxPopuli what is the test/feature/configuration/serializer_lookup/in_controllers_test.rb in the diff?

that's a good question -- must have gotten carried over from the serializer-lookup branch -- which I also don't remember editing that file.

lib/action_controller/serialization.rb
```diff
@@ -18,6 +18,10 @@ def serialization_scope(scope)
       self._serialization_scope = :current_user
     end
 
+    def namespace_for_serializer
+      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+    end
```
NullVoxPopuli: I'm sure people would appreciate it. And I don't think it would be a problem to support, as tons of people want namespaced-scoped serializer lookup -- but maybe that'd be better for the lookup customization PR?

NullVoxPopuli:
> Where do you think docs should go? Any sense of where people would look for this?

I'd say under rendering -- I'll add something real quick

.travis.yml
```diff
@@ -20,7 +20,8 @@ cache:
 
 script:
   - bundle exec rake ci
-
+after_success:
+  - codeclimate-test-reporter
```
bf4: Is this working?

docs/general/rendering.md
```diff
@@ -243,6 +243,28 @@ This will be rendered as:
 ```
 Note: the `Attributes` adapter (default) does not include a resource root. You also will not be able to create a single top-level root if you are using the :json_api adapter.
 
+#### namespace
+
+The namespace for serializer lookup is based on the controller, via `ActionController::Serialization#namespace_for_serializer`.
+
+To configure the implicit namespace, in your controller, define
+
```
bf4: > replace the implementation of `namespace_for_serializer` 
> or pass in the render option `:namespace`

?

docs/general/rendering.md
```diff
@@ -243,6 +243,28 @@ This will be rendered as:
...
+def namespace_for_serializer
+  @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+end
+```
+
+`namespace` can also be passed in as a render option:
+
+
+```ruby
+render json: @post, namespace: Api::V2
```
bf4: > or `namespace: 'Api::V2'` or `namespace: :'Api::V2'` but not `namespace: :api_v2'`

docs/general/rendering.md
bf4: i.e. below is the current implementation.  ooh,, this would be a good reason to add a setter

```
attr_writer :namespace_for_serializer
```

makes it cleaner and can do in a before filter ?

lib/action_controller/serialization.rb
bf4: > Once we add it, we need to support it.

:)

.travis.yml
NullVoxPopuli: It doesn't error -- I'm not familiar enough with code climate to see if the results are published per-pull-request. But it works locally

lib/active_model_serializers/serializable_resource.rb
bf4: 👍

test/serializers/serializer_for_with_namespace_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+module ActiveModel
+  class Serializer
+    class SerializerForWithNamespaceTest < ActiveSupport::TestCase
+      class Book < ::Model; end
+      class Page < ::Model; end
+      class Publisher < ::Model; end
+
+      module Api
```
bf4: can save space by 

```ruby
module Api; end
module Api::V3; end
class Api::V3::BookSerializer < ApplicationSerializer
  # etc.
```

lib/active_model/serializer/reflection.rb
```diff
@@ -106,6 +106,10 @@ def value(serializer, include_slice)
       #
       def build_association(parent_serializer, parent_serializer_options, include_slice = {})
         reflection_options = options.dup
+
+        # Pass the parent's namespace onto the child serializer
+        reflection_options[:namespace] ||= parent_serializer_options[:namespace]
```
NullVoxPopuli: nah, without coping the namespace from the parent to the reflection_options, the namespace isn't passed to all the children

test/serializers/serializer_for_with_namespace_test.rb
```diff
@@ -0,0 +1,85 @@
...
+      #   attributes :title, :author_name
+      # end
+      # test 'resource without a namespace' do
+      #   book = Book.new(title: 'A Post', author_name: 'hello')
+      #
+      #   result = ActiveModelSerializers::SerializableResource.new(book).serializable_hash
+      #
+      #   expected = { title: 'A Post', author_name: 'hello' }
+      #   assert_equal expected, result
+      # end
```
bf4: looks about right except for the `::`

test/serializers/serializer_for_with_namespace_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+module ActiveModel
```
bf4: looks good. needs controller test for implicit lookup as well as option passing, which is really covered below, here

NullVoxPopuli: :+1:

test/serializers/serializer_for_with_namespace_test.rb
NullVoxPopuli: The problem I was having with this test is that I couldn't get a root level BookSerializer to be used if a namespace wasn't present.

docs/general/rendering.md
```diff
@@ -243,6 +243,36 @@ This will be rendered as:
...
+
+The namespace for serializer lookup is based on the controller.
+
+To configure the implicit namespace, in your controller, create a before filter
+
+```ruby
+before_filter :use_my_namespace
+def use_my_namespace
+  self.namespace_for_serializer = Api::V2
+end
```
bf4: might as well just `before_action {  self.namespace_for_serializer = Api::V2 }`  (before_filter is such old news :( :))

NullVoxPopuli: excellent!

docs/general/rendering.md
```diff
@@ -243,6 +243,36 @@ This will be rendered as:
...
+
+
+```ruby
+@post = Post.first
+render json: @post, namespace: Api::V2
+```
+
+This tells the serializer lookup to check for the existence of `Api::V2::PostSerializer`, and if any relations are rendered with `@post`, they will also utilize the `Api::V2` namespace.  
+
+The namespace can _only_ be in one of the following formats:
```
bf4: Any object whose namespace can be represented by string interpolation (i.e. by calling`to_s`)

- Module: `Api::V2`
- String: `'Api::V2'`
- Symbol: `:'Api::V2'`

lib/action_controller/serialization.rb
```diff
@@ -16,6 +16,12 @@ def serialization_scope(scope)
     included do
       class_attribute :_serialization_scope
       self._serialization_scope = :current_user
+
+      attr_writer :namespace_for_serializer
+    end
+
+    def namespace_for_serializer
+      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
```
bf4: 👍

lib/active_model/serializer.rb
```diff
@@ -84,11 +85,12 @@ def self.serializers_cache
     #   1. class name appended with "Serializer"
     #   2. try again with superclass, if present
     #   3. nil
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(klass, namespace = nil)
       return nil unless config.serializer_lookup_enabled
-      serializers_cache.fetch_or_store(klass) do
+      serializers_cache.fetch_or_store("#{klass}-#{namespace}") do
```
bf4: this will look weird in some case.

```ruby
ActiveSupport::Cache.expand_cache_key(klass, namespace)
```

also used to be in AMS...

NullVoxPopuli: cool, thanks

lib/active_model/serializer.rb
```diff
@@ -84,11 +85,14 @@ def self.serializers_cache
     #   1. class name appended with "Serializer"
     #   2. try again with superclass, if present
     #   3. nil
-    def self.get_serializer_for(klass)
+    def self.get_serializer_for(klass, namespace = nil)
       return nil unless config.serializer_lookup_enabled
-      serializers_cache.fetch_or_store(klass) do
+
+      cache_key = ActiveSupport::Cache.expand_cache_key(klass, namespace)
```
bf4: for now, is ok, but really needs to be in  https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model/serializer/concerns/caching.rb

bf4: https://github.com/rails-api/active_model_serializers/blob/v0.10.2/lib/active_model/serializer/caching.rb wow we're behind on releases

NullVoxPopuli: :-(

NullVoxPopuli: cache updates and namespacing / configurable lookup chain sounds like a good set of updates for a release? ;-)

PR #1969, opened 2016-11-08
Add object equality test for model serializers

semenyukdmitriy:
#### Purpose
Make `SomeSerializer.new(some_object)` == `SomeSerializer.new(some_object)`

#### Changes
Add `ActiveModel::Serializer#==` that checks if `object` is the same.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -130,6 +130,10 @@ module ActiveModel
       define_singleton_method scope_name, -> { scope }
     end
 
+    def ==(other)
+      other.respond_to?(:object) && other.object == object
+    end
+
     def success?
       true
     end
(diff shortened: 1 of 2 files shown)


beauby:
Could you point to an example where this would be desired?

semenyukdmitriy:
@beauby I needed it when I tested a certain code to make sure it returns correct set of serializers like this:

```ruby
proper_result = [MySerializer(johns_post), MySerializer(jacks_post)]
expect(Posts.recent.decorate).to eq(proper_result)
```

beauby:
Wouldn't it be possible to have your code return an array of serializer classes instead?

bf4:
The way to do this is to include Comparable and define <=>

In any case, you'd need to take options into account as well

However, there's not really a natural case for sorting a renderee

PR #1976, opened 2016-11-16
make serialization_context available in link

f-mer:
#### Purpose
```ruby
module MyEngine
  class Engine < Rails::Engine
    isolate_namespace MyEngine
  end
end
```

```ruby
class MySerializer < ActiveModel::Serializer
  link :self do
    context.url_helpers.api_resource_url(object)
  end
end
```

```ruby
module MyEngine
  class MyController < ApplicationController
    def index
      render json: @resources, url_helpers: my_engine
    end
  end
end
```

Makes it possible to build urls of engines with isolated namespace by passing the `ActionDispatch::Routing::RoutesProxy` (`my_engine`).

#### Changes
Pass serialization_context to link.

### DIFF
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -467,7 +467,7 @@ module ActiveModelSerializers
       #   }.reject! {|_,v| v.nil? }
       def links_for(serializer)
         serializer._links.each_with_object({}) do |(name, value), hash|
-          result = Link.new(serializer, value).as_json
+          result = Link.new(serializer, value, instance_options).as_json
           hash[name] = result if result
         end
       end
--- a/lib/active_model_serializers/adapter/json_api/link.rb
+++ b/lib/active_model_serializers/adapter/json_api/link.rb
@@ -41,11 +41,12 @@ module ActiveModelSerializers
       class Link
         include SerializationContext::UrlHelpers
 
-        def initialize(serializer, value)
+        def initialize(serializer, value, adapter_options)
           @_routes ||= nil # handles warning
           # actionpack-4.0.13/lib/action_dispatch/routing/route_set.rb:417: warning: instance variable @_routes not initialized
           @object = serializer.object
           @scope = serializer.scope
+          @context = adapter_options[:serialization_context]
           # Use the return value of the block unless it is nil.
           if value.respond_to?(:call)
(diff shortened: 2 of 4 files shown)


lib/active_model_serializers/adapter/json_api/link.rb
```diff
@@ -41,11 +41,12 @@ class JsonApi
       class Link
         include SerializationContext::UrlHelpers
 
-        def initialize(serializer, value)
+        def initialize(serializer, value, adapter_options)
           @_routes ||= nil # handles warning
           # actionpack-4.0.13/lib/action_dispatch/routing/route_set.rb:417: warning: instance variable @_routes not initialized
           @object = serializer.object
           @scope = serializer.scope
+          @context = adapter_options[:serialization_context]
```
bf4: any reason you can't use the scope?

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -467,7 +467,7 @@ def relationships_for(serializer, requested_associations, include_slice)
       #   }.reject! {|_,v| v.nil? }
       def links_for(serializer)
         serializer._links.each_with_object({}) do |(name, value), hash|
-          result = Link.new(serializer, value).as_json
+          result = Link.new(serializer, value, instance_options).as_json
```
bf4: should probably mirror usage in pagination links?

- https://github.com/f-mer/active_model_serializers/blob/a6a8f53cadc1205b19f0ead41ab5c3f2d03bed10/lib/active_model_serializers/adapter/json_api/pagination_links.rb
- https://github.com/f-mer/active_model_serializers/blob/a6a8f53cadc1205b19f0ead41ab5c3f2d03bed10/lib/active_model_serializers/adapter/json_api.rb#L507

f-mer:
Thanks for getting back at this! Yep that was the idea.
While exploring the problem it may be possible to just pass the `view_context`:

```ruby
render json: @resources, url_helpers: view_context
```

Links could be build like this:

```ruby
link :self do
  scope.api_resource_url(object)
end
```

and the `current_user` could also be accessed via `scope.current_user`.
This would make the change obsolet. Have not tested this yet.

bf4:
@f-mer how's it going?  Have you seen #1981

f-mer:
Closing this pull request because `view_context` could be used.
Thank you heaps!

Issue #1971, opened 2016-11-11
url_helpers non-functional in links method?


michael-reeves:
#### Expected behavior vs actual behavior
I was trying to implement the links behavior described in how_to/add_relationship_links.md

When I implement def links as described I get 
`undefined method 'api_v1_entity_url'`

Same error when using `api_v1_entity_path`

#### Steps to reproduce
My serializer example:
```ruby
class IpAddressSerializer < ActiveModel::Serializer
  # type :ip_address

  attributes :id, :address, :gateway, :netmask, :ip_use, :ip_type, :links

  def links
    {
      entity: api_v1_entity_url(object.entity.name)
    }
  end
  # link(:entity) { api_v1_entity_url(object.entity.name) }
  

  def ip_use
    object.ip_use.name
  end

  def ip_type
    object.ip_type.name
  end
end
```

If I remove the links attr and uncomment link(:entity) piece, I get correct response.
```json
{
    "data": {
        "id": "1",
        "type": "ip-addresses",
        "attributes": {
            "address": "192.168.10.10",
            "gateway": "192.168.10.1",
            "netmask": "24",
            "ip-use": "IPMI",
            "ip-type": "IPv4"
        },
        "links": {
             "entity": "http://localhost/api/v1/entities/server1"
        }
    }
}
```
#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 0.10.2

Output of `ruby -e "puts RUBY_DESCRIPTION"`: 
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]

OS Type & Version: OS X 10.11.6

Integrated application and version *(e.g., Rails, Grape, etc)*: 
Rails 4.2.6

#### Backtrace

```sh
Completed 500 Internal Server Error in 51ms (ActiveRecord: 1.4ms)

NoMethodError - undefined method `api_v1_entity_url' for #<IpAddressSerializer:0x007fe3028e40a8>:
  app/serializers/ip_address_serializer.rb:8:in `links'
  active_model_serializers (0.10.2) lib/active_model/serializer.rb:195:in `read_attribute_for_serialization'
  active_model_serializers (0.10.2) lib/active_model/serializer/field.rb:23:in `value'
  active_model_serializers (0.10.2) lib/active_model/serializer/attributes.rb:22:in `block in attributes'
  active_model_serializers (0.10.2) lib/active_model/serializer/attributes.rb:19:in `attributes'
[...]

PR #1981, opened 2016-11-20
Fix relationship links doc

groyoh:
#### Purpose
The relationship links contained incorrect information that mislead users in using AMS improperly. More precisely, JSONAPI does not allow an attribute to be named `links`. This documentation was fixed and reworked at certain parts to match the latest API.

#### Changes

Only documentation.

#### Related GitHub issues

#1971 
#1935

I haven't reviewed everything yet but I wanted some feedback to see if I'm going in the right direction with theses changes.

@michael-reeves if you have some time, maybe you could also tell me if this documentation is clearer or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Fixes:
 
 Misc:
 
+- [#1981](https://github.com/rails-api/active_model_serializers/pull/1981) Fix relationship link documentation. (@groyoh)
 - [#1984](https://github.com/rails-api/active_model_serializers/pull/1984) Make test attributes explicit. Test models have 'associations' support. (@bf4)
 - [#1993](https://github.com/rails-api/active_model_serializers/pull/1993) Swap out KeyTransform for CaseTransform gem for the possibility of native extension use (@NullVoxPopuli)
 
--- a/docs/howto/add_relationship_links.md
+++ b/docs/howto/add_relationship_links.md
@@ -15,40 +15,42 @@ class Api::V1::UsersController < ApplicationController
       serializer: Api::V1::UserSerializer,
       include: []
   end
+end
 ```
 
 Bear in mind though that ActiveModelSerializers are [framework-agnostic](outside_controller_use.md), Rails is just a common example here.
 
 ### Links as an attribute of a resource
-**This is applicable to JSONAPI, JSON and Attributes adapters**
+**This is applicable to JSON and Attributes adapters**
 
 You can define an attribute in the resource, named `links`.
 
 ```ruby
 class Api::V1::UserSerializer < ActiveModel::Serializer
-  attributes :id, :name, :links
+  include Rails.application.routes.url_helpers
 
-  def links
+  attributes :id, :name
+
+  attribute :links do
+    id = object.id
     {
-      self: api_v1_user_path(object.id),
-      microposts: api_v1_microposts_path(user_id: object.id)
+      self: api_v1_user_path(id),
+      microposts: api_v1_microposts_path(user_id: id)
     }
   end
 end
 ```
 
-This will result in (example is in JSONAPI adapter):
+Using the `JSON` adapter, this will result in:
+
 ```json
 {
-  "data": {
+  "user": {
     "id": "1",
-    "type": "users",
-    "attributes": {
-      "name": "Example User",
-      "links": {
-        "self": "/api/v1/users/1",
-        "microposts": "/api/v1/microposts?user_id=1"
-      }
+    "name": "John",
+    "links": {
+      "self": "/api/v1/users/1",
+      "microposts": "/api/v1/microposts?user_id=1"
     }
   }
 }
@@ -58,7 +60,7 @@ This will result in (example is in JSONAPI adapter):
 ### Links as a property of the resource definiton
 **This is only applicable to JSONAPI adapter**
 
-You can use the `links` class method to define the links you need in the resource's primary data.
+You can use the `link` class method to define the links you need in the resource's primary data.
 
 ```ruby
 class Api::V1::UserSerializer < ActiveModel::Serializer
@@ -69,7 +71,8 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 end
 ```
 
-This will result in (example is in JSONAPI adapter):
+Using the `JSONAPI` adapter, this will result in:
+
 ```json
 {
   "data": {
@@ -104,12 +107,12 @@ class Api::V1::UserSerializer < ActiveModel::Serializer
 
   has_many :microposts, serializer: Api::V1::MicropostSerializer do
     link(:related) { api_v1_microposts_path(user_id: object.id) }
-  end
 
-  #this is needed to avoid n+1, gem core devs are working to remove this necessity
-  #more on: https://github.com/rails-api/active_model_serializers/issues/1325
-  def microposts
-    object.microposts.loaded ? object.microposts : object.microposts.none
+    microposts = object.microposts
+    # The following code is needed to avoid n+1 queries.
+    # Core devs are working to remove this necessity.
+    # See: https://github.com/rails-api/active_model_serializers/issues/1325
+    microposts.loaded? ? microposts : microposts.none
   end
 end
 ```


michael-reeves:
@groyoh 
Just my 2 cents, but the usage described here is very confusing to me.

```ruby 
user = User.new(id: 1, name: "John")
resource = ActiveModelSerializer::SerializableResource.new(user, adapter: :json)
resource.to_json
```

This code (or something similar, sometimes with meta: ) appears in several spots.  But it's not entirely obvious where that code belongs in a Rails app.  In these docs, it's just hanging out outside of the Serializer class.

Is it supposed to be inside the model, or controller?  It's not entirely obvious.  I think being as specific as possible keeps it clearer for newbs like me.

In my case, I put the adapter in a active_model_serializer initialization file.

docs/howto/add_link.md
```diff
@@ -0,0 +1,337 @@
...
+    "self": "http://example.com/articles?page[number]=3&page[size]=1",
+    "first": "http://example.com/articles?page[number]=1&page[size]=1",
+    "prev": "http://example.com/articles?page[number]=2&page[size]=1",
+    "next": "http://example.com/articles?page[number]=4&page[size]=1",
+    "last": "http://example.com/articles?page[number]=13&page[size]=1"
+  }
+}
+```
+
+Note that `ActiveModelSerializers` pagination relies on a collection that has the methods `current_page`, `total_pages`, and `size`, such as are supported by both [Kaminari](https://github.com/amatsuda/kaminari) or [WillPaginate](https://github.com/mislav/will_paginate). If you want to have paginated links your collections but don't want to use neither Kaminari nor WillPaginate, make sure to define these methods.
```
bf4: Could use note on how to opt-out of this AMS behavior

docs/howto/handle_pagination.md
```diff
@@ -0,0 +1,75 @@
...
+users = [User.new(id: 1, name: "Kelly")]
+meta = {
+  page: 1,
+  total_pages: 1
+}
+resource = ActiveModelSerializers::SerializableResource.new(users, adapter: :json_api, meta: meta)
+resource.to_json
+```
+
+* For usage of top-level links see [add_links.md#pagination].
```
bf4: link missing `()`

bf4:
@michael-reeves The meaning and usage of `ActiveModel::Serializer` has changed over time.  Whereas it was once the primary object in AMS, it no longer is.  

```ruby
user = User.new(id: 1, name: "John")
resource = ActiveModelSerializer::SerializableResource.new(user, adapter: :json)
resource.to_json
```

does exactly what it appears to:  given a `user`, you can use AMS to serialize it JSON by using `SerializableResource`.  `SerializableResource` is the same object used in the controller. The equivalent controller code would be

```ruby
user = User.new(id: 1, name: "John")
render json: user, adapter: :json
```

Of course, if the default adapter were JSON, then the adapter option would not be required.

In addition, the code assumes there is a `UserSerializer` class defined.

I'd love if you could open a new issue or PR helping us improve the docs wherever they may be unclear.  (And yes, I agree there is plenty of room for improvement).

bf4:
@groyoh looks great!

docs/howto/add_link.md
groyoh: Oh true. I did not think about that. Will update it later today.

docs/howto/handle_pagination.md
groyoh: Will fix

groyoh:
@michael-reeves AMS is not framework specific. Therefore it can be used outside of Rails (in my previous job we did not use it with rails at all). That is why I tried to provide a rails version and a non rails version when I could. That can indeed be confusing but that's the AMS API. Maybe i could specify at the top that examples are given for both rails and non rails. Would that make it more clear? Also i specifed the adapter option because I described many adapter within the same file. Therefore if someone wants to copy paste the code, they do not need extra setup.

michael-reeves:
@groyoh 
That's completely fair.  I thought about it not being Rails specific, but figured it was worth a mention.   

I guess I didn't see the Rails specific piece, other than the pagination stuff.  And that was super useful, since I'm about to get into that code, so thanks for that.

And it's very helpful that you had code for different adapters.  Thanks.

I do think the example @bf4 put in his comment to me above helped clear it up quite a bit.

So, your call if you want to add something like that.

groyoh:
@bf4 I updated the PR based on your comments

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ Fixes:
 - [#1881](https://github.com/rails-api/active_model_serializers/pull/1881) ActiveModelSerializers::Model correctly works with string keys (@yevhene)
 
 Misc:
+
+- [#1981](https://github.com/rails-api/active_model_serializers/pull/1981) Improve link and pagination documentation(@groyoh)
```
bf4: `s/documentation(@/documentation. (@`

docs/README.md
```diff
@@ -22,10 +22,10 @@ This is the documentation of ActiveModelSerializers, it's focused on the **0.10.
 
 ## How to
 
+- [How to use ActiveModelSerializers outside of rails controllers](howto/outside_controller_use.md)
```
bf4: `s/rails/Rails`

> How ActiveModelSerializers Outside Of Rails Controllers

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
+[Back to Guides](../README.md)
+
+# How to add links
+
+`ActiveModelSerializers` offers you many ways to add links in your JSON reponse depending on the adapter you are using.
+
+Note that within the following examples, we set the `adapter` option, but that is not needed if the `ActiveModelSerializers.config.adapter` variable is properly set. Also, some examples are given for both Rails and [non-Rails usage](outside_controller_use.md).
```
bf4: `is configured with that adapter`

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+
+Unless stated otherwise, the examples given here will assume that the following classes are defined:
+
+```ruby
+class User < ActiveModelSerializers::Model
+  attributes :id, :name
```
bf4: assuming #1984 is merged.. :)

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+class UserSerializer < ActiveModel::Serializer
+  attributes :id, :name
+end
+```
+
+## JSON API adapter
+
+The JSON API specification allows the usage of the `links` member for representing links in three cases:
+* [Top level links](http://jsonapi.org/format/#document-top-level)
+* [Resource link](http://jsonapi.org/format/#document-resource-object-links)
```
bf4: `s/ link/ links`

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+* [Relationship links](http://jsonapi.org/format/#document-resource-object-relationships)
+
+`ActiveModelSerializers` provides tools to handle these links.
```
bf4: `ActiveModelSerializers supports the JSON API links specification.`

bf4: (or `complies with`) or (`meets`)

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+
+### [Top level links](http://jsonapi.org/format/#document-top-level)
+
+In order to provide top-level links, you will need to specify the `links` option to the serialization:
+
+```ruby
+user = User.new(id: 1, name: "John")
+links = { home: "https://www.example.com/" }
```
bf4: I'm thinking `home` can be a confusing name.  Maybe use an example from the spec?

bf4: can condense to 

```ruby
user = User.new(id: 1, name: "John")
links = { home: "https://www.example.com/" }
# Either in a controller:
render json: user, adapter: :json_api, links: links
# Or directly:
ActiveModelSerializers::SerializableResource.new(user, adapter: :json_api, links: links).to_json
```

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+
+#### Pagination
+
+`ActiveModelSerializers` will provide automatic pagination links for collections if you use [Kaminari](https://github.com/amatsuda/kaminari)
+or [WillPaginate](https://github.com/mislav/will_paginate) and the `ActiveModelSerializers.config.jsonapi_pagination_links_enabled` config is set to `true` (defaults to `true`). To disable this feature, you can simply set this config to `false`.
+
+```ruby
+class UserController < ActionController::Base
+  def kaminari
+    users = User.all.to_a # .all.to_a used as example
```
bf4: And example could just be, assuming we're rendering the third page, one item per page

```ruby
render json: users, adapter: :json_api
```

would render as:

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+    "self": "http://example.com/articles?page[number]=3&page[size]=1",
+    "first": "http://example.com/articles?page[number]=1&page[size]=1",
+    "prev": "http://example.com/articles?page[number]=2&page[size]=1",
+    "next": "http://example.com/articles?page[number]=4&page[size]=1",
+    "last": "http://example.com/articles?page[number]=13&page[size]=1"
+  }
+}
+```
+
+Note that it is possible to opt-out `ActiveModelSerializers` pagination relies on a collection that has the methods `current_page`, `total_pages`, and `size`, such as are supported by both [Kaminari](https://github.com/amatsuda/kaminari) or [WillPaginate](https://github.com/mislav/will_paginate). If you want to have paginated links your collections but don't want to use neither Kaminari nor WillPaginate, make sure to define these methods.
```
bf4: ? vs. `jsonapi_pagination_links_enabled`?

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+### [Resource link](http://jsonapi.org/format/#document-resource-object-links)
+
+You can use the `link` class method to define the links you need in the resource's primary data:
+
+```ruby
+class UserSerializer < ActiveModel::Serializer
+  attributes :name
+
+  link(:self) { user_path(object.id) }
+  link(:posts) { posts_path(user_id: object.id) }
```
bf4:
```
Using url helpers *requires*:
1) that the SerializationContext has the url helpers configured. This is done for you when rendering from a controller
2) that a block is passed where the url_helpers are used
```

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+end
+
+user = User.new(id: 1, name: "Kelly")
+resource = ActiveModelSerializers::SerializableResource.new(user, adapter: :json_api)
```
bf4: this won't work without a SerializationContext.  We need to verify all our docs contain valid code.

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+### [Relationship links](http://jsonapi.org/format/#document-resource-object-relationships)
+
+You can use the `link` method wihtin you relationship block in order to define the links you need in the relationships:
+
+```ruby
+class UserSerializer < ActiveModel::Serializer
+  attributes :id, :name
+  has_many :posts do
+    link(:self) do
+      user_posts_path(object.id) # Rails url helper automatically available
```
bf4: See https://github.com/rails-api/active_model_serializers/pull/1981/files#r89681745

i.e. has the same api and requirements as resource links, but `link` is called in the context of an association block. 

Might be good to note `include_data: false`

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+## JSON adapter
+
+The `json` is not able to handle links defined by the `link` class method. However, you can either use the `meta`  member or define a `links` attribute on the serializer.
```
bf4: JSON adapter doesn't respect the `link` macro in resources or relationships, however it does respect 'top-level meta', which can be used to include links.  See the implementation of the https://github.com/rails-api/active_model_serializers/blob/v0.10.3/lib/active_model_serializers/adapter/json_api/link.rb for more details

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+  end
+end
+```
+
+### Via an attribute
+
+```ruby
+class UserSerializer < ActiveModel::Serializer
+  attributes :id, :name, :links
+  attribute :links do
```
bf4: we really shouldn't allow this..  e.g. https://github.com/cerebris/jsonapi-resources/blob/df5a0f5f938f829eaba67dd79b79b820a7c58b74/lib/jsonapi/resource.rb#L1104-L1109

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+    }
+  }
+}
+```
+
+If you want to use the Rails' url/path helpers such as `user_path`, you need to `include` the `ActiveModelSerializers::SerializationContext::UrlHelpers` module into your serializer:
+
+```ruby
+class UserSerializer < ActiveModel::Serializer
+  include ActiveModelSerializers::SerializationContext::UrlHelpers
```
bf4: I would emphasize that this is not our recommended usage, and reference `include Rails.application.url_helpers` like you'd see in other docs.

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+      self: user_path(object.id),
+      posts: posts_path(user_id: object.id)
+    }
+  end
+end
+```
+
+## Attributes adapter
+
+The only way to provide pagination with the `attributes` adapter is to define a `links` attribute [as described for the JSON adapter above](#via-an-attribute).
```
bf4: which honestly, is kind of dumb of us.  Why do we *forbid* having a meta attribute? Attributes adapter really needs to be reabsorbed into JSON, since it's really the domain of the serializer.

docs/howto/add_relationship_links.md
```diff
@@ -1,137 +0,0 @@
...
-
-### Links that follow the JSONAPI spec
-**This is only applicable to JSONAPI adapter**
-
-If you have a JSONAPI-strict client that you are working with (like `ember-data`)
-you need to construct the links inside the relationships. Also the link to fetch the
-relationship data must be under the `related` attribute, whereas to manipulate the
-relationship (in case of many-to-many relationship) must be under the `self` attribute.
-
-You can find more info in the [spec](http://jsonapi.org/format/#document-resource-object-relationships).
```
bf4: why was all this removed?

docs/howto/handle_pagination.md
```diff
@@ -0,0 +1,75 @@
+[Back to Guides](../README.md)
+
+# How to handle pagination
+
+## JSON API adapter
+
+If you want to paginate a collection with the `json_api` adapter, you can either use the top-level `links` member to provide pagination links or the top-level `meta` member to provide pagination information:
```
bf4: this should be cross-referenced in the `links` section or in it altogether, no?

bf4:
@groyoh PR is pretty ambitious.  Maybe good to break into smaller pieces? Reorganizing vs. rewriting vs. fixing bugs?

docs/howto/add_link.md
groyoh: Why not? If people use JSON or Attributes adapters, they have every right to use a `links` attribute.

docs/howto/add_link.md
groyoh: Wasn't the `attributes` method provided by #1982 ?

docs/howto/add_link.md
groyoh: Looks like this was a mix between two different docs. Not sure how that happened.

docs/howto/add_link.md
groyoh: Seeing users' reaction about the docs always makes me think that having an explicit example for Rails is usually easier to understand as it creates less confusion.

docs/howto/add_link.md
groyoh: Would it make more sense to only gives examples with hand-crafted urls and then add a `Using Rails url helpers` section?

docs/howto/add_link.md
```diff
@@ -0,0 +1,336 @@
...
+}
+```
+
+If you want to use the Rails' url/path helpers such as `user_path`, you need to `include` the `ActiveModelSerializers::SerializationContext::UrlHelpers` module into your serializer:
+
+```ruby
+class UserSerializer < ActiveModel::Serializer
+  include ActiveModelSerializers::SerializationContext::UrlHelpers
+
+  attributes :id, :name, :links
```
groyoh:
```diff
-  attributes :id, :name, :links
+  attributes :id, :name
```

docs/howto/add_link.md
groyoh: Why don't we recommend that? What's wrong with doing so?

docs/howto/add_link.md
groyoh: How would you add a `meta` to:

```json
[
  {
    "id": 1,
    "name": "John"
  },
  {
    "id": 2,
    "name": "Margaret"
  }
]
```
?

I still think Attributes has its own place here.

docs/howto/add_relationship_links.md
groyoh: Because I removed the documentation which described how `links` can be defined as attribute for the JSONAPI adapter as it is not part of the specs. Therefore this part of the doc is no more necessary in my opinion.

docs/howto/handle_pagination.md
groyoh: I already reference the link documentation below. I don't see why we should put `meta` in the `links` documentation though. That's two different things in my opinion.

docs/howto/add_link.md
groyoh: Should be

```diff
-Note that it is possible to opt-out `ActiveModelSerializers` pagination relies on a collection that has the methods `current_page`, `total_pages`, and `size`, such as are supported by both [Kaminari](https://github.com/amatsuda/kaminari) or [WillPaginate](https://github.com/mislav/will_paginate). If you want to have paginated links your collections but don't want to use neither Kaminari nor WillPaginate, make sure to define these methods.
+Note that `ActiveModelSerializers` automatic pagination relies on a collection that has the methods `current_page`, `total_pages`, and `size`, such as are supported by both [Kaminari](https://github.com/amatsuda/kaminari) or [WillPaginate](https://github.com/mislav/will_paginate). If you want to have paginated links for your collections but don't want to use neither Kaminari nor WillPaginate, make sure to define these methods.
```

docs/howto/add_link.md
bf4: oh, touche

docs/howto/add_link.md
bf4: I sort of feel if they don't know how to use `render` this isn't the place to learn it

docs/howto/add_link.md
bf4: well, that goes back to our design goals of making all serializers/adapters support links

docs/howto/add_link.md
bf4: So, we don't do this because it couples the serializer too tightly to both Rails and to the routes.

We also provide a way to use url helpers via the serialization context.

And depending on how this is written, we're going to get a ton of questions all the time why don't we just mix it in like this?

Also, just feels messy :(

docs/howto/add_link.md
bf4: Good point re: collections, but in terms of 'resource' meta/links... 


`ActiveModel::Serializers::JSON.include_root_in_json == false` -> 'Attributes' adapter

`ActiveModel::Serializers::JSON.include_root_in_json == true` -> 'JSON' adapter

only difference is presence of root, then what that implies. https://github.com/rails/rails/blob/5-0-stable/activemodel/lib/active_model/serializers/json.rb#L88-L101

```ruby
      def as_json(options = nil)
        root = if options && options.key?(:root)
          options[:root]
        else
          include_root_in_json
        end

        if root
          root = model_name.element if root == true
          { root => serializable_hash(options) }
        else
          serializable_hash(options)
        end
      end
```

docs/howto/add_relationship_links.md
bf4: But it has some good info in it, like about ember-data and n+1 queries, no?

docs/howto/add_link.md
groyoh: Yeah I get your point. AFAIK serialization context is not provided to links or is it?

docs/howto/add_relationship_links.md
groyoh: Well that's indeed a good note, but I don't feel like this should be part of this documentation. I mean, it's right in the middle of the link documentation. How is anyone suppose to find that out. Maybe I could add another section which discuss the use of `include_data false` and the how to fix n+1 queries issue.

groyoh:
@bf4 I reworked this PR to only include the fix to the documentation about JSONAPI (plus minor typos). I will create a different PR for reorganizing the docs.

docs/howto/add_relationship_links.md
```diff
@@ -15,40 +15,42 @@ class Api::V1::UsersController < ApplicationController
...
 ### Links as an attribute of a resource
-**This is applicable to JSONAPI, JSON and Attributes adapters**
+**This is applicable to JSON and Attributes adapters**
 
 You can define an attribute in the resource, named `links`.
 
 ```ruby
 class Api::V1::UserSerializer < ActiveModel::Serializer
-  attributes :id, :name, :links
+  include Rails.application.routes.url_helpers
```
bf4: > `You can define an attribute in the resource, named `links`.

Isn't this true for all the adapters? It's just especially confusing for JSONAPI. Also, the name `links` here is arbitrary once you're mixing in the url helpers and just making an attribute return an custom object (hash).

I'd also rather encourage usage of the serialization context over mixing in ` include Rails.application.routes.url_helpers` but that's me. I sorta think using the `include Rails.application.routes.url_helpers` should be a way of doing it until we unify the adapters or write better docs for serialization_context. 


Needs tests. That's part of the reason the doc fix is needed, right?  we could put that off into a followup pr?

groyoh: It is *working* for all the adapter but not *correct* for JSONAPI as the spec forbid to have an attribute named `links`:

> The value of the attributes key MUST be an object (an “attributes object”). Members of the attributes object (“attributes”) represent information about the resource object in which it’s defined.
>
> Attributes may contain any valid JSON value.
>
> Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute MUST NOT contain a relationships or links member, as those members are reserved by this specification for future use.

Also I had the url helper because within the `links` method there was a call to `api_v1_user_path`. And that call would not be correct without the url helper. I simply "fixed" the existing doc as I did not want to modify it as to avoid having too many changes.

bf4: I suppose they should do it like [ActiveModelSerializers::Adapter::JsonApi::Link](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/lib/active_model_serializers/adapter/json_api/link.rb) and include [ActiveModelSerializers::SerializationContext::UrlHelpers](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/lib/active_model_serializers/serialization_context.rb#L14-L22)

```ruby
        include ActiveModelSerializers::SerializationContext::UrlHelpers
```

bf4: oh dear god. I guess I... good eyes @groyoh

bf4:
@groyoh Merged it. Is improvement :)

PR #1982, opened 2016-11-21
Better AMS Model attributes interface

bf4:
First step for #1903

PR #1984, opened 2016-11-21
Make test attributes explicit

bf4:
Follows https://github.com/rails-api/active_model_serializers/pull/1982

### DIFF
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -3,42 +3,58 @@
 # serializable non-activerecord objects.
 module ActiveModelSerializers
   class Model
-    include ActiveModel::Model
     include ActiveModel::Serializers::JSON
+    include ActiveModel::Model
+
+    class_attribute :attribute_names
+    # Initialize +attribute_names+ for all subclasses.  The array is usually
+    # mutated in the +attributes+ method, but can be set directly, as well.
+    self.attribute_names = []
 
     def self.attributes(*names)
-      attr_accessor(*names)
+      self.attribute_names |= names.map(&:to_sym)
+      # Silence redefinition of methods warnings
+      ActiveModelSerializers.silence_warnings do
+        attr_accessor(*names)
+      end
     end
 
-    attr_reader :attributes, :errors
+    attr_reader :errors
+    # NOTE that +updated_at+ isn't included in +attribute_names+,
+    # which means it won't show up in +attributes+ unless a subclass has
+    # either <tt>attributes :updated_at</tt> which will redefine the methods
+    # or <tt>attribute_names << :updated_at</tt>.
+    attr_writer :updated_at
+    # NOTE that +id+ will always be in +attributes+.
+    attributes :id
 
     def initialize(attributes = {})
-      @attributes = attributes && attributes.symbolize_keys
       @errors = ActiveModel::Errors.new(self)
       super
     end
 
-    # Defaults to the downcased model name.
-    def id
-      attributes.fetch(:id) { self.class.name.downcase }
+    # The the fields in +attribute_names+ determines the returned hash.
+    # +attributes+ are returned frozen to prevent any expectations that mutation affects
+    # the actual values in the model.
+    def attributes
+      attribute_names.each_with_object({}) do |attribute_name, result|
+        result[attribute_name] = public_send(attribute_name).freeze
+      end.with_indifferent_access.freeze
     end
 
-    # Defaults to the downcased model name and updated_at
+    # To customize model behavior, this method must be redefined. However,
+    # there are other ways of setting the +cache_key+ a serializer uses.
     def cache_key
-      attributes.fetch(:cache_key) { "#{self.class.name.downcase}/#{id}-#{updated_at.strftime('%Y%m%d%H%M%S%9N')}" }
+      ActiveSupport::Cache.expand_cache_key([
+        self.class.model_name.name.downcase,
+        "#{id}-#{updated_at.strftime('%Y%m%d%H%M%S%9N')}"
+      ].compact)
     end
 
-    # Defaults to the time the serializer file was modified.
+    # When no set, defaults to the time the file was modified.
+    # See NOTE by attr_writer :updated_at
     def updated_at
-      attributes.fetch(:updated_at) { File.mtime(__FILE__) }
-    end
-
-    def read_attribute_for_serialization(key)
-      if key == :id || key == 'id'
-        attributes.fetch(key) { id }
-      else
-        attributes[key]
-      end
+      defined?(@updated_at) ? @updated_at : File.mtime(__FILE__)
     end
 
     # The following methods are needed to be minimally implemented for ActiveModel::Errors
--- a/test/action_controller/adapter_selector_test.rb
+++ b/test/action_controller/adapter_selector_test.rb
@@ -15,7 +15,7 @@ module ActionController
         end
 
         def render_skipping_adapter
-          @profile = Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+          @profile = Profile.new(id: 'render_skipping_adapter_id', name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
           render json: @profile, adapter: false
         end
       end
@@ -46,7 +46,7 @@ module ActionController
 
       def test_render_skipping_adapter
         get :render_skipping_adapter
-        assert_equal '{"name":"Name 1","description":"Description 1","comments":"Comments 1"}', response.body
+        assert_equal '{"id":"render_skipping_adapter_id","name":"Name 1","description":"Description 1"}', response.body
       end
     end
   end
--- a/test/action_controller/json_api/transform_test.rb
+++ b/test/action_controller/json_api/transform_test.rb
@@ -5,9 +5,17 @@ module ActionController
     class JsonApi
       class KeyTransformTest < ActionController::TestCase
         class KeyTransformTestController < ActionController::Base
-          class Post < ::Model; end
-          class Author < ::Model; end
-          class TopComment < ::Model; end
+          class Post < ::Model
+            attributes :title, :body, :publish_at
+            associations :author, :top_comments
+          end
+          class Author < ::Model
+            attributes :first_name, :last_name
+          end
+          class TopComment < ::Model
+            attributes :body
+            associations :author, :post
+          end
           class PostSerializer < ActiveModel::Serializer
             type 'posts'
             attributes :title, :body, :publish_at
--- a/test/action_controller/serialization_test.rb
+++ b/test/action_controller/serialization_test.rb
@@ -135,7 +135,7 @@ module ActionController
           like = Like.new(id: 1, likeable: comment, time: 3.days.ago)
 
           generate_cached_serializer(like)
-          like.likable = comment2
+          like.likeable = comment2
           like.time = Time.zone.now.to_s
 
           render json: like
--- a/test/active_model_serializers/model_test.rb
+++ b/test/active_model_serializers/model_test.rb
@@ -4,7 +4,7 @@ module ActiveModelSerializers
   class ModelTest < ActiveSupport::TestCase
     include ActiveModel::Serializer::Lint::Tests
 
-    def setup
+    setup do
       @resource = ActiveModelSerializers::Model.new
     end
 
@@ -17,6 +17,53 @@ module ActiveModelSerializers
       model_instance = klass.new('key' => value)
 
       assert_equal model_instance.read_attribute_for_serialization(:key), value
+    end
+
+    def test_attributes_can_be_read_for_serialization
+      klass = Class.new(ActiveModelSerializers::Model) do
+        attributes :one, :two, :three
+      end
+      original_attributes = { one: 1, two: 2, three: 3 }
+      instance = klass.new(original_attributes)
+
+      # Initial value
+      expected_attributes = { id: nil, one: 1, two: 2, three: 3 }.with_indifferent_access
+      assert_equal expected_attributes, instance.attributes
+      assert_equal 1, instance.one
+      assert_equal 1, instance.read_attribute_for_serialization(:one)
+
+      # Change via accessor
+      instance.one = :not_one
+
+      expected_attributes = { id: nil, one: :not_one, two: 2, three: 3 }.with_indifferent_access
+      assert_equal expected_attributes, instance.attributes
+      assert_equal :not_one, instance.one
+      assert_equal :not_one, instance.read_attribute_for_serialization(:one)
+    end
+
+    def test_id_attribute_can_be_read_for_serialization
+      klass = Class.new(ActiveModelSerializers::Model) do
+        attributes :id, :one, :two, :three
+      end
+      self.class.const_set(:SomeTestModel, klass)
+      original_attributes = { id: :ego, one: 1, two: 2, three: 3 }
+      instance = klass.new(original_attributes)
+
+      # Initial value
+      expected_attributes = { id: :ego, one: 1, two: 2, three: 3 }.with_indifferent_access
+      assert_equal expected_attributes, instance.attributes
+      assert_equal 1, instance.one
+      assert_equal 1, instance.read_attribute_for_serialization(:one)
+
+      # Change via accessor
+      instance.id = :superego
+
+      expected_attributes = { id: :superego, one: 1, two: 2, three: 3 }.with_indifferent_access
+      assert_equal expected_attributes, instance.attributes
+      assert_equal :superego, instance.id
+      assert_equal :superego, instance.read_attribute_for_serialization(:id)
+    ensure
+      self.class.send(:remove_const, :SomeTestModel)
     end
   end
 end
--- a/test/cache_test.rb
+++ b/test/cache_test.rb
@@ -34,6 +34,7 @@ module ActiveModelSerializers
     end
 
     class Article < ::Model
+      attributes :title
       # To confirm error is raised when cache_key is not set and cache_key option not passed to cache
       undef_method :cache_key
     end
@@ -46,6 +47,16 @@ module ActiveModelSerializers
     class InheritedRoleSerializer < RoleSerializer
       cache key: 'inherited_role', only: [:name, :special_attribute]
       attribute :special_attribute
+    end
+
+    class Comment < ::Model
+      attributes :body
+      associations :post, :author
+
+      # Uses a custom non-time-based cache key
+      def cache_key
+        "comment/#{id}"
+      end
     end
 
     setup do
@@ -271,7 +282,7 @@ module ActiveModelSerializers
         ended_at: nil,
         updated_at: alert.updated_at,
         created_at: alert.created_at
-      }
+      }.with_indifferent_access
       expected_cached_jsonapi_attributes = {
         id: '1',
         type: 'alerts',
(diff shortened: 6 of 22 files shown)


lib/active_model_serializers/model.rb
```diff
@@ -51,5 +59,43 @@ def self.lookup_ancestors
...
+    private
+
+    def _assign_attributes(attributes)
+      attributes.each do |k, v|
+        _assign_attribute(k, v)
+      end
+    end
+
+    def _assign_attribute(k, v)
+      fail UnknownAttributeError.new(self, k) unless respond_to?("#{k}=")
```
bf4: TODO: check vs. `attribute_names`

lib/active_model_serializers/model.rb
```diff
@@ -51,5 +59,43 @@ def self.lookup_ancestors
...
+      public_send("#{k}=", v)
+    end
+
+    def persisted?
+      false
+    end
+
+    # Raised when unknown attributes are supplied via mass assignment.
+    class UnknownAttributeError < NoMethodError
```
bf4: copied from Rails 5 ActiveModel::AttributeAssignment

test/action_controller/adapter_selector_test.rb
```diff
@@ -15,7 +15,7 @@ def render_using_adapter_override
         end
 
         def render_skipping_adapter
-          @profile = Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+          @profile = Profile.new(id: 'render_skipping_adapter_id', name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
```
bf4: probably a good thing that `id`, which is required, is now included

lib/active_model_serializers/model.rb
```diff
@@ -3,42 +3,50 @@
 # serializable non-activerecord objects.
 module ActiveModelSerializers
   class Model
-    include ActiveModel::Model
     include ActiveModel::Serializers::JSON
+    include ActiveModel::Validations
+    include ActiveModel::Conversion
+    extend ActiveModel::Naming
+    extend ActiveModel::Translation
```
bf4: code adapted from ActiveModel::Model which is has good logic in it, but doesn't still has a few cases we want to handle differently

NullVoxPopuli: what are those use cases?

NullVoxPopuli:
LGTM, ready for merge?

test/action_controller/serialization_test.rb
```diff
@@ -135,7 +135,7 @@ def render_fragment_changed_object_with_relationship
           like = Like.new(id: 1, likeable: comment, time: 3.days.ago)
 
           generate_cached_serializer(like)
-          like.likable = comment2
+          like.likeable = comment2
```
bf4: typo!

test/fixtures/poro.rb
```diff
@@ -3,23 +3,19 @@
...
-  # Convenience when not adding @attributes readers and writers
-  def method_missing(meth, *args)
-    if meth.to_s =~ /^(.*)=$/
-      attributes[Regexp.last_match(1).to_sym] = args[0]
-    elsif attributes.key?(meth)
-      attributes[meth]
-    else
-      super
-    end
+  # At this time, just for organization of intent
```
bf4: probably good to promote to parent class once we add some additional behavior to the `attributes` method

test/active_model_serializers/model_test.rb
```diff
@@ -18,5 +18,52 @@ def test_initialization_with_string_keys
...
+      assert_equal 1, instance.read_attribute_for_serialization(:one)
+
+      # Change via accessor
+      instance.one = :not_one
+
+      expected_attributes = { id: nil, one: :not_one, two: 2, three: 3 }.with_indifferent_access
+      assert_equal expected_attributes, instance.attributes
+      assert_equal :not_one, instance.one
+      assert_equal :not_one, instance.read_attribute_for_serialization(:one)
+    end
```
bf4: is fixed https://github.com/rails-api/active_model_serializers/pull/1903 

ref https://github.com/rails-api/active_model_serializers/issues/1877

bf4:
cc @richmolj

test/active_model_serializers/model_test.rb
bf4: bug ref https://github.com/rails-api/active_model_serializers/pull/1857#r75954677

lib/active_model_serializers/model.rb
```diff
@@ -3,42 +3,50 @@
...
-    include ActiveModel::Model
     include ActiveModel::Serializers::JSON
+    include ActiveModel::Model
+
+    class_attribute :attribute_names
+    self.attribute_names = []
 
     def self.attributes(*names)
       attr_accessor(*names)
+      self.attribute_names = attribute_names | names.map(&:to_sym)
```
beauby:
```ruby
self.attribute-names |= names.map(&:to_sym)
```

lib/active_model_serializers/model.rb
```diff
@@ -3,42 +3,50 @@
...
     def initialize(attributes = {})
-      @attributes = attributes && attributes.symbolize_keys
       @errors = ActiveModel::Errors.new(self)
       super
     end
 
     # Defaults to the downcased model name.
     def id
-      attributes.fetch(:id) { self.class.name.downcase }
+      @id ||= self.class.name && self.class.name.downcase
```
beauby: The `id` method should return a unique identifier for a resource, no? Not sure if providing such a default is more useful than dangerous.

lib/active_model_serializers/model.rb
```diff
@@ -3,42 +3,50 @@
...
     end
 
     # Defaults to the downcased model name and updated_at
     def cache_key
-      attributes.fetch(:cache_key) { "#{self.class.name.downcase}/#{id}-#{updated_at.strftime('%Y%m%d%H%M%S%9N')}" }
+      ActiveSupport::Cache.expand_cache_key([
+        self.class.name && self.class.name.downcase,
```
beauby: How about defaulting to `self.class.object_id` for anonymous classes? That way you avoid colliding all anonymous classes in the cache.

lib/active_model_serializers/model.rb
```diff
@@ -3,42 +3,50 @@
...
+        "#{id}-#{updated_at.strftime('%Y%m%d%H%M%S%9N')}"
+      ].compact)
     end
 
     # Defaults to the time the serializer file was modified.
     def updated_at
-      attributes.fetch(:updated_at) { File.mtime(__FILE__) }
+      defined?(@updated_at) ? @updated_at : File.mtime(__FILE__)
```
beauby: It is safe to do `@updated_at || File.mtime(__FILE__)`.

lib/active_model_serializers/model.rb
bf4: @beauby I agree this is a somewhat problematic implementation.  Also, I think only JSON API requires an 'id'.

lib/active_model_serializers/model.rb
bf4: I thought about that, or `hash`, but I'm not sure that's any better, as it's not really a unique identifier. We could hash the attributes, I guess.  If we just used `ActiveModel::Name` then it would blow up on the anonymous class and we could rely on it's exception message.

lib/active_model_serializers/model.rb
bf4: @beauby yeah, but give ruby warning undefined ivar

lib/active_model_serializers/model.rb
beauby: In that case, I'd suggest trying to read the `:id` attribute, and if none available, fall back to `object_id`.

lib/active_model_serializers/model.rb
beauby: `self.class.object_id` **is** a unique identifier for the class. For the the instances, I don't see any better solution than hashing the attributes, which would suck because it would either be complicated or fail for hashes. I'd vote for no caching for those models.

lib/active_model_serializers/model.rb
```diff
@@ -1,44 +1,48 @@
...
-    def read_attribute_for_serialization(key)
-      if key == :id || key == 'id'
-        attributes.fetch(key) { id }
-      else
-        attributes[key]
-      end
+    def attributes
+      attribute_names.each_with_object({}) do |attribute_name, result|
+        result[attribute_name] = public_send(attribute_name)
+      end.with_indifferent_access
```
bf4: `.freeze`

lib/active_model_serializers/model/caching.rb
```diff
@@ -0,0 +1,25 @@
...
+      included do
+        attr_writer :updated_at
+        attributes :id
+      end
+
+      # Defaults to the downcased model name and updated_at
+      def cache_key
+        ActiveSupport::Cache.expand_cache_key([
+          self.class.model_name.name.downcase,
+          "#{id}-#{updated_at.strftime('%Y%m%d%H%M%S%9N')}"
```
bf4: This will blow up if `updated_at` isn't a 'Time', but that will only happen when users set updated_at, since the default is `File.mtime`

test/active_model_serializers/model_test.rb
```diff
@@ -18,5 +27,59 @@ def test_initialization_with_string_keys
 
       assert_equal model_instance.read_attribute_for_serialization(:key), value
     end
+
+    def test_attributes_can_be_read_for_serialization
+      klass = subclass_without_caching_support do
```
bf4: specifically, this prevents the `id` from being added

test/active_model_serializers/model_test.rb
```diff
@@ -18,5 +27,59 @@ def test_initialization_with_string_keys
...
+      # Change via accessor
+      instance.one = :not_one
+
+      expected_attributes = { one: :not_one, two: 2, three: 3 }.with_indifferent_access
+      assert_equal expected_attributes, instance.attributes
+      assert_equal :not_one, instance.one
+      assert_equal :not_one, instance.read_attribute_for_serialization(:one)
+    end
+
+    def test_id_attribute_can_be_read_for_serialization
```
bf4: no longer necessary now that the `read_attribute_for_serialization` method has been removed

test/active_model_serializers/model_test.rb
bf4: But, since `id` no no longer has special behavior, subclassing without caching methods isn't necessary here and can be moved to another PR, if we like

test/cache_test.rb
```diff
@@ -34,6 +34,7 @@ class UncachedAuthor < Author
     end
 
     class Article < ::Model
+      attributes :title
       # To confirm error is raised when cache_key is not set and cache_key option not passed to cache
       undef_method :cache_key
```
bf4: seems like we want to make  a `NonCachingModel` poro base class for cases like this.

test/fixtures/poro.rb
```diff
@@ -3,23 +3,24 @@
...
+
+  def self.associations(*names)
+    attr_accessor(*names)
+    self.association_names = association_names | names.map(&:to_sym)
+  end
+
+  def associations
+    association_names.each_with_object({}) do |association_name, result|
+      result[association_name] = public_send(association_name)
+    end.with_indifferent_access
```
bf4: `.freeze` but this isn't actually called right now...

test/fixtures/poro.rb
```diff
@@ -30,67 +31,74 @@ def respond_to_missing?(method_name, _include_private = false)
...
 end
-
 class ProfileSerializer < ActiveModel::Serializer
   attributes :name, :description
 
-  # TODO: is this used anywhere?
+  # ActiveModel::Serializer::OptionsTest#test_options_are_accessible:
+  # test/serializers/options_test.rb:12:in `test_options_are_accessible'
+  # test/serializers/options_test.rb:17:in `test_no_option_is_passed_in'
   def arguments_passed_in?
```
bf4: TODO: move to `test/serializers/options_test.rb`

test/fixtures/poro.rb
```diff
@@ -30,67 +31,74 @@ def respond_to_missing?(method_name, _include_private = false)
...
+  attributes :id
+  has_many :posts
+end
+
 class Comment < Model
+  attributes :body, :date
+  associations :post, :author, :likes
+
   # Uses a custom non-time-based cache key
   def cache_key
```
bf4: still used?

test/fixtures/poro.rb
```diff
@@ -30,67 +31,74 @@ def respond_to_missing?(method_name, _include_private = false)
...
-
-class ObjectTag < ActiveRecord::Base
-  belongs_to :poly_tag
-  belongs_to :taggable, polymorphic: true
-end
-
-class Picture < ActiveRecord::Base
-  belongs_to :imageable, polymorphic: true
-  has_many :object_tags, as: :taggable
+  def custom_options
```
bf4: necessary?

test/fixtures/poro.rb
```diff
@@ -103,143 +111,135 @@ def blog
...
-  # TODO: is this used anywhere?
+  # ActiveModel::Serializer::ArraySerializerTest#test_each_object_should_be_serialized_with_appropriate_serializer:
+  # test/collection_serializer_test.rb:43:in `test_each_object_should_be_serialized_with_appropriate_serializer'
+  # ActiveModel::Serializer::ArraySerializerTest#test_serializer_option_not_passed_to_each_serializer:
+  # test/collection_serializer_test.rb:49:in `test_serializer_option_not_passed_to_each_serializer'
+  # ActiveModel::Serializer::CollectionSerializerTest#test_each_object_should_be_serialized_with_appropriate_serializer:
+  # test/collection_serializer_test.rb:43:in `test_each_object_should_be_serialized_with_appropriate_serializer'
+  # ActiveModel::Serializer::CollectionSerializerTest#test_serializer_option_not_passed_to_each_serializer:
+  # test/collection_serializer_test.rb:49:in `test_serializer_option_not_passed_to_each_serializer'
   def custom_options
```
bf4: TODO: move to `test/collection_serializer_test.rb`

test/fixtures/poro.rb
```diff
@@ -103,143 +111,135 @@ def blog
...
-  has_one :blog, key: :site
+class Tag < Model
+  attributes :name
 end
 
+class VirtualValue < Model; end
 class VirtualValueSerializer < ActiveModel::Serializer
   attributes :id
-
   has_many :reviews, virtual_value: [{ type: 'reviews', id: '1' },
```
bf4: I wonder if anyone actually uses this

NullVoxPopuli: I think we should remove it. It's not as clean as defining a reviews method, or using the block form.

test/fixtures/poro.rb
groyoh: I guess it is used to test that options are passed from collection serializer -> serializer and serializer -> association_serializer but it could definitely benefit from creating a specific serializer within the file testing this case.

lib/active_model_serializers/model.rb
```diff
@@ -1,44 +1,48 @@
...
+    class_attribute :add_caching_support
+    self.add_caching_support = true
+    class_attribute :attribute_names
+    self.attribute_names = []
 
     def self.attributes(*names)
-      attr_accessor(*names)
+      self.attribute_names |= names.map(&:to_sym)
+      # Silence redefinition of methods warnings
+      ActiveModelSerializers.silence_warnings do
```
groyoh: Why is this actually needed? Which method redefinition are you actually trying to silence?

Wouldn't

```ruby
names.map!(&to_sym)
attr_accessor(names - self.attribute_name)
self.attribute_names |= names
```

work instead of messing up with $VERBOSE?

test/action_controller/serialization_test.rb
bf4: well, caught it because it raised an error, undefined attribute.. makes me wonder if it was being used...

test/fixtures/poro.rb
bf4: It makes more sense internally... we could deprecate it, but first look through issues and google group for anyone using it

lib/active_model_serializers/model.rb
bf4: the `attr_accessor`.  It still yields warnings when, e.g. redefining `updated_at`, which isn't an `attribute` on the model by default

lib/active_model_serializers/model/caching.rb
bf4: note that `updated_at` isn't included in `attribute_names` and has no defined `updated_at=` setter.

lib/active_model_serializers/model/caching.rb
```diff
@@ -0,0 +1,25 @@
+module ActiveModelSerializers
+  class Model
+    module Caching
+      extend ActiveSupport::Concern
+
+      included do
+        attr_writer :updated_at
+        attributes :id
```
bf4: `id` will always be in `attributes`

test/action_controller/adapter_selector_test.rb
bf4: ref https://github.com/rails-api/active_model_serializers/pull/1984/files#r89365145

groyoh:
LGTM! Any more work to be done for this?

bf4:
@groyoh Just looks for consensus since there are some important changes in here.

bf4:
@rails-api/ams @rails-api/commit ready for final review.  Should I break this up? Good enough?

lib/active_model_serializers/model/caching.rb
```diff
@@ -4,19 +4,26 @@ module Caching
...
+      # there are other ways of setting the +cache_key+ a serializer uses.
       def cache_key
         ActiveSupport::Cache.expand_cache_key([
           self.class.model_name.name.downcase,
           "#{id}-#{updated_at.strftime('%Y%m%d%H%M%S%9N')}"
         ].compact)
       end
 
-      # Defaults to the time the serializer file was modified.
+      # When no set, defaults to the time the file was modified.
```
groyoh: `s/no/not/` ?

groyoh:
Isn't there too many changes here?

bf4:
@groyoh 
> Isn't there too many changes here?

Yes, but I'm not sure how I'd like to break it up, since they go well together... part of the issue is that making the test attributes explicit, we need to decide when `id` or `updated_at` should be in the attributes hash, IIRC

I'm open to any suggestions how best to do it. That's why I'm not comfortable merging just yet.

test/serializers/associations_test.rb
```diff
@@ -62,7 +62,13 @@ def test_serializer_options_are_passed_into_associations_serializers
...
                       .detect { |assoc| assoc.key == :comments }
 
-        assert association.serializer.first.custom_options[:custom_options]
+        comment_serializer = association.serializer.first
+        class << comment_serializer
+          def custom_options
+            instance_options
+          end
+        end
+        assert comment_serializer.custom_options
```
bf4: thanks, you're right! I lost the `custom_options[:custom_options]`

bf4:
@rails-api/ams @rails-api/commit ready for review again

test/action_controller/json_api/transform_test.rb
```diff
@@ -5,9 +5,17 @@ module Serialization
     class JsonApi
       class KeyTransformTest < ActionController::TestCase
         class KeyTransformTestController < ActionController::Base
-          class Post < ::Model; end
-          class Author < ::Model; end
-          class TopComment < ::Model; end
+          class Post < ::Model
+            attributes :title, :body, :publish_at
+            associations :author, :top_comments
```
NullVoxPopuli: i'm certainly a fan of this being more explicit

test/active_model_serializers/model_test.rb
```diff
@@ -18,5 +18,52 @@ def test_initialization_with_string_keys
 
       assert_equal model_instance.read_attribute_for_serialization(:key), value
     end
+
+    def test_attributes_can_be_read_for_serialization
```
NullVoxPopuli: should new tests be using `test '...' do`?

test/cache_test.rb
NullVoxPopuli: :+1:

test/fixtures/poro.rb
NullVoxPopuli: depracte in another PR or this one while we're talkin' about it?

NullVoxPopuli:
I'm mostly :+1:

test/active_model_serializers/model_test.rb
bf4: oh, I guess so.  I guess I haven't completely switch to activesupport form.  I mostly just wanted the stable interface vs. minitest

test/fixtures/poro.rb
bf4: @NullVoxPopuli maybe we can make a project and put this in the queue as a 'note'?

NullVoxPopuli: I dig it

Issue #1927, opened 2016-09-14
Use jsonapi gem for Deserialization


richmolj:
#### Expected behavior vs actual behavior

`ActiveModelSerializers::Adapter::JsonApi::Deserialization` has its own logic for deserialization. As we already depend on [https://github.com/beauby/jsonapi], let's use that gem for this purpose instead.

#### Steps to reproduce

N/A

#### Environment

N/A

#### Backtrace

N/A

#### Additonal helpful information

Relevant method for validating documents: https://github.com/beauby/jsonapi/blob/master/lib/jsonapi/document.rb#L37

This came up in https://github.com/rails-api/active_model_serializers/issues/1925#issuecomment-247036142

NullVoxPopuli:
:+1:

NullVoxPopuli:
I'm going to grab this one

NullVoxPopuli:
@richmolj added you, cause @remear told me you want to tackle this. I can just work on native extensioning key transforms. I've extracted key transforms into a gem already, as it needs to be shared between AMS and whatever ends up doing deserialization.

richmolj:
>  @remear told me you want to tackle this

@NullVoxPopuli @remear not sure this is coming from, I was glad to see you pick it up 😄 I can take it, but probably won't get a chance until next week/end.

NullVoxPopuli:
That's cool. Lets just stay in contact then, cause I've also started a few gems this week:

Released
 - https://github.com/NullVoxPopuli/case_transform - needs native extensions

In Progress
 - https://github.com/NullVoxPopuli/rails_module_unification
 - https://github.com/NullVoxPopuli/has_integrations

Placeholder I had made until I heard you were interested :-)
 - https://github.com/NullVoxPopuli/jsonapi_deserialization

PR #1928, opened 2016-09-15
Utilize JSONAPI family of gems for deserialization.

NullVoxPopuli:
#### Purpose

To further break apart AMS into functional sections that can be easily swapped with native extension gems to allow for much faster serialization/deserialization.

This PR cleans up a lot of Deserialization functionality, and offloads it to [jsonapi-deserialization](https://github.com/jsonapi-rb/jsonapi-deserializable).

#### Changes
- Remove existing deserialization / replace with jsonapi-deserialization
-
#### Caveats
- only / except have been removed in favor of either creating a whitelisting custom deserializer, or using strong parameters.

#### Related GitHub issues

https://github.com/rails-api/active_model_serializers/issues/1927 - use JSON API gem for deserialization
https://github.com/rails-api/active_model_serializers/issues/1925 - data can be an array
#### Additional helpful information

/cc @remear @richmolj @beauby

PR #1986, opened 2016-11-22
Underscore incoming params before filtering during Deserialization

aliaporci:
#### Purpose

**TL;DR** Ensure general use case of using snake case for permitted params i.e. `jsonapi_parse(params, only: [:first_name])` even when deserializing dasherized or camel cased params.

**Long version**: In Rails-land, we generally default to snake case when describing variables, symbols, etc. Right now, if we have incoming params with kebab or camel cased fields, we have to match the case when deserializing in the controller. 

E.g.

If we send dasherized parameters such as
```
{
  'data' => {
    'type' => 'users',
    'id' => 'blah',
    'attributes' => {
      'first-name' => 'Freddie',
      'last-name' => 'Mercury'
    }
  }
}
```
to a UserController containing this (fairly standard) method
```
def user_params
  ActiveModelSerializers::Deserialization.jsonapi_parse(params,
    only: [:first_name, :last_name])
end
```
the result will be empty `user_params` as both 'first-name' and 'last-name' are filtered out by the `only` option before the params are ever parsed.

The current (ugly and unsatisfying) workaround is to list out the kebabed or camel cased parameters as strings (i.e. `only: ['first-name', 'last-name']` or to convince whoever is building a front end for your API to patch their Ember adapter or whatever to match whatever you're using (snake case FTW). 

I'd much rather be able to use snake case for permitted params consistently across all of my Rails apps and not worry about remembering how one particular front end is configured to send params over (and not worry about having to change the casing of all of my permitted params in all of my controllers in the event the status quo changes!).

#### Changes

Call existing underscore function on incoming parameter field keys when filtering against options passed into the `parse` method

#### Caveats

Might be overkill.

#### Related GitHub issues

Possible made obsolete by #1927 if there's ever any movement on it.

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/deserialization.rb
+++ b/lib/active_model_serializers/adapter/json_api/deserialization.rb
@@ -141,6 +141,7 @@ module ActiveModelSerializers
 
         # @api private
         def filter_fields(fields, options)
+          KeyTransform.send(:underscore, fields)
           if (only = options[:only])
             fields.slice!(*Array(only).map(&:to_s))
           elsif (except = options[:except])


bf4:
@aliaporci Thanks so much for the writeup and link!

I'm kind of surprised this issue hasn't come up before.  This must have been a problem for all sorts of people.  (I've generally been of the opinion that Ember does a better job at the adapter layer of handling Rails snake case than Rails does in producing and consuming kebab-case to avoid these issues :))

bf4:
we'll need a before/after benchmark of this too.  Do you feel comfortable running `bin/bench` locally on master and on your branch?

axsuul:
As an Ember developer, this needs to be implemented!

bf4:
Needs tests

hardywu:
using this piece of codes for now. 
```ruby
props = ActiveModelSerializers::Deserialization.jsonapi_parse params
props = ActionController::Parameters.new(**props)
props.permit :attr1, :attr2
```
 Hope the pull request #1928 get merged soon.

hardywu:
any news?

bf4:
@hardywu I guess not :(

PR #1987, opened 2016-11-22
Preserve path when string is camelized

sineed:
#### Purpose
In my project I have some cases when I need to make a `POST` request with nested data. The following data may be provided:

```
{
  "data": {
    "type": "offers",
    "attributes": {
      "bid": { "amount": 1.12, "currency": "USD" }
    }
  }
}
```

Bid attribute may have validation errors so the response should provide this information:

```
{
  "errors": [
    { "source": { "pointer": "/data/attributes/bid/amount" }, "detail": "too_low" }
  ]
}
```

But when I use camel key transformation I receive `/data/attributes/bid::Amount` instead of `/data/attributes/bid/amount`

This PR fixes this behavior for `camel` and `camel_lower` key transformations

#### Changes

In `camel` and `camel_lower` key transformations 

#### Caveats

I saw work at https://github.com/rails-api/active_model_serializers/pull/1928 and introduction of [case_transform](https://github.com/NullVoxPopuli/case_transform). I can make PR to this gem also.

#### Related GitHub issues
#1928

### DIFF
--- a/lib/active_model_serializers/key_transform.rb
+++ b/lib/active_model_serializers/key_transform.rb
@@ -4,6 +4,7 @@ module ActiveModelSerializers
   module KeyTransform
     module_function
 
+    SLASH_SYMBOL = "/".freeze
     # Transforms values to UpperCamelCase or PascalCase.
     #
     # @example:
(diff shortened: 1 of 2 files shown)


bf4:
So the value is being transformed? that doesn't seem right

PR #1991, opened 2016-12-02
Add jsonapi_include_toplevel_object option to serializer

nilsding:
#### Purpose

Sometimes, I have the need to include the JSON-API top level object while it's disabled in the config.  This PR adds the functionality to include it on the fly using an option to the serializer:

```
 ActiveModelSerializers::SerializableResource.new(
  UserSerializer.new(obj),
  adapter:                         :json_api,
  jsonapi_include_toplevel_object: true
).as_json
```

#### Changes

Pass instance options to the `ActiveModelSerializers::Adapter::JsonApi::Jsonapi` model, and modify the `include_object?` method to use the value of the instance option `:jsonapi_include_toplevel_object` instead of the global config, if given.

#### Caveats

-

#### Related GitHub issues

-

-

### DIFF
--- a/lib/active_model_serializers/serializable_resource.rb
+++ b/lib/active_model_serializers/serializable_resource.rb
@@ -2,7 +2,7 @@ require 'set'
 
 module ActiveModelSerializers
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :serialization_context, :key_transform])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :serialization_context, :key_transform, :jsonapi_include_toplevel_object])
     include ActiveModelSerializers::Logging
 
     delegate :serializable_hash, :as_json, :to_json, to: :adapter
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
@beauby do you have opinions on this?

beauby:
I agree this should be overridable locally (and although it is weird that `jsonapi_include_toplevel_object` would be an option on the resource, since it is a property of the document, not of the primary resource, I don't see an other way to do it currently).

lib/active_model_serializers/serializable_resource.rb
```diff
@@ -2,7 +2,7 @@
 
 module ActiveModelSerializers
   class SerializableResource
-    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :serialization_context, :key_transform])
+    ADAPTER_OPTION_KEYS = Set.new([:include, :fields, :adapter, :meta, :meta_key, :links, :serialization_context, :key_transform, :jsonapi_include_toplevel_object])
```
bf4: A question I have is what should be different between wanting to pass in your own `jsonapi` object vs. telling the adapter to include the default one?

PR #1995, opened 2016-12-07
STI type deserialization

gauthier-delacroix:
#### Purpose
Allows STI type deserialization from JSONAPI attributes.

The type key has to be customized (`:type`for Active Record, `:_type` for Mongoid) to enable the feature so there's no breaking change.

#### Changes
Added `sti_key_type` global config key and `parse` option to enable/customize the feature.
Added `parse_type` private method to deserialize class name.
Added doc
Added 2 tests:

- Simple type deserialization
- Namespaced type deserialization

#### Caveats
Requires `jsonapi_namespace_separator` to be different than `key_transform` which is not the current default (note added in the doc).

#### Related GitHub issues
None found

### DIFF
--- a/test/adapter/json_api/parse_test.rb
+++ b/test/adapter/json_api/parse_test.rb
@@ -130,6 +130,25 @@ module ActiveModelSerializers
             }
             assert_equal(expected, parsed_hash)
           end
+
+          def test_sti_type
+            parsed_hash = ActiveModelSerializers::Adapter::JsonApi::Deserialization.parse!(@hash, sti_type_key: :_type)
+            assert_equal(@expected.merge(_type: 'Photo'), parsed_hash)
+
+            with_config(sti_type_key: :_type) do
+              parsed_hash = ActiveModelSerializers::Adapter::JsonApi::Deserialization.parse!(@hash)
+              assert_equal(@expected.merge(_type: 'Photo'), parsed_hash)
+            end
+          end
+
+          def test_namespaced_sti_type
+            hash = @hash
+            hash['data']['type'] = 'content--user-photos'
+            with_namespace_separator '--' do
+              parsed_hash = ActiveModelSerializers::Adapter::JsonApi::Deserialization.parse!(hash, sti_type_key: :_type)
+              assert_equal(@expected.merge(_type: 'Content::UserPhoto'), parsed_hash)
+            end
+          end
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Breaking changes:
 Features:
 
 - [#1982](https://github.com/rails-api/active_model_serializers/pull/1982) Add ActiveModelSerializers::Model.attributes to configure PORO attributes (@bf4).
+- [#1995](https://github.com/rails-api/active_model_serializers/pull/1995) Allow STI type deserialization (@gauthier-delacroix).
 
 Fixes:
 
--- a/docs/general/deserialization.md
+++ b/docs/general/deserialization.md
@@ -95,6 +95,24 @@ ActiveModelSerializers::Deserialization
 # }
 ```
 
+A [Single Table Inheritance](http://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html) type key can be set via the options or config key `sti_key_type` to get the original class name:
+
+```ruby
+ActiveModelSerializers::Deserialization
+  .jsonapi_parse(document, sti_key_type: :type)
+#=>
+# {
+#   type: 'Post',
+#   title: 'Title 1',
+#   date: '2015-12-20',
+#   author_id: 2,
+#   second_author_id: nil
+#   comment_ids: [3, 4]
+# }
+```
+
+Note that [jsonapi_namespace_separator](configuration_options.md#jsonapi_namespace_separator) has to be changed to something different than [key_transform](configuration_options.md#key_transform) to allow deserialization of namespaced STI models.
+
 ## Attributes/Json
 
 There is currently no deserialization for those adapters.
--- a/lib/active_model/serializer/concerns/configuration.rb
+++ b/lib/active_model/serializer/concerns/configuration.rb
@@ -22,6 +22,7 @@ module ActiveModel
         config.default_includes = '*'
         config.adapter = :attributes
         config.key_transform = nil
+        config.sti_type_key = nil
         config.jsonapi_pagination_links_enabled = true
         config.jsonapi_resource_type = :plural
         config.jsonapi_namespace_separator = '-'.freeze
--- a/lib/active_model_serializers/adapter/json_api/deserialization.rb
+++ b/lib/active_model_serializers/adapter/json_api/deserialization.rb
@@ -97,6 +97,7 @@ module ActiveModelSerializers
           filter_fields(relationships, options)
 
           hash = {}
+          hash.merge!(parse_type(primary_data['type'], options))
           hash.merge!(parse_attributes(attributes, options))
           hash.merge!(parse_relationships(relationships, options))
 
@@ -200,6 +201,17 @@ module ActiveModelSerializers
           transform_keys(relationships, options)
             .map { |(k, v)| parse_relationship(k, v['data'], options) }
             .reduce({}, :merge)
+        end
+
+        # @api private
+        def parse_type(type, options)
+          sti_type_key = options.fetch(:sti_type_key, ActiveModelSerializers.config.sti_type_key)
+          if sti_type_key && type
+            type = type.gsub(ActiveModelSerializers.config.jsonapi_namespace_separator, '/')
+            type = transform_keys(type, options).classify
+            { sti_type_key => type }
+          else {}
+          end
         end
 
         # @api private


test/adapter/json_api/parse_test.rb
```diff
@@ -130,6 +130,25 @@ def test_polymorphic
...
+
+            with_config(sti_type_key: :_type) do
+              parsed_hash = ActiveModelSerializers::Adapter::JsonApi::Deserialization.parse!(@hash)
+              assert_equal(@expected.merge(_type: 'Photo'), parsed_hash)
+            end
+          end
+
+          def test_namespaced_sti_type
+            hash = @hash
+            hash['data']['type'] = 'content--user-photos'
```
bf4: Thanks for the PR.  So, if I can understand the issue:

Given the resource

```ruby
{
  id: '1',
  type: 'content--user-photos',
  attributes: { title: 'A day at the beach' }
}
```

When we deserialize it we get

```ruby
{
  id: 1,
  title: 'A day at the beach' 
}
```

and lose the resource type in the resulting attributes.

In Rails, there is an single-table-inheritance (sti) attribute used to determine the class to instantiate from the database. The default name for this field is `type`, which is an `attribute` on the ActiveRecord object.

When AMS serializes a User it uses a UserSerializer with the default JSON API resource type of `users`.  Similarly, the default behavior would be UserPhoto to use the UserPhotoSerializer and have the JSON API resource type of `user-photos`. In both cases, unless specified in the serializer, the `type` field would not be serialized as an attribute.

So, I'm thinking that part of the use case here is what you do with the deserialized JSON, and if you can or should use the resource type vs. the attribute 'type'. Is that right?  

How are you using the deserialized JSON document to interact with ActiveRecord?

gauthier-delacroix: Disclaimer: I'm using Mongoid, but the behavior seems to be quite the same than ActiveRecord except Mongoid uses the `:_type` attribute, which is why I allowed customization.

This type field deserialization is necessary to create STI objects from the API.

When creating object, the JSON payload has the same format than serialized output, including the `type` root data key: http://jsonapi.org/format/#crud-creating

Why would you add a `data['attributes']['type']` when it's already supposed to be set in `data['type']` ?

gauthier-delacroix:
Disclaimer: I'm using Mongoid, but the behavior seems to be quite the same than ActiveRecord except Mongoid uses the `:_type` attribute, which is why I allowed customization.

This type field deserialization is necessary to create STI objects from the API.

When creating object, the JSON payload has the same format than serialized output, including the `type` root data key: http://jsonapi.org/format/#crud-creating

Why would you add a `data['attributes']['type']` when it's already supposed to be set in `data['type']` ?

bf4:
The resource type field has a definition in jsonapi.

The rails sti field is a record's attribute.  I usually name it something other than type to avoid this sort of confusion 

Could you share an example of how you're using the deserialized resource? It's important to understand what you're trying to do, not just what you want ams to do. 

Thanks!

B mobile phone

gauthier-delacroix:
The type field is a record attribute since it's the only way to store the object type in the DB.

JsonAPI has a resource type field which is the same information so there's no confusion IMHO.

Do you mean than to create STI objects you'd use this kind of payload ?

````ruby
{
  id: '1',
  type: 'content--user-photos',
  attributes: {
    type: 'Content::UserPhotos',
    title: 'A day at the beach'
  }
}
````

Aren't `content--user-photos` and `Content::UserPhotos` the same information in a different format ?

Here is a simple example of what I'm doing:

I send this create request:

````
POST /pets HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": {
    "type": "pets--cat",
    "attributes": {
      "name": "Casper",
    },
  }
}
````

My request params are deserialized to:
````ruby
{
  type: "Pet::Cat",
  name: "Casper",
}
````

I still need to include this into my original params to match the standard Rails format, where the STI type is a record attribute as you noted:
````ruby
{
  controller: 'pets',
  action: 'create',
  pet: {
    type: "Pet::Cat",
    name: "Casper",
  }
}
````

Then something like this to create my object:

````ruby
def create
  pet_params = params[:pet]
  klass = pet_params.delete(:type).safe_constantize
  if klass && klass <= Pet
    klass.create(pet_params)
  end
end
````

But I'm taking any simpler way to create STI object without deserializing the JsonAPI resource type !

**Comment by beauby** [comment]
```ruby
{
  id: '1',
  type: 'content--user-photos',
  attributes: {
    type: 'Content::UserPhoto',
    title: 'A day at the beach'
  }
}
```
This is not valid per the spec. A field cannot be named `type`.

gauthier-delacroix:
@beauby I aggree, this was an example of what I **don't** want to do.

That said, `type` is the default STI type field. My proposal is to deserialize this field from the JSONAPI resource type.

beauby:
@gauthier-delacroix My bad, I had only skimmed through the issue.

bf4:
@gauthier-delacroix 

> The type field is a record attribute since it's the only way to store the object type in the DB.

This is not true.  Rails let's you configure the STI field and set it whatever you want. Like I said: 

> The rails sti field is a record's attribute.  I usually name it something other than type to avoid this sort of confusion

You also wrote:

> JsonAPI has a resource type field which is the same information so there's no confusion IMHO.

This is true, but only because you have an record that has a field that is the class of the record, and that we often use the record's class to derive the JSONAPI `type`.

Now, I understand exactly what you are trying to do and why.  However, what I don't fully understand yet is if your API design makes sense in the first place.  That's why I'm asking what the resource, route, and controller look like.

The API example you gave is that you have a (I assume contrived) `/pets` resource that lets you create any kind of pet you want.  I have a hard time imagining what business rule that could possibly represent.  The resource I think you actually have is something like `/photos`, which is why you might have 'Content::UserPhoto'. I realize that I am now repeating what I wrote in my original review: https://github.com/rails-api/active_model_serializers/pull/1995#discussion_r91287755

I'm spending time on this issue because it's interesting, but I'm not sure it's a problem AMS should solve or even a problem at all.

Now, assuming there's a `/photos` endpoint that lets you upload different kinds of photos. And, as it turns out, no matter what type of photo it is, you persist the same fields.  So, you can write all the photos to the same table, and use polymorphism to model the different behavior.

Someone is uploading a photo. Let's imagine the data:

```http
POST /photos HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": {
    "type": "photos",
    "attributes": {
      "title": "A day at the beach",
      "file": "someblob"
      "attached_id": "5",
      "attached_type": "user"
    },
  }
}
```

and then in the controller

```ruby
def create
    deserialized_params = ActionController::Parameters.new(
        ActiveModelSerializers::Deserialization.jsonapi_parse!(params)
      )
    photo_params = deserialized_params.permit(:id, :title, :attached_id, :attached_type)
    photo = Photo.new(photo_params)
    if photo.save
      render jsonapi: photo, status: :created
    else
      render jsonapi: photo, status: :bad_request, serializer: ActiveModel::Serializer::ErrorSerializer
    end
end
```

So, I don't see a need for special behavior.

beauby:
@gauthier-delacroix If you need configurable deserialization you could give a try to the `jsonapi-deserializable` gem, which is the deserialization component of a lib I wrote ([jsonapi-rb](http://jsonapi-rb.org)). I believe it covers your use-case via configuration.

gauthier-delacroix:
@bf4 I better understand our misunderstanding when you write this:

> as it turns out, no matter what type of photo it is, you persist the same fields.

As I said, I'm using Mongoid, and I'm using it for one single reason: The ability to have STI models with different attributes/relations than the root one.

I understand that it's not an AR use case (this gem isn't called "ActiveRecordSerializers" anyway) but STI could be useful with AR models too, bringing specific methods or features, and it's necessary to specify the type (actually the class) somewhere when we create them.

My real use case is a CDN management app, which includes lots of namespaced models, some with STI like `CDN::Purge` which is declined to:

- `CDN::Purge::Full` => No additional attribute
- `CDN::Purge::Custom` => `:inclusions` and `:exclusions` array attributes
- `CDN::Purge::URLCollection` => `:collection` relation to a `CDN::URLCollection` object that is also used for cache rules.

In your example, you're assuming a single controller always create the same kind of objects, which is not true in my case. The model has to be provided in the POST payload, and I don't see any interest in using something else than the JSONAPI resource attribute, especially since it's exactly the role [JSONAPI spec](http://jsonapi.org/format/#document-resource-object-identification) gives to the `type` member:

> The `type` member is used to describe resource objects that share common attributes and relationships.

About this:

> The rails sti field is a record's attribute. I usually name it something other than type to avoid this sort of confusion

That's why it's named `:_type` in Mongoid, and why I've made it configurable in my PR ;o)

@beauby I'm using AMS for years and I can't change my serialisation process now but anyway, AMS is highly configurable, perfectly handles JSONAPI serialization, and implements deserialization. I'm currently using my fork with success and my only options if my PR isn't merged will be to monkey patch the deserialization feature (I don't like staying on a fork).

beauby:
@gauthier-delacroix I was only talking about deserialization (which is very recent in AMS), but whatever floats your boat :+1:.

bf4:
@gauthier-delacroix 

> I'm currently using my fork with success and my only options if my PR isn't merged will be to monkey patch the deserialization feature (I don't like staying on a fork).

So, the reason I'm asking all these questions is because I don't want to increase the complexity of AMS without making sure it's necessary.  It's good to be more conservative in a library than in your own app. :)

While we're discussing this, is there a change that would remove your need for the monkey patch?

Can you give me an example of a payload containing something like  `CDN::Purge::Full`, `CDN::Purge::Custom`, `CDN::Purge::URLCollection` and what you're doing with the deserialized resources?

gauthier-delacroix:
Sorry, I thought my STI models description was enough.

Here are some JASONAPI + STI payloads:

````ruby
{
	"data": {
		"type": "cdn--purge--full",
		"attributes": {
			"name": "Full Purge",
			"persistent": true
		}
	}
}
````


````ruby
{
	"data": {
		"type": "cdn--purge--custom",
		"attributes": {
			"name": "Custom Purge",
			"persistent": true,
			"inclusions": [
				"\\.jpg$",
				"\\.gif$",
				"\\.png$"
			],
			"exclusions": []
		}
	}
}
````

````ruby
{
	"data": {
		"type": "cdn--purge--url-collection",
		"attributes": {
			"name": "URL Collection Purge",
			"persistent": true
		},
		"relationships": {
			"url-collection": {
				"data": [{
					"id": "55eea7a3676f6b5589040000",
					"type": "coreye--cdn--url-collection"
				}]
			}
		}
	}
}
````

Now, with my PR, I can deserialize them like this:

````ruby
  def create
    deserialized = ActiveModelSerializers::Deserialization.jsonapi_parse!(params, sti_type_key: :_type)
    klass = deserialized.delete(:_type).safe_constantize
    if klass && klass <= CDN::Purge
      if purge = klass.create(deserialized)
        [...rendering actions...]
      end
    end
  end
````
I have to admit that I add the class name in `:_type` to delete it after deserialization because it's not directly handled by Mongoid on instanciation (I wish it does...).

My real need is to deserialize the resource type from the payload, but it's not necessary to deserialize it into the attributes hash.

Maybe you'd prefer having a distinct method to do this, but I'd really like to have this in the derserialization class to access the `transform_keys` method which does half of the job.

**Comment by gauthier-delacroix** [comment]
I've been able to avoid a monkey patch using this callback as `transform_keys` is documented as private but is actually public:

````ruby
prepend_before_action(only: :create) do
  deserialize = ActiveModelSerializers::Adapter::JsonApi::Deserialization
  separator   = ActiveModelSerializers.config.jsonapi_namespace_separator
  klass = params[:data][:type].gsub(separator, '/') rescue nil
  klass = deserialize.transform_keys(klass, {})
  klass = klass.to_s.classify.safe_constantize
  if klass && klass <= CDN::Purge
    parsed = deserialize.parse!(params)
    params[:cdn_purge] = parsed
  end
end
````
(my real callback is actually in `ApplicationController`, with dynamic root class and params key)

Of course, I rescue `ActiveModelSerializers::Adapter::JsonApi::Deserialization::InvalidDocument` to an `:unprocessable_entity` error.

Now it's up to you to know if you want to handle the STI pattern or not!

**Comment by bf4** [comment]
> is documented as private but is actually public:
usually that means not to rely on it :)

> Now it's up to you to know if you want to handle the STI pattern or not!

@gauthier-delacroix Actually, what you've described has nothing to do with STI.  STI is an implementation detail of how you manage persistence.  What you've describe is an endpoint, perhaps `/cdn_purge`, that you POST different resources to.  You're posting resources of different types: `cdn--purge--url-collection`, `cdn--purge--full`, `cdn--purge--custom`.  If this were an STI scenario, I would have expected you to be using polymorphism by posting only one resource type `cdn--purge` and the attributes would tell the controller which behavior (class) to use.  The fact that this sentinel is your 'inheritance column' is kind of besides the point.

It appears to me that what you're trying to do isn't something AMS should be doing, though you do have the tools to do it if you want, so I'll be closing this PR.  Please feel free to continue discussion.

You wrote:

```ruby
 klass = params[:data][:type].gsub(separator, '/') rescue nil
  klass = deserialize.transform_keys(klass, {})
  klass = klass.to_s.classify.safe_constantize
  if klass && klass <= CDN::Purge
    parsed = deserialize.parse!(params)
    parsed[:_type]     = klass.name
    params[:cdn_purge] = parsed
  end
```

A little code review:

- `rescue nil` is a huge anti-pattern.  What are you trying to rescue (presumably not all of StandardError) and why do you want to continue if there's an exception? If type isn't set, you're payload is invalid.
- We have case transform utilities in AMS that you might want to use directly. see https://github.com/rails-api/active_model_serializers/blob/v0.10.3/lib/active_model_serializers/key_transform.rb
- `klass.to_s.classify.safe_constantize` is a security risk. you shouldn't be eval'ing user-submitted data.  A better pattern would be to have a lookup of valid resources.  e.g. `klass = lookup.fetch('cdn--purge--url-collection') # => CDN::Purge::UrlCollection`

NullVoxPopuli:
what's the TL;DR on this?

gauthier-delacroix:
@NullVoxPopuli JSONAPI resource type deserialization to handle STI/polymorphic object creation. Probably my use case wasn't as wise as I thought or I didn't explain it clearly enough but I'll go on with a callback that allows the same behavior.

@bf4 This is STI because `CDN::Purge::Custom`, `CDN::Purge::Full`, etc. inherits from `CDN::Purge`. Each final class only has a few different fields from the root one, and doesn't justify a specific controller.

Quickly about the callback review (I didn't PR this):

- `rescue nil` is a shortcut to avoid checking presence of `params[:data]` and `params[:data][:type]`. `transform_keys` and  `safe_constantize` would return `nil` too and `if klass` would stop execution. But I should actually re-raise an `ActiveModelSerializers::Adapter::JsonApi::Deserialization::InvalidDocument` since I've a global rescue on it.
- I'll change to use KeyTransform directly. I've just took my PR code from inside the class but it needs a little more cleanup.
- `klass.to_s.classify.safe_constantize` is safe since I check that the constant inherits from `CDN::Purge` (I have other similar cases so I need to be DRY, but it's an app specific concern).

Thank you for your time anyway !

bf4:
>  This is STI because 
yes, I know what STI is. But STI is a persistance detail that should be be reflected in the API design.  If this is a polymorphic endpoint, fine, that makes sense, but if if STI shouldn't be an HTTP  layer consideration. This kind of thing is why I think ActiveRecord encourages bad design.
> `rescue nil` is a shortcut to avoid checking presence
really, don't do that. I'm surprised you got it through code review
> klass.to_s.classify.safe_constantize is safe since I check that the constant
yeah, I regret not mentioning that I saw that.  Still, not a good idea or practice. That's how you get hacked.
>  have other similar cases so I need to be DRY
not sure what you mean by DRY in this context. this summarizes my thoughts on the matter, and a common misunderstanding of 'Don't Repeat Yourself'. https://speakerdeck.com/bf4/david-chelimsky-rubyconf-2010-maintaining-balance-while-reducing-duplication

gauthier-delacroix:
`rescue nil` wasn't in the PR, only in my app specific workaround callback ;o)

bf4:
@gauthier-delacroix I know. I was just offering a friendly code review.

Issue #943, opened 2015-06-05
Is it possible to pass dynamic value into cache key?


eliduke:
I am trying to cache an AppLabel serializer, but I'm running into a problem with locales. The app labels should return different values depending on the current user's app_language, so that needs to be factored in with regard to cache keys.

Here is my serializer:

```
class AppLabelSerializer < ActiveModel::Serializer
  cache key: 'app_label', expires_in: 3.hours
  attributes :id, :key, :label, :label_plural

  def key
    object.app_label_dictionary.key
  end
end
```

I tried interpolating into the key like this:

`cache key: "#{scope.app_language.name}/app_label", expires_in: 3.hours`

And I keep getting an `undefined method 'scope'` error. I double checked that the interpolation is working with this:

`cache key: "#{ 1 + 1 }/app_label", expires_in: 3.hours`

And that works just fine. I double checked that the scope is being passed into the serializer with this: 

```
class AppLabelSerializer < ActiveModel::Serializer
  cache key: "app_labels", expires_in: 3.hours
  attributes :id, :key, :label, :label_plural, :cache_key_language

  def key
    object.app_label_dictionary.key
  end

  def cache_key_language
    scope.app_language.name
  end
end
```

And the app language is correctly being returned. So, I tried passing the `cache_key_language` into the cache key like this:

`cache key: "#{ cache_key_language }/app_label", expires_in: 3.hours`

But that returns `undefined method 'cache_key_language'`.

I'm stumped. Does anyone have any ideas?

groyoh:
I could provide a PR to allow cache key to be a Proc but I would like to know @joaomdmoura thoughts on this. Until then, monkey patch is the only solution here IMO. Here are two way that you could try:
* Proc monkey patch
```ruby
class Proc
  def to_s
    call
  end
end
class WhateverSerializer < ActiveModel::Serializer
 cache key: ->(){ @serializer.scope.whatever }
end
```
* Adapter monkey patch
```ruby
module ActiveModel
  class Serializer
    class Adapter
      def cache_key
        key = @klass._cache_key
        key = @cached_serializer.instance_eval { key.call } if key.is_a?(Proc)
        key ? "#{key}/#{@cached_serializer.object.id}-#{@cached_serializer.object.updated_at}" : @cached_serializer.object.cache_key
      end
    end
  end
end
class WhateverSerializer < ActiveModel::Serializer
 cache key: ->(){ scope.whatever }
end
```
[...]

PR #1997, opened 2016-12-09
Allow passing a proc to the :key option of ActiveModel::Serializer::c…

tinbka:
#### Purpose
Allow overriding cache key prefix using ActiveModel::Serializer::cache (it won't work in the current version).  
Allow respecting serializer instance options in caching mechanism.

#### Changes
Allowed passing a proc to the :key option of ActiveModel::Serializer::cache  
Made use of this value as a priority over object.cache_key which, in fact, is always defined in ActiveRecord making  
Added an example to ::cache method.

#### Caveats
Although I've functionally tested the result in the context of my project, I couldn't have found a similar example in /test/cache_test.rb and couldn't have understand how to make a valid test for it.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/issues/943

### DIFF
--- a/lib/active_model/serializer/concerns/caching.rb
+++ b/lib/active_model/serializer/concerns/caching.rb
@@ -7,7 +7,7 @@ module ActiveModel
       included do
         with_options instance_writer: false, instance_reader: false do |serializer|
           serializer.class_attribute :_cache         # @api private : the cache store
-          serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key.  Ignored if the serializable object defines #cache_key.
+          serializer.class_attribute :_cache_key     # @api private : when present, is first item in cache_key. If is a Proc, will be evaluated in context of a serializer instance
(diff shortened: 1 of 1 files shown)


lib/active_model/serializer/concerns/caching.rb
```diff
@@ -270,15 +281,17 @@ def expand_cache_key(parts)
         ActiveSupport::Cache.expand_cache_key(parts)
       end
 
-      # Use object's cache_key if available, else derive a key from the object
-      # Pass the `key` option to the `cache` declaration or override this method to customize the cache key
+      # Use a custom cache key defined via serializer's ::cache method if available,
+      # else use object's #cache_key.
+      # Pass the `key` option to the `cache` declaration or override this method to customize the cache key.
       def object_cache_key
-        if object.respond_to?(:cache_key)
```
bf4: The order here is intentional

bf4:
@tinbka I'm closing this since I think your intention is really just to define an instance method `cache_key` on your serializer, and I haven't heard back.

PR #1998, opened 2016-12-12
WIP: Restore legacy ams::model behavior

bf4:
Needed to make #1982 and #1984 not a breaking change.  

- Successfully tested in an app that broke in various ways on master without this PR, but works with it
- Why this is wip?  now we need tests for with and without the prepended module, for both the 'new' and 'legacy' behavior
- Implementation-wise: Why it's prepended on inherited? That way, it can be easily disabled.
- Next steps: I intend to do something similar for caching, so that it is optional to include caching concerns in the poro, where it doesn't usually make sense.

### DIFF
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -1,6 +1,5 @@
-# ActiveModelSerializers::Model is a convenient
-# serializable class to inherit from when making
-# serializable non-activerecord objects.
+# ActiveModelSerializers::Model is a convenient superclass when making serializable non-activerecord objects.
+# It also serves as documentation of an implementation that satisfies ActiveModel::Serializer::Lint::Tests.
 module ActiveModelSerializers
   class Model
     include ActiveModel::Serializers::JSON
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -9,7 +9,7 @@ module ActiveModelSerializers
     # Configuration to avoid a breaking change with older versions of this class which lacked defined attributes.
     # Previous behavior was: 1) initialized attributes were the
     class_attribute :attributes_are_always_the_initialization_data, instance_writer: false, instance_reader: false
-    self.attributes_are_always_the_initialization_data = true
+    self.attributes_are_always_the_initialization_data = false # remove this commit, just for demonstration
     module AttributesAreAlwaysTheInitializationData
       def initialize(attributes = {})
         @initialized_attributes = attributes && attributes.symbolize_keys
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -6,8 +6,28 @@ module ActiveModelSerializers
     include ActiveModel::Serializers::JSON
     include ActiveModel::Model
 
-    # Configuration to avoid a breaking change with older versions of this class which lacked defined attributes.
-    # Previous behavior was: 1) initialized attributes were the
+    # @method :attributes
+    #   - unset attributes will appear in nil values in the attributes hash
+    #   - instance attributes can ONLY be changed by accessor methods.
+    #   - the return value of the attributes method is now frozen. since it is always derived, trying to mutate it is strongly discouraged :)
+    #
+    # @method :attributes_are_always_the_initialization_data
+    #
+    # When true, the deprecate attribute behavior is mixed in.
+    # Defaults to +true+, in order to avoid breaking changes with older versions of this class which lacked defined attributes.
+    #
+    # However, new apps probably want to change the default to false, and old apps want to create a special new model subclass where it is false for use in new models.
+    #
+    # Specifically, the deprecated behavior this re-adds is:
+    #   - the attributes hash returned by `#attributes` will always be the value of the attributes passed in at initialization,
+    #     regardless of what attr_* methods were defined.
+    #   - unset attributes, rather than being nil, will not be included in the attributes hash, even when an `attr_accessor` exists.
+    #   - the only way to change the change the attributes after initialization is to mutate the `attributes` directly.  Accessor methods
+    #     will NOT mutate the attributes.  (This is the bug that led to the change).
+    #
+    # For now, the Model only supports the notion of 'attributes'. In the tests, we do support 'associations' on the PORO,
+    # in order to adds accessors for values that should not appear in the attributes hash, as we model associations.
+    # However, it is not yet clear if it makes sense for a PORO to have associations outside of the tests.
     class_attribute :attributes_are_always_the_initialization_data, instance_writer: false, instance_reader: false
     self.attributes_are_always_the_initialization_data = true
 
@@ -20,6 +40,9 @@ module ActiveModelSerializers
       super
     end
 
+    # @method :attribute_names
+    # Is only available as a class-method since the ActiveModel::Serialization mixin in Rails
+    # uses an +attribute_names+ local variable, which may conflict if we were to add instance methods here.
     class_attribute :attribute_names, instance_writer: false, instance_reader: false
     # Initialize +attribute_names+ for all subclasses.  The array is usually
     # mutated in the +attributes+ method, but can be set directly, as well.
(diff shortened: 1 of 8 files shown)


lib/active_model_serializers/model.rb
```diff
@@ -6,7 +6,36 @@ class Model
     include ActiveModel::Serializers::JSON
     include ActiveModel::Model
 
-    class_attribute :attribute_names
+    # Configuration to avoid a breaking change with older versions of this class which lacked defined attributes.
+    # Previous behavior was: 1) initialized attributes were the
+    class_attribute :attributes_are_always_the_initialization_data, instance_writer: false, instance_reader: false
+    self.attributes_are_always_the_initialization_data = false # remove this commit, just for demonstration
```
NullVoxPopuli: remove this commit?

test/action_controller/serialization_test.rb
```diff
@@ -456,7 +456,7 @@ def use_adapter?
         end
       end
 
-      def test_render_event_is_emmited
```
NullVoxPopuli: so many spelling errors, nice finds.

lib/active_model_serializers/model.rb
NullVoxPopuli: So it looks like if someone wanted the old functionality, they'd just set `attributes_are_always_the_initialization_data` to true at the top of their class inheriting from AMS::Model?

lib/active_model_serializers/model.rb
```diff
@@ -6,7 +6,36 @@ class Model
...
+
+      def attributes
+        @initialized_attributes
+      end
+    end
+
+    def self.inherited(subclass)
+      if subclass.attributes_are_always_the_initialization_data
+        unless subclass.included_modules.include?(AttributesAreAlwaysTheInitializationData)
+          subclass.prepend(AttributesAreAlwaysTheInitializationData)
```
NullVoxPopuli: This is curious. Is this changing the default `include Module` functionality to make the module be a kind of superclass to the class using the module?

lib/active_model_serializers/model.rb
NullVoxPopuli: But, I don't see an `include AttributesAreAlwaysTheInitializationData` anywhere. Do you want to add tests for this?

lib/active_model_serializers/model.rb
beauby: `prepend` is basically the same as `include` except it inserts the module below instead of above the class in the inheritance chain, such that the module can override the class' methods.

NullVoxPopuli: oh ok, I thought it went the other way.

lib/active_model_serializers/model.rb
bf4: sorry to be unclear, I just added [Hey look it passes](bba724c211384f13682323d15424afb7e56ad63d) to show that I didn't break the existing test suite when 'false'. I'm going to remove that commit :)

lib/active_model_serializers/model.rb
bf4: @NullVoxPopuli this is just a compatibility layer to avoid a breaking change.  I should add some docs around setting it true vs. false and how to do both at same time and why :)

lib/active_model_serializers/model.rb
```diff
@@ -6,7 +6,21 @@ class Model
     include ActiveModel::Serializers::JSON
     include ActiveModel::Model
 
-    class_attribute :attribute_names
+    # Configuration to avoid a breaking change with older versions of this class which lacked defined attributes.
+    # Previous behavior was: 1) initialized attributes were the
```
bf4: what in the world happened to this comment?

bf4: I think what it's trying to say is that the attributes hash was *always* the value of the attributes passed in, regardless of what `attr_*` methods were defined.  This change now lets you define the attributes so that 1) unset attributes are now nil and 2) you can change the attributes after initialization by using the attr_writer, rather than hackily mutating the return value of the `attributes` method 3) the return value of the `attributes` method is now frozen. since it is always derived, trying to mutate it is strongly discouraged :)

test/fixtures/poro.rb
```diff
@@ -1,3 +1,4 @@
+ActiveModelSerializers::Model.attributes_are_always_the_initialization_data = false
```
bf4: see https://github.com/rails-api/active_model_serializers/pull/1998/files#r93893962 but the test poro superclass can't have the legacy behavior included because there's no simple way to remove it.  Instead, the test poro superclass has the 'new' behavior, but, subclasses of it, by default, will have the 'old' behavior.  Basically, it kind of extends the AMS::Model in the test context so that we can test it with the legacy behavior on and off.

test/support/poro.rb
```diff
@@ -0,0 +1,15 @@
+module ActiveModelSerializersWithoutLegacyModelSupport
+  module_function
+
+  def poro_without_legacy_model_support(superklass = ActiveModelSerializers::Model, &block)
```
bf4: see https://github.com/rails-api/active_model_serializers/pull/1998/files#r93894191

test/active_model_serializers/model_test.rb
```diff
@@ -20,7 +20,7 @@ def test_initialization_with_string_keys
     end
 
     def test_attributes_can_be_read_for_serialization
-      klass = Class.new(ActiveModelSerializers::Model) do
+      klass = poro_without_legacy_model_support do
```
bf4: ref https://github.com/rails-api/active_model_serializers/pull/1998/files#r93894191

test/action_controller/adapter_selector_test.rb
```diff
@@ -3,6 +3,14 @@
 module ActionController
   module Serialization
     class AdapterSelectorTest < ActionController::TestCase
+      Profile = poro_without_legacy_model_support(::Model) do
```
bf4: ref https://github.com/rails-api/active_model_serializers/pull/1998/files#r93894191

lib/active_model_serializers/model.rb
```diff
@@ -67,5 +81,21 @@ def self.lookup_ancestors
...
+
+    module AttributesAreAlwaysTheInitializationData
+      def initialize(attributes = {})
+        @initialized_attributes = attributes && attributes.symbolize_keys
+        super
+      end
+
+      # Defaults to the downcased model name.
+      def id
+        @initialized_attributes.fetch(:id) { self.class.model_name.name && self.class.model_name.name.downcase }
```
bf4: probably not a good default, but that was the old behavior.

lib/active_model_serializers/model.rb
```diff
@@ -67,5 +81,21 @@ def self.lookup_ancestors
...
+      end
+
+      def attributes
+        @initialized_attributes
```
bf4: yup, they can be mutated https://github.com/rails-api/active_model_serializers/pull/1998/files#r93893962

lib/active_model_serializers/model.rb
```diff
@@ -37,7 +51,7 @@ def initialize(attributes = {})
     # +attributes+ are returned frozen to prevent any expectations that mutation affects
     # the actual values in the model.
     def attributes
-      attribute_names.each_with_object({}) do |attribute_name, result|
+      self.class.attribute_names.each_with_object({}) do |attribute_name, result|
```
bf4: removed the instance reader/writer since the ActiveModel::Serialization mixin in Rails uses an `attribute_names` local variable, which I found because it conflicted in some scenario I've forgotten

test/support/poro.rb
bf4: An example test helper [I derived from one I wrote in some unmerged commits](https://github.com/rails-api/active_model_serializers/blob/ed1904358767697883f6832ac4c9eb833b82978a/test/support/poro.rb)

bf4:
Closed by https://github.com/rails-api/active_model_serializers/pull/2022

PR #2010, opened 2016-12-21
Updated isolated tests to assert correct behavior.

akshah123:
Updated isolated tests to assert correct behavior and removed a test case that wasn't really testing anything.

#### Purpose
Tighten up some test cases

#### Changes
Fixed assertion criteria for test cases to look for correct behavior

#### Caveats
Removed test isn't really needed

#### Related GitHub issues
N/A

### DIFF
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -12,7 +12,7 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
     end
 
     def render_with_jsonapi_renderer
-      unlocked_params = ActiveSupport::HashWithIndifferentAccess.new(params)
+      unlocked_params = Rails::VERSION::MAJOR >= 5 ? params.to_unsafe_h : params
       attributes = unlocked_params[:data].present? ? unlocked_params[:data][:attributes] : {}
       author = Author.new(attributes)
       render jsonapi: author
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -12,7 +12,8 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
     end
 
     def render_with_jsonapi_renderer
-      attributes = params[:data].present? ? params[:data][:attributes] : {}
+      unlocked_params = ActiveSupport::HashWithIndifferentAccess.new(params)
+      attributes = unlocked_params[:data].present? ? unlocked_params[:data][:attributes] : {}
       author = Author.new(attributes)
       render jsonapi: author
     end
@@ -64,6 +65,7 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
       assert_equal 500, response.status
+      assert response.request.env['action_dispatch.exception'].is_a?(ActionView::MissingTemplate)
     end
 
     def test_jsonapi_parser
@@ -109,7 +111,6 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      puts "RESPONSE #{response.body}"
       assert_equal expected, JSON.parse(response.body)['data']['attributes']
     end
 
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -60,14 +60,6 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
     end
 
     def test_jsonapi_renderer_not_registered
-      expected = {
-        'data' => {
-          'attributes' => {
-            'name' => 'Johnny Rico'
-          },
-        'type' => 'authors'
-        }
-      }
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -65,7 +65,7 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
       assert_equal 500, response.status
-      assert response.request.env['action_dispatch.exception'].is_a?(ActionView::MissingTemplate)
+      assert response.request.env['action_dispatch.exception'].is_a?(ActionView::MissingTemplate) if response.request.present?
     end
 
     def test_jsonapi_parser
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -109,6 +109,7 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
+      puts "RESPONSE #{response.body}"
       assert_equal expected, JSON.parse(response.body)['data']['attributes']
     end
 
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -2,6 +2,7 @@ require 'support/isolated_unit'
 require 'minitest/mock'
 require 'action_dispatch'
 require 'action_controller'
+require 'pry'
 
 class JsonApiRendererTest < ActionDispatch::IntegrationTest
   include ActiveSupport::Testing::Isolation
@@ -65,6 +66,7 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
       assert_equal 500, response.status
+      assert_equal '', response.body
       assert response.request.env['action_dispatch.exception'].is_a?(ActionView::MissingTemplate) if response.request.present?
     end
 
@@ -106,12 +108,23 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
     end
 
     def test_jsonapi_renderer_registered
-      expected = { 'name' => 'Johnny Rico' }
+      expected = {
+        'data' => {
+          'id' => 'author',
+          'type' => 'authors',
+          'attributes' => { 'name' => 'Johnny Rico' },
+          'relationships' => {
+            'posts' => { 'data' => nil },
+            'roles' => { 'data' => nil },
+            'bio' => { 'data' => nil }
+          }
+        }
+      }
 
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert_equal expected, JSON.parse(response.body)['data']['attributes']
+      assert_equal expected.to_json, response.body
     end
 
     def test_jsonapi_parser
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -2,7 +2,6 @@ require 'support/isolated_unit'
 require 'minitest/mock'
 require 'action_dispatch'
 require 'action_controller'
-require 'pry'
 
 class JsonApiRendererTest < ActionDispatch::IntegrationTest
   include ActiveSupport::Testing::Isolation
--- a/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
+++ b/test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
@@ -16,6 +16,11 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       render jsonapi: author
     end
 
+    def render_with_jsonapi_fields
+      post = Post.new(params[:data][:attributes])
+      render jsonapi: post, fields: {posts: ['title']}
+    end
+
     def parse
       self.class.last_request_parameters = request.request_parameters
       head :ok
@@ -58,21 +63,6 @@ class JsonApiRendererTest < ActionDispatch::IntegrationTest
       assert_nil parsers[Mime[:jsonapi]]
     end
 
-    def test_jsonapi_renderer_not_registered
-      expected = {
-        'data' => {
-          'attributes' => {
-            'name' => 'Johnny Rico'
-          },
-          'type' => 'users'
-        }
-      }
-      payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
-      headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
-      post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert expected, response.body
-    end
-
     def test_jsonapi_parser
       assert_parses(
         {},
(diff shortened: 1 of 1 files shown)


bf4:
rubocop

```
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:21:37: C: Style/SpaceInsideHashLiteralBraces: Space inside { missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      render jsonapi: post, fields: {posts: ['title']}
                                    ^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:21:54: C: Style/SpaceInsideHashLiteralBraces: Space inside } missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      render jsonapi: post, fields: {posts: ['title']}
                                                     ^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:104:18: C: Style/SpaceInsideHashLiteralBraces: Space inside { missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      expected = {'name' => 'Johnny Rico'}
                 ^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:104:42: C: Style/SpaceInsideHashLiteralBraces: Space inside } missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      expected = {'name' => 'Johnny Rico'}
                                         ^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:109:56: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)
      assert_equal expected, JSON.parse(response.body)["data"]["attributes"] 
                                                       ^^^^^^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:109:64: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)
      assert_equal expected, JSON.parse(response.body)["data"]["attributes"] 
                                                               ^^^^^^^^^^^^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:109:77: C: Style/TrailingWhitespace: Trailing whitespace detected. (https://github.com/bbatsov/ruby-style-guide#no-trailing-whitespace)
      assert_equal expected, JSON.parse(response.body)["data"]["attributes"] 
                                                                            ^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:113:18: C: Style/SpaceInsideHashLiteralBraces: Space inside { missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      expected = {'title'  => 'This is a test' }
                 ^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:113:28: C: Style/SpaceAroundOperators: Operator => should be surrounded by a single space. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      expected = {'title'  => 'This is a test' }
                           ^^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:117:56: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)
      assert_equal expected, JSON.parse(response.body)["data"]["attributes"]
                                                       ^^^^^^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:117:64: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)
      assert_equal expected, JSON.parse(response.body)["data"]["attributes"]
                                                               ^^^^^^^^^^^^
```

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -58,21 +63,6 @@ def test_jsonapi_parser_not_registered
...
-          'attributes' => {
-            'name' => 'Johnny Rico'
-          },
-          'type' => 'users'
-        }
-      }
-      payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
-      headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
-      post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert expected, response.body
```
bf4: so, this should have raised an exception when hitting `render_with_jsonapi_renderer` which calls` render jsonapi: json` but there's no `jsonapi` renderer registered. good test to have

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -111,18 +101,20 @@ def test_jsonapi_parser_registered
...
-          'type' => 'users'
-        }
-      }
+      expected = {'name' => 'Johnny Rico'}
+
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert expected, response.body
+      assert_equal expected, JSON.parse(response.body)["data"]["attributes"]
```
bf4: why not just `assert_equal expected, JSON.parse(respond.body)`?

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -111,18 +101,20 @@ def test_jsonapi_parser_registered
...
-      assert expected, response.body
+      assert_equal expected, JSON.parse(response.body)["data"]["attributes"] 
+    end
+
+    def test_jsonapi_renderer_registered_withfields
+      expected = {'title'  => 'This is a test' }
+      payload = '{"data": {"attributes": {"title": "This is a test","body": "Of an emergency alert system."}, "type": "posts"},"fields": { "posts" : ["title"] } }'
+      headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
+      post '/render_with_jsonapi_fields?fields[posts][]=title', params: payload, headers: headers
+      assert_equal expected, JSON.parse(response.body)["data"]["attributes"]
```
bf4: I don't think this test belongs here.  (typo in name as well).  This test appears to be testing usage of the jsonapi renderer, where all this file is doing is testing registering it, which is why it's isolated

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: That was because actual response has tons of other things added to it like meta, relationships which was not part of this test and would break anytime a new element is added.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: the problem is that it raises an exception while trying to create a new author since params are not parsed correctly without this.  So, I can setup another api method.

I think i have a way to change controller to test this case.  Let me take a shot.

akshah123: @bf4 i can't get the test controller to return the actual exception.  I tried using rescue_from but i think since the error is with render it just returns 500 with blank body.

I think for now, I am just going to assert that response.status = 500

akshah123:
@bf4 I made requested changes. Let me know what you think about failure check.

bf4:
@akshah123 

```
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:63:7: W: Lint/UselessAssignment: Useless assignment to variable - expected. (https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars)
      expected = {
      ^^^^^^^^
test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:68:9: C: Style/AlignHash: Align the elements of a hash literal if they span more than one line.
        'type' => 'authors'
```

:(  did you check the CI failures?

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -12,7 +12,8 @@ class << self
     end
 
     def render_with_jsonapi_renderer
-      author = Author.new(params[:data][:attributes])
+      attributes = params[:data].present? ? params[:data][:attributes] : {}
```
bf4: why? how could this not be set and why would we want that fallback if not set?

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -64,13 +65,13 @@ def test_jsonapi_renderer_not_registered
           'attributes' => {
             'name' => 'Johnny Rico'
           },
-          'type' => 'users'
+        'type' => 'authors'
```
bf4: why? type looks correct to me

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -111,18 +112,12 @@ def test_jsonapi_parser_registered
...
-          'type' => 'users'
-        }
-      }
+      expected = { 'name' => 'Johnny Rico' }
+
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert expected, response.body
+      assert_equal expected, JSON.parse(response.body)['data']['attributes']
```
bf4: Why only checking part of response. did this fail when changing `assert` to `assert_equal`? perhaps should have be `expected.to_json`

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -64,13 +65,13 @@ def test_jsonapi_renderer_not_registered
...
           },
-          'type' => 'users'
+        'type' => 'authors'
         }
       }
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert expected, response.body
+      assert_equal 500, response.status
```
bf4: can we make an assertion about the nature of the failure by matching on the response.body?

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: this is because when adapter is not set as jsonapi, params are not set.  This is there so that it fails on render and not on trying to set Author object.  Since we use the same controller in case where it should succeed and fail, I wanted to make sure this part doesn't cause a failure.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: I just wanted it to match what was being returned from controller.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: I couldn't get response.body to show the exception.  It kept returning empty body.  I think this probably has something to do with it failing on render.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: it did fail. The problem is that response.body contains a lot of stuff then just data.attributes and type.  It includes relationships, id and other things.

akshah123: here is the actual response.body
```json
{
    "data": {
        "id": "author",
        "type": "authors",
        "attributes": {
            "name": "Johnny Rico"
        },
        "relationships": {
            "posts": {
                "data": null
            },
            "roles": {
                "data": null
            },
            "bio": {
                "data": null
            }
        }
    }
}
```

akshah123:
@bf4 i have made some changes so that the tests pass. It took a while since i ran into issues with different versions of rails and had a hard time trying to reproduce the failing tests locally.  I also made modification to check for exact exception when that information is available.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -59,18 +61,11 @@ def test_jsonapi_parser_not_registered
...
-          },
-          'type' => 'users'
-        }
-      }
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
       headers = { 'CONTENT_TYPE' => 'application/vnd.api+json' }
       post '/render_with_jsonapi_renderer', params: payload, headers: headers
-      assert expected, response.body
+      assert_equal 500, response.status
+      assert response.request.env['action_dispatch.exception'].is_a?(ActionView::MissingTemplate) if response.request.present?
```
bf4: when is `response.request.present?` not true?

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
bf4: looks good. 

can you add an assertion for the empty body?

 If you know what the exception actually, was, would love for you to paste it and stack trace here.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
bf4: Oh, this is because of the PORO `Author` having relationships that are explicit now, whereas before they were sort of squishy...

Fix is we should be explicit what our model and serializer are:

```ruby
class JsonApiRendererTest < ActionDispatch::IntegrationTest
  include ActiveSupport::Testing::Isolation
  class Author < ::Model
    attributes :name
  end
    type: 'users'
    attribute :id
    attribute :name
  end
```

I think that's a better expectation, no?

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -12,7 +12,9 @@ class << self
     end
 
     def render_with_jsonapi_renderer
-      author = Author.new(params[:data][:attributes])
+      unlocked_params = Rails::VERSION::MAJOR >= 5 ? params.to_unsafe_h : params
+      attributes = unlocked_params[:data].present? ? unlocked_params[:data][:attributes] : {}
```
bf4: So, I gather `Author.new(params[:data][:attributes])` is failing for some reason. perhaps we should instead `author = Author.new(params.require(:data).require(:attributes))`?  I'm not sure what the need is to cast the ActionController::Parameters or when `data` wouldn't be present. 

In actual usage I would use the JsonApi deserializer here.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: From what I can tell, it changes on a specific rails version.  At some 4.x version, we start getting response.request but we dont' have that for 4.1.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: i think so. But I can't get your code to work.  The dependency required is getting a bit too much.  For now, i have updated expected object to include all relationships so that we check for exact output.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: You are right that params[:data] is not set at all and the reason is that app doesn't know how to parse provided data. 

`.require` will not help since it doesn't see data at all. Also, require will cause problem when jsonapi is not registered since it doesn't see the actual value.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
akshah123: actual exception message:

```
Missing template json_api_renderer_test/test/render_with_jsonapi_renderer with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby]}. Searched in:
```

backtrace 

```
["test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:19:in `render_with_jsonapi_renderer'",
 "test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb:68:in `test_jsonapi_renderer_not_registered'"]
```

akshah123:
@bf4 made some suggested changes. However, now the build is failing for something unrelated to my code. I can't reproduce this failure locally even with travis-ci's docker image.  Do you have any idea on what I can do to get it passing?

bf4:
@akshah123 re: https://github.com/rails-api/active_model_serializers/pull/2010#issuecomment-269057408 these failures are resolved on master. If you rebase, they should go away.

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
bf4: oh, so params is `{}` ? Then that's probably still a good failure :)

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -113,16 +109,21 @@ def test_jsonapi_parser_registered
...
-          },
-          'type' => 'users'
+          'id' => 'author',
+          'type' => 'authors',
+          'attributes' => { 'name' => 'Johnny Rico' },
+          'relationships' => {
+            'posts' => { 'data' => nil },
+            'roles' => { 'data' => nil },
+            'bio' => { 'data' => nil }
+          }
```
bf4: https://github.com/rails-api/active_model_serializers/pull/2010#discussion_r93787539

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
bf4: awesome! thanks!

test/active_model_serializers/register_jsonapi_renderer_test_isolated.rb
```diff
@@ -113,16 +109,21 @@ def test_jsonapi_parser_registered
...
         }
       }
+
       payload = '{"data": {"attributes": {"name": "Johnny Rico"}, "type": "authors"}}'
```
bf4: I guess we should include a uuid here. how about `"id": ""36c9c04e-86b1-4636-a5b0-8616672d1765"` will make the expectation simpler. JSON API recommends that if a client wants to create an id on post, to use a uuid.

You've been working on this PR a while now, and the main goal of it is done. Thanks! I can do some of the other things I'd like in a followup :)

PR #2022, opened 2017-01-09
Fix model attribute accessors

bf4:
#### Purpose

Replaces https://github.com/rails-api/active_model_serializers/pull/1998 and #1984 with an opt-in change, rather that breaking change with backwards compatibility.

#### Changes

Add new opt-in behavior for fixing the attributes change via accessors bug.

#### Caveats

It will still need to be reconciled with master, which I'll take care of once this is reviewed.

### DIFF
--- a/lib/active_model_serializers/model.rb
+++ b/lib/active_model_serializers/model.rb
@@ -6,6 +6,17 @@ module ActiveModelSerializers
     include ActiveModel::Serializers::JSON
     include ActiveModel::Model
 
+    # Declare names of attributes to be included in +sttributes+ hash.
+    # Is only available as a class-method since the ActiveModel::Serialization mixin in Rails
+    # uses an +attribute_names+ local variable, which may conflict if we were to add instance methods here.
+    #
+    # @overload attribute_names
+    #   @return [Array<Symbol>]
+    class_attribute :attribute_names, instance_writer: false, instance_reader: false
+    # Initialize +attribute_names+ for all subclasses.  The array is usually
+    # mutated in the +attributes+ method, but can be set directly, as well.
+    self.attribute_names = []
+
     # Easily declare instance attributes with setters and getters for each.
     #
     # All attributes to initialize an instance must have setters.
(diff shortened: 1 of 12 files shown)


lib/active_model_serializers/model.rb
```diff
@@ -25,10 +36,57 @@ class Model
     #   @param names [Array<String, Symbol>]
     #   @param name [String, Symbol]
     def self.attributes(*names)
+      self.attribute_names |= names.map(&:to_sym)
       # Silence redefinition of methods warnings
       ActiveModelSerializers.silence_warnings do
         attr_accessor(*names)
       end
+      return if included_modules.include?(DeriveAttributesFromNamesAndFixAccessors)
```
bf4: warning code needs some thought. might be better in a separate pr

lib/active_model_serializers/model.rb
```diff
@@ -25,10 +36,57 @@ class Model
...
+      # rubocop:enable Style/ClassVars
+      msg = ['NOTE: The next major version of ActiveModelSerializers will change how attributes are computed.',
+             "\nThe new behavior is available now by calling `derive_attributes_from_names_and_fix_accessors`",
+             "\nin any subclass. #{target} is still using the deprecated behavior.",
+             "\nCalled from #{ActiveModelSerializers.location_of_caller.join(':')}.\n"]
+      warn msg.join.to_s
+    end
+
+    # Opt-in to breaking change
+    def self.derive_attributes_from_names_and_fix_accessors
```
bf4: Anyone think of a better name for this?

lib/active_model_serializers/model.rb
```diff
@@ -25,10 +36,57 @@ class Model
...
+      end
+    end
+
+    module DeriveAttributesFromNamesAndFixAccessors
+      def self.included(base)
+        # NOTE that +id+ will always be in +attributes+.
+        base.attributes :id
+      end
+
+      # Override the initailize method so that attributes aren't processed.
```
bf4: typo

test/cache_test.rb
```diff
@@ -27,44 +32,83 @@ def cache_key
...
       cache_store.clear
       @comment        = Comment.new(id: 1, body: 'ZOMG A COMMENT')
-      @post           = Post.new(title: 'New Post', body: 'Body')
+      @post           = Post.new(id: 'post', title: 'New Post', body: 'Body')
       @bio            = Bio.new(id: 1, content: 'AMS Contributor')
-      @author         = Author.new(name: 'Joao M. D. Moura')
+      @author         = Author.new(id: 'author', name: 'Joao M. D. Moura')
       @blog           = Blog.new(id: 999, name: 'Custom blog', writer: @author, articles: [])
       @role           = Role.new(name: 'Great Author')
       @location       = Location.new(lat: '-23.550520', lng: '-46.633309')
```
bf4: A bunch of the associations here aren't used and could probably be removed both in this file and in the poro fixtures file

test/cache_test.rb
```diff
@@ -330,7 +374,7 @@ def test_uses_file_digest_in_cache_key
     end
 
     def test_cache_digest_definition
-      assert_equal(::Model::FILE_DIGEST, @post_serializer.class._cache_digest)
+      assert_equal(FILE_DIGEST, @post_serializer.class._cache_digest)
```
bf4: not sure how I feel about this assertion.. maybe this should just be ` Digest::MD5.hexdigest(File.open(__FILE__).read)`

test/fixtures/poro.rb
```diff
@@ -1,5 +1,6 @@
 class Model < ActiveModelSerializers::Model
   FILE_DIGEST = Digest::MD5.hexdigest(File.open(__FILE__).read)
+  rand(2).zero? && derive_attributes_from_names_and_fix_accessors
```
bf4: make sure the test suite works the same with and without the accessors fix

Issue #1868, opened 2016-08-02
0.9.x meta_key has to be a symbol and will not work for a string


vinodlala:
For `meta_key`, use a symbol instead of a string.

The docs on http://www.rubydoc.info/gems/active_model_serializers/0.9.5 has this:

```
render json: @posts, serializer: CustomArraySerializer, meta_object: {total: 10}, meta_key: 'meta_object'
```

But that doesn't work.

This will work:

```
render json: @posts, serializer: CustomArraySerializer, meta_object: {total: 10}, meta_key: :meta_object
```

In `lib/active_model/serializer.rb`, check out the `initialize` method:

```
      @meta_key      = options[:meta_key] || :meta
      @meta          = options[@meta_key]
```

If `options[:meta_key]` is `nil`, `@meta_key` is `:meta` (a symbol, not a string).

Using the example in the docs, if `options[:meta_key]` is a string like `'meta_object'`, `@meta_key` is `'meta_object'` (a string, not a symbol).  

However, `options['meta_object']` does not exist. Only `options[:meta_object]` exists.

If `options[:meta_key]` is a symbol like `:meta_object`, `@meta_key` is `:meta_object`, which works.

bf4:
Thanks @vinodlala would you mind making a PR (against the 0-9-stable branch) to fix the docs? Would be much appreciated.

vinodlala:
Done - https://github.com/rails-api/active_model_serializers/pull/1869

Issue #1328, opened 2015-11-12
Meta key not being picked up !!


NeoElit:
Custom meta not showing up even with root on active_model_serializers (0.9.3).
```
render :json => @place, :serializer => PlaceSerializer, :meta => {:content => "10", :type => "success"}
```
is working fine and meta is retrieved, however with:
```
render :json => @place, :serializer => PlaceSerializer, :meta => {:content => "10", :type => "success"}, :meta_key => 'message'
```
it's not giving any meta or message.

NeoElit:
tried same with version 0.8.3 and worked fine.
Seems like an issue with 0.9.3

LoganD:
I'm using 0.9.3 with the following code and it works:
```      render json: data,
             root: 'results',
             serializer: serializer,
             meta: meta(request, @query, error),
             status: code```

My meta function returns an object, so trying making the object before you render.

**Comment by bf4** [comment]
@NeoElit thanks for the issue. Have you taken a look at the code to see what might be happening?

**Comment by NeoElit** [comment]
@bf4 following is the render code and response in test
```
$ render code:
render json: @place,
                 serializer: PlaceSerializer,
                 meta: {code: 200, message: "successful"},
                 meta_key: "status" and return

$ response: 
{"place"=>{"id"=>903798785, "title"=>"new toit"}}
```

**Comment by NeoElit** [comment]
@LoganD 
How did meta({hash}) works for you? It's throwing no method error for me.
Following is the render code and reponse
```
$ render code:
render json: @place,
                 serializer: PlaceSerializer,
                 meta: meta({code: 200, message: "successful"}), status: 200 and return

$ response:
NoMethodError:         NoMethodError: undefined method `meta' for #<Client::Api::V1::PlacesController:0xac5b764>
```

**Comment by LoganD** [comment]
@NeoElit Sorry for the confusion, meta was a function that I wrote.
```
def make_meta(message, code)
{
  code: code,
  message: message
}
end
```

then in the render `meta: make_meta(code, message)`.
Since you can return a status code as a separate param, you don't need to return that in the meta, unless that has a different meaning and is specific to your api. I made that function so that I can do customized formatting for the meta block since that render function is generic.

NeoElit:
@LoganD it doesn't solve the problem rt?  This method just returns the hash to render. However what i need is to change root/name of meta.

LoganD:
[...]

PR #2025, opened 2017-01-13
Fix doc about meta_key example

wangboguo:
#### Purpose
In the documentation, the example for changing the meta key seems incorrect. The example has a string, but only a symbol will work, and only the 'custom_meta'  works well rather than 'meta'. The version 0.9 had already fixed this little mistake.

#### Related GitHub issues
[#1868](https://github.com/rails-api/active_model_serializers/issues/1868)
[#1328](https://github.com/rails-api/active_model_serializers/issues/1328)

### DIFF
--- a/docs/general/rendering.md
+++ b/docs/general/rendering.md
@@ -95,7 +95,7 @@ render json: @post, meta: { total: 10 }
 The key can be customized using `meta_key` option.
 
 ```ruby
-render json: @post, meta: { total: 10 }, meta_key: "custom_meta"
+render json: @post, custom_meta: { total: 10 }, meta_key: :custom_meta
 ```
 
 `meta` will only be included in your response if you are using an Adapter that


groyoh:
This doc seems to be correct to me (ref. https://gist.github.com/groyoh/97438d130f4c27e227969f9747740054). Please open an issue and provide a failling snippet/app if this doesn't work for you. Thanks! :v:

bf4:
@wangboguo @groyoh It looks like this was intended to be a PR against the 0-9 branch?

groyoh:
>  `wangboguo  wants to merge 1 commit into rails-api:master`

> The version 0.9 had already fixed this little mistake.

@bf4 I don't think that was the intention but let's let @wangboguo answer that question. In the meantime, I'd rather let it closed as we do not have more info.

wangboguo:
I feel so sorry about this PR, I found that I made a mistake. My project is using version 0.9 but I forgot about it. Really sorry about it.

wangboguo:
And I've tested that everything goes well in 0.10.4

Issue #2029, opened 2017-01-14
Add an option to remove log messages


MSathieu:
I would like an option to turn this 
`[active_model_serializers] Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModelSerializers::Adapter::JsonApi (136.7ms)`
 log message of.

bf4:
@MSathieu Of course, we should reference that in https://github.com/rails-api/active_model_serializers/blob/93ca27fe444c3a940890cb7a4dedbd8369f481f3/docs/general/logging.md along with the [log subscriber](http://api.rubyonrails.org/v4.2/classes/ActiveSupport/LogSubscriber.html) and [active support notifications api](http://api.rubyonrails.org/v4.2/classes/ActiveSupport/Notifications.html).

It's unclear to me if you want to set AMS logger to `Logger.new(IO::NUL)` or just unsubscribe `ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)`  and maybe `ActiveSupport::LogSubscriber.log_subscribers.delete_if {|subscriber| subscriber.instance_of?(ActiveModelSerializers::Logging::LogSubscriber) }` though I haven't tried these myself.

MSathieu:
I want to set `ActiveModelSerializers.logger = Logger.new(nil)`

bf4:
@MSathieu did that resolve your issue?

MSathieu:
`ActiveModelSerializers.logger = Logger.new(nil)` doesnt works,
 `ActiveModelSerializers.logger = Logger.new(IO::NUL)` raises an error,
but `ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)` works perfectly

bf4:
@MSathieu Would you like to make a PR to improve the docs, especially based on how you went about looking for the info?

MSathieu:
I will do that.

PR #2035, opened 2017-01-17
Update logging documentation

MSathieu:
Include documentation on how to disable the logging.
#2029 @bf4

### DIFF
--- a/docs/general/logging.md
+++ b/docs/general/logging.md
@@ -13,7 +13,7 @@ You may customize the logger in an initializer, for example:
 ActiveModelSerializers.logger = Logger.new(STDOUT)
 ```
 
-You can also disable the logger, just put this in `config/application.rb`:
+You can also disable the logger, just put this in `config/initializers/active_model_serializers.rb`:
 
 ```ruby
 ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)
--- a/docs/general/logging.md
+++ b/docs/general/logging.md
@@ -12,3 +12,9 @@ You may customize the logger in an initializer, for example:
 ```ruby
 ActiveModelSerializers.logger = Logger.new(STDOUT)
 ```
+
+You can also disable the logger, just put this in `config/application.rb`:
+
+```ruby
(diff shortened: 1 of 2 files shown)


docs/general/logging.md
```diff
@@ -12,3 +12,9 @@ You may customize the logger in an initializer, for example:
 ```ruby
 ActiveModelSerializers.logger = Logger.new(STDOUT)
 ```
+
+You can also disable the logger, just put this in `config/application.rb`:
+
+```ruby
+ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)
```
bf4: It's a better practice to put this in an initializer in a Rails app.

`config/initializers/active_model_serializers.rb` is the usual location.

```ruby
ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)
```

and perhaps reference the relevant railtie https://github.com/rails-api/active_model_serializers/blob/0-10-stable/lib/active_model_serializers/railtie.rb#L23-L27 ?

bf4:
@MSathieu Thanks so much for this and the quick turnaround!

docs/general/logging.md
MSathieu: Why requiring it

MSathieu:
@bf4 Updated it to use an initializer

docs/general/logging.md
bf4: @MSathieu not exactly. It's actually the call to `Bundle.require` at the top of most `config/application.rb` and then only if the gem doesn't have `require: false`.

Also, relying on autoloading of your project's classes is pretty much required for code reloading to work. But that doesn't really apply to gems.  It's a good idea to require a gem before you use it. But this opinion is controversial enough that I won't tell you to include it :)

MSathieu:
@bf4 I have updated it to require active_model_serializers in the initializer

bf4:
@MSathieu would you mind adding a changelog entry for yourself under 'misc'?

MSathieu:
@bf4 Done

PR #2037, opened 2017-01-19
Document Model declared attributes

bf4:


docs/howto/serialize_poro.md
```diff
@@ -42,4 +42,32 @@ end
...
+Existing applications can create can use the fix and avoid breaking changes
+by making a superclass for new models. For example:
+
+```ruby
+class SerializablePoro < ActiveModelSerializers::Model
+  derive_attributes_from_names_and_fix_accessors
+end
+```
+
+So that `MyModel` above would inherit from `SerializablePoro`.
```
kurko: > MyModel above

there's no MyModel

docs/howto/serialize_poro.md
```diff
@@ -42,4 +42,32 @@ end
 
 The default serializer would be `MyModelSerializer`.
 
+*IMPORTANT*: There is a surprising behavior (bug) in the current implementation of ActiveModelSerializers::Model that
+prevents an accessor from modifying attributes on the instance.  The fix for this bug
+is a breaking change, so we have made an opt-in configuration.
```
kurko: Do we agree that 1.0 can have as many breaking changes as required? I think this method is a nasty duct tape that we should remove.

bf4: @kurko absolutely!  This is *only* to prevent a breaking change within `0.10.x`

bf4: I actually played with having a deprecation warning when the module *isn't* prepended, but didn't want to complicate the PR.

docs/howto/serialize_poro.md
```diff
@@ -42,4 +42,32 @@ end
...
+
+New applications should set:
+
+```ruby
+ActiveModelSerializers::Model.derive_attributes_from_names_and_fix_accessors
+```
+
+Existing applications can create can use the fix and avoid breaking changes
```
bf4: ~~can create~~ can use

docs/howto/serialize_poro.md
bf4: @kurko any objections to this PR

docs/howto/serialize_poro.md
```diff
@@ -42,4 +42,32 @@ end
...
+
+```ruby
+class SerializablePoro < ActiveModelSerializers::Model
+  derive_attributes_from_names_and_fix_accessors
+end
+```
+
+So that `MyModel` above would inherit from `SerializablePoro`.
+
+`derive_attributes_from_names_and_fix_accessors` prepends the `DeriveAttributesFromNamesAndFixAccessors`
```
bf4: Needs comment that this will be the default  behavior and the option removed in the next major release

PR #1947, opened 2016-10-17
Adds :have_valid_schema RSpec matcher

leonelgalan:
#### Purpose

Provides built-in matchers for RSpec users. Example:

``` ruby
require 'rails_helper'
require 'active_model_serializers/rspec'
ActiveModelSerializers.config.schema_path = 'spec/support/schemas'

RSpec.describe V1::PostsController do
  describe '#index' do
    before { get :index }

    describe 'response' do
      subject { response }
      it { is_expected.to have_valid_schema }
    end
  end
end
```
#### Changes

Adds modules:
- `ActiveModelSerializers::RSpecMatchers::Schema`
#### Caveats
- ~~No tests, I'm still trying to figure out how to run rspec next to minitest and translate _test/active_model_serializers/test/schema_test.rb_ to RSpec.~~
- ~~Couldn't use `autoload`, expected path would be _/active_model_serializers/r_spec/matchers_, as opposed to: _/active_model_serializers/rspec/matchers_~~. Doesn't apply anymore, see below.
- ~~Had to `require 'minitest/autorun'` on _lib/active_model_serializers/test/schema.rb_~~. Not anymore.

I could use some help solving this issues.
#### Related GitHub issues
- [#1470 - Rspec tests & assert_response_schema](https://github.com/rails-api/active_model_serializers/issues/1470)

PR #2046, opened 2017-02-06
Fixes bug in Test::Schema when using filter_parameters

leonelgalan:
By default all Rails apps are created with an initializer called _filter_parameter_logging.rb_ and its default contents are:

```ruby
Rails.application.config.filter_parameters += [:password]
```

If this line is executed, as it is on this PR, the class of `request.filtered_parameters` changes from `ActiveSupport::HashWithIndifferentAccess` to `Hash`. This causes `Schema::Test` not to be able to find schemas by controller path and action:

```
     ActiveModelSerializers::Test::Schema::MissingSchema:
       No Schema file at spec/support/schemas//.json
```

Not properly "mocking" the Rails app properly caused the bug to be hidden in this and [leonelgalan/rspec-active_model_serializers](https://github.com/leonelgalan/rspec-active_model_serializers) test suite.

This bug was first reported by me in another in https://github.com/rails-api/active_model_serializers/pull/1947#pullrequestreview-5497390, but at the time I couldn't produce failing tests.

### DIFF
--- a/test/active_model_serializers/test/schema_test.rb
+++ b/test/active_model_serializers/test/schema_test.rb
@@ -1,5 +1,7 @@
 require 'test_helper'
 
+Rails.application.config.filter_parameters += [:password]
+
 module ActiveModelSerializers
   module Test
     class SchemaTest < ActionController::TestCase
--- a/test/active_model_serializers/test/schema_test.rb
+++ b/test/active_model_serializers/test/schema_test.rb
@@ -1,7 +1,5 @@
 require 'test_helper'
 
-Rails.application.config.filter_parameters += [:password]
-
 module ActiveModelSerializers
   module Test
     class SchemaTest < ActionController::TestCase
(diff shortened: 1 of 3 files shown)


test/active_model_serializers/test/schema_test.rb
```diff
@@ -1,5 +1,7 @@
 require 'test_helper'
 
+Rails.application.config.filter_parameters += [:password]
```
bf4: @leonelgalan so, this is enough to make the test fail? (this should be set in the test app, I think)

leonelgalan: I moved into a configure block inside _test/support/rails_app.rb_

test/support/rails_app.rb
```diff
@@ -6,6 +6,8 @@ module ActiveModelSerializers
       config.active_support.test_order = :random
       config.action_controller.perform_caching = true
       config.action_controller.cache_store = :memory_store
+
+      config.filter_parameters += [:password]
```
bf4: do the tests fail with just this change, and not the fix in `lib/active_model_serializers/test/schema.rb` ?

leonelgalan: Yes, they will fail with the message above. I'm not sure how to change this setting for a single test to write a failure case.

PR #2055, opened 2017-02-17
jsonapi is deprecated, just use jsonapi-renderer

jaredbeck:
From the author of jsonapi:

> .. The jsonapi gem was previously just a bundle of jsonapi-serializer and jsonapi-renderer, and AMS is using only a helper class of jsonapi-renderer (namely JSONAPI::IncludeDirective). The AMS dependency was previously not properly pinned to a specific version, which I saw as a risk for many users, so I avoided updating this gem. Moreover, the name jsonapi being somewhat too generic for what this gem evolved into (namely jsonapi-rb, which bundles jsonapi-renderer and jsonapi-parser, along with serializers and deserializers, with tight integrations with various frameworks), I decided to stay away from it for fairness.
> TL;DR: Yes, people should use jsonapi-parser and jsonapi-renderer directly (or give a try to jsonapi-rb, depending on their needs).
> https://github.com/beauby/jsonapi/issues/54#issuecomment-280537900

We should also update jsonapi-renderer to the latest, currently 0.1.2, but I thought that should be a separate PR.

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -42,7 +42,8 @@ Gem::Specification.new do |spec|
   # 'minitest'
   # 'thread_safe'
 
-  spec.add_runtime_dependency 'jsonapi', '0.1.1.beta6'
+  # TODO: Latest jsonapi-renderer is 0.1.2
+  spec.add_runtime_dependency 'jsonapi-renderer', '0.1.1.beta1'
   spec.add_runtime_dependency 'case_transform', '>= 0.2'
 
   spec.add_development_dependency 'activerecord', rails_versions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Fixes:
 
 Misc:
 
+- [#2055](https://github.com/rails-api/active_model_serializers/pull/2055)
+  Replace deprecated dependency jsonapi with jsonapi-renderer. (@jaredbeck)
 - [#2021](https://github.com/rails-api/active_model_serializers/pull/2021) Make test attributes explicit. Tests have Model#associations. (@bf4)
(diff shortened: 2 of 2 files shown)


active_model_serializers.gemspec
```diff
@@ -42,7 +42,8 @@ Gem::Specification.new do |spec|
   # 'minitest'
   # 'thread_safe'
 
-  spec.add_runtime_dependency 'jsonapi', '0.1.1.beta6'
+  # TODO: Latest jsonapi-renderer is 0.1.2
+  spec.add_runtime_dependency 'jsonapi-renderer', '0.1.1.beta1'
```
bf4: oh cool. why lock it to this version?

jaredbeck: i didn't want to change the version in this PR. i'd be happy to do another PR that changes the version constraint.

bf4: ah, so that's the version in the gem. Sure, let's do this. Can you add a changelog under misc?

jaredbeck:
> Can you add a changelog under misc?

Added.

jaredbeck:
@bf4 Hi Benjamin, can you please release a new version with this change? Thanks!

bf4:
Sure. I assume you're running on master? @jaredbeck

jaredbeck:
> Sure. I assume you're running on master? @jaredbeck

No, but I can try master with my app's test suite, just a sec.

bf4:
Me too, just passed.

bf4:
Released 

B mobile phone

jaredbeck:
My app's test suite (admittedly limited) passes on 0.10.5, FWIW. Thanks for the release, Benjamin.

PR #2057, opened 2017-02-20
Update version constraint for jsonapi-renderer

jaredbeck:
Now that we are no longer using the deprecated `jsonapi` (https://github.com/rails-api/active_model_serializers/pull/2055) we can discuss updating the version constraint of `jsonapi-renderer`.

Currently (2017-02-20) the latest version is 0.1.2.

Why not use a version constraint like '~> 0.1.1'? Because we know of no reason why 0.1.1.beta1 cannot still be used. That said, we have done no research looking for such a reason.


bf4:
Great. @jaredbeck could you add a changelog under feature?

jaredbeck:
> Great. @jaredbeck could you add a changelog under feature?

Added.

Sorry Benjamin, I keep forgetting the changelog entry. I don't know why, as I keep changelogs on all my own projects!

PR #2075, opened 2017-03-13
BUG: Expose id writer not being set

bf4:


test/fixtures/poro.rb
```diff
@@ -1,7 +1,9 @@
 class Model < ActiveModelSerializers::Model
-  rand(2).zero? && derive_attributes_from_names_and_fix_accessors
-
-  attr_writer :id
+  if rand(2).zero?
+    derive_attributes_from_names_and_fix_accessors
+  else
+    attr_writer :id
```
bf4: I found when using `ActiveModelSerializers::Model.derive_attributes_from_names_and_fix_accessors ` and `class SomeName < ActiveModelSerializers::Model; attributes :something; end` in my own code, that `SomeName.new(id: 'uuid')` failed, as there was no attr_writer

PR #2076, opened 2017-03-13
Testing collection serializer is broken ?

romaintb:
Hey guys.

This is not a proper pull request, this is more a test to show the issue I am facing right now.

WDYT ?

#### Purpose

Contrary to what is stated here : https://github.com/rails-api/active_model_serializers/blob/a032201a91cbca407211bca0392ba881eef1f7ba/docs/howto/test.md#controller-serializer-usage : `assert_serializer` does not work on collection.

Error of the test implemented : 
```
ActiveModelSerializers::Test::SerializerTest#test_supports_specifying_serializers_with_a_multi_serializer_class [/home/romain/work/ams/test/active_model_serializers/test/serializer_test.rb:28]:
expecting <ProfileSerializer> but rendering with <["ActiveModel::Serializer::CollectionSerializer"]>
```

### DIFF
--- a/test/active_model_serializers/test/serializer_test.rb
+++ b/test/active_model_serializers/test/serializer_test.rb
@@ -6,6 +6,12 @@ module ActiveModelSerializers
       include ActiveModelSerializers::Test::Serializer
 
       class MyController < ActionController::Base
+        def render_multi_using_serializer
+          render json: [Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1'),
+                        Profile.new(name: 'Name 2', description: 'Description 2', comments: 'Comments 2')],
+                 each_serializer: ProfileSerializer
+        end
+
         def render_using_serializer
           render json: Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
         end
@@ -16,6 +22,11 @@ module ActiveModelSerializers
       end
 
       tests MyController
+
+      def test_supports_specifying_serializers_with_a_multi_serializer_class
+        get :render_multi_using_serializer
+        assert_serializer ProfileSerializer
+      end
 
       def test_supports_specifying_serializers_with_a_serializer_class
         get :render_using_serializer


test/active_model_serializers/test/serializer_test.rb
```diff
@@ -17,6 +23,11 @@ def render_some_text
 
       tests MyController
 
+      def test_supports_specifying_serializers_with_a_multi_serializer_class
+        get :render_multi_using_serializer
+        assert_serializer ProfileSerializer
```
bf4: Perhaps what you want is `assert_each_serializer`?  I think it is correct to say the serializer is `ActiveModel::Serializer::CollectionSerializer`. In any case, you don't need any assertion when you provide the serializer explicitly.

romaintb: `assert_each_serializer` may be cool, yeah.
My test is dumb on purpose, and sure if I declare it explicitely 5 lines earlier it looks silly. But in our process, we usually don't have code yet when writing tests. And anyway we test it, declared explicitly or not.

(anyway, the doc was misleading to me, do you think it should be edited ?)

Thanks

bf4:
@romaintb Yes, the doc is wrong.  Honestly, I'd love to remove this feature. I think it's better to test the behavior rather than which serializer was called.

romaintb:
@bf4 here at $DAYJOB, we test both ways. We test the serializer on its own, and we tests controllers calls the right serializer. This let us have controller tests without "display" concerns.
Do you have something else in mind to replace the current behavior ? Or do you plan to remove this entirely ?

bf4:
@romaintb My plan is to make the serializer output a schema you can use in the request/integration test to validate the response.  (Prefer request/integration tests to controller tests if you can.  Controller tests can miss things in the stack and give you false confidence, and Rails is making them equally fast)

romaintb:
usually, something alongs those lines : 

```
class SdServiceSerializerTest < ActiveSupport::TestCase                                                                                                                                                                                        
  setup do                                                                                                                                                                                                                                     
    @sds = sd_services(:two)                                                                                                                                                                                                                   
    @serializer = SdServiceSerializer.new(@sds)                                                                                                                                                                                                
    @serialization = ActiveModelSerializers::Adapter.create(@serializer)                                                                                                                                                                       
  end                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                               
  test 'it should not expose secret field' do                                                                                                                                                                                                  
    res = JSON.parse(@serialization.to_json)                                                                                                                                                                                                   
    assert_not res.key?('secret')                                                                                                                                                                                                              
  end                                                                                                                                                                                                                                          
end

**Comment by romaintb** [comment]
usually something along those lines : 

```

romaintb:
usually something along those lines : 

```ruby
class FoobarSerializerTest < ActiveSupport::TestCase
  setup do
    @record = fixtures(:one)
    @serializer = FoobarSerializer.new(@record)
    @serialization = ActiveModelSerializers::Adapter.create(@serializer)
  end                                                                                                                                                                                                                                          

  test 'it should not expose secret field' do
    res = JSON.parse(@serialization.to_json)
    assert_not res.key?('secret')
  end
end
```

not very classy, but does the job

PR #2079, opened 2017-03-15
Backport caching of the constant lookup

dylanahsmith:
#### Purpose

Backport https://github.com/rails-api/active_model_serializers/pull/833 to the 0-9-stable branch

> Currently, each time a model is serializer is tries to look up the Serializer class using safe_constantize. This can be slow when serializing a large number of models.

#### Changes

> This adds a cache to return the serializer class for each class, which reduced the amount of time looking up the serializer classes.

This difference is that the cache is around `ActiveMode::Serializer::Utils._const_get` and uses Concurrent::Map (the successor to ThreadSafe::Cache) from concurrent-ruby.

### DIFF
--- a/lib/active_model/serializable/utils.rb
+++ b/lib/active_model/serializable/utils.rb
@@ -4,11 +4,13 @@ module ActiveModel
       extend self
 
       def _const_get(const)
-        begin
-          method = RUBY_VERSION >= '2.0' ? :const_get : :qualified_const_get
-          Object.send method, const
-        rescue NameError
-          const.safe_constantize
+        Serializer.serializers_cache.fetch_or_store(const) do
+          begin
+            method = RUBY_VERSION >= '2.0' ? :const_get : :qualified_const_get
+            Object.send method, const
+          rescue NameError
+            const.safe_constantize
+          end
         end
       end
     end
--- a/lib/active_model/serializable/utils.rb
+++ b/lib/active_model/serializable/utils.rb
@@ -4,13 +4,11 @@ module ActiveModel
       extend self
 
       def _const_get(const)
-        Serializer.serializers_cache.fetch_or_store(const) do
-          begin
-            method = RUBY_VERSION >= '2.0' ? :const_get : :qualified_const_get
-            Object.send method, const
-          rescue NameError
-            const.safe_constantize
-          end
+        begin
+          method = RUBY_VERSION >= '2.0' ? :const_get : :qualified_const_get
+          Object.send method, const
+        rescue NameError
+          const.safe_constantize
         end
       end
     end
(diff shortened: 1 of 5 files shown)


lib/active_model/serializer.rb
```diff
@@ -4,6 +4,7 @@
 require 'active_model/serializer/config'
 
 require 'thread'
+require 'concurrent/map'
```
rafaelfranca: concurrent is not an dependency of this project in this branch, so we need to add it as a dependency.

lib/active_model/serializable/utils.rb
```diff
@@ -4,13 +4,15 @@ module Utils
       extend self
 
       def _const_get(const)
-        begin
-          method = RUBY_VERSION >= '2.0' ? :const_get : :qualified_const_get
-          Object.send method, const
-        rescue NameError
-          const.safe_constantize
+        Serializer.serializers_cache.fetch_or_store(const) do
```
rafaelfranca: Can we cache in the call of `_const_get`? It doesn't  seems right to Utils know about `serializers_cache`.

dylanahsmith: There are two places where _const_get is called, so [ActiveModel::Serializable#namespace](https://github.com/rails-api/active_model_serializers/blob/c2565ed72623eb4ecc9d21149778aa160a9ad9b2/lib/active_model/serializable.rb#L36) and [ActiveModel::Serializer.serializer_for](https://github.com/rails-api/active_model_serializers/blob/c2565ed72623eb4ecc9d21149778aa160a9ad9b2/lib/active_model/serializer.rb#L68).  That is why I decided to move it into _const_get instead of doing it in ActiveModel::Serializer.serializer_for.  Should I be duplicating the code between both call sites?  Or only worry about the latter? Since I am not sure if the former is actually a performance problem.

rafaelfranca: Yeah. I'd duplicate the cache call to avoid coupling between the utils model and the `Serializer` constant.

dylanahsmith: ok, done

yosiat:
Hi!

I did something like this but I did two changes:
- the cache key wasn't the string that represents the type, it was a hash of resource and options in order to reduce allocations (not building strings) - https://github.com/yosiat/active_model_serializers/blob/serializer-cache/lib/active_model/serializer.rb#L65
- I added local cache to array serializer - if the user didn't specify with serializer we do lookup in the serializer cache everything, so I added cache inform of `Hash`
- Instead of `fetch_and_store`, I used `compute_if_absent`, which does the same but more atomic.

I can submit for those changes if you would like.

Issue #2067, opened 2017-03-04
[0.9] ActiveSupport payload is incorrect


yosiat:
Hi,

By reading the code (and then verifying this), I think I found a bug.
When I am subscribing to the serialize event I only get "serializer" in the payload and not all other instrumentation keys (object, scope, root, meta_key).

### Bug description

By looking at the instrument code:
https://github.com/rails-api/active_model_serializers/blob/0-9-stable/lib/active_model/serializable.rb#L51
```
payload = instrumentation_keys.inject({ serializer: self.class.name }) do |payload, key|
  payload[:payload] = self.instance_variable_get(:"@#{key}")
  payload
end
```
It keeps overriding the payload object with the current instrumentation key's value.

### Perf Impact

I found this code by looking at the object's allocations and found this method as the source of 50% of the allocations.
By replacing the payload to only be `{serializer: self.class.name}` - we get drop off 50% allocations and more ip/s.

Before:
```
Array: DefaultSerializer 726.6811271730311 ip/s  2657 objects
Array: PostFlatSerializer 975.9174166658823 ip/s  2059 objects
```

After:
```
Array: DefaultSerializer 819.9266362390441 ip/s  1957 objects
Array: PostFlatSerializer 1,321.0953937395007 ip/s  1009 objects
```

### Proposal
I suggest to remove instrumentation keys support and sending only the serializer.
Why?

- It's not working today, so nobody relies on it
- In 0.10 we send only `serializer` and `adapter` (If I am not mistaken)

What do you think?

yosiat:
I found a conversation about this in the original commit -  https://github.com/rails-api/active_model_serializers/commit/eaedcefa4e186f60907f50b0980d9cc07fb3a1d3#commitcomment-14831558

bf4:
@yosiat even if it's wrong I think it would be a breaking change... need to take a look

yosiat:
@bf4 how this is going to be a breaking change?
Currently, the payload is nil.
If you remove the payload you preserve your behavior.

bf4:
@yosiat good point

bf4:
strange that no one would have noticed. Been years, no?

yosiat:
@bf4 it looks like that someone noticed it.
https://github.com/rails-api/active_model_serializers/commit/eaedcefa4e186f60907f50b0980d9cc07fb3a1d3#commitcomment-14831558

bf4:
[...]

PR #2080, opened 2017-03-17
[0.9] Removing `instrumentation_keys` in order to fix the payload

yosiat:
Fixes - https://github.com/rails-api/active_model_serializers/issues/2067

By the way, if the instrumentation is always the same - `!serialize.active_model_serializers`,
I think a nice change would be to create a const for `INSTRUMENTATION_KEY` that will have `!serialize.active_model_serializers` as a frozen string value.
This improves performance & reduces allocations by a ton.

I can do that if the author wants that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### [0-9-stable](https://github.com/rails-api/active_model_serializers/compare/v0.9.5...0-9-stable)
 
+- [#2080](https://github.com/rails-api/active_model_serializers/pull/2080) remove `{ payload: nil }` from `!serialize.active_model_serializers` ActiveSupport::Notification. `payload` never had a value.  Changes, for example `{ serializer: 'ActiveModel::DefaultSerializer', payload: nil }` to be `{ serializer: 'ActiveModel::DefaultSerializer' }` (@yosiat)
+
 ### [v0.9.6 (2017-01-10)](https://github.com/rails-api/active_model_serializers/compare/v0.9.5...v0.9.6)
 
 - [#2008](https://github.com/rails-api/active_model_serializers/pull/2008) Fix warning on Thor. (@kirs)
--- a/test/unit/active_model/serilizable_test.rb
+++ b/test/unit/active_model/serilizable_test.rb
@@ -1,0 +1,50 @@
+require 'test_helper'
+
+module ActiveModel
+  class SerializableTest
+    class InstrumentationTest < Minitest::Test
+      def setup
+        @events = []
+
+        @subscriber = ActiveSupport::Notifications.subscribe('!serialize.active_model_serializers') do |name, start, finish, id, payload|
+          @events << { name: name, serializer: payload[:serializer] }
+        end
+      end
+
+      def teardown
+        ActiveSupport::Notifications.unsubscribe(@subscriber) if defined?(@subscriber)
+      end
+
+      def test_instruments_default_serializer
+        DefaultSerializer.new(1).as_json
+
+        assert_equal [{ name: '!serialize.active_model_serializers', serializer: 'ActiveModel::DefaultSerializer' }], @events
+      end
+
+      def test_instruments_serializer
+        profile = Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+        serializer = ProfileSerializer.new(profile)
+
+        serializer.as_json
+
+        assert_equal [{ name: '!serialize.active_model_serializers', serializer: 'ProfileSerializer' }], @events
+      end
+
+      def test_instruments_array_serializer
+        profiles = [
+          Profile.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1'),
+          Profile.new(name: 'Name 2', description: 'Description 2', comments: 'Comments 2')
+        ]
+        serializer = ArraySerializer.new(profiles)
+
+        serializer.as_json
+
+        assert_equal [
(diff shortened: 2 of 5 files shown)


bf4:
Looks fine to me. However, any thought of using `render.active_model_serializers` instead of `!serialize.active_model_serializers`? As you said, it's currently broken in 0.9.

Would you mind adding tests and yourself in the changelog?

cc @wagenet ref https://github.com/skylightio/skylight-ruby/blob/v1.1.0/lib/skylight/probes/active_model_serializers.rb#L34-L47

```ruby
          # End users could override as_json without calling super, but it's likely safer
          # than overriding serializable_array/hash/object.

          [::ActiveModel::Serializer, ::ActiveModel::ArraySerializer].each do |klass|
            klass.class_eval do
              alias as_json_without_sk as_json
              def as_json(*args)
                payload = { serializer: self.class }
                ActiveSupport::Notifications.instrument('render.active_model_serializers', payload) do
                  as_json_without_sk(*args)
                end
              end
            end
          end
```

yosiat:
@bf4 The instrumentation today works, you only get the `serializer` in the payload.
so changing the name is a breaking behaviour.

bf4:
Would you mind adding tests and yourself in the changelog?

yosiat:
@bf4 yes, I will add tests at the weekend.

yosiat:
@bf4 I saw that today we use the instrumentation in assert_serializer.
So If I broke somehow the instrumentation than assert_serializer will get broken, so I don't think there is a need for tests here.
If you still me want me to add some sanity tests, I will add.

By the way, here is the status of the benchmarks:

**0-9-stable:**
```
Single: DefaultSerializer 27187/ips; 76 objects
ArraySerializer 14801/ips; 130 objects
ArraySerializer: each_serializer: DefaultSerializer 18183/ips; 122 objects
FlatAuthorSerializer 27428/ips; 85 objects
ArraySerializer: each_serializer: FlatAuthorSerializer 19851/ips; 114 objects
AuthorWithDefaultRelationshipsSerializer 320/ips; 6190 objects
ArraySerializer: each_serializer: AuthorWithDefaultRelationshipsSerializer 331 ip/s; 6219 objects
```

**this branch:**
```
Single: DefaultSerializer          32682 ip/s         67 allocs/op
ArraySerializer    20902 ip/s         89 allocs/op
ArraySerializer: each_serializer: DefaultSerializer        24746 ip/s         81 allocs/op
FlatAuthorSerializer       36517 ip/s         62 allocs/op
ArraySerializer: each_serializer: FlatAuthorSerializer     29672 ip/s         68 allocs/op
AuthorWithDefaultRelationshipsSerializer             397 ip/s       5258 allocs/op
ArraySerializer: each_serializer: AuthorWithDefaultRelationshipsSerializer           397 ip/s       5264 allocs/op
```

bf4:
@yosiat Yes, I'd love tests!

test/unit/active_model/serilizable_test.rb
```diff
@@ -0,0 +1,48 @@
+require 'test_helper'
+
+module ActiveModel
+  class SerializableTest
+    class InstrumentationTest < Minitest::Test
+      def setup
+        @events = []
+
+        ActiveSupport::Notifications.subscribe("!serialize.active_model_serializers") do |name, start, finish, id, payload|
```
bf4: Better:

`@subscriber = ActiveSupport::Notifications.subscribe(.....` in the setup

and

`ActiveSupport::Notifications.unsubscribe(@subscriber) if defined?(@subscriber)` in the teardown.  less global.

test/unit/active_model/serilizable_test.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def teardown
+        ActiveSupport::Notifications.unsubscribe("!serialize.active_model_serializers")
+      end
+
+      def test_default_serializer
+        DefaultSerializer.new(1).as_json
+
+        assert_equal @events, [{ name: '!serialize.active_model_serializers', serializer: 'ActiveModel::DefaultSerializer' }]
```
bf4: expectation in assert_equal should be on the left http://docs.seattlerb.org/minitest/Minitest/Assertions.html

test/unit/active_model/serilizable_test.rb
```diff
@@ -0,0 +1,48 @@
...
+        ActiveSupport::Notifications.subscribe("!serialize.active_model_serializers") do |name, start, finish, id, payload|
+          @events << { name: name, serializer: payload[:serializer] }
+        end
+      end
+
+      def teardown
+        ActiveSupport::Notifications.unsubscribe("!serialize.active_model_serializers")
+      end
+
+      def test_default_serializer
```
bf4: maybe `test_instruments_default_serializer` here and below, I think would better description of the test

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### [0-9-stable](https://github.com/rails-api/active_model_serializers/compare/v0.9.5...0-9-stable)
 
+- [#2080](https://github.com/rails-api/active_model_serializers/pull/2080) remove `{ payload: nil }` from `!serialize.active_model_serializers` ActiveSupport::Notification. `payload` never had a value.  Changes, for example `{ serializer: 'ActiveModel::DefaultSerializer', payload: nil }` to be `{ serializer: 'ActiveModel::DefaultSerializer' }` (@yosiat)
```
bf4: 👍

PR #2085, opened 2017-03-24
Fix example code in `doc/general/getting_started.md`

tricknotes:
The `belongs_to` method should take relation name, not a foreign_key property.


bf4:
@tricknotes thanks!

PR #2089, opened 2017-03-28
Improve reflection internal interface

bf4:
Extracted from #2026

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -14,6 +14,19 @@ module ActiveModel
     #     end
     #     has_many :secret_meta_data, if: :is_admin?
     #
+    #     has_one :blog do |serializer|
+    #       meta count: object.roles.count
+    #       serializer.cached_blog
+    #     end
+    #
+    #     private
+    #
+    #     def cached_blog
+    #       cache_store.fetch("cached_blog:#{object.updated_at}") do
+    #         Blog.find(object.blog_id)
+    #       end
+    #     end
+    #
     #     def is_admin?
     #       current_user.admin?
     #     end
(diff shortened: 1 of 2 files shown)


lib/active_model/serializer/reflection.rb
```diff
@@ -14,6 +14,19 @@ class Serializer
...
+    #       serializer.cached_blog
+    #     end
+    #
+    #     private
+    #
+    #     def cached_blog
+    #       cache_store.fetch("cached_blog:#{object.updated_at}") do
+    #         Blog.find(object.blog_id)
+    #       end
+    #     end
```
bf4: I added a bunch of code comments here and elsewhere in the file re the actually interface since I'm always finding them a little hard to track down.. not sure if they're all good example or even belong here

lib/active_model/serializer/reflection.rb
```diff
@@ -113,8 +173,8 @@ def build_association(parent_serializer, parent_serializer_options, include_slic
         association_value = value(parent_serializer, include_slice)
         serializer_class = parent_serializer.class.serializer_for(association_value, reflection_options)
         reflection_options[:include_data] = include_data?(include_slice)
-        reflection_options[:links] = @_links
-        reflection_options[:meta] = @_meta
+        reflection_options[:links] = options[:links]
+        reflection_options[:meta] = options[:meta]
```
bf4: Removing dependency on instance variables makes it possible to just use reflection_options instead of the explicit passing of @_links and @_meta

bf4: ❓ confirm the reflection meta instance value changes on block eval before this PR

lib/active_model/serializer/reflection.rb
```diff
@@ -32,43 +45,82 @@ class Serializer
...
         super
-        @_links = {}
-        @_include_data = Serializer.config.include_data_default
-        @_meta = nil
+        options[:links] = {}
+        options[:include_data_setting] = Serializer.config.include_data_default
+        options[:meta] = nil
       end
 
+      # @api public
```
NullVoxPopuli: I'm a fan of these

lib/active_model/serializer/reflection.rb
bf4: Looked into it https://github.com/rails-api/active_model_serializers/pull/2092

PR #2092, opened 2017-03-31
Add reflection tests

bf4:
I wanted to test that evaluating reflection blocks mutates the reflection

Try to distinguish between the reflection and what happens when building the association (which evaluates the associated object)

- [ ] test reflection block methods
  - [x] link
  - [x] link block
  - [x] meta
  - [x] meta block
  - [x] meta in link block
  - [x] reflection link block does not hand href when not evaluated by json api adapter 'Link' object
  - [x] include_data
- [ ] test other instance methods
  - [ ] value
  - [ ] excluded? (`:if`, `:unless` options)
- [ ] test options
  - [ ] key 
  - [ ] polymorphic 
  - [ ] serializer 
  - [ ] virtual_value
  - [ ]  namespace

### DIFF
--- a/test/serializers/reflection_test.rb
+++ b/test/serializers/reflection_test.rb
@@ -213,6 +213,37 @@ module ActiveModel
         end
         assert_match(/undefined method `href'/, exception.message)
       end
+
+      def test_mutating_reflection_block_is_not_thread_safe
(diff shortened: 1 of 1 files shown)


bf4:
Merging as it provides value and I'm not coming back to it.

Issue #2002, opened 2016-12-15
Incorrect behavior with fields named after object methods


idoa01:
#### Expected behavior vs actual behavior
When using an attribute with an object method name (in my case `display`, but any other will cause the bug) the serializers will get the method's output instead of the value sent for serialization.

#### Steps to reproduce
*(e.g., detailed walkthrough, runnable script, example application)*

```rb
require 'active_model_serializers'

class SomeResource < ActiveModelSerializers::Model
  attr_accessor :id, :display, :first_name, :hash
end

class SomeSerializer < ActiveModel::Serializer
  attributes :id, :display, :first_name, :hash
end

> resource = SomeResource.new(id: 1, display: "Chuck Norris", first_name: "Chuck", hash: "1234")
> puts SomeSerializer.new(resource).to_json

<SomeSerializer:0x007f898fb03860>{"id":1,"display":null,"first_name":"Chuck","hash":-2648756963900511786}

> puts resource.attributes
{:id=>1, :display=>"Chuck Norris", :first_name=>"Chuck", :hash=>"1234"}
```

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 
ActiveModel::Serializer::VERSION => "0.10.3"

Output of `ruby -e "puts RUBY_DESCRIPTION"`: 
ruby 2.1.5p273 (2014-11-13 revision 48405) [x86_64-darwin14.0]

OS Type & Version: 
MacOS Sierra  10.12.1

Integrated application and version *(e.g., Rails, Grape, etc)*: 
None

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*

#### Additonal helpful information
*(e.g., Gemfile.lock, configurations, PR containing a failing test, git bisect results)*

looking into github, the bug is here:
https://github.com/rails-api/active_model_serializers/blob/95f832fd0e57ad5c960076c851bf553b88fb4430/lib/active_model/serializer.rb#L191-L192
inserted in commit 6370e5c72a71060feeb9e16b67fd0e1a5742bedc

maybe switching the order of the if/else will fix this?
Thanks!

PR #2093, opened 2017-03-31
Undef problematic Object methods

bf4:
Needs tests and references to issues

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -12,6 +12,9 @@ require 'active_model/serializer/lint'
 # reified when subclassed to decorate a resource.
 module ActiveModel
   class Serializer
+    undef_method :select, :display # These IO methods, which are mixed into Kernel,
+    # sometimes conflict with attribute names. We don't need these IO methods.
+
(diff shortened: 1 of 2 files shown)


idoa01:
resurfacing an old issue which I think can be solved here: https://github.com/rails-api/active_model_serializers/issues/2002

bf4:
@idoa01 not sure I follow

idoa01:
Oh, sorry, I thought that I didn't comment in the end :( 
(I didn't want to comment since I didn't check if the bug is still active in the latest release, but I see that the code is still there)

There is an old bug that I think can be addressed in the same manner - you can't have an attribute named `hash` since your code checks if the object `respond_to?` the attribute. with basic ruby objects method names, respond_to will always be true.

my bug originated with the `display` attribute that this pull request handles, but doesn't handle other methods like `hash` (see example in #2002)

you might be able to fix this with `undef_method` like in this pull request (but it's a more complicated issue, since undefing the `.hash` method might cause issues with equality)

a better solution would be to ignore all Object methods, something in the lines of:
```ruby
if respond_to?(attr) && method(attr).owner != Kernel
  send(attr)
else
end
```

PR #2094, opened 2017-03-31
Move association serialization to association

bf4:


bf4:
Merging as it is an improvement and I haven't come back to it

PR #2100, opened 2017-04-06
Fieldset support in Attributes/JSON adapters

kosmatov:
Adds the ability to specify the returned relation fields in Attributes and JSON adapters

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -379,6 +379,7 @@ module ActiveModel
     def serializable_hash(adapter_options = nil, options = {}, adapter_instance = self.class.serialization_adapter_instance)
       adapter_options ||= {}
       options[:include_directive] ||= ActiveModel::Serializer.include_directive_from_options(adapter_options)
+      options[:fields] ||= adapter_options[:fieldset] && adapter_options[:fieldset].fields_for(json_key)
       cached_attributes = adapter_options[:cached_attributes] ||= {}
       resource = fetch_attributes(options[:fields], cached_attributes, adapter_instance)
       relationships = resource_relationships(adapter_options, options, adapter_instance)
--- a/lib/active_model_serializers/adapter/attributes.rb
+++ b/lib/active_model_serializers/adapter/attributes.rb
@@ -2,7 +2,7 @@ module ActiveModelSerializers
   module Adapter
     class Attributes < Base
       def serializable_hash(options = nil)
-        options = serialization_options(options)
+        options ||= serialization_options(options)
         options[:fields] ||= instance_options[:fields]
         serialized_hash = serializer.serializable_hash(instance_options, options, self)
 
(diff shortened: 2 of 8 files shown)


bf4:
I like this.. will need to review a little more closely

lib/active_model_serializers/adapter/base.rb
```diff
@@ -78,6 +83,32 @@ def serialization_options(options)
       def root
         serializer.json_key.to_sym if serializer.json_key
       end
+
+      def fields_to_hash(fields)
+        case fields
+        when Hash
+          fields
+        when Array
```
BenMorganIO: To follow the ruby philosophy, I think it would be of a greater benefit if we used `respond_to?`. This way you can send in object that implement `Hash` or `Array`, but aren't necessarily those objects themselves.

lib/active_model_serializers/adapter/base.rb
```diff
@@ -46,6 +46,7 @@ def self.fragment_cache(cached_hash, non_cached_hash)
 
       def initialize(serializer, options = {})
         @serializer = serializer
+        @fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(fields_to_hash(options.delete(:fields)))
```
bf4: in jsonapi was

```
@fieldset = options[:fieldset] || ActiveModel::Serializer::Fieldset.new(options.delete(:fields))
```

why was `fields_to_hash` introduced here?

lib/active_model_serializers/adapter/json.rb
```diff
@@ -3,6 +3,7 @@ module Adapter
     class Json < Base
       def serializable_hash(options = nil)
         options = serialization_options(options)
+        instance_options[:fieldset] = fieldset
```
bf4: why is this here and not in the base adapter? 

why in the attributes adapter is it `options[:fields] ||= fields || instance_options[:fields]` both adapters do basically the same thing, no?

lib/active_model_serializers/adapter/attributes.rb
```diff
@@ -2,8 +2,8 @@ module ActiveModelSerializers
   module Adapter
     class Attributes < Base
       def serializable_hash(options = nil)
-        options = serialization_options(options)
-        options[:fields] ||= instance_options[:fields]
+        options ||= serialization_options(options)
```
bf4: why was `options =` changed to `options ||=`?

lib/active_model/serializer.rb
```diff
@@ -379,6 +379,7 @@ def associations(include_directive = ActiveModelSerializers.default_include_dire
     def serializable_hash(adapter_options = nil, options = {}, adapter_instance = self.class.serialization_adapter_instance)
       adapter_options ||= {}
       options[:include_directive] ||= ActiveModel::Serializer.include_directive_from_options(adapter_options)
+      options[:fields] ||= adapter_options[:fieldset] && adapter_options[:fieldset].fields_for(json_key)
```
bf4: This seems to duplicate logic in the adapter.

```
def fields
   fieldset && fieldset.fields_for(serializer.json_key)
end
```

which should it be?

Is the problem here that the adapter does `options.delete(:fields)`?

wasifhossain:
closing this as https://github.com/rails-api/active_model_serializers/pull/2223 is merged

PR #2101, opened 2017-04-06
Add another strategy for finding Serializers

tijn:
This strategy looks at the nesting modules of the current namespace.
For example:

    Api::V3::Frontend::AuthorsController => [
      Api::V3::AuthorSerializer, Api::AuthorSerializer]

As the example shows, this strategy will return ALL the module names
that surround the current namespace providing a straightforward way
to...

- use different serializers for specialized cases (for when you have
  similar controllers for the same resource but for different user
  roles);
- use a specialized serializer in certain controllers and a generic
  one in all the rest;
- a combination of the above.

#### Purpose

Allow having multiple specialized serializers while also allowing to reuse generic serializers. 

#### Changes

Added another loopup proc that returns an array of all the namespaces that surround the current one.

#### Caveats

You have to remember to change the lookup-setting in the config.

I'm sorry for not writing any tests for this. I don't really understand the format of the current lookup-chain test. All help is welcome. 😄


tijn:
closing this PR because I stopped caring 🤣

PR #2114, opened 2017-04-18
Fix to_prepare warning

kirs:
Fixes

```
DEPRECATION WARNING: to_prepare is deprecated and will be removed from Rails 5.1 (use ActiveSupport::Reloader.to_prepare instead) (called from block in <main> at
 /Users/kir/Projects/shopify/vendor/bundle/bundler/gems/active_model_serializers-1e04d1128bc7/lib/active_model_serializers.rb:17)
```

review @rafaelfranca

### DIFF
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -14,9 +14,6 @@ begin
       ActionController::Base.send(:include, ::ActionController::Serialization)
       ActionController::TestCase.send(:include, ::ActionController::SerializationAssertions)
     end
-    ActionDispatch::Reloader.to_prepare do
-      ActiveModel::Serializer.serializers_cache.clear
-    end
   end
 rescue LoadError
   # rails not installed, continuing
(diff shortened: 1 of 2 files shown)


lib/active_model_serializers.rb
```diff
@@ -14,7 +14,7 @@
       ActionController::Base.send(:include, ::ActionController::Serialization)
       ActionController::TestCase.send(:include, ::ActionController::SerializationAssertions)
     end
-    ActionDispatch::Reloader.to_prepare do
+    ActiveSupport::Reloader.to_prepare do
```
rafaelfranca: Can we move this to a Railtie and use the `to_prepare` in the application object to do this? This way it can support any Rails version, since ActiveSupport::Reloader doesn't exist on Rails 4.2 and 0.9 supports 4.2.

bf4:
@kirs @rafaelfranca OT: will you be at RailsConf next week? Would love to chat about AMS and friends and Rails

kirs:
I won't be there unfortunately. Have fun at Railsconf!

rafaelfranca:
I'll!

kirs:
@rafaelfranca I updated the PR 👌

rafaelfranca:
Thank you for fixing this

PR #2119, opened 2017-05-01
Return null resource object identifier for blank id

bf4:
Fix regression from 6e4152851554b33d8cdf607fe4e175dd1aafd1c7 where relationships with null data shows up with `:data=>{:id=>"", :type=>"locations"}`

```
       -:data => {:id=>"123", :type=>"places", :attributes=>{:status=>"live"}, :relationships=>{:location=>{:data=>nil}}},
       +:data => {:id=>"123", :type=>"places", :attributes=>{:status=>"live"}, :relationships=>{:location=>{:data=>{:id=>"", :type=>"locations"}}}},
```

Also, fix test where attributes were included when id was ""

```
  1) Failure:
  ActionController::Serialization::AdapterSelectorTest#test_render_using_adapter_override
  [test/action_c$ntroller/adapter_selector_test.rb:53]:
  --- expected
  +++ actual
  @@ -1 +1 @@
  -"{\"data\":{\"id\":\"\",\"type\":\"profiles\",\"attributes\":{\"name\":\"Name 1\",\"description\":\"Description 1\"}}}"
  +"{\"data\":null}"
```

### DIFF
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -295,20 +295,8 @@ module ActiveModelSerializers
 
       # {http://jsonapi.org/format/#document-resource-objects Document Resource Objects}
       def resource_object_for(serializer, include_slice = {})
-        resource_object = serializer.fetch(self) do
-          resource_object = ResourceIdentifier.new(serializer, instance_options).as_json
+        resource_object = data_for(serializer, include_slice)
 
-          requested_fields = fieldset && fieldset.fields_for(resource_object[:type])
-          attributes = attributes_for(serializer, requested_fields)
-          resource_object[:attributes] = attributes if attributes.any?
-          resource_object
-        end
-
-        requested_associations = fieldset.fields_for(resource_object[:type]) || '*'
-        relationships = relationships_for(serializer, requested_associations, include_slice)
-        resource_object[:relationships] = relationships if relationships.any?
-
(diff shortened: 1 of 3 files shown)


lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -295,20 +295,8 @@ def attributes_for(serializer, fields)
 
       # {http://jsonapi.org/format/#document-resource-objects Document Resource Objects}
       def resource_object_for(serializer, include_slice = {})
-        resource_object = serializer.fetch(self) do
```
bf4: refactored to `data_for` to decrease method size, per style

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -322,7 +310,10 @@ def resource_object_for(serializer, include_slice = {})
         # prs:
         #   https://github.com/rails-api/active_model_serializers/pull/1247
         #   https://github.com/rails-api/active_model_serializers/pull/1018
-        resource_object[:links] = links if links.any?
+        if (links = links_for(serializer)).any?
+          resource_object ||= {}
```
bf4: resource_object may be `nil`

InteNs:
what if i want to serialize an object via json_api adapter that doesn't have an ID?
I need this functionality for testing my PATCH requests.
Isn't it possible to just leave the ID part out of the serialized json, or is this against the spec?

PR #1082, opened 2015-08-25
Add option to select adapter by the request header

akaKuruma:
It resolves [#1062](https://github.com/rails-api/active_model_serializers/issues/1062)

PR #2122, opened 2017-05-02
Introduces per-request adapter switching based on the request mime.

fletcher91:
#### Purpose
The current behaviour is to always use the same adapter, this is not very desirable in more formal hypermedia conforming systems where content-negotiation determines the format which should be used. This PR introduces the option to let the system choose the adapter based on the negotiation results.

#### Changes
Content-type determination is left to ActionController::MimeResponds, so that users won't have to specify the adapter in every render call.

#### Caveats
* This implementation uses a non-existent virtual adapter `:mime` to let the subsystem know to look up the adapter dynamically, which may be considered inconsistent.
* By using MimeResponds, an implicit dependency between mime types and serializers is created, which might not be suitable for projects needing to use different adapters under the same mime type.
* I'm not aware how these changes will affect view caching.

#### Related GitHub issues
@remear's comment on PR https://github.com/rails-api/active_model_serializers/pull/1082 was the cause to reimplement this feature.


docs/general/adapters.md
```diff
@@ -36,6 +36,20 @@ The `Attributes` adapter does not include a root key. It is just the serialized
...
+To automatically switch between adapters based on the request, set the adapter to `:mime`
+
+```ruby
+  # In configuration
+  ActiveModelSerializers.config.adapter = :mime
+  
+  # In controller action
+  respond_to do |format|
+    format.json { render json: Model.first }     # Uses the JSON adapter
+    format.json_api { render json: Model.first } # Users the JSON:API adapter
```
bf4: 👍 

should be `jsonapi` not `json_api`, and `render jsonapi: model` rather than `render json: model`.

But really, this is just using the content type negotiation that respond_to already does. This would work without any changes to the PR, assuming mime -> adapter is 1:1

fletcher91: Would you prefer just a correction:

>To automatically switch between adapters based on the request, set the adapter to `:mime`:
>
> ```ruby
>   # In configuration
>   ActiveModelSerializers.config.adapter = :mime
>   
>   # In controller action
>   respond_to do |format|
>     format.json { render json: Model.first }     # Uses the JSON adapter
>     format.jsonapi { render jsonapi: Model.first } # Users the JSON:API adapter
>     format.bug { render json: Model.first } # Will raise `UnknownAdapterError`
>   end
> ```

or to just leave it  implicit:
> 
>To automatically switch between adapters [based on the request](http://api.rubyonrails.org/classes/ActionController/MimeResponds.html), set the adapter to `:mime`;
>
> ` ActiveModelSerializers.config.adapter = :mime`

bf4: Really, if you want to render different mime types, each mime type should be registered to its own formatter.  That way, you can do the below, which is the Rails-way IMHO and requires no changes to the codebase. 

```
 # In configuration
ActiveModelSerializers.config.adapter = :json
  
# In controller action
respond_to do |format|
    format.json { render json: model, serializer: ModelSerializer }     # Uses the JSON adapter
    format.jsonapi { render jsonapi: model, serializer: ModelSerializer } # Users the JSON:API adapter
    format.bug { render json: model, serializer: ModelSerializer } # Will raise something about bug not being a known mime type
  end
```

The respond_to block does the content type negotiation for you and usually we use a renderer format with the same name as the negotiated / acceptable mime types.

I think I'm making different assumptions than you, so please continue to discuss.

PR #2123, opened 2017-05-04
Fix readme in 0.9

jeffrey008:
There is no `@options` in 0.9, but `@serialization_options` instead. Changed readme to stop confusing people.

#### Changes
 `@options` => `@serialization_options`


bf4:
Thanks!

Issue #2125, opened 2017-05-06
0.10.6: type attribute not respected on belongs to relations


rmachielse:
When using a custom `type` on a serializer, the type is not used on nested belongs to relations.
This only happens with belongs_to relationships, has_many relationships are working fine.

```ruby
class Foo < ActiveRecord::Base
end

class Bar < ActiveRecord::Base
  belongs_to :foo
end

class FooSerializer < ActiveModel::Serializer
  type 'custom-type'
end

class BarSerializer < ActiveModel::Serializer
end
```

#### Expected behavior vs actual behavior

Expected:

```json
{ "data": { "relationships": { "foo": { "data": { "id": 1, "type": "custom-type" } } } }
```

Actual:

```json
{ "data": { "relationships": { "foo": { "data": { "id": 1, "type": "foos" } } } }
```

As it's working fine on 0.10.5, it seems to be a regression of 0.10.6

bf4:
@rmachielse Thanks for the bug report. Would you be interested in making a PR to fix this? I know what the problem is, but have limited bandwidth.

rmachielse:
@bf4 hmm so that logic would need to be reverted entirely? As you cannot know upfront if a custom type is used or not

bf4:
> As you cannot know upfront if a custom type is used or not

well, since the serializer specifies the type, and prior to the changes, the serializer was being inferred from the (associated) object, your options are 

1. specify the type on the association
1. specify the serializer on the association (my preference)
1. make the belongs_to behavior that avoids instantiating the associated object (just to look up the serializer) opt-in or opt-out. (This was added to reduce db queries)

rmachielse:
I understand options 1 and 2 but I do not prefer those as it involves some duplication. I assume it is a bug anyway, as this is a breaking change in a patch release. When it comes to option 3, wouldn't it be possible to instantiate the serializer without loading the associated object?

sdornan:
This also seems to happen with `id` on relationships. Even if you specify a serializer that defines an `id` method on the relationship, it doesn't seem to honor that method in the `relationships` block, but does in the `included` block.

Example:

```
...
"relationships" =>
{
    "brand"=>{"data"=>{"id"=> 3, "type"=>"brands"}},
},
"included" =>
[
[...]

PR #2121, opened 2017-05-02
Clear out master, move 0.10.x dev to the stable branch

bf4:
Status:

- ❗️ All existing PRs against master will need to be closed and re-opened against 0-10-stable, if so desired
- ❗️ Master, for the moment, won't have any released version of AMS on it.

Changes to 0.10.x maintenance:

- The 0.10.x version has become a huge maintenance version.  We had hoped to get it in shape for a 1.0 release, but it is clear that isn't going to happen.  Almost none of the maintainers from 0.8, 0.9, or earlier 0.10 are still working on AMS. We'll continue to maintain 0.10.x on the 0-10-stable branch, but maintainers won't otherwise be actively developing on it
  - We may choose to make a 0.11.x ( 0-11-stable) release based on 0-10-stable that just removes the deprecations.

Changes in this PR:

- The code and docs have been removed from master to signify that development against master has been suspended.  When choosing ways that people might be surprised by the current state of 0.10.x development, we thought it would be simpler to make master simply incompatible with 0.10.x PRs than to ensure none are merged.

What's happening to AMS:

- There's been a lot of churn around AMS since it began back in Rails 3.2 and a lot of new libraries are around and the JSON:API spec has reached 1.0.
- If there is to be a 1.0 release of AMS, it will need to address the general needs of serialization in much the way ActiveJob can be used with different workers.
- The next major release *is* in development. I am building and extracting it from an app I'm working on and thereby dogfooding it myself to ensure it to see if it makes sense.  I'm starting simple and avoiding, at least at the outset, all the complications in AMS version, especially all the implicit behavior from guessing the serializer, to the association's serializer, to the serialization type, etc.
- The basic idea is that models to serializers are a one to many relationship.  Everything will need to be explicit.  If you want to serialize a User with a UserSerializer, you'll need to call it directly.  The serializer will essentially be for defining a basic JSON:API resource object: id, type, attributes, and relationships. The serializer will have an as_json method and can be told which fields (attributes/relationships) to serialize to JSON and will likely *not* know serialize any more than the relations id and type.  Serializing anything more about the relations would require code that called a serializer. (This is still somewhat in discussion).
- If this works out, the idea is to get something into Rails that existing libraries can use.

Steps:

1. review this PR, clearing out master, moving 0-10 dev to 0-10-stable
1. merge it
1. make a PR from dev to master and work on it
 - add rails rubocop https://github.com/rails/rails/search?utf8=%E2%9C%93&q=rubocop&type=
[...]

PR #2128, opened 2017-05-09
Allow relationship ID to be overwritten for belongs_to relationships

greysteil:
#### Purpose
If the `id` attribute for a class isn't taken directly from the object when serializing it, it may be desirable for other classes that serialize a relationship with that class to overwrite the relationship IDs they serialize.

For example, suppose we have:

```(ruby)
class Repo < Model
  attributes :id, :github_id, :name
  associations :configs
end

class Config < Model
  attributes :id
  belongs_to :repo
end

class RepoSerializer < ActiveModel::Serializer
  attributes :id, :name

  has_many :update_configs

  def id
    object.github_id
  end
end

class ConfigSerializer < ActiveModel::Serializer
  attributes :id
  belongs_to :repo
end
```

In the above example, serializing a list of `Repo`s will give the `github_id` for each one, but serializing a `Config` will give the `id` for its parent repo.

#### Changes
I've updated the `belongs_to?` path of `Relationship#data_for_one` to use `read_attribute_for_serialization` instead of `object.send` to fetch the serialized foreign key. This allows the serialized relationship ID to be
overwritten using:

```(ruby)
class ConfigSerializer < ActiveModel::Serializer
  ...

  def repo_id
    object.repo.github_id
  end
end
```

#### Caveats
Ideally AMS would inspect the `RepoSerializer` when serializing the `Config`, and realise it can't just output the foreign key. Unfortunately, getting the serialization class for the child repo currently requires loading the record (via evaluating `lazy_assocation`), and loses the performance benefit of the existing `belongs_to?` path.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/issues/2125. (Not resolved by this, but the ID issue is at least improved, for no performance cost.)

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/relationship.rb
+++ b/lib/active_model_serializers/adapter/json_api/relationship.rb
@@ -45,7 +45,7 @@ module ActiveModelSerializers
         def data_for_one(association)
           if association.belongs_to? &&
               parent_serializer.object.respond_to?(association.reflection.foreign_key)
-            id = parent_serializer.object.send(association.reflection.foreign_key)
+            id = parent_serializer.read_attribute_for_serialization(association.reflection.foreign_key)
             type = association.reflection.type.to_s
             ResourceIdentifier.for_type_with_id(type, id, serializable_resource_options)
           else
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -165,6 +165,36 @@ module ActiveModel
         assert_equal expected, actual
       end
 
+      class ExternalBlog < Blog
+        attributes :external_id
+      end
+      class BelongsToExternalBlogModel < ::Model
+        attributes :id, :title, :external_blog_id
+        associations :external_blog
+      end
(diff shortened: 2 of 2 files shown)


greysteil:
@bf4 - hadn't see #2125 when I put this together and didn't realise this behaviour was considered a bug. I can change approach if you'd rather?

test/fixtures/poro.rb
```diff
@@ -146,6 +146,9 @@ class CustomBlogSerializer < ActiveModel::Serializer
   attribute :special_attribute
   has_many :articles
 end
+class ExternalBlog < Blog
+  attributes :external_id
+end
```
bf4: This can go in the `test/serializers/associations_test.rb`.  (This file is big enough :) )

bf4:
This looks fine to me. I don't mind hiding the object behind the serializer a little more :)

Just the one change

bf4:
Would mind closing this PR and opening a new one against 0-10-stable? per https://github.com/rails-api/active_model_serializers/pull/2121

greysteil:
Done. #2130.

PR #2135, opened 2017-05-16
WIP - Add ability to pass block to `type` directive

zeiv:
#### Purpose
This will allow you to dynamically set the serialized type of an object, like so:

    class ThingamajigSerializer
      ...
      type do
        object.special_type
      end
      ...
    end

This is especially helpful if you're using a gem like [active_record-acts_as](https://github.com/hzamani/active_record-acts_as) where in order to get the actual type of an object you have to call `object.actable_type`.

#### Changes
Add `&block` param to `type` and handle accordingly.
Add tests.

#### Caveats
WIP - Need some guidance here.  `self._type` doesn't seem to behave quite like `self._meta` or `self._links`.

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -109,6 +109,21 @@ module ActiveModel
       # @example
       #   has_one :blog do
       #     include_data false
+      #     meta(id: object.blog.id)
+      #     meta liked: object.likes.any?
+      #     link :self do
+      #       href object.blog.id.to_s
+      #       meta(id: object.blog.id)
+      #     end
+      def set_type(value = nil)
+        @type = block_given? ? Proc.new : value
+        :nil
+      end
+
+      # @api public
+      # @example
+      #   has_one :blog do
+      #     include_data false
       #     link :self, 'a link'
       #     link :related, 'another link'
       #   end
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -295,8 +295,13 @@ module ActiveModel
     # @example
     #   class AdminAuthorSerializer < ActiveModel::Serializer
     #     type 'authors'
-    def self.type(type)
-      self._type = type && type.to_s
+    # @example
+    #   class AdminAuthorSerializer < ActiveModel::Serializer
+    #     type do
+    #       object.type
+    #     end
+    def self.type(type = nil, &block)
+      self._type = (type && type.to_s) || block
     end
 
     # END SERIALIZER MACROS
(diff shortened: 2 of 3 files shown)


zeiv:
I hoping I can get some pointers on this one.  I thought this would be fairly straightforward, but when I run the tests now I'm getting an error in `active_model/serializer/fieldset.rb:13` that says `undefined method `singularize' for #<Proc:0x007fd657a88ee8>`.

If I change the method to say `self._type = (type && type.to_s) || block.call` then I get an `undefined method or local variable 'object'`, meaning I can't access the resource in this scope.

bf4:
@zeiv would you consider just adding an instance method?

```ruby
def json_key
  object.actable_type
end
```

alternatively, you could use inheritance to make a general serializer and then subclass for each actable type.

zeiv:
Yeah, that's basically what I'm doing right now.  I just thought it might be nice to add this feature since every other directive like `meta` and `links` all take blocks as well.  I was surprised when I tried to pass in a block without checking if I could do so and got an error.

bf4:
> I just thought it might be nice to add this feature since every other directive like meta and links all take blocks as well

@zeiv fair point

zeiv:
In case anyone comes here with the same problem I had (using active_record-acts_as with AMS), I ended up using this workaround in order to "dynamically" set the type on my relationships:

```ruby
  belongs_to :workable, type: 'projects', if: :workable_is_project? do
    object.workable.actable
  end
  belongs_to :workable, type: 'service_orders', if: :workable_is_service_order? do
    object.workable.actable
  end
```

Of course you then have to define the conditionals as methods in the Serializer

zeiv:
Actually, the above does not work.  The last `belongs_to :whatever` of the same `:whatever` is king.  It will override previous declarations even if a prior one evaluates true and the last one evaluates false.  Might be a bug?  Either way it doesn't work.

zeiv:
The original feature here (block in type) shouldn't be too difficult... I just don't know where these blocks are being evaluated or where I can access `object`.  If anyone can point me to where this happens I should be able to finish this PR fairly quickly.

lib/active_model/serializer/reflection.rb
```diff
@@ -109,6 +109,21 @@ def meta(value = nil)
       # @example
       #   has_one :blog do
       #     include_data false
+      #     meta(id: object.blog.id)
+      #     meta liked: object.likes.any?
+      #     link :self do
+      #       href object.blog.id.to_s
+      #       meta(id: object.blog.id)
+      #     end
+      def set_type(value = nil)
```
NullVoxPopuli: where is this called?

zeiv: This would be called in a serializer, something like this:
```
belongs_to :commentable do
  set_type object.commentable.type.to_s
end
```

I tried using `type` instead of `set_type` but it wasn't really working. Probably because it is used elsewhere for something else.

PR #2136, opened 2017-05-16
Allow referencing sideloaded `include` by `key`.

caomania:
#### Purpose
When an association is set to be `sideloaded` and the association has a `key` defined we should be able to include this association by referencing the `key`.

#### Changes
When checking `include_slice` for associations to include, check the association's `key` option first, then fallback to the association `name`.

Added tests to demonstrate that attempting to reference an association by `key` when including associations currently fails.

### DIFF
--- a/test/adapter/json_api/include_data_if_sideloaded_test.rb
+++ b/test/adapter/json_api/include_data_if_sideloaded_test.rb
@@ -135,7 +135,6 @@ module ActiveModel
             assert_nil(hash[:included])
           end
 
-
           def test_nested_relationship
             expected = {
               data: [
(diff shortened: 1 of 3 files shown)


test/adapter/json_api/include_data_if_sideloaded_test.rb
```diff
@@ -105,6 +110,32 @@ def test_sideloads_included
...
+            hash = result(include: :granted_roles)
+            assert_equal(expected, hash[:included])
+          end
+
+          def test_sideloads_not_included_when_using_name_when_key_defined
+            hash = result(include: :roles)
+            assert_nil(hash[:included])
+          end
+
+
```
mtwentyman: this line is what rubocop doth complain about

bf4:
@caomania merged!

PR #2140, opened 2017-05-22
Adds a configuration option belongs_to_uses_id_on_self

### LINKED ISSUE #2125: 0.10.6: type attribute not respected on belongs to relations

cassidycodes:
#### Purpose

Adds a configuration option `belongs_to_uses_id_on_self ` to allow for automatic serializer lookup. This solves an issue where the custom `type` attribute was not being used as the type was inferred from the associated object.

#### Changes

- Adds `ActiveModelSerializers.config.belongs_to_uses_id_on_self `
- Adds documentation for usage

#### Caveats

- Setting `belongs_to_uses_id_on_self` to true will result in a lookup on the associated object,

#### Related GitHub issues

Closes #2125

### DIFF
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -28,6 +28,31 @@ Possible values:
 
 When `false`, serializers must be explicitly specified.
 
+##### belongs_to_uses_id_on_self
+
+Enable automatic serializer lookup when using a `belongs_to` without `has_many`.
+
+Possible values:
+
+- `true`
+- `false` (default)
+
+When `false`, serializers must be explicitly specified in order to use a custom `type` field in the serializer.
+
+```ruby
+class PostSerializer <  ActiveModel::Serializer
+  belongs_to :blog, type: 'custom-type'
+end
+```
+
+or to explicitly state the serializer:
+
+```ruby
+class PostSerializer <  ActiveModel::Serializer
+  belongs_to :foo, serializer: FooSerializer
+end
+```
+
 ##### key_transform
 
 The [key transform](key_transforms.md) to use.
--- a/lib/active_model_serializers/adapter/json_api/relationship.rb
+++ b/lib/active_model_serializers/adapter/json_api/relationship.rb
@@ -43,7 +43,8 @@ module ActiveModelSerializers
         end
 
         def data_for_one(association)
-          if association.belongs_to? &&
+          if !ActiveModelSerializers.config.belongs_to_uses_id_on_self &&
+              association.belongs_to? &&
               parent_serializer.object.respond_to?(association.reflection.foreign_key)
             id = parent_serializer.read_attribute_for_serialization(association.reflection.foreign_key)
             type = association.reflection.type.to_s
--- a/test/serializers/associations_test.rb
+++ b/test/serializers/associations_test.rb
@@ -212,22 +212,20 @@ module ActiveModel
       end
 
       def test_belongs_to_allows_type_overwriting
-        begin
-          ActiveModelSerializers.config.belongs_to_uses_id_on_self = true
+        ActiveModelSerializers.config.belongs_to_uses_id_on_self = true
 
-          attributes = {
-            id: 1,
-            title: 'Belongs to Blog with Custom Type',
-            blog_with_custom_type: BlogWithCustomType.new(id: 1)
-          }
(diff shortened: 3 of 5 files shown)


docs/general/configuration_options.md
```diff
@@ -28,6 +28,31 @@ Possible values:
 
 When `false`, serializers must be explicitly specified.
 
+##### belongs_to_uses_id_on_self
+
+Enable automatic serializer lookup when using a `belongs_to` without `has_many`.
```
bf4: > Tells the JSON:API adapter to build `belongs_to` relationship id and type using the foreign key on self, rather than loading the association.
> For example, a Comment that belongs to a user will have a `user_id` attribute. 
> This means the relationship will the call `object.user_id` instead of `object.user.id`, which can prevent unneeded queries.
> IMPORTANT: Because the association is no longer be loaded,  any `type` must come from the `belongs_to` options as `type: :comment_users` or `serializer: CommentUserSerializer` (which would call CommentUserSerializer._type, i.e. no instance methods)

or something to that effect. Examples below are good.

docs/general/configuration_options.md
```diff
@@ -28,6 +28,31 @@ Possible values:
...
+class PostSerializer <  ActiveModel::Serializer
+  belongs_to :blog, type: 'custom-type'
+end
+```
+
+or to explicitly state the serializer:
+
+```ruby
+class PostSerializer <  ActiveModel::Serializer
+  belongs_to :foo, serializer: FooSerializer
```
bf4: examples should both be belongs to :blog, but one uses the `type` option and one uses the `serializer` option, and then show how they differ in their serialized relationship given a Post with blog_id = 1 and different BlogSerializer configurations, and what happens when both are given, or if the relationship is included...

test/serializers/associations_test.rb
```diff
@@ -195,6 +195,39 @@ def test_belongs_to_allows_id_overwriting
...
+        attributes :id
+        type 'custom-type'
+      end
+      class BelongsToBlogWithCustomTypeModel < ::Model
+        attributes :id, :title, :blog_with_custom_type_id
+        associations :blog_with_custom_type
+      end
+      class BelongsToBlogWithCustomTypeModelSerializer < ActiveModel::Serializer
+        type :posts
+        belongs_to :blog_with_custom_type
```
bf4: also good to include an instance method `json_key` to show it's not used

test/serializers/associations_test.rb
```diff
@@ -195,6 +195,39 @@ def test_belongs_to_allows_id_overwriting
...
+      end
+
+      def test_belongs_to_allows_type_overwriting
```
bf4: `test_belongs_to_doesnt_load_record` should also change, no, since it requires `belongs_to_uses_id_on_self` to be true, no?

lib/active_model_serializers/adapter/json_api/relationship.rb
```diff
@@ -43,7 +43,8 @@ def data_for(association)
         end
 
         def data_for_one(association)
-          if association.belongs_to? &&
+          if !ActiveModelSerializers.config.belongs_to_uses_id_on_self &&
```
bf4: should be 

```diff
-          if !ActiveModelSerializers.config.belongs_to_uses_id_on_self &&
+          if ActiveModelSerializers.config.belongs_to_uses_id_on_self &&
```

cassidycodes:
@bf4 Just to clarify, do we want`belongs_to_uses_id_on_self to be `false` by default? If it's false by default, the less performant behaviour will be the default.

Issue #1849, opened 2016-07-18
Provide a clear way to include nested associations and fields


iMacTia:
#### The Situation
I encourage to have a read to #1333 and #1845.

Let's assume we have the current serializers:

```ruby
class BlogSerializer < ActiveModel::Serializer
  attributes :id
  attribute :name, key: :title

  has_many :posts
end

class PostSerializer < ActiveModel::Serializer
  attributes :id, :title

  has_one :author
  has_one :category
end

class AuthorSerializer < ActiveModel::Serializer
  attributes :id, :name

  has_one :address 
end

class CategorySerializer < ActiveModel::Serializer
  attributes :id, :name
end
```

#### The issue
The only way to specify nested associations inclusion, at the moment, is through the controller's render `include` option, as shown in the current examples:

```ruby
# blogs_controller.rb
render json: @blog, include: 'posts.category, posts.author.address', fields: { posts: { category: [:name], author: [:id, :name] } }

# posts_controller.rb
render json: @posts, include: 'category, author.address', fields: { category: [:name], author: [:id, :name] }
```

The issue is that in this case (I admit is probably a stupid example, but stay with me), when rendering a post, we always want to render also its category name, its author attributes (except for the bio) and its author address.
(@bf4, I would like to specify the fields for author's address as well, but I'm not sure about how to do it in the controller's include.)

#### The proposed solution
The same result could be achieved without filling controller's with common serialisation logics, moving them into serialisers. For example, the above issue could have a more elegant solution like:

```ruby
class BlogSerializer < ActiveModel::Serializer
  attributes :id
  attribute :name, key: :title

  has_many :posts, include: :all
end

class PostSerializer < ActiveModel::Serializer
  attributes :id, :title

  has_one :author, exclude: :bio
  has_one :category, include: :name
end

class AuthorSerializer < ActiveModel::Serializer
  attributes :id, :name, :bio

  has_one :address, include: :all
end

class CategorySerializer < ActiveModel::Serializer
  attributes :id, :name
end

# blogs_controller.rb
render json: @blog

# posts_controller.rb
render json: @posts
```

#### To be decided

We'll probably need some team discussion together with @bf4 and @NullVoxPopuli 
Open questions:

1. When should `include` be taken into consideration? Should we use only the ones in the class we're using to serialise? Does the above example make sense or should the Author's and Post's `include`s be ignored when serialising the Blog because they are on a deeper level?

[...]

PR #2146, opened 2017-05-29
Support association nested fields without json api

charlesvallieres:
#### Purpose
We currently don't use JSON API and we need to support `fields` for root resource _and_ nested associations. I implemented it so it reads like strong_parameters : `fields: { [:attribute, { relation: [:nested_attribute] }] }`. This is different from how it currently works in JSON API : https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/adapters.md#json-api
I am not sure how do you whitelist root object AND nested associations using the JSON API adapter? I'll be happy to change this PR so it works the same way once I get it.

#### Changes
We need to pass options around to associations.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/issues/1849

I read the whole discussion but didn't find any consensus, maybe this is a bad idea for some reason I am not aware. Since we need this, I thought it might be useful for some other people too. 🙂  I didn't include any documentation changes because I'd like to know if this is something you would like to see merged beforehand.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -335,7 +335,7 @@ module ActiveModel
     # @param [JSONAPI::IncludeDirective] include_directive (defaults to the
     #   +default_include_directive+ config value when not provided)
     # @return [Enumerator<Association>]
-    def associations(include_directive = ActiveModelSerializers.default_include_directive, include_slice = nil)
+    def associations(include_directive = ActiveModelSerializers.default_include_directive, include_slice = nil, options = {})
       include_slice ||= include_directive
       return Enumerator.new unless object
 
@@ -344,7 +344,7 @@ module ActiveModel
           next if reflection.excluded?(self)
           next unless include_directive.key?(key)
 
-          association = reflection.build_association(self, instance_options, include_slice)
+          association = reflection.build_association(self, instance_options.merge(options), include_slice)
           y.yield association
         end
       end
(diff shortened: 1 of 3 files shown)


johan-smits:
Why did this not get merged?

bf4:
@johan-smits I'm not sure exactly what I was thinking at the time.  I know I implemented a fields selector for all adapters that I left in PR-land. From reading the related issue, it seems I may have thought it was a better solution in userland than in the gem.  I'm not sure what you're looking for.

johan-smits:
@bf4 I want to be able to select fields of the relation. Example:

Normal object
```json
{
  "id": 1,
  "name": "some name",
  "nicknames": [
    {
      "id": 1,
      "name": "dummy"
    }
  ]
}
```

What I want to pass as a filter to have only these fields is:
```ruby
{:name, nicknames: [:name]}
```

The end result would be:
```json
{
  "name": "some name",
  "nicknames": [
    {
      "name": "dummy"
    }
  ]
}
```

bf4:
@charlesvallieres like I said, it looks like I closed it around when I wrote a PR to address the issue https://github.com/rails-api/active_model_serializers/pull/2223 and see https://github.com/rails-api/active_model_serializers/pull/2278 and #2100 

I'm going to lock this issue. Please look at #2223 and try it out and see if it meets your needs. Or maybe #2278.

PR #2145, opened 2017-05-29
Support options/conditions for multiple attributes

charlesvallieres:
#### Purpose
Support options/conditions on `attributes`
```
attributes :foo, :bar, if: -> { current_user.admin? }
```

#### Changes
serializer.rb#attributes

#### Related GitHub issues
We needed this so I built it. While reading contributing.md I started searching for issues and found this PR which is pretty much exactly the same thing : https://github.com/rails-api/active_model_serializers/pull/1714

Since this is +1 year ago and OP didn't respond I thought I would still open a new PR. Feel free to close it if I was wrong in doing so 🙂

PR #2148, opened 2017-06-03
Serializer attributes options

nicklandgrebe:
#### Purpose

Add options to `Serializer.attributes` so we can apply conditions to multiple attributes at once.

#### Changes

* Extract `options` hash from last arg of `Serializer.attributes` using `ActiveSupport::Array#extract_options!`
  * Exclude `options[:key]`
* Pass in `options` to the subsequent `each` call to `attribute` => `attribute(attr, options)` 

#### Related GitHub issues

Old PR: #1714 

Resubmitted against 0-10-stable


charlesvallieres:
Hi @nicklandgrebe I opened #2145 a couple days ago about the same issue. I think the extract options is good. If you could take the documentation update here : https://github.com/rails-api/active_model_serializers/pull/2145/files#diff-683b15b4d7c82f436aaeada1733d9eec and I'll close my PR so this can get merged upstream ASAP! :)

nicklandgrebe:
@charlesvallieres Awesome, thanks so much for adding that useful doc

lib/active_model/serializer.rb
```diff
@@ -199,10 +199,12 @@ def self._attributes
     #   class AdminAuthorSerializer < ActiveModel::Serializer
     #     attributes :id, :name, :recent_edits
     def self.attributes(*attrs)
```
beauby: Why not `def self.attributes(*attrs, **options)`?

nicklandgrebe: That works! The things you learn. Thanks

nicklandgrebe:
@bf4 The original use case of this feature was to be able to use one serializer in two different contexts of our API, separated between being queries made in a private environment (our own backend) vs. a public environment (a third party user). For example, a `Product`:

```
class ProductSerializer < ActiveModel::Serializer
  attributes :price, :number_available

  attributes :number_sold, :other_private_information, unless: :public?

  def public?
    # does the scope say we are in a public API environment?
  end
end

maxKimoby:
Please post an update on this PR, not being able to use conditions on attributes makes life hard right now.

nerfologist:
Hello, this would still be useful to us. Any reason why it has been closed?

Thanks

tfwright:
Bump 🙏

hashwin:
@bf4 Can this be followed up on? Looks like the changes were approved, but for some reason this was closed.

btrd:
@bf4 Hi, I saw that you just released a new version, can you add this also in the next version ?

Thanks for your work !

nakwa:
@bf4 @nicklandgrebe @beauby was there ever any update on this? 
I have this "temporarily" patched in my code for the last 4 years ... 😆

sergioisidoro:
This was approved, but then closed? What happened?

PR #2166, opened 2017-07-25
For nil return DefaultSerializer

yosiat:
There is no need to check the serializer for nil serializer,
since it should be DefaultSerializer.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -59,6 +59,8 @@ end
       def serializer_for(resource, options = {})
         if resource.respond_to?(:serializer_class)
           resource.serializer_class
+        elsif resource.nil?
+          DefaultSerializer
         elsif resource.respond_to?(:to_ary)
           if Object.constants.include?(:ArraySerializer)
             ::ArraySerializer


yosiat:
@bf4 ?

bf4:
@yosiat would you mind describing the problem this is solving and add a test?

also, would be helpful if you could guess why you think this hasn't come up before

yosiat:
@bf4 It will be hard to add test and I explain.
I am using a mocking library (RR) which adds to the mocked object a `to_ary?` method, and for nil as well.

In my application tests, we have situations (wrong ones) that mock nil, which causes an issue here, because then ArraySerializer is trying to serialize nil.
While it's possible to fix this issue, I think a fix is needed in AMS -

- Handling nil right away
- Improving the array detection

I am afraid of changing the array detection since it can break compatibility.

Adding a test to this is possible, but the test will require me to class_eval NilClass and add it to_ary.

bf4:
> Adding a test to this is possible, but the test will require me to class_eval NilClass and add it to_ary.

Honestly, that's seems kind of nuts to me that anything would want to add the implicit array conversation method to nil.  That's not really the fault of AMS.  What does it even return?  Are you saying that in your tests `nil` is treated as `[]` but not in prod?

yosiat:
@bf4 what I am saying, that in tests only, nil responds to_ary (and not to_a, for example) because of our test framework

bf4:
>  nil responds to_ary (and not to_a, for example) because of our test framework

Right, and that's super bizarro. 

For example:

```irb
>> x = 1
>> "string" + x
TypeError: no implicit conversion of Fixnum into String
 "string#{x}"
=> "string1"
>> x.to_str
NoMethodError: undefined method `to_str' for 1:Fixnum
>> x.to_s
=> "1"
```

the `to_ary` is the implicit conversion function. Implementing the implicit array conversion (i.e. is it 'really' an array) but not the explicit one (i.e. can we force it into an array) doesn't make sense to me.

```ruby
>> x = Object.new
=> #<Object:0x007fa4d4d805d8>
>> [] + x
TypeError: no implicit conversion of Object into Array
        from (irb):9
        from /Users/benjamin/.rvm/rubies/ruby-2.2.3/bin/irb:15:in `<main>'
>> x.to_a
NoMethodError: undefined method `to_a' for #<Object:0x007fa4d4d805d8>
>> x.to_ary
NoMethodError: undefined method `to_ary' for #<Object:0x007fa4d4d805d8>


>> def x.to_ary; []; end
=> :to_ary
>> [] + x
=> []

=> []
>> x.to_a
NoMethodError: undefined method `to_a' for #<Object:0x007fa4d4d805d8>

>> Array(x)
=> []

>> x.concat([])
NoMethodError: undefined method `concat' for #<Object:0x007fa4d4d805d8>

>> x + []
NoMethodError: undefined method `+' for #<Object:0x007fa4d4d805d8>
```

Sure, you can make it kind of quack like an array in some cases, but it's still not.

I mean I'd love to help, but you're asking a prod lib to change it's behavior because of some bizarre test monkeypatch of nil to be enumerable.

```ruby
>> [].map
=> #<Enumerator: []:map>
>> [].map.to_a
=> []
=> []
>> x.map
NoMethodError: undefined method `map' for #<Object:0x007fa4d4d805d8>
```

Issue #1923, opened 2016-09-11
Compatibility of Rails 5 API-only mode and AMS 0.9x


oyeanuj:
Hi folks, I was trying to upgrade my current project to Rails 5, API-only mode while continuing to use AMS 0.9 (due to pending #1845 being blocker for us) and ran into the following issue which goes away on upgrading to 0.10x (which unfortunately, I can't do for my project).

While generating scaffolds with a Rails 5 API project, the project complains about `api_controller.rb`: 

```ruby
→ rails g scaffold User
Running via Spring preloader in process 20090
      invoke  active_record
      create    db/migrate/20160911170034_create_users.rb
      create    app/models/user.rb
      invoke    test_unit
      create      test/models/user_test.rb
      create      test/fixtures/users.yml
      invoke  resource_route
       route    resources :users
      create  app/serializers/user_serializer.rb
      invoke  scaffold_controller
Could not find "api_controller.rb" in any of your source paths. Your current source paths are:
/Users/anuj/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/active_model_serializers-0.9.5/lib/active_model/serializer/generators/serializer/templates
```

Is this expected behavior and is there any workaround for this? Are there any other issues that I should expect to encounter in this process? I'd love to be a part of the Rails 5 party :)

The only issue related to Rails 5 and AMS 0.9 that I found was #600, but that is slightly unconnected to this one (and I've added their fix for the other problem).

Thank you!

remear:
It's undesirable, but not entirely unexpected. This is a symptom of development efforts shifting to 0.10.x and support for 0.9.x being reduced and deprecated. Problems like these will become more frequent as 0.9 ages. I'd recommend migrating to 0.10.x and would be interested to hear why you're currently unable to do so.

oyeanuj:
@remear Thanks for the response! While I totally understand not wanting to invest more on the 0.9x and 0.8x versions, I am hoping that atleast having them compatible with Rails-5 would be worth the effort and fits under 'reduced support'.

For our case, I'd love to upgrade to 0.10x but #1845 is a blocker which I know @bf4 is mulling over and working on. (_In general, and please correct me if I'm wrong, 0.10x doesn't have the same feature compatibility with 0.8/0.9 especially in terms of supporting Rails `as_json` behavior_). Until some of those issues are resolved, I would propose that 0.8 and 0.9 <-> Rails 5 be something that the team looks into.

[...]

PR #2167, opened 2017-07-27
Add missing api_controller that Rails 5 generator expects

oyeanuj:
#### Purpose
This allows AMS 0.9x to be used with Rails 5, without any errors.

#### Changes
Added an `api_controller.rb` template that the Rails 5 Generator expects.

#### Caveats
This fixes the main issue that I was facing in #1923 but as I experiment more, there might be more issues getting Rails 5, AMS 0.9x to play well.

#### Related GitHub issues
#1923 

Thanks @bf4 @remear for your help debugging this issue.

### DIFF
--- a/lib/active_model/serializer/generators/serializer/templates/api_controller.rb
+++ b/lib/active_model/serializer/generators/serializer/templates/api_controller.rb
@@ -1,0 +1,93 @@
+<% if namespaced? -%>
+require_dependency "<%= namespaced_file_path %>/application_controller"
+
+<% end -%>
+<% module_namespacing do -%>
+class <%= controller_class_name %>Controller < ApplicationController
+  before_action :set_<%= singular_table_name %>, only: [:show, :edit, :update, :destroy]
+
+  # GET <%= route_url %>
+  # GET <%= route_url %>.json
+  def index
+    @<%= plural_table_name %> = <%= orm_class.all(class_name) %>
+
+    respond_to do |format|
+      format.html # index.html.erb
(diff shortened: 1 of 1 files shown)


bf4:
@oyeanuj Sorry for the delay in responding to this:  what's the error? I'm not sure why AMS should be adding a controller generator template

oyeanuj:
@bf4 So, it seems that Rails 5 in API mode [expects there to be these controller generators](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/scaffold_controller/scaffold_controller_generator.rb#L21) which were missing and causing errors mentioned in #1923.

Also, to point out that 0.9x already had controller template but it was named `controller.rb`. In API mode, it seems to expect the controller generator to be named as `api_controller.rb`.

bf4:
ref 
- rails template https://github.com/rails/rails/blob/5-1-stable/railties/lib/rails/generators/rails/scaffold_controller/templates/api_controller.rb
- devise template https://github.com/plataformatec/responders/commit/8f443b2f185b04816cb0e990d50832a015ca2594

bf4:
@oyeanuj would you be up to rebase and force push this?

bf4:
I think it might make more sense to avoid the api controller template. There's nothing different in it from the Rails default, right?

```diff
module Rails
  module Generators
    class ScaffoldControllerGenerator
-      if Rails::VERSION::MAJOR >= 4
+      if Rails::VERSION::MAJOR >= 4 && ApplicationController < ActionController::Base
```

PR #2169, opened 2017-08-01
Add notes about what's going on with AMS

NullVoxPopuli:
[rendered](https://github.com/NullVoxPopuli/active_model_serializers/blob/notice-of-jsonapi-rb/README.md)

personally, I have opinions, and one of those is that anyone using a json api should use JSONAPI.org style json. 

I don't think it makes sense for for AMS to continue with the adapter flexability either -- unless someone had some very specific scenario where they needed multiple adapters in the same project.

Update:
 - if AMS is to continue, it needs to be A LOT simpler. Current plan is to do the heavy lifting with jsonapi-rb and transform based on chosen adapter.

TODO:
- [x] Create Orphan Branch including the benchmark.
- [x] Discuss next Iteration of AMS

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ ams                        :   12565796 allocated - 3.53x more
 Every scenario builds and renders [JSONAPI.org](jsonapi.org) documents of 301 records.
 
 eager means eager loaded data (no db hits).
-The benchmark for this can be found [here](https://github.com/NullVoxPopuli/rails-NPlusOneTests/blob/2a5ebfec262e53d8bcb7f3308388fc5ba64f599d/serialization_benchmark.rb)
+The benchmark for this can be found [here](https://github.com/rails-api/active_model_serializers/blob/43c1518cfff680781b943075ceeb6ca61a3536dd/benchmarks/serialization_libraries/benchmark.rb)
 
 
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ ams                        :   12565796 allocated - 3.53x more
 Every scenario builds and renders [JSONAPI.org](jsonapi.org) documents of 301 records.
 
 eager means eager loaded data (no db hits).
-The benchmark for this can be found [here](https://github.com/NullVoxPopuli/rails-NPlusOneTests/blob/master/serialization_benchmark.rb)
+The benchmark for this can be found [here](https://github.com/NullVoxPopuli/rails-NPlusOneTests/blob/2a5ebfec262e53d8bcb7f3308388fc5ba64f599d/serialization_benchmark.rb)
 
 ## About
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,56 @@
-# ActiveModelSerializers
+# ActiveModelSerializers (Deprecated)
+
+Please use [jsonapi-rb](https://github.com/jsonapi-rb) ([docs](http://jsonapi.org))
+
+Comparison of jsonapi-rb vs AMS 0.10.6
+
+```
+Calculating -------------------------------------
+ams                        
+                          5.509  (± 1.8%) i/s -     55.000  in  10.102302s
+jsonapi-rb                 
+                         15.872  (± 0.6%) i/s -    159.000  in  10.042557s
+ams        eager           
+                          6.166  (± 1.3%) i/s -     62.000  in  10.097632s
+jsonapi-rb eager           
+                         32.555  (± 0.8%) i/s -    327.000  in  10.079955s
+                   with 95.0% confidence
+
+Comparison:
+jsonapi-rb eager           :       32.6 i/s
+jsonapi-rb                 :       15.9 i/s - 2.05x  (± 0.02) slower
+ams        eager           :        6.2 i/s - 5.28x  (± 0.08) slower
+ams                        :        5.5 i/s - 5.90x  (± 0.12) slower
+                   with 95.0% confidence
+
+Calculating -------------------------------------
+ams                        
+                        12.566M memsize (   945.066k retained)
+                       153.201k objects (    12.896k retained)
+                        50.000  strings (    50.000  retained)
+jsonapi-rb                 
+                         5.670M memsize (     0.000  retained)
+                        66.887k objects (     0.000  retained)
+                        50.000  strings (     0.000  retained)
+ams        eager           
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,56 @@
...
+Comparison:
+jsonapi-rb eager           :    3564036 allocated
+jsonapi-rb                 :    5670156 allocated - 1.59x more
+ams        eager           :   11316060 allocated - 3.18x more
+ams                        :   12565796 allocated - 3.53x more
+```
+Every scenario builds and renders [JSONAPI.org](jsonapi.org) documents of 301 records.
+
+eager means eager loaded data (no db hits).
+The benchmark for this can be found [here](https://github.com/NullVoxPopuli/rails-NPlusOneTests/blob/master/serialization_benchmark.rb)
```
bf4: should ref a specific commit.

README.md
```diff
@@ -1,4 +1,56 @@
-# ActiveModelSerializers
+# ActiveModelSerializers (Deprecated)
```
bf4: Do we want to declare AMS dead?

NullVoxPopuli: I updated the description for some clarity in my intent.

like, it sucks, cause it has high visibility, but @beauby did a GREAT job with jsonapi-rb. 

I just don't feel like _another_ re-write is going to give people confidence. 
migrating to jsonapi-rb is pretty straight forward, too. 

tonight, I'm working on some compatibility stuff to allow smooth migration from AMS to jsonapi-rb.

README.md
NullVoxPopuli: thanks

README.md
```diff
@@ -1,4 +1,56 @@
...
+
+Please use [jsonapi-rb](https://github.com/jsonapi-rb) ([docs](http://jsonapi.org))
```
bf4: This isn't a totally fair recommendation since not everyone wants JSON:API, and jsonapi-rb isn't the most-obvious choice for JSON:API.  If we want to be fair, we should suggest ROAR, jsonapi-resources, etc. 

I'd be happy to add an alternatives section down where we mention dev might not be released, but generally I think this recommendation is too specific for one library.

README.md
bf4: Instructions on migration would be fantastic

bf4: Should also mention https://jsonapi-suite.github.io/jsonapi_suite/

README.md
```diff
@@ -1,4 +1,56 @@
...
+Comparison:
+jsonapi-rb eager           :    3564036 allocated
+jsonapi-rb                 :    5670156 allocated - 1.59x more
+ams        eager           :   11316060 allocated - 3.18x more
+ams                        :   12565796 allocated - 3.53x more
+```
+Every scenario builds and renders [JSONAPI.org](jsonapi.org) documents of 301 records.
+
+eager means eager loaded data (no db hits).
+The benchmark for this can be found [here](https://github.com/NullVoxPopuli/rails-NPlusOneTests/blob/2a5ebfec262e53d8bcb7f3308388fc5ba64f599d/serialization_benchmark.rb)
```
bf4: maybe this could be an (orphan) branch on AMS since it's a recommendation from AMS -> something else?

NullVoxPopuli:
@bf4 I guess, I'm back to struggling with what the goal of AMS is?

README.md
NullVoxPopuli: ROAD and jsonapi-resources provide a layer on top of rails, whereas jsonapi-rb and ams are just (de)serialization

README.md
NullVoxPopuli: that's also more than serialization.

but yeah, I'll submit a separate PR for transition instructions to both AMS and jsonapi-rb when I get that stuff working. :-)

beauby:
My 2cts: jsonapi-rb was intended as a jsonapi-only faster/smaller/more flexible alternative to AMS, which I think it succeeded being. That said, quite a lot of people still use AMS for non-jsonapi stuff, so I don't think AMS should be declared dead. 
However, the initial idea that one would have a common interface to declare serializers for various formats has been implemented in a much better way in ROAR IMO, so if AMS is to keep going forward, I believe its scope should be narrowed down to simple serialization (i.e. attributes/json adapters).

Regarding suggesting alternatives for JSONAPI, here are those I have worked with:
+ (de)serialization only:
    - jsonapi-rb (or jsonapi-rails for rails bindings)
    - jsonapi-serializers (small and efficient, but has a few bugs, plus a dependency on activesupport)
    - roar-jsonapi (the jsonapi adapter is not production-ready and has many bugs)
+ higher level frameworks (with a higher-level notion of resource, automatic handling of query parameters, etc.)
    - jsonapi_suite (which uses jsonapi-rb under the hood)
    - jsonapi-resources

NullVoxPopuli:
I gave up on the translator for jsonapi-rb <-> AMS. It's easy enough to just convert things / add a second set of serializers gradually. (I also used this as an opportunity to make [a little more configurable](https://github.com/NullVoxPopuli/drawers/releases/tag/1.1.0)

In case anyone else finds this, the only real differences I had to deal with:
 -  serializers need a type explicitly set
 - `attribute :method_name` must be replaced by by `attribute(:method_name) { @object.get_data_or_whatever }`
 - It's better to specify explicit serializers as the string form to avoid circular dependency issues

README.md
bf4: should link down to https://github.com/rails-api/active_model_serializers#status-of-ams

NullVoxPopuli: updated

NullVoxPopuli:
@bf4, I created the orphan branch, and updated this PR to reference that :-)

bf4:
Ah, missed it https://github.com/rails-api/active_model_serializers/tree/benchmarks

README.md
```diff
@@ -1,5 +1,68 @@
 # ActiveModelSerializers
 
+## Status of AMS
+AMS has had quite the rocky history. 3 rewrites, and possibly a 4th. Currently AMS is open to Proof of Concept Design PRs.
+
+### Motivation for yet another re-write
+ - The maintainers of AMS have cycled in and out a few times, and the code base has become unwieldy
+ - 0.10.x's architecture could not handle all the feature requests in an elegant way.
+ - performance of AMS is bad/slow.
```
bf4: These notes are good

PR #2170, opened 2017-08-02
Proof of Concept for new AMS Design.

NullVoxPopuli:
I don't think this needs to be anything complicated, as most of the functionality leans on @beauby's jsonapi-rb.

Disclaimer: I haven't ran this code, just kinda wanted to show what I was thinking. :-)

/cc @bf4 @beauby 

The gist is that you'd define your serializers as:

```ruby
# initializer somewhere
ActiveModelSerializers.adapter = :json

# serializer
class WhateverSerializer < ActiveModelSerializers::Serializer
 # jsonapi-rb usage in here
  type 'whatevers'

  attribute :name

  has_many :mehs, class: MehSerializer do
    data do
      @object.meh_records
    end
  end
end

# and then render with:
WhateverSerializer.render(my_resource, options)
# => {
#      whatever: {
#        name: 'something',
#        mehs: [
#          # meh attributes and relations here
#        ]
#      }
#    }
```

### DIFF
--- a/lib/active_model_serializers.rb
+++ b/lib/active_model_serializers.rb
@@ -1,0 +1,23 @@
+require 'jsonapi/serializable'
+
+require 'active_model_serializers/version'
+require 'active_model_serializers/formatters/attributes'
+require 'active_model_serializers/formatters/json'
+require 'active_model_serializers/serializer'
+
+module ActiveModelSerializers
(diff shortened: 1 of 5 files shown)


NullVoxPopuli:
Cleaning up my PRs. If this work is still desired, feel free to cherry-pick

PR #2171, opened 2017-08-02
Prevent out of bounds pages from returning a next link.

bcaplan:
#### Purpose

In the JSON:API adapter, this will prevent `next` link URLs from being generated for nonexistent pages if the page requested is out of bounds.

#### Changes

Now checking `current_page` against `total_pages` using `>=` instead of `==`.

This also makes sure the `prev` link is a real page.

The `self` link is unmodified, it will return a link to the current page even if it doesn't exist. Seemed to best conform to the JSON:API spec.


bcaplan:
@bf4 Is there anything more I can do help get this merged? I think tests were failing on the original branch I forked which may have prevented this from getting looked at. This is now fixed since I just pulled the upstream changes.

bcaplan:
@bf4 Would you be willing to provide any info on why this was closed and not merged? This was my first attempt at contributing to open source software, so the information would be much appreciated.

bf4:
@bcaplan Not sure how I might have closed it

bcaplan:
@bf4 Haha, no worries. Thanks for reopening.

bf4:
This looks fine to me.  Does the spec comment on this scenario? I don't recall. hashtag 'lazy'

bcaplan:
@bf4 I named them `test_out_of_bounds_page_pagination_links_using_{serializer here}`, but there isn't a specific comment outlining the reason for the test. I didn't see any other specs with comments outlining expected behavior. Is that something you'd like me to add?

bf4:
> I didn't see any other specs with comments outlining expected behavior

Sorry, I meant http://jsonapi.org/format/#fetching-pagination

> Keys MUST either be omitted or have a null value to indicate that a particular link is unavailable.

So, nothing about out of bounds.

lib/active_model_serializers/adapter/json_api/pagination_links.rb
```diff
@@ -53,11 +53,16 @@ def last_page_url
 
         def prev_page_url
           return nil if collection.current_page == FIRST_PAGE
-          url_for_page(collection.current_page - FIRST_PAGE)
+
+          if collection.current_page > collection.total_pages
+            url_for_page(collection.total_pages)
+          else
+            url_for_page(collection.current_page - FIRST_PAGE)
+          end
```
bf4: of

```diff
-          url_for_page(collection.current_page - FIRST_PAGE)
+          page_number = collection.current_page > collection.total_pages ? collection.total_pages : collection.current_page - FIRST_PAGE
+          url_for_page(page_number)
```

So, `prev` is the last page if `current page` is out of bounds,

lib/active_model_serializers/adapter/json_api/pagination_links.rb
```diff
@@ -53,11 +53,16 @@ def last_page_url
...
         end
 
         def next_page_url
-          return nil if collection.total_pages == 0 || collection.current_page == collection.total_pages
+          return nil if collection.total_pages == 0 || collection.current_page >= collection.total_pages
           url_for_page(collection.next_page)
```
bf4: So no `next page` if the `current page` is the last page or out of bounds.

test/adapter/json_api/pagination_links_test.rb
```diff
@@ -174,6 +193,18 @@ def test_last_page_pagination_links_using_will_paginate
           assert_equal expected_response_with_last_page_pagination_links, adapter.serializable_hash
         end
 
+        def test_out_of_bounds_page_pagination_links_using_kaminari
+          adapter = load_adapter(using_kaminari(4), mock_request)
```
bf4: might be good to replace `4` with `out_of_bounds_page_number = 4` here and below

lib/active_model_serializers/adapter/json_api/pagination_links.rb
bcaplan: That's correct. This way if you request out of bounds and rely on the next link, you don't paginate forever.

bcaplan:
Ah, I see what you meant now. Yeah, it doesn't mention this specific circumstance. I decided when out of bounds: omitting the next link, providing the last page as the previous link, and providing a current link even though you are out of bounds seemed to best conform to my reading of the spirit of the spec.

test/adapter/json_api/pagination_links_test.rb
bcaplan: Done.

lib/active_model_serializers/adapter/json_api/pagination_links.rb
bcaplan: Done.

jkburges:
It'd be good to see this merged. Is there anything I can do to help move things along?

bcaplan:
This appears to have been addressed by this newer PR from December 2020: https://github.com/rails-api/active_model_serializers/pull/2399

A little disappointed #2171 was never merged, but glad it's fixed nonetheless.

PR #2172, opened 2017-08-03
Include namespace when looking up superclass

dpogue:
#### Purpose
When falling back to the superclass for finding a serializer, the namespace was getting dropped.

#### Changes
Pass the namespace through when finding a serializer for the superclass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Fixes:
 - [#2223](https://github.com/rails-api/active_model_serializers/pull/2223) Support Fieldset in Attributes/JSON adapters documented in [docs/general/fields.md](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/fields.md) that worked partially before (@bf4)
 - [#2337](https://github.com/rails-api/active_model_serializers/pull/2337) fix incorrect belongs_to serialization when foreign_key on object and belongs_to is blank (@InteNs)
     - Fixes incorrect json-api generation when `jsonapi_use_foreign_key_on_belongs_to_relationship` is `true` and the relationship is blank
+- [#2172](https://github.com/rails-api/active_model_serializers/pull/2172) Preserve the namespace when falling back to a superclass serializer
 
 Misc:
 
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -94,7 +94,7 @@ module ActiveModel
         if serializer_class
           serializer_class
         elsif klass.superclass
-          get_serializer_for(klass.superclass)
+          get_serializer_for(klass.superclass, namespace)
         else
           nil # No serializer found
         end
--- a/test/serializers/serializer_for_with_namespace_test.rb
+++ b/test/serializers/serializer_for_with_namespace_test.rb
@@ -9,6 +9,7 @@ module ActiveModel
         attributes :title, :author_name
         associations :publisher, :pages
       end
+      class Ebook < Book; end
       class Page < ::Model; attributes :number, :text end
       class Publisher < ::Model; attributes :name end
 
@@ -84,6 +85,11 @@ module ActiveModel
           }
         }
         assert_equal expected, result
+      end
+
+      test 'follows inheritance with a namespace' do
+        serializer = ActiveModel::Serializer.serializer_for(Ebook.new, namespace: Api::V3)
+        assert_equal Api::V3::BookSerializer, serializer
       end
     end
   end


NullVoxPopuli:
can you provide a test demonstrating your scenario?
also, is this a problem that could be solved by:
```ruby
ActiveModelSerializers.config.serializer_lookup_chain.unshift(
  lambda do |resource_class, _, namespace|
    "#{namespace.name}::#{resource_class.name}Serializer" if namespace
  end
)
```

dpogue:
It was an issue with the controller namespace not getting passed up to the superclass.

```ruby
# Superclass model
class User < ActiveRecord::Base
end

# Subclass model
class AppUser < User
end

# Non-namespaced superclass serializer (Wrong)
end

# Namespaced superclass serializer (Correct)
class Admin::UserSerializer < ActiveModel::Serializer
end

# Controller
class Admin::UsersController < ApplicationController
  def show
    render :json => AppUser.first
  end
end
```

When Admin::UsersController tried to render a AppUser, the lookup chains were like this:
```
["Admin::AppUserSerializer", "AppUserSerializer", "::AppUserSerializer"]
["UserSerializer", "::UserSerializer"]
```

With this patch, it correctly preserves the controller namespace when searching for the superclass:
```
["Admin::AppUserSerializer", "AppUserSerializer", "::AppUserSerializer"]
["Admin::UserSerializer", "UserSerializer", "::UserSerializer"]
```

NullVoxPopuli:
What happens when you try using the config change to the lookup chain?

dpogue:
The config change to the lookup chain does not resolve the issue, because it is still trying to use a nil namespace for the superclass.

NullVoxPopuli:
Can you change it to: 

```ruby
ActiveModelSerializers.config.serializer_lookup_chain.unshift(
  lambda do |resource_class, iforgotwhatthisis, namespace|
    puts '-----------------------------------------------'
    puts resource_class
    puts iforgotwhatthisis
    puts namespace


    "#{namespace.name}::#{resource_class.name}Serializer" if namespace
  end
)
```

and then restart your server, try to load your resource, and copy the console output to here?
This'll help, cause I actually don't remember which namespace that is in the lookups.

dpogue:
```
-----------------------------------------------
AppUser
ActiveModel::Serializer
Admin
-----------------------------------------------
User
ActiveModel::Serializer

```

NullVoxPopuli:
So then, this should solve your use case?
```ruby
ActiveModelSerializers.config.serializer_lookup_chain.unshift(
  lambda do |resource_class, serializer_class, namespace|
    "#{namespace.name}::#{resource_class.name}Serializer" if namespace
  end
```
?
cause namespace is `Admin` it the first execution, and 'resource_class' is `AppUser`, so putting them together would be `Admin::AppUserSerializer`

So, I guess the root of this is that you want to use a Serializer based off the superclass of the _model_ rather than specifically add namespacing support.

So then, you could add another lookup chain entry that did something like this:

```ruby
ActiveModelSerializers.config.serializer_lookup_chain.unshift(
  lambda do |resource_class, serializer_class, namespace|
    if namespace && resource_class.superclass != ActiveRecord::Base # or whatever your top level class is
      "#{namespace.name}::#{resource_class.superclass.name}Serializer"
    end
  end
```

does that work?

NullVoxPopuli:
It seems we need another PR that fixes the tests on master, as at least one of them seems unrelated to this PR.

I 'think' I agree that this PR is valuable, though. the namespace should def continue getting passed through the recursion.

dpogue:
https://github.com/rails-api/active_model_serializers/pull/2185 makes the same change, but includes tests, so that would probably be a better one to merge

dpogue:
I have cherry-picked the commit from https://github.com/rails-api/active_model_serializers/pull/2185 that includes tests, and rebased it on 0-10-stable, as requested in https://github.com/rails-api/active_model_serializers/pull/2242.

This PR now supersedes both https://github.com/rails-api/active_model_serializers/pull/2242 and https://github.com/rails-api/active_model_serializers/pull/2185. The cherry-picked commit is attributed to all the authors of all 3 PRs.

wasifhossain:
@bf4 you might have had some reason to close a similar PR. can you please share the point with us?
but if you have changed your mind any way, I would like to merge it.

@dpogue can you please leave a changelog entry under the `Fixes` section. thanks

dpogue:
Changelog entry added :)

wasifhossain:
tests are looking good :ok_hand:  let's give it a go! also have an eye open for any regression in case

PR #2176, opened 2017-08-22
Add warnings about global adapter config to docs

mrpinsky:
#### Purpose
Clarify effects of global adapter configuration in relation on serializer behavior

#### Changes
Add warnings to docs

#### Caveats
No code changes, just documentation

This came up for me when trying to test serializer behavior. I wanted to the serialization to include a root key, but despite setting `ActiveModelSerializers.config.adapter = :json` in an initializer, my calls to `serializer.as_json` didn't have the root key. When I tried to repro in a fresh Rails app, I found that calls to `render` in a controller did add a root key, but direct calls to `serializer.as_json` and `serializer.serializable_hash` with no arguments did not. Digging in, I think the problem is in `ActiveModel::Serializer:112-115`:
```ruby
    # @api private
    def self.serialization_adapter_instance
      @serialization_adapter_instance ||= ActiveModelSerializers::Adapter::Attributes
    end
```
that is, `@serialization_adapter_instance` is never set, and so it defaults to the hard-coded `Attributes` adapter if it isn't passed in to `serializable_hash` (which it isn't if you just use `as_json`);
```ruby
    # @return [Hash] containing the attributes and first level
    # associations, similar to how ActiveModel::Serializers::JSON is used
    # in ActiveRecord::Base.
      adapter_options ||= {}
      options[:include_directive] ||= ActiveModel::Serializer.include_directive_from_options(adapter_options)
      resource = attributes_hash(adapter_options, options, adapter_instance)
      relationships = associations_hash(adapter_options, options, adapter_instance)
      resource.merge(relationships)
    end
    alias to_hash serializable_hash
    alias to_h serializable_hash

    # @see #serializable_hash
    def as_json(adapter_opts = nil)
      serializable_hash(adapter_opts)
    end
```

### DIFF
--- a/docs/general/adapters.md
+++ b/docs/general/adapters.md
@@ -36,6 +36,12 @@ The `Attributes` adapter does not include a root key. It is just the serialized 
 
 Use either the `JSON` or `JSON API` adapters if you want the response document to have a root key.
 
+***IMPORTANT***: Adapter configuration has *no effect* on a serializer instance
+being used directly. That is, `UserSerializer.new(user).as_json` will *always*
+behave as if the adapter were the 'Attributes' adapter. See [Outside Controller
+Usage](../howto/outside_controller_use.md) for more details on recommended
+usage.
+
 ## Built in Adapters
 
 ### Attributes - Default
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Fixes:
 
 Misc:
 
+- #2176 Documentation for global adapter config (@mrpinsky)
+
 ### [v0.10.6 (2017-05-01)](https://github.com/rails-api/active_model_serializers/compare/v0.10.5...v0.10.6)
 
 Fixes:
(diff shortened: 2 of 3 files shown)


docs/general/adapters.md
```diff
@@ -36,6 +36,10 @@ The `Attributes` adapter does not include a root key. It is just the serialized
 
 Use either the `JSON` or `JSON API` adapters if you want the response document to have a root key.
 
+Note that this wil not change the behavior of a serializer's `to_json`,
```
bf4: Good point. Maybe:

```
> **IMPORTANT**: Adapter configuration has *no effect* on a serializer instance being used directly.
> That is, `UserSerializer.new(user).as_json` will *always* behave as if the adapter were the 'Attributes' adapter.
> See [Outside Controller Usage](../howto/outside_controller_use.md) for more details on recommended usage.
```

mrpinsky:
@bf4 I've changed the language in both places to better reflect your intentions, but I'm curious as to why it's supposed to behave this way. It seems like it makes the behavior of the system less consistent, meaning that you'd need to write different tests for the serializer behavior in a serializer test vs a controller test. How do you envision testing the serializers?

bf4:
Looks good. want to add yourself to the changelog?

mrpinsky:
Changelog updated

PR #2177, opened 2017-08-25
Ignore logging event when the serializer is the Null serializer.

### LINKED ISSUE #2024: Status code 500 when rendering hash, Rendered ActiveModel::Serializer::Null with Hash

In my controller I still want to render an arbitrary Hash as json, so while in one action I'd like to use AMS implicitly with ```render json: @object``` , I'd still like to use something like ```render json: { message: message }, status: status``` without hitting the AMS or using some very simple 'passthrough' serializer. What I 

#### Steps to reproduce
*(e.g., detailed walkthrough, runnable script, example application)*

```ruby
def some_action
  ...
  message = 'Incorrect URL'
  status = :bad_request
  render json: { message: message }, status: status
end
```

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*:  0.10.4

Output of `ruby -e "puts RUBY_DESCRIPTION"`:  ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin15]

OS Type & Version:  Mac OS X 10.11.6

Integrated application and version *(e.g., Rails, Grape, etc)*:  Rails 5.0.1

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*

```
[active_model_serializers] Rendered ActiveModel::Serializer::Null with Hash (0.14ms)
Completed 500 Internal Server Error in 45ms (Views: 7.4ms | ActiveRecord: 0.0ms)
```

It should not respond with status 500, this is not an error. I tried a workaround like ```    render inline: { message: message }.to_json, status: status, content_type: 'application/json'
``` with no success.

### PR DESCRIPTION
**Pull request by calavera**
#### Purpose

Stop logging when there is no serialization for an object.

Fix #2024

#### Changes

Ignore render event when the serializer is the Null serializer.

#### Caveats

Maybe some people expect to see this in the log, I don't, personally.

#### Related GitHub issues

#2024

This preserves the render payload event as it is to prevent side effects in people expecting both values to not be nil.

### DIFF
--- a/lib/active_model_serializers/logging.rb
+++ b/lib/active_model_serializers/logging.rb
@@ -104,8 +104,10 @@ module ActiveModelSerializers
 
     class LogSubscriber < ActiveSupport::LogSubscriber
       def render(event)
+        serializer = event.payload[:serializer]
+        return if serializer == ActiveModel::Serializer::Null
+
         info do
-          serializer = event.payload[:serializer]
           adapter = event.payload[:adapter]
           duration = event.duration.round(2)
           "Rendered #{serializer.name} with #{adapter.class} (#{duration}ms)"
--- a/test/active_model_serializers/logging_test.rb
+++ b/test/active_model_serializers/logging_test.rb
@@ -72,6 +72,11 @@ module ActiveModel
         ActiveModelSerializers::SerializableResource.new(@post).serializable_hash
         assert_match(/\(\d+\.\d+ms\)/, @logger.messages)
       end
+
+      def test_logs_without_serializer
+        ActiveModelSerializers::SerializableResource.new(foo: :bar).to_json
+        assert_equal('', @logger.messages)
+      end
     end
   end
 end

### CODE REVIEW

File: lib/active_model_serializers/logging.rb [hunk 1]
```diff
@@ -104,8 +104,10 @@ def logger_tagged_by_active_model_serializers?
 
     class LogSubscriber < ActiveSupport::LogSubscriber
       def render(event)
+        serializer = event.payload[:serializer]
+        return if serializer == ActiveModel::Serializer::Null
+
         info do
-          serializer = event.payload[:serializer]
```
bf4: I could see people disagreeing with this behavior.. not logging at all when logger is null... thoughts?

test/active_model_serializers/logging_test.rb
```diff
@@ -72,6 +72,11 @@ def test_logs_the_duration
         ActiveModelSerializers::SerializableResource.new(@post).serializable_hash
         assert_match(/\(\d+\.\d+ms\)/, @logger.messages)
       end
+
+      def test_logs_without_serializer
```
bf4: should be logs nothing when serializer is not found (i.e. is null)

and then assert that the serializable resource serializer is the Null serializer

bf4:
@calavera would you mind elaborating why you closed this?

PR #2184, opened 2017-09-01
correct spelling mistake

EdwardBetts:


bf4:
Normally I'd advise not changing historical changelogs, since but since you're a first-time contributor... :)

PR #2181, opened 2017-08-29
Fix CI on 0-10-stable

bf4:
#### Purpose

Fix CI

I'll accept PRs into the fix_ci branch if anyone wants to help

#### Changes

A bunch because it's been broken for a while.

Each change is isolated in a commit, though, so should be easy enough to follow

#### Additional helpful information

e.g. in 

https://github.com/rails-api/active_model_serializers/pull/2179 

- [Travis Failures](https://travis-ci.org/rails-api/active_model_serializers/builds/269424966)
- [Appveyor Failures](https://ci.appveyor.com/project/bf4/active-model-serializers/build/1.0.1032-0-10-stable)

PR #2191, opened 2017-09-17
Fix RuboCop offence

tagliala:
#### Purpose
Fix a RuboCop offence which is preventing travis from running tests


bf4:
Related: https://github.com/rails-api/active_model_serializers/pull/2181

bf4:
Thanks!

PR #2192, opened 2017-09-18
Prevent loading has_one relation with block, if it is not included

furkanayhan:
Before this PR; when setting a has_one relation with block, it loads content of it even if the relation is not included

### DIFF
--- a/lib/active_model/serializer/reflection.rb
+++ b/lib/active_model/serializer/reflection.rb
@@ -154,8 +154,9 @@ module ActiveModel
         @object = serializer.object
         @scope = serializer.scope
 
-        block_value = instance_exec(serializer, &block) if block
         return unless include_data?(include_slice)
+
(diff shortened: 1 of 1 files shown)


bf4:
Thanks for this... 

We really gotta get CI fixed so we can review PR more easily..

furkanayhan:
@bf4 It fails for the test "test_relationship_not_including_data"
It does not expect to include relationship when doing with "include_data false"

I don't know how to proceed this.
Can you help?

bf4:
Can you rebase this?

furkanayhan:
@bf4 did it. But we still have the same failing test

bf4:
@furkanayhan would you be up to looking at the failing tests? They're caused by the change in this PR, right?

furkanayhan:
I tried to look at them but since I didn't know much of this repo, I couldn't find why it failed.
And yes, this pr causes failing tests.

PR #2193, opened 2017-09-18
Adjusts Deserializer relationship parsing to allow links/meta keys

nikz:
#### Purpose

The JSON API spec allows for the [relationships key](http://jsonapi.org/format/#document-resource-object-relationships) to include one of `data`, `links` or `meta`.

The current parsing logic for the deserialiser raises an exception unless a `data` key is present (this may have been changed after the code was written 🚄), which means that we're getting an exception on a valid JSON:API document 😭 

#### Changes

* Adjusts the document validation inside the Deserializer to allow for one of `data`, `links`, or `meta` as per the spec.
* Ignores the `links` or `meta` keys when parsing the relationship into something to be used by Rails.

#### Caveats

It may not be the best behaviour to ignore the keys silently - however, I'm not really sure what a useful thing to do here would be. Perhaps print a warning?

#### Related GitHub issues

None 😺 

Thanks for the 💎, and your time reviewing this folks!

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/deserialization.rb
+++ b/lib/active_model_serializers/adapter/json_api/deserialization.rb
@@ -7,6 +7,9 @@ module ActiveModelSerializers
       module Deserialization
         InvalidDocument = Class.new(ArgumentError)
 
+        # http://jsonapi.org/format/#document-resource-object-relationships
+        VALID_RELATIONSHIP_KEYS = %w(links data meta).freeze
+
         module_function
 
         # Transform a JSON API document, containing a single data object,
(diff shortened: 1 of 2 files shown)


bf4:
@beauby any objections?

nikz:
@bf4 I think this still a change that's needed - is there a reason why you closed the PR? I'm not sure the test failures are related to the changes I've made but I could update to latest `0-10-stable` and attempt to get them passing?

bf4:
@nikz Not sure how or why I closed it

nikz:
@bf4 slip of the mouse! No worries at all. Is there anything you need me to do on this patch?

bf4:
I think I'd rather replace our deserializer with the jsonapi-deserializer.  There are prs for it. See https://github.com/rails-api/active_model_serializers/pull/1928#issuecomment-298298106

PR #2194, opened 2017-09-19
Prepare changelog for 0.8.4 release

ebeigarts:
The changelogs are a bit confusing, there are really outdated changelogs in 0-8-stable, 0-9-stable branches, I can open a PR to also update the changelog in 0-8-stable and bumb the version.

Maybe it would be better to remove `CHANGELOG-0-*.md` from master and put them only in stable branches, so that there is a single place for each changelog? At least `rails` does it this way - https://github.com/rails/rails/blob/master/activemodel/CHANGELOG.md

Changes:

- Added 0.8.4
- Formatted dates in the same format for all releases
- Removed 0.0.1 since it is not published to rubygems

### DIFF
--- a/CHANGELOG-0-08.md
+++ b/CHANGELOG-0-08.md
@@ -1,10 +1,22 @@
 ## 0.08.x
 
-### v0.8.3 (2014/12/10 14:45 +00:00)
+### v0.8.4 (2017-09-19)
+
+- [#1675](https://github.com/rails-api/active_model_serializers/pull/1675) Fix memory leak with :scope_name.
+  Supplying :scope_name causes `ActiveModel::Serializer#initialize` to
+  define a method on the class, which retains a reference to the
+  serializer instance. (@Botje)
+- [#1276](https://github.com/rails-api/active_model_serializers/pull/1276) Fix serializers were never including associations on cache hits (@kieran)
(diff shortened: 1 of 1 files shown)


bf4:
@ebeigarts I had actually meant for this to just be a commit against the 0-8-stable branch... 

also, let's separate cleanup from just adding the missing items.

bf4:
I applied it there as 3125069

bf4:
I've now released 0.8.4

ebeigarts:
Thanks, I see that https://github.com/rails-api/active_model_serializers/blob/master/CHANGELOG-0-08.md still shows 0.8.3. What is the reason to keep changelogs in two separate places/branches? Do I need to still change something or this PR can be closed?

bf4:
@ebeigarts The two changelogs is an artifact of having multiple 'stable' branches.  I'd be ok removing the changelog from master and linking to the v0.8.4 tag (since branches are unreliable). Thoughts? We never really discussed this.

NullVoxPopuli:
what if we tag twice on version releases?

0.8.x <- version tag
0-8-latest <- current release tag
and then the readme can point to 0-8-latest?

bf4:
Tags aren't mutable... perhaps I misunderstand you

B mobile phone

ebeigarts:
@bf4 why branches (0-x-stable) are unreliable?

bf4:
They aren't in practice 

B mobile phone

PR #2198, opened 2017-09-19
Test against latest stable Ruby and Rails versions

tagliala:
Also skips Rails 4.1 tests against Ruby 2.4.2 and ruby-head,
because Rails 4.2.8 is the first version of the 4.x series that
officially	support Ruby 2.4.

#### Purpose

Test against latest stable Ruby version: 2.4.2
Also skips tests that are going to fail (Rails 4.1 with Ruby >= 2.4)

#### Changes

Just the Travis CI configuration

Travis CI tests were previously skipping the latest stable Ruby version (2.4.2), jumping from 2.3 to 2.5.0.dev. Same for Rails, jumping from 5.0.6 to 5.2.0alpha and skipping 5.1.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,12 +20,14 @@ env:
     - "RAILS_VERSION=4.1"
     - "RAILS_VERSION=4.2"
     - "RAILS_VERSION=5.0"
+    - "RAILS_VERSION=5.1"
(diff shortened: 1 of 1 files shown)


bf4:
Looks good to me

PR #2200, opened 2017-09-20
Fix deserialization of polymorphic relationships

dennis95stumm:
#### Purpose
The type of polymorphic relationships must be classified. Otherwise a `NameError` will be raised by ruby.

#### Changes
The type of polymorphic relationships gets now classified.


bf4:
@dennis95stumm want to add an item under the changelog for fixes?

dennis95stumm:
@bf4 I've added an item under changelog.

PR #2206, opened 2017-10-10
use config.default_includes for JsonApi adapter relationships

lanej:
#### Purpose

Fix a raw `'*'` assumption which causes unrequested relationships to evaluated and added with only a `meta` member.

#### Changes

Use `config.default_includes` as intended.

#### Caveats

None?

#### Related GitHub issues

Couldn't find any.


lanej:
This incorrectly solved a different issue I had.  Closing.

Issue #2210, opened 2017-10-26
Rendering an empty array with JSON Adapter


jmeredith16:
Hello all,
#### Explanation and Expected behavior vs actual behavior
I have an Array of objects called `clusters`.  A Cluster does not have a serializer, as I am using only AMS to serialize objects in my API, not this part of my codebase.

I expect that when I render my empty clusters array, I get back an empty array `[]`, but I am actually getting back `{"":[]}`.  This is causing some frontend issues, resulting in an error unrelated to AMS.

Removing `ActiveModelSerializers.config.adapter = :json` from my initializer causes the empty clusters Array to respond with `[]` as expected.

With ActiveModelSerializers.config.adapter = :json
```
[2] pry(#<ClustersController>)> @clusters
=> []
[3] pry(#<ClustersController>)> render json: @clusters
I, [2017-10-26T15:37:33.504075 #67524]  INFO -- : Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModelSerializers::Adapter::Json (2.63ms)
=> ["{\"\":[]}"]
```

With no AMS adapter specified in my initializer (ie, :attributes)
```
[2] pry(#<ClustersController>)> render json: @clusters
I, [2017-10-26T15:41:04.020091 #67607]  INFO -- : Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModelSerializers::Adapter::Attributes (0.25ms)
=> ["[]"]
```

- Is `{"":[]}` the expected response when rendering an empty Array while using AMS?
- Is there another way to pass this variable through untouched by AMS?

#### Environment 
active_model_serializers: 0.10.6 - dff621e1744d37672ccba13f8e4b64dcc5b135c3
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]
OS X El Capitan, V10.11.3
Rails 4.2.6

#### Additional helpful information
As a solution/workaround, I used `render json: @clusters.to_json`, which returns the empty Array as expected.
Regardless of adapter type, if I have data in my `clusters` variable it will render correctly.

Possibly related: 
- https://github.com/rails-api/active_model_serializers/issues/945
- https://github.com/rails-api/active_model_serializers/tree/0-10-stable#primitive-handling

Thanks for your time

bf4:
@jmeredith16  Sorry for the bug.

[...]

PR #2214, opened 2017-10-30
Fail if collection type cannot be inferred with json adapter (#2210)

jmeredith16:
#### Purpose
This issue was found while rendering an empty array while using the `:json` adapter in https://github.com/rails-api/active_model_serializers/issues/2210.  Without a known ActiveModel object, the root element cannot be determined, so what should have been `{ my_obj: [] }` was being represented by `{ "": [] }`.

#### Changes

- Returns an ArgumentError if the `json_key` method cannot determine the root key.  Previously, this method would just return `nil`.
- Updates two tests to verify the ArgumentError is returned instead of `nil`.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/issues/2210
https://github.com/rails-api/active_model_serializers/issues/1745

### DIFF
--- a/lib/active_model/serializer/collection_serializer.rb
+++ b/lib/active_model/serializer/collection_serializer.rb
@@ -46,7 +46,10 @@ module ActiveModel
         # 3. get from collection name, if a named collection
         key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
         # 4. key may be nil for empty collection and no serializer option
-        key && key.pluralize
+        key &&= key.pluralize
+        # 5. fail if the key cannot be determined
+        key || fail(ArgumentError, 'Cannot infer root key from collection type. Please
+                 specify the root or each_serializer option, or render a JSON String')
       end
       # rubocop:enable Metrics/CyclomaticComplexity
 
--- a/test/collection_serializer_test.rb
+++ b/test/collection_serializer_test.rb
@@ -93,12 +93,16 @@ module ActiveModel
         resource = []
         resource.define_singleton_method(:name) { nil }
         serializer = collection_serializer.new(resource)
-        assert_nil serializer.json_key
+        assert_raise ArgumentError do
+          serializer.json_key
+        end
       end
 
       def test_json_key_with_resource_without_name_and_no_serializers
         serializer = collection_serializer.new([])
-        assert_nil serializer.json_key
+        assert_raise ArgumentError do
+          serializer.json_key
+        end
       end
 
       def test_json_key_with_empty_resources_with_serializer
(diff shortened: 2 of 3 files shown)


lib/active_model/serializer/collection_serializer.rb
```diff
@@ -46,7 +46,13 @@ def json_key
...
         key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
         # 4. key may be nil for empty collection and no serializer option
-        key && key.pluralize
+        key &&= key.pluralize
+        # 5. fail if the key cannot be determined
+        unless key
+          fail ArgumentError, "Cannot infer root key from collection type. Please specify the root or each_serializer option, or render a JSON String"
+        end
+
+        key
```
bf4: how about?

`key or fail ArgumentError,  "Cannot infer root key from collection type. Please specify the root or each_serializer option, or render a JSON String"`

test/collection_serializer_test.rb
```diff
@@ -93,12 +93,16 @@ def test_json_key_with_resource_with_nil_name_and_no_serializers
...
+        end
       end
 
       def test_json_key_with_resource_without_name_and_no_serializers
         serializer = collection_serializer.new([])
-        assert_nil serializer.json_key
+        assert_raise ArgumentError do
+          serializer.json_key
+        end
       end
```
bf4: :+1:

bf4:
Looks great. want to add a note to the changelog?

lib/active_model/serializer/collection_serializer.rb
bf4: Also note rubocop failure here ```lib/active_model/serializer/collection_serializer.rb:52:31: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)
          fail ArgumentError, "Cannot infer root key from collection type. Please specify the root or each_serializer option, or render a JSON String"
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^```

jmeredith16:
Thanks for the review, I'll make the changes shortly.

lib/active_model/serializer/collection_serializer.rb
```diff
@@ -46,7 +46,10 @@ def json_key
         # 3. get from collection name, if a named collection
         key ||= object.respond_to?(:name) ? object.name && object.name.underscore : nil
         # 4. key may be nil for empty collection and no serializer option
-        key && key.pluralize
+        key &&= key.pluralize
+        # 5. fail if the key cannot be determined
+        key || fail(ArgumentError, 'Cannot infer root key from collection type. Please
```
bf4: I would have thought this required `or`.  This is fine

beauby: FYI the only difference between `||` and `or` is operator precedence: `a = false || true` is equivalent to `(a = false) || true`.

mvz:
I'm getting this error on a line that does have `:each_serializer` specified 😞

mvz:
What seems to be missing from this pull request is an actual test for the case where collection without inferable type is passed but the each_serializer option is specified (like the error message tells you to do).

Issue #2144, opened 2017-05-26
Set serialization_scope and scope is nil in serializer


crustyratfink:
#### Expected behavior vs actual behavior
Expect to set
serialization_scope :view_context
in application_controller.rb

And in serializer:
delegate :current_user, to: :scope

and then access current_user in serializer.

Actual:
IssueSerializer#current_user delegated to scope.current_user, but scope is nil:...

#### Steps to reproduce
See above:
Set serialization_scope :view_context in application_controller.rb, delegate current_user to scope in serializer, try to access current_user in serializer.

#### Environment
ActiveModelSerializers Version *(commit ref if not on tag)*: 0.10.6

Output of `ruby -e "puts RUBY_DESCRIPTION"`: 2.3.1p112

OS Type & Version: OSX 10.12.4

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails 5.1

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*
Module::DelegationError (IssueSerializer#current_user delegated to scope.current_user, but scope is nil: #<IssueSerializer:0x007fbd370c8bf8 @object=#<Issue id: 22, title: ...data removed>, @scope=nil, @root=nil, @polymorphic=false, @meta_key=:meta, @meta=nil, @wrap_in_array=nil, @only=nil, @except=nil, @key_format=nil, @context=nil, @namespace=nil, @serialization_options={}>):

smeshko:
Hello, I have the same issue, any updates?

bf4:
How are you rendering?

kurko:
I got this same issue, except that mine is `nil` only within `has_many do end` blocks. 

```ruby
def title
  puts scope.inspect # => present
  object.title
end

has_many :comments do
  puts scope.inspect #=> nil
  object.comments
end
```

stratigos:
On _0.10.5_, I have the same issue as the OP, except unlike @kurko, my scope us always nil, regardless of where it is inspected (in the Serializer).

bf4:
@crustyratfink @stantona To what extent does the documentation at https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/serializers.md#controller-authorization-context help?

stratigos:
@bf4 I have emulated that example you posted exactly, and `view_context` is always `nil` for me in _0.10.5_.

However, if I set no `serialized_scope`, then `current_user` is always present, and is the current authenticated User. 

[...]

PR #2215, opened 2017-10-30
Update serializers.md

stratigos:
* Add note about `ActionController::API` and `view_context` set to `nil`, with respect to accessing helpers in a Serializer.

#### Purpose

Clarify a use case of `scope` that could present a problem when using a Rails 5 API controller. Refs  #2144. If using the `view_context` for scope with an API controller, the `view_context` will always be `nil`, and helpers can not be accessed in a Serializer this way.

#### Changes

Added a note to the documentation, and demonstrated an example workaround.

#### Related GitHub issues

 * #2144 

* [Related StackOverflow post on accessing a view helper inside a Presenter](https://stackoverflow.com/questions/5176718/how-to-use-the-number-to-currency-helper-method-in-the-model-rather-than-view/5176779).

### DIFF
--- a/docs/general/serializers.md
+++ b/docs/general/serializers.md
@@ -294,6 +294,8 @@ end
 Whether you write the method as above or as `object.comments.where(created_by: scope)`
 is a matter of preference (assuming `scope_name` has been set).
 
+Keep in mind that the scope can be set to any available controller reference. This can be utilized to provide access to any other data scopes or presentation helpers. 
+
 ##### Controller Authorization Context
 
 In the controller, the scope/scope_name options are equal to
@@ -311,7 +313,7 @@ current authorization scope when you call `render :json`.
 called on every request.  This was [also a problem](https://github.com/rails-api/active_model_serializers/pull/1252#issuecomment-159810477)
 in [`0.9`](https://github.com/rails-api/active_model_serializers/tree/0-9-stable#customizing-scope).
 
-We can change the scope from `current_user` to `view_context`.
+We can change the scope from `current_user` to `view_context`, which is included in subclasses of `ActionController::Base`.
 
 ```diff
 class SomeController < ActionController::Base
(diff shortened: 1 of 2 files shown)


docs/general/serializers.md
```diff
@@ -379,6 +379,31 @@ class PostsController < ActionController::Base
   end
 end
 ```
+Note that `view_context` is made available through `ActionController::Base`. The `view_context` will be `nil` when accessed from a controller that inherits from `ActionController::API`. In such a case, a helper will need to be instantiated as needed, e.g.:
```
bf4: Isn't there anything else that might substitute for `view_context`? Maybe `helpers`? or just `self`?

bf4:
Thanks! 💯 for the PR. Question about whether the proposed solution is best. See there.

We should add a Rails5 test for this, too, I think

docs/general/serializers.md
stratigos: Long story short - I can remove the example I posted if it helps keep the documentation more clear and coherent. 

When I was working on this for my own project, I found that `Rails.application.helpers`  only included the helpers defined by the current application, (i.e., helpers you wrote), and does not include other view helpers, like `NumberHelper` or `CsrfHelper` etc. 

I only proposed a generic solution to imply that "some class that has these modules included needs instantiated, somehow" and that `view_context` wont include them. 

I understand if the example I provided is out of scope for the documentation, and also something you'd rather not officially represent. Would it be simpler if I just removed the example, and only left the note itself? I.e.,

```ruby
Note that `view_context` is made available through `ActionController::Base`. The `view_context` will be `nil` when accessed from a controller that inherits from `ActionController::API`.
```

bf4: How about we edit anywhere on the page that says view_context (for subclasses of ActionController:Base), and adds another example or note that any controller reference which provides the desired scope is acceptable, for example `:self` may be useful for subclasses of `ActionController::API` which does not include ActionView::ViewContext ?

Thoughts?

stratigos: Im not sure if I follow the use of `:self`, as it too is always `nil` for:

```ruby
class SomeApiController < ActionController::API
  serialization_scope :self
end
```

I think I can change the wording to match something similar to what you have above, though.

bf4:
Awesome! Want to add yourself to the changelog?

stratigos:
@bf4 Hows it look now?

bf4:
> Awesome! Want to add yourself to the changelog?

:) If not, I'll merge when it passes

stratigos:
Sorry, I update the _git_ log instead ;)

stratigos:
Tysm! Happy to contribute something!

PR #2224, opened 2017-11-22
Omit empty pagination links

mschneider85:
#### Purpose
I wanted to validate my API against the schema provided by [json-api](https://github.com/json-api/json-api). There were validation errors caused by pagination links having "Null" values instead of strings. To make it schema compliant, I made this small modification.

#### Changes
An empty pagination link is serialized as nil, those are getting omitted now. This change affects the JsonApi adapter.

#### Caveats
Maybe omitting or returning empty pagination links should rather be optional?

#### Related GitHub issues
PR #1668 addresses a similar issue for links.

There's an open issue on the json-api repo pointing out the contradiction for pagination links within the specs. https://github.com/json-api/json-api/issues/887


bf4:
> I wanted to validate my API against the schema provided by json-api. 

The schema isn't authoritative.  You shouldn't rely on it too strongly.  Are null values allowed by the specs? I believe we recently added this behavior for the opposite reason.

mschneider85:
The spec says 
> Each member of a links object is a “link”. A link MUST be represented as either:
a string containing the link’s URL.
an object (“link object”) [...]
For the pagination links (subset of links) it is: 
> Keys MUST either be omitted or have a null value
So I'm not quite sure how it is meant to be. In doubt I tended to follow the provided schema.

lib/active_model_serializers/adapter/json_api/pagination_links.rb
```diff
@@ -26,7 +26,7 @@ def as_json
             prev:  prev_page_url,
             next:  next_page_url,
             last:  last_page_url
-          }
+          }.compact
```
bf4: did not know about this on hash. thanks!

bf4:
I assume you're vendoring the schema, in which case I suggest you modify it and consider submitting a PR to json schema to allow a nil value.

PR #2237, opened 2018-01-29
0 10 stable

PolymathWhiz:
#### Using JSONAPI Spec

Create a file named `active_model_serializer.rb` under config/intializer and type `ActiveModel::Serializer.config.adapter = :json_api`


bf4:
Not sure what this is

PR #2239, opened 2018-02-02
Fix link in README.md

iacobus:
#### Purpose
Fix link to `fast_jsonapi`, which is pointing at the `active_model_serializers` repo.

#### Changes
Point the link to the correct repo.

PR #2240, opened 2018-02-05
Fix wrong link of fast_jsonapi gem in README

dopin:


dopin:
Oops, it's duplicated with #2239 sorry

PR #2253, opened 2018-05-09
[docs] disable logger via /dev/null instead of AS::N.unsubscribe

zvkemp:
#### Purpose

The advice given to unsubscribe from `ActiveSupport::Notifications` is a bit over-eager, as `#unsubscribe` will remove listeners other than the `ActiveModelSerializers::LogSubscriber`. A better solution to disable logging is to redirect the logger to /dev/null.

#### Changes

Alter logger documentation

### DIFF
--- a/docs/general/logging.md
+++ b/docs/general/logging.md
@@ -16,6 +16,5 @@ ActiveModelSerializers.logger = Logger.new(STDOUT)
 You can also disable the logger, just put this in `config/initializers/active_model_serializers.rb`:
 
 ```ruby
-require 'active_model_serializers'
-ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)
+ActiveModelSerializers.logger = Logger.new(IO::NULL)
 ```


bf4:
@zvkemp that surprises me.  Are you saying that if I unsubscribe from `"sql.active_record"` it unsubscribes all listeners? Something somewhere isn't what it seems.  The content of the PR, I think, is incorrect, but would you mind looking into this more?

zvkemp:
@bf4 Yes, passing a string to `ActiveSupport::Notifications.unsubscribe` unsubscribes all matching listeners, including those that matched via regex (although the regex issue is only somewhat related to the point of this change— I've opened another pr to address that: https://github.com/rails/rails/pull/32861).

Source: https://github.com/rails/rails/blob/375a4143cf5caeb6159b338be824903edfd62836/activesupport/lib/active_support/notifications/fanout.rb#L34

Docs:
http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#method-c-unsubscribe
> Note that this will unsubscribe all subscriptions with the given name

Given that the purpose of AS::N is to decouple the handling of notifications from their senders, I also believe it is a mistake to encourage the usage of a blanket `unsubscribe` to disable **one** specific handler (in this case, the logger), because there may be handlers registered other than this logger.

Issue #1865, opened 2016-08-02
How to exclude an association in AMS 10 comparing to AMS 9


MartinKoerner:
Using rails 4, I have a domain object with a Serializer
```ruby
class DomainObject < ActiveRecord::Base
  ...
end

class DomainObjectSerializer < ActiveModel::Serializer
  attributes :attr1, :attr2, :attr3
  has_many :foos
  has_many :bars
end
```
Using active_model_serializer v0.9.4 I could do

```ruby
record = DomainObject.find(..)
serializer = DomainObjectSerializer.new(record, root: false, except: [:foos])
serializer.to_json
```
to get a JSON-String containing all the attributes and the association `:bars`, but not the association `:foos`.

After switching to active_model_serializer v0.10.2 this does not work anymore. The string contains all attributes. It seems, like the following gives me the same output:

```ruby
record = DomainObject.find(..)
serializer = DomainObjectSerializer.new(record)
serializer.to_hash.to_json(except: [:foos])
```
But as the `to_json` call typically is done in the view layer, while the serializer is build inside a model or service, it seems not to be a good solution. Did I miss some configuration option?

bf4:
AMS 0.10.2 doesn't (yet) support the Rails `as_json` semantics. 

Related https://github.com/rails-api/active_model_serializers/issues/1851 and maybe https://github.com/rails-api/active_model_serializers/issues/1849 and https://github.com/rails-api/active_model_serializers/pull/1845

Probably you'd want something like:

```ruby
serializer.to_json(fields: DomainObjectSerializer._attributes - :foos)
```

This is definitely an area that needs work for people coming from 0.8/0.9 used to working directly with serializers. (0.10 would say to work with ActiveModelSerializers::SerializableResource, but in this case, where you're not using an adapter, it wouldn't make a difference )

bf4:
Related https://github.com/rails-api/active_model_serializers/issues/1643

PR #2259, opened 2018-06-06
Optional relationships

jacobbullock:
#### Purpose
With models that have many relationships, especially nested relationships, automatically including those relationships by default can make an otherwise fast query take a long time. Allowing an easy way to make relationships optional unless specifically requested through `include` is essential for performance

#### Changes
Only include relationships in the `associations` method if they are not defined as optional OR if they are optional and specified in the include request

There is also a small update to the `.gitignore` for excluding intelliJ project files  

#### Caveats
This has worked perfectly for our existing needs including a number of polymorphic relationshisp, but may need further testing.

#### Related GitHub issues
#1865 

In the following serializer:

        has_many :comments, :optional => true
        attributes :id, :title, :body
    end

`comments` will not be the response unless asked for

    render  :json => :post, :include  => "comments" # will include comments

    render  :json => :post, # will only include post data

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ tags
 Icon?
 ehthumbs.db
 Thumbs.db
+
+# sdk
+.idea
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -343,6 +343,7 @@ module ActiveModel
       Enumerator.new do |y|
         self.class._reflections.each do |key, reflection|
           next if reflection.excluded?(self)
+          next if reflection.options[:optional] && ( instance_options[:included].nil? || !instance_options[:included].include?(key.to_s) )
           next unless include_directive.key?(key)
 
           association = reflection.build_association(self, instance_options, include_slice)


bf4:
Interesting idea. I assume this is for JSON:API?  Would you consider using a 'minimal' serializer?

richmolj:
Not sure if it still exists in master, but I actually added exactly this in [this PR](https://github.com/rails-api/active_model_serializers/pull/1931).

(or use [jsonapi suite](https://jsonapi-suite.github.io/jsonapi_suite) and it's automatic 🙂 )

jacobbullock:
@richmolj that PR looks interesting,  that looks more thought out than this.   I couldn't find anything about that while I was searching for how to do this.  I'll take  a look and  see if this works in lieu of what i  posted here.  I  was going to use the jsonapi suite, but i like the auto serializing that AMS does that I  believe is not in any of the other serializers.

jacobbullock:
@bf4 I am not quite sure I understand what you mean  by  'minimal' serializer

bf4:
> @bf4 I am not quite sure I understand what you mean by 'minimal' serializer

e.g.

```ruby
  attribute :name
end
class UserPostsSerializer < UserSerializer
  has_many :posts
end
class UserAllTheThingsSerializer < UserSerializer
  has_many :posts
  belongs_to: :blog
  has_many :roles
  # etc
end

PR #2286, opened 2018-09-24
Fix media type parameter (charset) in JSON:API responses (#2285)

### LINKED ISSUE #2285: A media type parameter (charset) is set for JSON API, againsst the standard

Expected the `jsonapi` media type responses to have the header `Content-Type: application/vnd.api+json` _without_ any media type parameters, as per the JSON:API specification. Instead, `; charset=utf-8` is present in the response `Content-Type`

#### Steps to reproduce
Load `active_model_serializers/register_jsonapi_renderer.rb` and make a response in the `jsonapi` format.

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 0.10.7

Output of `ruby -e "puts RUBY_DESCRIPTION"`: `ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]`

OS Type & Version: GNU/Linux 4.17 (Debian Stretch)

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails 5.2

tkalliom:
#### Purpose
Fixes #2285

#### Changes
The renderer sets the response Content-Type headerstring instead of using the higher-level interface where the media type is passed as a Mime.

#### Caveats
Perhaps the `headers` interface is a bit too low-level?

#### Related GitHub issues
#2285

### DIFF
--- a/lib/active_model_serializers/register_jsonapi_renderer.rb
+++ b/lib/active_model_serializers/register_jsonapi_renderer.rb
@@ -43,7 +43,7 @@ module ActiveModelSerializers
       # ref https://github.com/rails/rails/pull/21496
       ActionController::Renderers.add :jsonapi do |json, options|
         json = serialize_jsonapi(json, options).to_json(options) unless json.is_a?(String)
-        self.content_type ||= Mime[:jsonapi]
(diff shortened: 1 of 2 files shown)


bf4:
I'm pretty sure this is fine, but thanks for the issue!

PR #2294, opened 2018-10-19
Removes null links from pagination

langalex:
#### Purpose

Become (more) compliant with JSON-API.

#### Changes

Remove null links from pagination.

#### Caveats

None.

#### Related GitHub issues

None.

I ran the pagination links generated by active_model_serializers against the [JSON-API spec](https://jsonapi.org/format/#document-links), and it turns out links must not be null. See also the [schema](https://github.com/json-api/json-api/blob/gh-pages/schema#L176). Only strings and objects are allowed.

### DIFF
--- a/lib/active_model_serializers/adapter/json_api/pagination_links.rb
+++ b/lib/active_model_serializers/adapter/json_api/pagination_links.rb
@@ -26,7 +26,7 @@ module ActiveModelSerializers
             prev:  prev_page_url,
             next:  next_page_url,
             last:  last_page_url
-          }
+          }.compact
         end
 
         protected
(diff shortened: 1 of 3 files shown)


bf4:
@langalex Thanks for the issue

The behavior of AMS is actually correct.  You linked to document links.  If you scroll down to https://jsonapi.org/format/#fetching-pagination  you'll see that null is allowed.

In addition, the schema is not authoritative.

langalex:
Ok sorry about that. I ran the responses of my tails API tests against the schema and that failed, so I guess the schema is wrong in this case. Do you have any idea how much sense it would make to try to get it updated?

langalex:
Never mind. I'll do some proper research before asking any more uninformed questions.

bf4:
@langalex No need to be hard on yourself.  You're asking the same questions I and many others have asked :)

langalex:
Turns out the schema does in fact not allow null links even though it probably should. But [there's already an issue for that](https://github.com/json-api/json-api/issues/887)

bf4:
see #2224 
and https://github.com/rails-api/active_model_serializers/pull/1668

PR #2297, opened 2018-10-24
Remove unnecessary line break from exception message

vnbrs:
This removes the unnecessary line break from the exception message. Example:

```
Cannot infer root key from collection type. Please\n
specify the root or each_serializer option, or render a JSON String
```


bf4:
Thanks!

PR #2307, opened 2018-12-29
Reevaluation of object

faucct:
Object gets reevaluated when `value` is `nil` or `false`. I don't know where to put a test on this.

### DIFF
--- a/lib/active_model/serializer/lazy_association.rb
+++ b/lib/active_model/serializer/lazy_association.rb
@@ -11,7 +11,8 @@ module ActiveModel
       end
 
       def object
-        @object ||= reflection.value(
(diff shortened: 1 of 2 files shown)


lib/active_model/serializer/lazy_association.rb
```diff
@@ -11,7 +11,8 @@ def reflection_options
       end
 
       def object
-        @object ||= reflection.value(
+        return @object if defined?(@object)
+        @object = reflection.value(
```
bf4: looks fine to me. I write code like this all the time.

mind adding an entry to the changelog under fixes?

faucct: done

Issue #2316, opened 2019-02-24
Deprecation warning in Rails 6.0.0 beta


iggant:
DEPRECATION WARNING: `Module#parent` has been renamed to `module_parent`. `parent` is deprecated and will be removed in Rails 6.1. (called from namespace_for_serializer at /usr/local/bundle/gems/active_model_serializers-0.10.9/lib/action_controller/serialization.rb:26)

xyingsoft:
I have deprecation warning after upgrading rails,
So I tried to use master branch, but its not working with rails 6 at all.

```

Failure/Error: require File.expand_path("../../config/environment", __FILE__)
FrozenError:
  can't modify frozen Array
# ./config/environment.rb:5:in `<top (required)>'
# ./spec/rails_helper.rb:5:in `<top (required)>'

```
When will be next version released for rails 6?

iggant:
don't use master branch use 


`gem 'active_model_serializers', '~> 0.10.0', require: true, github: 'rails-api/active_model_serializers', branch: '0-10-stable'`

xyingsoft:
Thanks @iggant, I will try now

xyingsoft:
Awesome it works, thanks @iggant

PR #2319, opened 2019-02-27
Fix deprecation warnings on Rails 6

kylekeesling:
#### Purpose
Fix depreciation warnings in Rails 6.0.0.beta2

```
DEPRECATION WARNING: `Module#parent` has been renamed to `module_parent`. `parent` is deprecated and will be removed in Rails 6.1.
~/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/active_model_serializers-0.10.9/lib/action_controller/serialization.rb:26:in `namespace_for_serializer'
```
#### Changes
changes `parent` calls to `module_parent`

#### Related GitHub issues
#2316

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Breaking changes:
 Features:
 
 Fixes:
+- [#2319](https://github.com/rails-api/active_model_serializers/pull/2319) Fixes #2316. (@kylekeesling)
+  - Fix Rails 6.0 deprication warnings
+  - update test fixture schema to use `timestamps` instead of `timestamp`
 
 Misc:
 
--- a/lib/action_controller/serialization.rb
+++ b/lib/action_controller/serialization.rb
@@ -23,7 +23,15 @@ module ActionController
     end
 
     def namespace_for_serializer
-      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+      @namespace_for_serializer ||= namespace_for_class(self.class) unless namespace_for_class(self.class) == Object
+    end
+
+    def namespace_for_class(klass)
+      if Module.method_defined?(:module_parent)
+        klass.module_parent
+      else
+        klass.parent
+      end
     end
 
     def serialization_scope
--- a/test/action_controller/namespace_lookup_test.rb
+++ b/test/action_controller/namespace_lookup_test.rb
@@ -125,7 +125,12 @@ module ActionController
       tests Api::V3::LookupTestController
 
       setup do
-        @test_namespace = self.class.parent
+        @test_namespace =
+          if Module.method_defined?(:module_parent)
+            self.class.module_parent
+          else
+            self.class.parent
+          end
       end
 
       test 'uses request headers to determine the namespace' do
--- a/test/fixtures/active_record.rb
+++ b/test/fixtures/active_record.rb
@@ -19,28 +19,28 @@ ActiveRecord::Schema.define do
     t.text :contents
     t.references :author
     t.references :post
-    t.timestamp null: false
+    t.timestamps null: false
   end
   create_table :employees, force: true do |t|
     t.string :name
     t.string :email
-    t.timestamp null: false
+    t.timestamps null: false
   end
   create_table :object_tags, force: true do |t|
     t.string :poly_tag_id
     t.string :taggable_type
     t.string :taggable_id
-    t.timestamp null: false
+    t.timestamps null: false
   end
   create_table :poly_tags, force: true do |t|
     t.string :phrase
-    t.timestamp null: false
+    t.timestamps null: false
   end
   create_table :pictures, force: true do |t|
     t.string :title
     t.string :imageable_type
     t.string :imageable_id
-    t.timestamp null: false
+    t.timestamps null: false
   end
 end
 


wasifhossain:
Looks like a good issue to be solved. So we can keep using `Module#parent` until Rails 6.1, after which we have to rename it to `module_parent` as listed in https://github.com/rails/rails/blob/master/activesupport/CHANGELOG.md#rails-600beta1-january-18-2019

@bf4, according to AMS policy of supporting **existing** rails versions and the **new** ones to come, what would be the best way to update our codebase? something like this?
```
self.class.respond_to?(:module_parent) ? self.class.module_parent : self.class.parent
```
or make a monkey-patch inside AMS?

kylekeesling:
Happy to make this tweak once we figure out the best way to ensure backward compatibility.

bf4:
Since this is in a test, I don't mind defining it inline. However, it might be nice to add a rails6 shim test helper like we have for rails5, and add a `def module_parent` method or something like that


If it were used at runtime, I'd have suggested something ilke

Something like

```ruby
require 'active_support/core_ext/module'
if Module.method_defined?(:module_parent)
  def module_parent(klass)
    klass.module_parent
  end
else
  def module_parent(klass)
      klass.parent
  end
end
```

kylekeesling:
Sorry - I fixed the test but there were two more calls in `action_controller/serialization.rb` - just force pushed an update. Happy to make any more tweaks you think are necessary to get this working for older versions.

iggant:
we can use something like 
`if AS_GEM_VERSION > Gem::Version.new("5.3")`
in the method definition, and this will be resolved in require (just once) 

`AS_GEM_VERSION = ::ActiveSupport.gem_version.dup.freeze`

lib/action_controller/serialization.rb
```diff
@@ -23,7 +23,7 @@ def serialization_scope(scope)
     end
 
     def namespace_for_serializer
-      @namespace_for_serializer ||= self.class.parent unless self.class.parent == Object
+      @namespace_for_serializer ||= self.class.module_parent unless self.class.module_parent == Object
```
bf4:
```suggestion
      @namespace_for_serializer ||= namespace_for_class(self.class) unless namespace_for_class(module_parent) == Object
    end
    
    if Module.method_defined?(:parent)
      def namespace_for_class(klass)
        klass.parent
      end
    elsif Module.method_defined?(:module_parent)
      def namespace_for_class(klass)
        klass.module_parent
      end
    else
      fail "Module must define either parent or module_parent"
    end
```

bf4: I mean it could be a utility, but in any case, it's just ducky, and only runs once at runtime.  Other gems I've worked on like RSpec are full of this stuff

test/action_controller/namespace_lookup_test.rb
```diff
@@ -125,7 +125,7 @@ def namespace_set_by_request_headers
       tests Api::V3::LookupTestController
 
       setup do
-        @test_namespace = self.class.parent
+        @test_namespace = self.class.module_parent
```
bf4: It could call `Api::V3::LookupTestController.new.namespace_for_serializer` I guess

lib/action_controller/serialization.rb
kylekeesling: I changed `namespace_for_class(module_parent)` to `namespace_for_class(self.class)`

kylekeesling:
@bf4 I think I've got it. I changed the order of the `Module.method_defined?` check. It made more sense in my mind to check for the new, updated method name, and then fallback to the older method name. 

Tested in my app and it seems to be working just fine w/o any deprivation warnings

test/action_controller/namespace_lookup_test.rb
iggant: but to make tests pass for previous rails version you also have to use if Module.method_defined?(:module_parent)

wasifhossain: `Api::V3::LookupTestController.new.namespace_for_serializer # => ActionController::Serialization::NamespaceLookupTest::Api::V3` 

But

`self.class.parent # => ActionController::Serialization`. 

@bf4 Will it make any issue if we hardcode the namespace for this case only?

or we can define `namespace_for_class` in a global namespace so that we can call it from here?

kylekeesling: good catch - just pushed an update that fixes this

lib/action_controller/serialization.rb
```diff
@@ -23,7 +23,19 @@ def serialization_scope(scope)
...
+    if Module.method_defined?(:module_parent)
+      def namespace_for_class(klass)
+        klass.module_parent
+      end
+    elsif Module.method_defined?(:parent)
+      def namespace_for_class(klass)
+        klass.parent
+      end
+    else
+      fail "Module must define either parent or module_parent"
```
iggant: ok, last change by rubocop `fail 'Module must define either parent or module_parent'`

iggant:
@kylekeesling  last change left

lib/action_controller/serialization.rb
kylekeesling: sorry - for some reason I didn't see this - just changing double quotes to singles and pushing back up shortly

kylekeesling:
@iggant - just pushed the style changes - let me know if I missed anything else

wasifhossain:
@bf4 there has been an issue with how the timestamp is added to the db tables resulting in https://travis-ci.org/rails-api/active_model_serializers/jobs/504045421#L771 against Rails `master` branch.

do you suggest adding the fix (`t.timestamps`) in https://github.com/rails-api/active_model_serializers/blob/0-10-stable/test/fixtures/active_record.rb in a separate PR?

bf4:
Might as well here if OP is up for it

kylekeesling:
@bf4 @wasifhossain I just pushed another update that swaps out all of the old calls to `t.timestamp` to `t.timestamps`

iggant:
@wasifhossain  probably now in the appveyor/pr we have to uninstall bundler 2 and install bundle 1.17 

```
gem update --system 2.7.8
gem install bundler -v '1.17.3'
```

or maybe just relax bundler dependency here: active_model_serializers.gemspec
` spec.add_development_dependency 'bundler', '> 1.6'`

wasifhossain:
@iggant there is a conflicting case between appveyor/travis on bundler version. Last time when I tried to relax it to use the latest version (2.0.1), appveyor got happy but travis failed some tests due to < 2 dependency for some other gem (i can't remember the exact gem name, pardon me).

@bf4 I am not quite sure how to go forward with this conflict between appveyor/travis. do you have any suggestion?

kylekeesling:
Happy to apply any fixes as soon as we figure out what's needed.

iggant:
@kylekeesling @wasifhossain bundle problem seems unrelated with this PR, and will occur in any PR
I propose to merge this PR and fix bundle issue in another.

lib/action_controller/serialization.rb
```diff
@@ -23,7 +23,19 @@ def serialization_scope(scope)
...
+    if Module.method_defined?(:module_parent)
+      def namespace_for_class(klass)
+        klass.module_parent
+      end
+    elsif Module.method_defined?(:parent)
+      def namespace_for_class(klass)
+        klass.parent
+      end
+    else
+      fail 'Module must define either parent or module_parent'
```
wasifhossain: I would prefer to have it this way:
```
def namespace_for_class(klass)
  if Module.method_defined?(:module_parent)
    klass.module_parent
  else
    klass.parent
  end
end
```
I dont think the `fail` case would ever happen as Rails is supposed to provide either `Module.parent` upto Rails 6.0 at least and `Module.module_parent` since Rails 6.1 afaik

kylekeesling: I agree - adjusting this now. Look for the commit shortly.

kylekeesling:
just pushed the update suggested via @wasifhossain

iggant:
@kylekeesling lib/action_controller/serialization.rb:37:1: C: Style/EmptyLines: Extra blank line detected.

kylekeesling:
@iggant - sorry should be fixed now

wasifhossain:
@kylekeesling The PR LGTM! can you please add an entry under fixes in the changelog?

After that, I have np to merge!

kylekeesling:
ok cool - done. Hopefully this is what you're looking for:
https://github.com/kylekeesling/active_model_serializers/blob/rails6/CHANGELOG.md

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Breaking changes:
 Features:
 
 Fixes:
+- [#2319](https://github.com/rails-api/active_model_serializers/pull/2319)
+  Fix Rails 6.0 deprication warnings and update test fixture schema to use `timestamps` instead of `timestamp`
+  Fixes #2319. (@kylekeesling)
```
wasifhossain: @kylekeesling you're almost done, but I guess this PR would fix #2316, no? :smile:

iggant:
@kylekeesling last one change to merge PR

CHANGELOG.md
kylekeesling: lol - oops - I guess I fixed my own PR! Changing now.

kylekeesling:
just pushed the updated changelog entry

wasifhossain:
awesome! there we go :+1: great work! @kylekeesling

cesarjr:
Hi!

Is there any chance of this pull request be backported to 0.9.x? 😬

bf4:
@cesarjr it's actually just about done https://github.com/rails-api/active_model_serializers/pull/2373

cesarjr:
@bf4 thanks guy! I used the commit and it's working well!!!

PR #2223, opened 2017-11-20
Fieldset support in Attributes/JSON adapters

bf4:
Supersedes stale PR https://github.com/rails-api/active_model_serializers/pull/2100
Closes: https://github.com/rails-api/active_model_serializers/pull/2100, https://github.com/rails-api/active_model_serializers/pull/2278
Addresses partially: https://github.com/rails-api/active_model_serializers/issues/2273

PR #2321, opened 2019-03-11
 Stick to bundle < 2

iggant:


bf4:
@iggant you're welcome to keep plugging at this!

wasifhossain:
[Travis issue with `timestamp`](https://github.com/rails-api/active_model_serializers/pull/2319#issuecomment-471217980) appeared in this PR again and will continue the same for each new PR until we merge [this commit](https://github.com/rails-api/active_model_serializers/pull/2319/commits/44efb10b89eb5333b25d13187e18b696dc5244f8) by @kylekeesling separately on `0-10-stable`.

If we do that, new PRs will have at least 1 less problem to solve :)

wasifhossain:
@iggant would you mind rebasing the PR on latest `0-10-stable`? If you have some time, lets make some progress with this in coming days.

iggant:
@wasifhossain yes will do this on Friday

iggant:
@wasifhossain rebased and squashed 
please have a look

appveyor.yml
```diff
@@ -13,6 +14,9 @@ cache:
 
 install:
   - SET PATH=C:\%ruby_version%\bin;%PATH%
+  - gem uninstall bundler -a -x -I
+  - gem update --system 2.7.8
```
wasifhossain: @iggant thanks a lot! btw can we push the version to 2.7.9, which is known to be the last rubygems version before 3.0. We have to keep using that until we drop support for Ruby < 2.3.

But still bundler version needs to be < 2, as long as we support Rails < 5.

test/cache_test.rb
```diff
@@ -164,7 +164,7 @@ def test_expiring_of_cache_at_update_of_record
 
       assert_equal 'Foo', author_json[:name]
 
-      author.update_attributes(name: 'Bar')
+      author.update(name: 'Bar')
```
wasifhossain: :+1:

test/cache_test.rb
```diff
@@ -202,10 +202,12 @@ def test_cache_expiration_in_collection_on_update_of_record
       end
 
       bar = 'Bar'
-      author.update!(name: bar)
+      Timecop.travel(10.seconds.from_now) do
```
wasifhossain: @iggant can you please elaborate the case why you need **Timecop** in this case?

wasifhossain:
@iggant thanks a lot for your effort. Appveyor became green again :raised_hands:

appveyor.yml
iggant: done

test/cache_test.rb
iggant: before this change there were random test failture 
probably this is because cache version relay on updated_at column which doesn't contain millisecond part, so time to time it doesn't invalidate cache. 
and with travel to 10 second ahead we eliminate this possibility 
probably current check in the 172 line is also related, but I will thy to change that test in the next PR

iggant:
@wasifhossain made requested changed
also can I ask to release new version of the gem, because we already using rails rc1 and ams from github version, but want to come back to release

test/cache_test.rb
wasifhossain: perfect reason :ok_hand: can't wait to see you apply the same logic for the other test too on line 167. You can do it in your next PR. Then we may also be able to remove these weird checks
```ruby
if ENV['APPVEYOR'] && actual != expected
  skip('Cache expiration tests sometimes fail on Appveyor. FIXME :)')
```

wasifhossain:
I would like to have a release too. Btw we got another important PR https://github.com/rails-api/active_model_serializers/pull/2329, which is supposed to fix an issue introduced by https://github.com/rails-api/active_model_serializers/pull/2223.

I will try to focus on that one after merging this PR. Once I am done with that, I will request @bf4 to consider for a new release.

Thanks a lot! @iggant

PR #2323, opened 2019-03-21
Fix serializer class link on guides

feliperenan:
The link was pointing out to the master branch instead of the v-10-stable branch.


wasifhossain:
thanks @feliperenan :+1:

PR #2149, opened 2017-06-05
Always include self, first, last pagination links

mecampbellsoup:
Found a bug in jsonapi adapter. Basically, when there’s one page, all links are omitted. The specs says

> Keys MUST either be omitted or have a null value to indicate that a *particular link* is *unavailable*.

It doesn’t say “remove pagination entirely”. A book with one page still has a first/current/last page. It doesn’t contain `next` or `prev`.

PR #2326, opened 2019-04-12
option to omit 'last' page link from json:api responses

alexdean:
#### Purpose

Allow jsonapi users to opt out of calculating the `last` page link. 

For complex record sets, the `COUNT(*)` query required to determine the total number of pages adds additional overhead which isn't entirely necessary. Clients can still consume all records by continuing to request pages until the 'next' link is `nil`.

In local testing, this changed one of my endpoints as such (with rails 5.2.3)

```
# without patch
Completed 200 OK in 1536ms (Views: 103.6ms | ActiveRecord: 1431.0ms)

# with patch
Completed 200 OK in 994ms (Views: 100.1ms | ActiveRecord: 892.8ms)
```

The difference is all attributable to the performance of the `COUNT(*)` query.

#### Changes

Adds a new configuration `ActiveModelSerializers.config.jsonapi_omit_total_pages`.
Defaults value is `false`, which maintains current behavior.

When this option is `true`:

  1. the `last` page link will always be `nil`
  2. a different method (which does not use `total_pages`) is used to calculate the `next` page

#### Caveats

If a collection contains a number of records which is evenly divisible by the page size, this option will cause the client to see one additional 'ghost' page (due to the way that the `next` page is calculated). That final page will contain 0 records and will have the expected `next: nil` link.

example:

if a record set contains 9 records, and the page size is 3, the 3rd page (records 7-9) will include a `next` link to page 4, which will have 0 records.

#### Related GitHub issues

  - #2149 ensured that all pagination links are always added to each response
  - json-api/json-api#887. that one is still open, but consensus of commenters seems to be that `null` pagination links are acceptable.

I believe this is compatible with the json:api pagination specification, which says

> Keys MUST either be omitted or have a null value to indicate that a particular link is unavailable.
https://jsonapi.org/format/#fetching-pagination

in other words, servers may provide a `last` link but are not obligated to.


alexdean:
CI failures seem unrelated to this PR.

## travis failure

on rub 2.5.3 + rails master with the following error

```
/home/travis/.rvm/gems/ruby-2.5.3/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:408:in `block (2 levels) in replace_gem': Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? can't activate sqlite3 (~> 1.4), already activated sqlite3-1.3.13. Make sure all dependencies are added to Gemfile. (LoadError)
```

https://travis-ci.org/rails-api/active_model_serializers/jobs/519398637

## appveyor failure

```
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 1.6) x86-mingw32
  Current Bundler version:
    bundler (2.0.1)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?
Could not find gem 'bundler (~> 1.6)' in any of the relevant sources:
  the local ruby installation
```

https://ci.appveyor.com/project/bf4/active-model-serializers/builds/23811402/job/64lu1fo8a4gh3rqc

lib/active_model_serializers/adapter/json_api/pagination_links.rb
```diff
@@ -59,6 +61,20 @@ def prev_page_url
...
+            next_page_url_without_total_pages
+          else
+            next_page_url_with_total_pages
+          end
+        end
+
+        # avoid the 'COUNT(*)' query generated by `collection.total_pages`
+        def next_page_url_without_total_pages
+          return nil if collection.size < per_page
+          url_for_page(collection.current_page + 1)
```
bf4: interesting

alexdean:
@bf4 any interest in moving forward with this?

alexdean:
@bf4 raising this again. any chance you'd be willing to merge this?

PR #2329, opened 2019-05-01
Prevent undefined method for Array:Class error

jpawlyn:
#### Purpose
Fix `NoMethodError: undefined method 'model_name' for Array:Class`. This occurs when serializing an Array object.

This originates  from adding fieldset support - https://github.com/rails-api/active_model_serializers/pull/2223.

#### Changes

Add greater defensiveness when calling `json_key`.


test/serializers/serialization_array_test.rb
```diff
@@ -0,0 +1,40 @@
...
+        attributes :id, :name
+      end
+      class AuthorSerializer < ActiveModel::Serializer
+        attributes :id, :name
+      end
+
+      setup do
+        @authors = [Author.new(id: 1, name: 'Blog Author')]
+        @serializer_instance = ArraySerializer.new(@authors)
+        @serializable = ActiveModelSerializers::SerializableResource.new(@authors, serializer: ArraySerializer, adapter: :attributes)
```
wasifhossain: if we change the `adapter` to `:json` and pass an option like `root: 'items'` to meet https://github.com/rails-api/active_model_serializers/pull/2329/files#diff-d81a4fb04ad7026e9924afee4099ddbfR6, then **the tests will pass without any change in the lib/active_model/serializer.rb**
```ruby
@serializable = ActiveModelSerializers::SerializableResource.new(
  @authors,
  adapter: :json,
  root: 'items'
)
```

If the serializer class cannot be determined for the instances of the array `@authors`, e.g. the name of the serializer is `AuthorPreviewSerializer` instead of `AuthorSerializer`, then we need to pass the name using `each_serializer` option like this way:
```ruby
@serializable = ActiveModelSerializers::SerializableResource.new(
  @authors,
  each_serializer: AuthorPreviewSerializer,
  adapter: :json,
  root: 'items'
)
```
More information on the `root` key can be found here: https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/rendering.md#overriding-the-root-key

wasifhossain:
Thank you @jpawlyn for your PR. would like to know your feedback on my comment above.

jpawlyn:
Thanks @wasifhossain for the explanation and example. What I did for the array serialization test was to simplify a serializer from our app but I think I oversimplified it! 

I've added another top level attribute to better show the problem we're facing. If there's a solution that doesn't involve changing the code that would be really great.

test/serializers/serialization_array_test.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module ActiveModel
+  class Serializer
+    class SerializationArrayTest < ActiveSupport::TestCase
+      class ArraySerializer < ActiveModel::Serializer
+        attributes :total, :items
```
wasifhossain: @jpawlyn can you put a `type` option here with any name you like and rerun all the tests again. I believe it should pass the tests.
```ruby
type 'array'

attributes :total, :items # you can add any number of fields you like
```

The `type` value is used at the top level of the json if you use the `:json` adapter, but skipped for the `:attributes` adapter.

You can find the details of this option here: https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/serializers.md#type

Though it says this option is not useful with the `:attributes` adapter, but providing it in the non-resource serializers like yours will help overcome the issue that you are having.

Let me know if you still find any issue after providing this option. Thanks.

test/serializers/serialization_array_test.rb
```diff
@@ -0,0 +1,44 @@
...
+        attributes :id, :name
+      end
+      class AuthorSerializer < ActiveModel::Serializer
+        attributes :id, :name
+      end
+
+      setup do
+        @authors = [Author.new(id: 1, name: 'Blog Author')]
+        @serializer_instance = ArraySerializer.new(@authors)
+        @serializable = ActiveModelSerializers::SerializableResource.new(@authors, serializer: ArraySerializer, adapter: :attributes)
```
wasifhossain: https://github.com/rails-api/active_model_serializers/pull/2223 introduces the `fields` option for the `:attributes` adapter. So we can further trim down the fields that we want to render.
```ruby
@serializable = ActiveModelSerializers::SerializableResource.new(
  @authors,
  serializer: ArraySerializer,
  adapter: :attributes,
  fields: [:items]
)
```
But now this https://github.com/rails-api/active_model_serializers/blob/d182e60898d0fad956afbed4d9e0a5824baef06d/lib/active_model/serializer.rb#L369 will not be enough to prevent the issue:
> NoMethodError: undefined method `model_name'

test/serializers/serialization_array_test.rb
jpawlyn: That's great, thanks a lot for your help with this @wasifhossain, much appreciated. I'll close the PR now.

jpawlyn:
No need for PR since using `type` option solves the problem

test/serializers/serialization_array_test.rb
wasifhossain: you are most welcome! glad that it worked for you :raised_hands:

PR #2331, opened 2019-05-13
Fixes error on Rails 6 with Kaminari 1.1.1 paginated collection

mauriciozaffari:
This commit fixes NoMethodError (undefined method `model_name' for ::ActiveRecord_AssociationRelation:Class)

Checks for `model_name` in the object first and fallback to `class.model_name` if that's not found.

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -384,7 +384,11 @@ module ActiveModel
     def json_key
       root || _type ||
         begin
-          object.class.model_name.to_s.underscore
+          if object.respond_to?(:model_name)
+            object.model_name
+          else
+            object.class.model_name
+          end.to_s.underscore
         rescue ArgumentError
           'anonymous_object'
         end


bf4:
Any idea what changed?

mauriciozaffari:
For some reason, it looks like ActiveRecord classes doesn't respond to `model_name` on Rails 6, ActiveRecord instances do.

wasifhossain:
`model_name` is provided by `ActiveModel::Naming` module, which is **extended** in `ActiveRecord::Base`.

Btw I don't see any change in how the module is extended in Rails 6 rc1 vs older versions:
Rails 5.2.3: https://github.com/rails/rails/blob/v5.2.3/activerecord/lib/active_record/base.rb#L278
Rails 6.0.0.rc1: https://github.com/rails/rails/blob/v6.0.0.rc1/activerecord/lib/active_record/base.rb#L278

which means that any AR model class will still have the ability to call `model_name`, i.e. `object.class.model_name`, where `object` is an instance of an AR model.

wasifhossain:
@mauriciozaffari can you please give us detailed information on why you have had to make this change.

Thanks.

mauriciozaffari:
@wasifhossain just created a dummy Rails application to demonstrate the issue:
https://github.com/mauriciozaffari/active_model_serializers_issue_demonstration

mauriciozaffari:
@wasifhossain did you had the chance to look at this?

wasifhossain:
@mauriciozaffari  i tried to follow the project, but found it hard to jump into the exact scope where I could find the issue that you are trying to solve with your patch that is present in this PR.

Thanks for your efforts, anyways :+1:

mauriciozaffari:
@wasifhossain how so? You just need to run the specs.

wasifhossain:
@mauriciozaffari got it. [the serializer](https://github.com/mauriciozaffari/active_model_serializers_issue_demonstration/blob/master/app/serializers/paginated_comments_serializer.rb) is not a regular one, so you need to set the [`type`](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/serializers.md#type), e.g. 
```
type 'comments'
```
and run the spec again. It should pass.

For detailed info on non-regular serializer, you may take a look at a very recent discussion here: https://github.com/rails-api/active_model_serializers/pull/2329

Hope that helps. Thanks.

mauriciozaffari:
Gotcha, I'll try that.

However, I think this PR still holds some value. I don't see the point of calling `instance.class.model_name` when `instance.model_name` is available.

wasifhossain:
**Note for future contribution**:

We highly appreciate if the PR description follows the [issues template](https://github.com/rails-api/active_model_serializers/issues/new) too. Strong test code along with the proposed change would make it a good candidate to be merged.

wasifhossain:
@bf4 you can merge it if you like.

wasifhossain:
closing the PR due to lack of proper test that would highlight the fail case

PR #2328, opened 2019-04-29
Add support for Rails 6.0.0.rc1 on travis

wasifhossain:
Configure **Travis** to
- **include** tests for `Rails 6.0.0.rc1` using `Ruby 2.5+`
- **allow failures** for 
   - `Rails 4.1` using `Ruby 2.4+`
   - `Rails master` using `Ruby 2.5+` (until we support Rails 6 in the gemspec that matches Rails master version)
   - `ruby-head` (currently 2.7) against `Rails 4.2+`
- **exclude** tests for
   - `Ruby 2.1` against `Rails 5+` ([Rails 5.0+ supports Ruby 2.2.2+ only](https://weblog.rubyonrails.org/2016/6/30/Rails-5-0-final/))
   - `Rails 6+` for `Ruby < 2.5` ([Rails 6+ requires Ruby 2.5+](https://weblog.rubyonrails.org/2018/12/20/timeline-for-the-release-of-Rails-6-0/))
- Install `rubygems 2.7.9` (keep rubygems version < 3.0 until we keep supporting ruby < 2.3)

PR #2334, opened 2019-05-29
Allow AMS to run on Rails < 6.1

sikachu:
#### Purpose
Given that Rails 6.0.0.rc1 is out, the stable release of Rails 6.0.0 is coming soon, and that Rails' `master` version [already defining 6.1.0.alpha](https://github.com/rails/rails/blob/master/version.rb), I would like to propose us to remove the upper bound of Rails requirement to be open-ended so AMS can be installed and tested against edge Rails.

#### Changes
This commit <del>removes</del> increase the upper bound of `rails_versions` to `< 6.1.0`, which is being used when defining runtime dependencies.

#### Caveats
Note that this means that there's a chance that AMS will break when using with future unsupported version of Rails. However, by leaving the dependency version to be unbounded like this, I believe this will encourage other contributors to be able to catch the bug early and contribute back to AMS in time before the future version of Rails.

I also believe that there are a few companies that do test against edge Rails, and by doing this we'll allow them to have AMS in their project without having to maintain their separate fork which removes the upper limit of Rails version.

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/pull/2328 (added 6.0.0.rc1 to Travis build matrix)

None

Thank you very much.

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.1'
 
-  rails_versions = ['>= 4.1', '< 6']
+  rails_versions = ['>= 4.1', '< 6.1']
   spec.add_runtime_dependency 'activemodel', rails_versions
   # 'activesupport', rails_versions
   # 'builder'


wasifhossain:
I would prefer the upper bound still present, supporting the rails master version:
```
rails_versions = ['>= 4.1', '< 6.1']
```

sikachu:
@wasifhossain I brought back the upper bound. Please take another look.

Sexual:
Any update regarding this? @wasifhossain

wasifhossain:
the PR is already merged. would you please clarify what would you like to know about this PR? thanks

Sexual:
@wasifhossain expectations as to when the master branch will be released? It's been 5 months since v0.10.9 was released

wasifhossain:
@bf4 can surely help you on this. I am yet to have privileges to release a version btw :smile:

bf4:
@Sexual master probably will never be released again. 0-10-stable is being maintained by @wasifhossain 

@wasifhossain what's your rubygems.org email address? I'll add you as an owner.

wasifhossain:
@bf4 i have sent you an email to your rubygems account. thanks

PR #2340, opened 2019-06-08
Configure `ActionDispatch::Response#content_type` behavior on Rails 6+

wasifhossain:
#### Purpose

This PR will affect all `#content_type` calls in the following files:

https://github.com/rails-api/active_model_serializers/blob/55a6b23cd84395029aa92605f8aae73f37f57c4a/test/action_controller/serialization_test.rb#L158

https://github.com/rails-api/active_model_serializers/blob/55a6b23cd84395029aa92605f8aae73f37f57c4a/test/action_controller/explicit_serializer_test.rb#L74

#### Background

Rails 6 changes the default behavior of `ActionDispatch::Response#content_type`, which now returns the **Content-Type header as it is**.

To retain old behavior, i.e. retrieve the mime type only, we can either set the following configuration in application settings:
```rb
config.action_dispatch.return_only_media_type_on_content_type = true
```

or set it inline as:
```rb
ActionDispatch::Response.return_only_media_type_on_content_type = true
```

or call `#media_type` instead of `#content_type`.

Rails 6.0 ACTIONPACK release notes: 
https://github.com/rails/rails/blob/6-0-stable/actionpack/CHANGELOG.md

https://github.com/rails/rails/pull/36490
https://github.com/rails/rails/pull/36946

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ ruby_supported_versions:
 
 jruby_supported_versions:
   - &jruby_9_1 jruby-9.1.17.0
-  - &jruby_9_2 jruby-9.2.7.0
+  - &jruby_9_2 jruby-9.2.8.0
   - &jruby_head jruby-head
 
 jdk_supported_versions:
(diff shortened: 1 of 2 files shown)


wasifhossain:
@bf4 I would like to have your thoughts on this. Not sure if this is the best way to handle the case.

Issue #2350, opened 2019-07-21
`root` key for :attributes adapter is pointless


LeKristapino:
Before upgrading from 0.10.5 to 0.10.10 this line worked without any errors 
```
render json: {first_key: "some string",  second_key: "another string" } , serializer: SomeCustomSerializer
```

Now it produces an error `undefined method 'model_name' for Hash:Class`. 

Adding a `root` parameter (for example `..., serializer: SomeCustomSerializer, root: ''` fixes the issue.

BUT from the documentation [here](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/serializers.md#root) I figure that `root` is not even used for `:attributes` adapter.

My question is, shouldn't it be ignored in the `Attribute` and `JSON API` adapters, thus avoiding breaking the example serialization. Or have I missed some extra functionality that `root`  provides?

wasifhossain:
Hi @LeKristapino that was unfortunate. but upgrading to the latest HEAD of the `0-10-stable` branch should resolve the issue.

```
gem 'active_model_serializers', :git => 'https://github.com/rails-api/active_model_serializers.git', :branch => '0-10-stable'
```

This PR fixes the case: https://github.com/rails-api/active_model_serializers/pull/2344

wasifhossain:
@LeKristapino  I think I am wrong here. that PR does not fix the case yet! I see it now :(

Actually that was kind of like a side-effect while implementing the `fields` option for the `attributes` adapter: [docs/general/fields.md](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/fields.md)

can you please write a test for the scenario that you described, so that we can figure out an effective solution to avoid further side-effects related to this feature fix

LeKristapino:
@wasifhossain  I think it would take me some time to understand the differences in mechanics for calling `SomeSerializer.new( {name: "John"}).to_json` and the differences with `render json: {name: "John"}, serializer: SomeSerializer` but in first case there is no problem (and wasn't before). The second case produces an error in 0.10.10 but was working fine in 0.5.10.

wasifhossain:
@LeKristapino can you share some code to reproduce the fail case. Specifically how `SomeCustomSerializer` looks like when you pass `{first_key: "some string",  second_key: "another string" }` in `render json:` call

bf4:
[...]

PR #2344, opened 2019-07-17
Lookup fieldset using either string or symbol

wasifhossain:
Fixes https://github.com/rails-api/active_model_serializers/issues/2341

#### Related GitHub issues
https://github.com/rails-api/active_model_serializers/pull/2223

Issue #2343, opened 2019-07-17
Versioning support


pirvudoru:
#### Expected behavior vs actual behavior
According to: https://semver.org/ #3:
```
PATCH version when you make backwards-compatible bug fixes.
```
I expect patches to not break the app. 

But...
![image](https://user-images.githubusercontent.com/2707601/61363491-c2499200-a88c-11e9-8d21-7c3cf69c2bfc.png)

This library is a nightmare to work with when it comes to updating the version, and it has been this way since v 0.7.x (when we started using it - huge mistake).

pethron:
Same problem.
It's not possible to break things in a minor version upgrade, at least run the goddamn tests.

pethron:
<img width="1378" alt="Screenshot 2019-07-17 at 15 40 33" src="https://user-images.githubusercontent.com/12462586/61380252-532e6680-a8a9-11e9-9ece-23cb166cee03.png">

What the f*** is this???
Next time use https://github.com/auchenberg/volkswagen

FeminismIsAwesome:
Y'all should calm down. A) your test suite caught them, and while annoying that's easy enough to not upgrade or make an issue about it, it's an open source project you can contribute to since it needs love and support.
B) If you're using an undocumented feature and the standard docs suggest root as string anyways; it was easy for us to correct the issues we saw and move on. Or are you saying we should be testing things we don't know people care about? Once the issue was seen, they fixed it and added a test so it's not like they don't test they just don't do it the way you do.

pirvudoru:
:wave:

wasifhossain:
nevertheless I agree with you @pirvudoru :+1: 

should have followed the semver strictly. unfortunately we are not quite sure at this point in time if v1 would be released in near future, from which it would be easier to maintain the semver policy

bf4:
@wasifhossain 0.10 is supposed to be stable.  AMS is older than semver.  There shouldn't be breaking changes with 0.10.x. Just like rake of old, think of 0.10.x and 10.x

bf4:
@pethron I think you and @pirvudoru may be looking at master https://github.com/rails-api/active_model_serializers/blob/bd40fe68f0ba7332a4b86c793149ebbcd207ca9f/.travis.yml  the 0.10.x release comes from the 0-10-stable branch

wasifhossain:
@bf4 thanks for the point.

in that case, should we consider a new version 0.10.x.y, where y would be for patch.

[...]

PR #2353, opened 2019-07-24
prevent ActiveModel::Serializer::Fieldset.fields_for crashing in case…

aovertus:
#### Purpose
Current AMS implementation allows things like this:

`render json: users, each_serializer: UserSerializer, root: :users`

but the new fieldset class don't accept anything else than string for `#fields_for`

#### Changes

Allow `#fields_for` to receive a symbol and parse it to string before any manipulation

value is coming from `#json_key` method

### DIFF
--- a/lib/active_model/serializer/fieldset.rb
+++ b/lib/active_model/serializer/fieldset.rb
@@ -12,7 +12,7 @@ module ActiveModel
       end
 
       def fields_for(type)
-        fields[type.singularize.to_sym] || fields[type.pluralize.to_sym]
+        fields[type.to_s.singularize.to_sym] || fields[type.to_s.pluralize.to_sym]
       end
 
       protected


wasifhossain:
Hi @aovertus thanks for your PR :+1:

btw it has already been merged as part of https://github.com/rails-api/active_model_serializers/pull/2344. closing it anyways

aovertus:
@wasifhossain should not it be part of a new tag ? fetching the v0.10.10 will not provide the fix 🤔

wasifhossain:
@bf4 can you please suggest the best step that we can take here. (https://github.com/rails-api/active_model_serializers/issues/2343#issuecomment-513364958 as a reference)

thanks

bf4:
@wasifhossain 0.10.11 is fine as is 0.10.10.1, which is a thing.  I don't have a strong feeling

aovertus:
@bf4 @wasifhossain  can we settle on a tag so the update is available ? Since its a fix I would go for `0.10.10.1`

wasifhossain:
I would also prefer to tag it as 0.10.10.1, but the fact is, unless we create a new branch from `0-10-stable`, it does not really matter whether we label it as any of those; the next tags released after this proposed tag will make way to all projects having this gem version defined as `'~> 0.10.10'`.

another point that I would like to address before we make the release is https://github.com/rails-api/active_model_serializers/issues/2350#issuecomment-514398432. otherwise we will continue to receive relevant issues
> I agree that the attributes adapter has no need to call root

also we received a good number of issues most recently resulted from [lib/active_model/serializer.rb#L387](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/lib/active_model/serializer.rb#L387), which I think is due to lack of guide that tells about the requirement on how to define the PORO.

[docs/howto/serialize_poro.md](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/howto/serialize_poro.md) explains it nicely, but [the linter](https://www.rubydoc.info/gems/active_model_serializers/0.10.10/ActiveModel/Serializer/Lint/Tests#test_model_name-instance_method) only mentions that its required if caching is enabled.

I think we should highlight the point for the PORO that it should respond to `model_name` and be an instance of `ActiveModel::Name`. In that case, `:attributes` adapter can work fine without the need to have [`root`](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/serializers.md#root) set.

@bf4 any thoughts, please?

bf4:
- [ ] `0.10.10.1` is fine
- [ ] Should make an issue that the attributes adapter shouldn't invoke `json_key`
- [ ] Shouldn't require the poro to include ActiveModel::Name as the linter intended.  I'd be in favor of rescuing the NoMethodError in `json_key` and re-raising it with a message with the gist "no root specified and none could be inferred"

wasifhossain:
regarding **item 2** above:

https://github.com/rails-api/active_model_serializers/blob/0-10-stable/lib/active_model_serializers/adapter/attributes.rb#L8
```
instance_options[:fieldset] ||= ActiveModel::Serializer::Fieldset.new(fields_to_fieldset(instance_options.delete(:fields)))
```
right now, `instance_options[:fieldset]` is always assigned an instance of `ActiveModel::Serializer::Fieldset` whether `instance_options[:fields]` is present or not, causing `if (fieldset = adapter_options[:fieldset])` to always pass and [fieldset.fields_for(json_key)](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/lib/active_model/serializer.rb#L369) to be called every time.

rather we can set an instance of `ActiveModel::Serializer::Fieldset` to `instance_options[:fieldset]` only when `instance_options[:fields]` is present. this will make the `json_key` called only when the `fields` option is present.

still that would make the `attributes` adapter to depend on either of `root`, `_type` or a class that responds to `model_name` in order to take the benefit of the Fieldset class.

wasifhossain:
i agree that we can work on a PR to make the `attributes` adapter not invoke `json_key` at all. should I give it a try?

aovertus:
Nice! Let me know if I can be of anyway help.

wasifhossain:
@aovertus thanks for your hands! 

btw I need to take a dive into the internals and lets see what I can come up with in next few days.

bf4:
I think a first-level fix could be

```diff
-        instance_options[:fieldset] ||= ActiveModel::Serializer::Fieldset.new(fields_to_fieldset(instance_options.delete(:fields)))
+        if (fields_options = instance_options.delete(:fields)))
+          instance_options[:fieldset] ||= ActiveModel::Serializer::Fieldset.new(fields_to_fieldset(fields_options))
+        end
```

so that it's only even possible to get a fieldset when there are field options present

secondly, we need to think through how a fieldset in should even work without a json_key.  My feeling is, you can't have your cake and eat it too. If you want to use a fieldset, you need a 'json_key'.

I think if we make the above change, then we can raise a runtime error

changing

```ruby
      if (fieldset = adapter_options[:fieldset])
        options[:fields] = fieldset.fields_for(json_key)
      end
```

to 

```ruby
      if (fieldset = adapter_options[:fieldset])
        options[:fields] = fields_for!(fieldset)
      end

.....

def fields_for!(fieldset)
  fieldset.fields_for(fieldset_type)
end

def fieldset_type
  json_key
rescue NoMethodError
  raise NoMethodError, "Cannot determine fieldset type for serializer instance class=#{self.class.name} object=#{object.class.name}"
end
```

wasifhossain:
> you can't have your cake and eat it too

:smile:

kddeisz:
We were burned by this as well when we tried to upgrade. In case anyone ends up here because there's no new version out yet, here's the fix I'm using locally in an initializer:

```ruby
ActiveModel::Serializer::Fieldset.prepend(
  Module.new do
    def fields_for(type)
      super(type.to_s)
    end
  end
)
```

PR #2361, opened 2019-10-15
Use SHA1 instead of MD5 for caching

### LINKED ISSUE #2358: FIPS 140-2 compatibility
We'd like AMS to be compatible with FIPS 140-2 mode. This is currently not working because of the use of MD5 [here](https://github.com/rails-api/active_model_serializers/blob/ab47e983f587e93ece7ab2552537d1475a5eb234/lib/active_model/serializer/concerns/caching.rb#L59).

#### Steps to reproduce
*(e.g., detailed walkthrough, runnable script, example application)*
- install FIPS-compatible OpenSSL binary (this probably differs by OS)
- turn on FIPS mode for OpenSSL with `OpenSSL.fips_mode = true`
- try to invoke an AMS serializer

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 
0.10.0.rc3, 8981683b9fa8d0bcff0e76b3702d699689b787b4

Output of `ruby -e "puts RUBY_DESCRIPTION"`: 
```
ruby 2.5.6p201 (2019-08-28 revision 67796) [x86_64-linux]
```
OS Type & Version: 
Amazonlinux 2

Integrated application and version *(e.g., Rails, Grape, etc)*: 
Rails 5.2.3

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*
```
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
```
(there is no ruby stack trace, but I verified that `caching.rb` is the culprit)

#### Additonal helpful information
*(e.g., Gemfile.lock, configurations, PR containing a failing test, git bisect results)*

Just replacing MD5 with SHA1 in that file seems to work.
Also note that Rails added FIPS compatibility in 5.2: https://github.com/rails/rails/pull/31651

alexzherdev:
#### Purpose
Using SHA1 as the hashing algorithm allows AMS to be [FIPS 140-2](https://en.wikipedia.org/wiki/FIPS_140-2) compatible.

#### Changes
Only had to change one place where serializer file contents are being hashed.

#### Related GitHub issues
Fixes #2358.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Breaking changes:
 
 Features:
 
+- [#2361](https://github.com/rails-api/active_model_serializers/pull/2361) Add `ActiveModelSerializers.config.use_sha1_digests` to allow customization of the hashing algorithm used for serializer caching (@alexzherdev)
+
 Fixes:
 
 - [#2344](https://github.com/rails-api/active_model_serializers/pull/2344) Fixes #2341 introduced since #2223 (@wasifhossain)
--- a/docs/general/configuration_options.md
+++ b/docs/general/configuration_options.md
@@ -111,6 +111,18 @@ ActiveModelSerializers.config.serializer_lookup_chain.unshift(
 
 See [lookup_chain.rb](https://github.com/rails-api/active_model_serializers/blob/master/lib/active_model_serializers/lookup_chain.rb) for further explanations and examples.
 
+#### use_sha1_digests
+
+Determines which hashing algorithm to use internally when caching serializers.
+
+Possible values:
+
+- `true`
+- `false` (default)
+
+When `true`, ActiveModelSerializers will use SHA1. Otherwise, it will default to using MD5.
+Be warned that changing this value may result in serializer caches being invalidated.
+
 ## JSON API
 
 ##### jsonapi_resource_type
--- a/lib/active_model/serializer/concerns/caching.rb
+++ b/lib/active_model/serializer/concerns/caching.rb
@@ -56,7 +56,8 @@ module ActiveModel
         def digest_caller_file(caller_line)
           serializer_file_path = caller_line[CALLER_FILE]
           serializer_file_contents = IO.read(serializer_file_path)
-          Digest::SHA1.hexdigest(serializer_file_contents)
+          algorithm = ActiveModelSerializers.config.use_sha1_digests ? Digest::SHA1 : Digest::MD5
+          algorithm.hexdigest(serializer_file_contents)
         rescue TypeError, Errno::ENOENT
           warn <<-EOF.strip_heredoc
             Cannot digest non-existent file: '#{caller_line}'.
--- a/lib/active_model/serializer/concerns/caching.rb
+++ b/lib/active_model/serializer/concerns/caching.rb
@@ -56,7 +56,7 @@ module ActiveModel
         def digest_caller_file(caller_line)
           serializer_file_path = caller_line[CALLER_FILE]
           serializer_file_contents = IO.read(serializer_file_path)
-          Digest::MD5.hexdigest(serializer_file_contents)
+          Digest::SHA1.hexdigest(serializer_file_contents)
         rescue TypeError, Errno::ENOENT
           warn <<-EOF.strip_heredoc
             Cannot digest non-existent file: '#{caller_line}'.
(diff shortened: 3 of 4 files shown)


wasifhossain:
https://github.com/rails-api/active_model_serializers/blob/5d3f56f0a4e3083f34fddcf943e54aed5f5aa5c6/lib/active_model/serializer/concerns/caching.rb#L275

as the digest is added as part of the generated cache key, changing the digest class from MD5 to SHA1 would invalidate all caches.

Hence I would prefer to introduce a **config** that users would use to set their choice of digest function (MD5/SHA). By default, the config would use `MD5` so that users, who do not need to change it, will have their cache intact. Doc should also clearly mention about the cache invalidation fact on changing the digest function.

alexzherdev:
Sure. How about something like a boolean `ActiveModelSerializers.config.use_sha1_digests` just parroting Rails?

alexzherdev:
Updated with the proposed configuration option.

bf4:
pasting rails pr code

railtie

```ruby
      config.after_initialize do
        if app.config.active_support.use_sha1_digests
          ActiveSupport::Digest.hash_digest_class = ::Digest::SHA1
        end
```

alexzherdev:
So with something like
```rb
ActiveModelSerializers.config.use_active_support_digest = true
```
we would read `ActiveSupport::Digest.hash_digest_class`, otherwise `Digest::MD5`?

wasifhossain:
i think `hash_digest_class` config can be set as `Digest::MD5` by default to keep cache intact and only set it to `Digest::SHA1` if the user sets the corresponding boolean field, as pointed in https://github.com/rails-api/active_model_serializers/pull/2361#issuecomment-542884090

alexzherdev:
I'm confused. The code in https://github.com/rails-api/active_model_serializers/pull/2361#issuecomment-542884090 is Rails and `hash_digest_class` is Rails configuration. What should we do in AMS? Should we have a boolean flag, or an option to set the digest class? If someone could post a sample of what you're envisioning, that would be very helpful.

wasifhossain:
My thought was to enable AMS to use a configured digest class across the whole codebase in uniform way. Setting it in railtie would be a good solution in that case.

As AMS is only using it in this specific cache class, I think we can keep it simple for now. This PR looks good in that sense :+1:

I am looking forward to testing the changes soon. In the mean time, can you please add a relevant entry in the CHANGELOG. Thank you!

alexzherdev:
Awesome. Are there any plans to do a release in the near future?

wasifhossain:
we definitely have plans to do that. but we also have to complete some todos before.

in the mean time, you may wish to use the new changes by adding the gem in this way:
```rb
gem 'active_model_serializers', github: 'rails-api/active_model_serializers', branch: '0-10-stable'
```

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Breaking changes:
 
 Features:
 
+- [#2361](https://github.com/rails-api/active_model_serializers/pull/2361) Add `ActiveModelSerializers.config.use_sha1_digests` to allow customization of the hashing algorithm used for serializer caching (@alexzherdev)
```
wasifhossain: :+1:

wasifhossain:
would you mind explaining the new config in the DOC too: https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/general/configuration_options.md

thanks!

docs/general/configuration_options.md
```diff
@@ -111,6 +111,18 @@ ActiveModelSerializers.config.serializer_lookup_chain.unshift(
...
+
+Determines which hashing algorithm to use internally when caching serializers.
+
+Possible values:
+
+- `true`
+- `false` (default)
+
+When `true`, ActiveModelSerializers will use SHA1. Otherwise, it will default to using MD5.
+Be warned that changing this value may result in serializer caches being invalidated.
```
wasifhossain: :+1:

alexzherdev:
@wasifhossain let me know if there is anything else to handle here 🙇

lib/active_model/serializer/concerns/caching.rb
```diff
@@ -56,7 +56,8 @@ def _cache_digest
         def digest_caller_file(caller_line)
           serializer_file_path = caller_line[CALLER_FILE]
           serializer_file_contents = IO.read(serializer_file_path)
-          Digest::MD5.hexdigest(serializer_file_contents)
+          algorithm = ActiveModelSerializers.config.use_sha1_digests ? Digest::SHA1 : Digest::MD5
```
bf4: this is fine

PR #2362, opened 2019-10-25
Add allow_wildcard configuration

Borzik:
#### Purpose
jsonapi-renderer provides an `allow_wildcard` configuration which can be used to configure how `*` and `**` include directives are treated. Since these wildcards are specific to AMS, in some cases it might be better to completely disable them instead of calling `gsub(/\*/, '')` on every include parameter passed to the API.

#### Changes
This PR adds a way to disable these wildcards by updating how `JSONAPI::IncludeDirective` is initialized. Default behaviour remains the same. When `config.allow_wildcard` is `false`, wildcards will be ignored by [jsonapi-renderer](https://github.com/jsonapi-rb/jsonapi-renderer/blob/master/lib/jsonapi/include_directive.rb#L43).

#### Caveats
`default_includes` configuration, which is currently `*`, will not be respected and will be handled as if it was an empty string


test/action_controller/json/include_test.rb
```diff
@@ -114,6 +126,54 @@ def test_render_resource_with_deep_include
...
+          }
+
+          response = JSON.parse(@response.body)
+
+          assert_equal(expected, response)
+        end
+
+        def test_render_resource_with_deep_wildcard_include
+          assert_raise SystemStackError do
+            get :render_resource_with_deep_wildcard_include
```
Borzik: `**` wildcard would cause stack overflow in this case. I guess this test can be removed, but left it here as a PoC for now.

lib/active_model_serializers/adapter/json_api.rb
```diff
@@ -52,7 +52,7 @@ def self.fragment_cache(cached_hash, non_cached_hash, root = true)
 
       def initialize(serializer, options = {})
         super
-        @include_directive = JSONAPI::IncludeDirective.new(options[:include], allow_wildcard: true)
+        @include_directive = JSONAPI::IncludeDirective.new(options[:include], allow_wildcard: ActiveModel::Serializer.config.allow_wildcard)
```
bf4: wrong namespace here. otherwise looks good.

Borzik: Fixed

PR #2369, opened 2020-01-02
Code cleanup

wasifhossain:


lib/active_model/serializer.rb
```diff
@@ -73,7 +73,7 @@ def self.serializer_lookup_chain_for(klass, namespace = nil)
     # Used to cache serializer name => serializer class
     # when looked up by Serializer.get_serializer_for.
     def self.serializers_cache
-      @serializers_cache ||= ThreadSafe::Cache.new
+      @serializers_cache ||= Concurrent::Map.new
```
bf4: kind of odd.  I'm not sure where the dependency is coming from.  I know later versions of Rails introduce concurrent-ruby.  I guess thread_safe was coming from active_support?

lib/active_model/serializer/reflection.rb
```diff
@@ -87,8 +87,8 @@ def initialize(*)
       #       meta ids: ids
       #     end
       #   end
-      def link(name, value = nil)
-        options[:links][name] = block_given? ? Proc.new : value
+      def link(name, value = nil, &block)
+        options[:links][name] = block_given? ? block : value
```
bf4: I'm install 2.7 locally so I can figure out what you're seeing. As I wrote in the other PR, this change doesn't make sense to me

bf4: ok, this makes more sense
> lib/active_model/serializer/reflection.rb:91: warning: Capturing the given block using Proc.new is
deprecated; use `&block` instead

bf4: per https://github.com/ruby/ruby/commit/9f1fb0a17febc59356d58cef5e98db61a3c03550

lib/active_model/serializer.rb
wasifhossain: `thread_safe` is needed by 2 different sources:
- `active_support`
- `grape` (through `virtus`)

PR #2368, opened 2019-12-26
Add Ruby 2.7 support

wasifhossain:
#### Purpose
https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/

Note: this PR is meant to experiment some cases that might arise while supporting Ruby 2.7, as it installs bundler 2.1.2 by default, which might cause dropping support for 
- Ruby 2.1/2.2
- Rails 4.1/4.2

PR #2371, opened 2020-01-10
Adjust TravisCI Ruby matrix

wasifhossain:
#### Purpose
- bring back changes that were removed from #2368 unintentionally
- make travis green

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ cache:
     - vendor/bundle
 
 before_install:
+  - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
   - gem install bundler -v 1.17.3
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,10 @@ cache:
     - vendor/bundle
 
 before_install:
-  - "travis_retry gem update --system 2.7.9"
-  - "travis_retry gem install bundler -v '1.17.3'"
-install: bundle install --path=vendor/bundle --retry=3 --jobs=3
+  - gem install bundler -v 1.17.3
+
+install:
+  - bundle install --path=vendor/bundle --retry=3 --jobs=3
 
 script:
   - bundle exec rake ci
@@ -65,6 +66,9 @@ rvm:
   - *ruby_2_1
   - *ruby_2_2
   - *ruby_2_3
+  - *ruby_2_4
+  - *ruby_2_5
+  - *ruby_2_6
   - *ruby_2_7
   - *ruby_head
 
@@ -72,10 +76,9 @@ branches:
   only: 0-10-stable
 
 jobs:
+  fast_finish: true
+
   include:
-    - { rvm: *ruby_2_4,  env: *rails_4_2 }
-    - { rvm: *ruby_2_5,  env: *rails_4_2 }
-    - { rvm: *ruby_2_6,  env: *rails_4_2 }
     - { rvm: *jruby_9_1,  jdk: *jdk_8, env: *rails_4_1_jruby }
     - { rvm: *jruby_9_1,  jdk: *jdk_8, env: *rails_4_2_jruby }
     - { rvm: *jruby_9_1,  jdk: *jdk_8, env: *rails_5_0_jruby }
@@ -97,7 +100,14 @@ jobs:
     - { rvm: *jruby_head, jdk: *jdk_8, env: *rails_master_jruby }
 
   exclude:
+    - { rvm: *ruby_2_4,  env: *rails_4_1 }
+    - { rvm: *ruby_2_5,  env: *rails_4_1 }
+    - { rvm: *ruby_2_6,  env: *rails_4_1 }
+    - { rvm: *ruby_2_7,  env: *rails_4_1 }
     - { rvm: *ruby_head, env: *rails_4_1 }
+
+    - { rvm: *ruby_2_7,  env: *rails_4_2 }
+    - { rvm: *ruby_head, env: *rails_4_2 }
 
     - { rvm: *ruby_2_1,  env: *rails_5_0 }
     - { rvm: *ruby_2_1,  env: *rails_5_1 }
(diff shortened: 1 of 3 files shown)


wasifhossain:
slightly different status from **travis**. lets see if the build is allowed to run this time

image

wasifhossain:
no change in appveyor since last green run, though its now failing for some reason

.travis.yml
```diff
@@ -97,26 +100,41 @@ jobs:
...
     - { rvm: *ruby_2_2,  env: *rails_6_0 }
     - { rvm: *ruby_2_3,  env: *rails_6_0 }
+    - { rvm: *ruby_2_4,  env: *rails_6_0 }
 
     - { rvm: *ruby_2_1,  env: *rails_master }
     - { rvm: *ruby_2_2,  env: *rails_master }
     - { rvm: *ruby_2_3,  env: *rails_master }
+    - { rvm: *ruby_2_4,  env: *rails_master }
 
   allow_failures:
```
bf4: I almost never look at the allowed failures. unless you are, there's really no reason to put the load on the CI

.travis.yml
```diff
@@ -65,17 +66,19 @@ rvm:
   - *ruby_2_1
   - *ruby_2_2
   - *ruby_2_3
+  - *ruby_2_4
+  - *ruby_2_5
+  - *ruby_2_6
```
bf4: it's fine that you want this, but  we're already testing 2.7, so the pr description is out of sync with the intent

.travis.yml
wasifhossain: I agree with you very much. The fact is, the build matrix, how its configured, runs tests for each Ruby version listed under `rvm` against each of the Rails versions under `jobs`
```yml
rvm:
  - *ruby_2_1
  - *ruby_2_2
  - *ruby_2_3
  - *ruby_2_4
  - *ruby_2_5
  - *ruby_2_6
  - *ruby_2_7
  - *ruby_head

jobs:
  - *rails_4_1
  - *rails_4_2
  - *rails_5_0
  - *rails_5_1
  - *rails_5_2
  - *rails_6_0
  - *rails_master
```

In order **to SKIP the CI**, we list the envs (rvm/jobs) under `exclude` section:
```diff
exclude:
  - { rvm: *ruby_2_4,  env: *rails_4_1 }
  - { rvm: *ruby_2_5,  env: *rails_4_1 }
  - { rvm: *ruby_2_6,  env: *rails_4_1 }
  - { rvm: *ruby_2_7,  env: *rails_4_1 }
  - { rvm: *ruby_head, env: *rails_4_1 }

  - { rvm: *ruby_2_7,  env: *rails_4_2 }
  - { rvm: *ruby_head, env: *rails_4_2 }

  - { rvm: *ruby_2_1,  env: *rails_5_0 }
  - { rvm: *ruby_2_1,  env: *rails_5_1 }
  - { rvm: *ruby_2_1,  env: *rails_5_2 }

  - { rvm: *ruby_2_1,  env: *rails_6_0 }
  - { rvm: *ruby_2_2,  env: *rails_6_0 }
  - { rvm: *ruby_2_3,  env: *rails_6_0 }
  - { rvm: *ruby_2_4,  env: *rails_6_0 }

  - { rvm: *ruby_2_1,  env: *rails_master }
  - { rvm: *ruby_2_2,  env: *rails_master }
  - { rvm: *ruby_2_3,  env: *rails_master }
+ - { rvm: *ruby_2_4,  env: *rails_master }
```

Moreover, the envs which are not run under default configuration (e.g. **jRuby**) have been listed under `include` section.

And the envs which are from (ruby/jruby)-head or rails-master, have been listed under `allow_failures`, so that we can see any upcoming errors in advance due to new changes on those envs

.travis.yml
wasifhossain: these lines have already been there for long. here is the last commit prior to our attempt to support ruby 2.7

https://github.com/rails-api/active_model_serializers/blob/6b093c965f75f87d5bbf0510e64b5193f4c6f157/.travis.yml#L65..L67

which got removed from #2368 by mistake

wasifhossain: I have updated the OP intent. Please let me know if you think we should move some of these changes out of this

.travis.yml
wasifhossain: here is the last successful build on Travis, where you can find all the envs that are using the CI.

https://travis-ci.org/rails-api/active_model_serializers/builds/604587180

let me know if we need to exclude any of the envs from the list

bf4: yeah, that's why I removed some from the rvm list and selectively added them to the include list

.travis.yml
bf4: I actually removed those lines on purpose. Maybe my execution failed, inasmuch as I intend to just test a few of them, in order to reduce the matrix size to what is likely to catch a problem

wasifhossain: could you please suggest the build matrix that you have in mind? we can try to configure it accordingly

wasifhossain:
good news is, travis is working again :tada: 

bad news is, not sure why simplecov is failing with
```
RuntimeError cannot change the measuring target during coverage measurement
```

wasifhossain:
update on simplecov:

0.18.x only allows ruby >= 2.4. 

latest version that supports ruby >= 2.1 is 0.17.1. it looks like putting this constraint on the gemspec should overcome the error
```
spec.add_development_dependency 'simplecov', '~> 0.17.1'
```

wasifhossain:
simplecov issue resolved :tada:

wasifhossain:
next issue with **grape**; looks slightly complex

the current bug is caused in grape tests due to **rack >= 2.1.0**. more info here: https://github.com/ruby-grape/grape/issues/1980

**grape 1.3.0** fixes the bug, though with the cost of requiring **ruby >= 2.4** :confused:

wasifhossain:
we can use a band-aid solution to have grape tests passed, by restricting **rack to <= 2.0.8**

wasifhossain:
Travis is looking green (partially) after quite some time. CRuby tests are all passed. JRuby 9.2.x are also passing, though 9.1.x failed due to already existing bundler > 2.

wasifhossain:
Removing bundler >= 2 didn't help either; it still remains (most probably due to rubygems)

image

bf4:
travis ci is dead

PR #2279, opened 2018-09-01
Support conditions in link statements

mkon:
Hello,

First time contributor here, however have been using AMS happily for quite some time.

#### Purpose

Currently hiding certain links is a bit cumbersome, since the serializer's instance options are not available when the link block is evaluated by the (JSON-API) adapter.

With this change the `link()` statement in serializers supports the same condition options as `attribute()` and is possible to hide certain links from being rendered.

Eg I can now write a serializer like this:

```ruby
link(:private, if: -> { internal? }) do
  object.private_link
end
```

#### Changes

- Added `ActiveModel::Serializer::Link` as a subclass of `Field` to store options and handle condition
- JSON-API apater's `links_for` detects `ActiveModel::Serializer::Link`, excludes them if necessary and converts them to the adapter's `Link` instances.

PR #2389, opened 2020-08-05
Value is not always a ActiveModel::Serializer::Link

christopherchild:
#### Purpose
This PR will fix an issue I was having ``NoMethodError: undefined method `excluded?' for "/nodes/746868371":String`` in ``active_model_serializers-0.10.10/lib/active_model_serializers/adapter/json_api.rb:487:in `block in links_for'``

The PR link below does have a comment that states 'value' is now always a Link, but this doesn't seem to be the case as the error above shows. 

I am new to the project, so this might not be an appropriate fix for the problem I am having.

#### Changes
Revert commit 572f11b7e0592eb0781006b9c364d8bd13474831.

#### Related
#2279

### DIFF
--- a/lib/active_model_serializers/adapter/json_api.rb
+++ b/lib/active_model_serializers/adapter/json_api.rb
@@ -484,8 +484,11 @@ module ActiveModelSerializers
       #   }.reject! {|_,v| v.nil? }
       def links_for(serializer)
         serializer._links.each_with_object({}) do |(name, value), hash|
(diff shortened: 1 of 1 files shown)


christopherchild:
This can be ignored.

The issue was with my code - I know!

I had this in my Serializer; presumably from an older version:

```
def _links
    { self: polymorphic_path(object) }
  end
```

this needed to be changed to:

```
link(:self) { polymorphic_path(object) }
```

Issue #2031, opened 2017-01-16
Changing cache key not working


chrismanderson:
#### Expected behavior vs actual behavior

Changing a serializer-level cache key with the `key option` does not change the cache key.
  
#### Steps to reproduce

With a basic controller

    class ZonesController < ApplicationController
      def index
        render json: Zone.all
      end
    end

and basic serializer

    class ZoneSerializer < ActiveModel::Serializer
      cache key: 'area', expires_in: 3.hours

      attributes :type, :key

      def type
        'Feature'
      end

      def cache_key
        @cache_key # for debug
      end
    end

I would expect the cache key to resemble something like:

    area/24-20160217165512052884/attributes/63c314f796ba5ea3898d4a639a9448f4

When instead the key is: 

    zone/24-20160217165512052884/attributes/63c314f796ba5ea3898d4a639a9448f4

I've verified that the expected key is not present in the cache (memcache in my case). The incorrect key that is being set instead is present in the cache and does contain the cached value, so the cache itself is working.

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*: 0.10.4

Output of `ruby -e "puts RUBY_DESCRIPTION"`: `ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]`

OS Type & Version: 10.12.2

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails 5.0.1, api mode

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*

#### Additonal helpful information

Occurs with both `memcache/dalli` and Rail's `memory_store`.

development.rb cache section: 

    # Enable/disable caching. By default caching is disabled.
    if Rails.root.join('tmp/caching-dev.txt').exist?
      config.action_controller.perform_caching = true

      config.cache_store = :mem_cache_store, "localhost"
      config.public_file_server.headers = {
        'Cache-Control' => 'public, max-age=172800'
      }
    else
      config.action_controller.perform_caching = false

      config.cache_store = :null_store
    end

EarlyZhao:
> cache key: 'some key' 
the key is not working.

I also met this problem, it cause me some problems,when i generate different format  json data for the same model in method Index and Show,  it will confuse the two formats, as the cache key set by me not working, the cache data of two json format was treated as one, store in redis. I can't get two json formats for one object.
 
I checked the sourse code of gem, the key set in the model ***Serializer was not used first.

[...]

PR #2390, opened 2020-08-25
prioritize serializer's cache key over ActiveRecord's cache key

ggmichaelgo:
#### Purpose
Allow serializers to override cache key by adding
```
class CompactUserSerializer
  cache key: 'compact-user'
end

class UserSerializer
  cache key: 'user'
end
```

Our Rails application has different version of serializers for a model like `CompactUserSerializer` and `UserSerializer`.
Since our `User` model has a cache key, active model serializer does not use our custom cache key names...

#### Changes

Prioritize Serializer's cache key attribute over Active Model's `cache_key` method.

### DIFF
--- a/lib/active_model/serializer/concerns/caching.rb
+++ b/lib/active_model/serializer/concerns/caching.rb
@@ -286,12 +286,12 @@ module ActiveModel
       def object_cache_key
         if object.respond_to?(:cache_key_with_version)
           object.cache_key_with_version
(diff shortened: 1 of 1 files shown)


ggmichaelgo:
Found a better solution to override the cache key name 🤦‍♂️

```
class CompactUserSerializer
  def object_cache_key
    "#{object.cache_key}/compact_user"
  end
```

https://github.com/rails-api/active_model_serializers/issues/2031

PR #2404, opened 2020-12-30
Add ability to disable using foreign key on belongs_to

pdud:
#### Purpose

The [jsonapi_use_foreign_key_on_belongs_to_relationship](https://github.com/rails-api/active_model_serializers/blob/v0.10.10/docs/general/configuration_options.md#jsonapi_use_foreign_key_on_belongs_to_relationship) option is very useful. In some rare scenarios we want to opt out of it.

In our scenario, we sometimes opt to use a token/ code rather than an id.

#### Changes

This enables users to opt out of using the foreign_key global setting at an association level.

#### Caveats
_None_

#### Related GitHub issues
_None_

* One current work around is to set `foreign_key` on the `belongs_to` to something that does not exist.
* I can update the docs to match


bf4:
sure, looks fine

Issue #2406, opened 2021-01-23
Raise ArgumentError  "Cannot infer root key from collection type. Please specify the root or each_serializer option, or render a JSON String" when collection is empty array in >= 10.0.7


tashirosota:
#### Expected behavior vs actual behavior
 - 【Expected behavior】※ <= version `10.0.6`
```
::ActiveModelSerializers::SerializableResource.new(
  collections, # [] empty array
  each_serializer: CollectionSerializer,
).serializable_hash
=> { collection_types: nil }

```
 - 【Actual behavior】※ >= version `10.0.7`
```
::ActiveModelSerializers::SerializableResource.new(
  collections, # [] empty array
  each_serializer: CollectionSerializer,
).serializable_hash
=> Raise ArgumentError  "Cannot infer root key from collection type. Please specify the root or each_serializer option, or render a JSON String"

```
Is this breaking change or bug?

#### Steps to reproduce
*(e.g., detailed walkthrough, runnable script, example application)*
```
# >= version 10.0.7
::ActiveModelSerializers::SerializableResource.new(
  collections, # [] empty array
  each_serializer: CollectionSerializer,
).serializable_hash
```

#### Environment

ActiveModelSerializers Version *(commit ref if not on tag)*:  `>= 10.0.7`

Output of `ruby -e "puts RUBY_DESCRIPTION"`: ruby `2.6.6p146` && `2.5.1`

OS Type & Version: alpine3.12 on docker

Integrated application and version *(e.g., Rails, Grape, etc)*: Rails `5.2.4.4` & `5.1.6`

#### Backtrace
*(e.g., provide any applicable backtraces from your application)*
 - 10.0.7 PR: https://github.com/rails-api/active_model_serializers/pull/2214

#### Additonal helpful information
*(e.g., Gemfile.lock, configurations, PR containing a failing test, git bisect results)*

In PR https://github.com/rails-api/active_model_serializers/pull/2214, @mvz  said「I'm getting this error on a line that does have :each_serializer specified 😞」. Me too.

PR #2412, opened 2021-02-11
Add raise_cannot_infer_root_key_error to config

tashirosota:
#### Purpose
 - Raise error when cannot infer root key by v10.0.7 . I think this is braking change. 
 - Added config that you can upgrade without changing existing code.

#### Changes
 - ArgumentError changes to CannotInferRootKeyError in CollectionSerializer#json_key
    - Custom error helps clarify the problem
 - Add config.raise_cannot_infer_root_key_error
 - Add test when serializer does not have type and collection is empty

#### Related GitHub issues
 - https://github.com/rails-api/active_model_serializers/issues/2406

#### 
Our project has many like this code😢

```
::ActiveModelSerializers::SerializableResource.new(
  collections, # [] empty array
  each_serializer: NonTypeCollectionSerializer,
).serializable_hash || []
※ v0.10.6
```

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -146,6 +146,8 @@ module ActiveModel
     config.jsonapi_include_toplevel_object = false
     config.jsonapi_use_foreign_key_on_belongs_to_relationship = false
     config.include_data_default = true
+    # Raise ActiveModel::Serializer::CollectionSerializer::CannotInferRootKeyError when cannot infer root key from collection type
+    config.raise_cannot_infer_root_key_error = true
 
     # For configuring how serializers are found.
     # This should be an array of procs.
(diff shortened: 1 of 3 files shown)


wasifhossain:
Looks good to me

tashirosota:
@wasifhossain 
Will this PR be merged?

wasifhossain:
let's do this

wasifhossain:
thanks for the reminder!

mvz:
This doesn't seem to resolve the problem that this error is *also* raised when `each_serializer` is specified, even though specifying an `each_serializer` is suggested as a way to resolve the error.

tashirosota:
@mvz 
When each_serializer is specified and don't want to make an error.
Can set `ActiveModelSerializers.config.raise_cannot_infer_root_key_error = true` and not raise error.
This error coused when each_serializer is specified and (serializer's type is null || collection is empty || root key is null), it is correct behavior, but braking change.
So I patched this option, resolve to this case < v0.10.6

```
::ActiveModelSerializers::SerializableResource.new(
  collections, # [] empty array
  each_serializer: NonTypeCollectionSerializer,
).serializable_hash || []
```

PR #2395, opened 2020-10-12
remove explicit require for thread_safe

ritikesh:
In Rails master, [tzinfo has been updated to 2.x](https://github.com/rails/rails/commit/e9425abe33924623b1dce62bd817eace757c2b4e), which no longer depends on the `thread_safe` gem and has moved to `concurrent-ruby` instead. Thereby, unless you're explicitly adding `thread_safe` to your Gemfile, `active_model_serializers` throws up a `cannot require thread_safe`. #2369 has already replaced `ThreadSafe::Cache` with `Concurrent::Map`. This require should no longer be required!(😅)

PR #1736, opened 2016-05-18
Correct ruby/rails version requirements in gemspec

bf4:
cc @remear wanna release as 0.10.0.1?

PR #2428, opened 2021-12-16
Relax gem requirement to allow Rails 7

jpawlyn:
#### Purpose

Follow on from https://github.com/rails-api/active_model_serializers/pull/2423 to enable AMS to be used with Rails 7

#### Changes

Updates `rails_versions` in Gemspec to `< 7.1`


delphaber:
Since Ruby 3.0 is "preferred" in Rails 7, should Ruby 3.0 be added to travis too? 🤔

wasifhossain:
I think AMS is currently using Github CI instead of Travis. could you please make the changes on https://github.com/rails-api/active_model_serializers/blob/0-10-stable/.github/workflows/ci.yml.

Thanks!

active_model_serializers.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.1'
 
-  rails_versions = ['>= 4.1', '< 7.0']
+  rails_versions = ['>= 4.1', '< 8.0']
```
wasifhossain, line 24: as the rails version on `main` branch is now having `7.1.0`, can we use `< 7.1` instead of `< 8.0`. thanks

jpawlyn: Good idea

active_model_serializers.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.1'
 
-  rails_versions = ['>= 4.1', '< 7.0']
+  rails_versions = ['>= 4.1', '< 7.1']
```
bf4, line 24: might as well remove the max version.  only real benefit is it gets people to ask us to add a new version to CI

jpawlyn: @wasifhossain had a different [idea](https://github.com/rails-api/active_model_serializers/pull/2428#discussion_r770411827) but happy to go with the consensus.

NickLaMuro:
Thought I should throw a couple of relevant links in here since this took me a bit too stumble across this PR and almost opened a duplicate PR:

A big reason most will need this fix pretty quickly is because without it, you will get forced into using an ancient version predating this change:

https://github.com/rails-api/active_model_serializers/pull/1736

This means that the errors mentioned in and fixed by https://github.com/rails-api/active_model_serializers/pull/2395 are not in the gem version that is bundled, leading to a very confusing upgrade experience.

For now, using the `jpawlyn:0-10-stable` branch for this should be enough for proof of concept purposes, but is probably not best for prime time.

* * *

Thanks for all the hard work and will be looking forward to this getting merged.  🙇

test/support/rails_7_patch.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# HACK: to prevent the resetting of instance variables after each request in Rails 7
+# see https://github.com/rails/rails/pull/43735
+if Rails::VERSION::MAJOR >= 7
+  module ActionController
+    module Testing
+      module Functional
+        def clear_instance_variables_between_requests; end
```
wasifhossain, line 9: my only concern is, whether we should go forward and adopt the new rails practice (resetting the instance variables) by rewriting the affected tests in this repo, or should we pull things behind by using this hack

jpawlyn: One other option is to target the hack at the one controller that causes the issue `ImplicitSerializationTestController` in order to limit the likelihood of instance variables being re-used across requests in future.

ie
```ruby
# frozen_string_literal: true

require 'test_helper'

module ActionController
  module Serialization
    class ImplicitSerializerTest < ActionController::TestCase
      class ImplicitSerializationTestController < ActionController::Base
        include SerializationTesting
        
        ...

        # HACK: to prevent the resetting of instance variables after each request in Rails 7
        # see https://github.com/rails/rails/pull/43735
        def clear_instance_variables_between_requests; end

        def update_and_render_object_with_cache_enabled
          @post.updated_at = Time.zone.now # requires hack above to prevent `NoMethodError: undefined method `updated_at=' for nil:NilClass`

          generate_cached_serializer(@post)
          render json: @post
        end

        ...
```

wasifhossain: I think that would be a more optimized approach. this way we can apply the hack on the affected controllers exclusively, at the same time influence adopting the new practice before writing new tests.

Also in future iteration, we may proactively work on getting rid of this hack from these controllers by rewriting the tests.

wasifhossain:
Hi @jpawlyn the PR looks good to me! thanks for your efforts :+1: 

as ruby 3.1.0 has just been out, would you mind adding this in github CI as well :slightly_smiling_face: thanks!

jpawlyn:
> Hi @jpawlyn the PR looks good to me! thanks for your efforts 👍
> 
> as ruby 3.1.0 has just been out, would you mind adding this in github CI as well 🙂 thanks!

Sure, easily done.

jpawlyn:
> > Hi @jpawlyn the PR looks good to me! thanks for your efforts 👍
> > as ruby 3.1.0 has just been out, would you mind adding this in github CI as well 🙂 thanks!

The CI error is due to an incompatibility I think between Rails 7.0 and ruby 3.1 (see https://github.com/rails/rails/issues/43998). It has been fixed on main `https://github.com/rails/rails/pull/43951` so once Rails 7.0.1 is released it should be fine.

bf4:
I figure this is fine to merge as it's going to keep failing on Rails 7.0 and Ruby 3.1 due to incompatibilities outside of AMS scope?

jpawlyn:
> I figure this is fine to merge as it's going to keep failing on Rails 7.0 and Ruby 3.1 due to incompatibilities outside of AMS scope?

I reckon so 👍🏼

wasifhossain:
Great! let's merge it.

bopm:
@bf4 I am going to tag you here, as this seems to be stalled for a week.

bf4:
I just released 0.10.13. I didn't check to make sure the change log is up to date. Help appreciated.

Issue #2443, opened 2023-03-27
 ActiveModel::Serializer.serializer_for unable to search for serializer with base model after Ruby upgrade from 2.5 to 2.7


Physium:
Current I have a model that references a base model  and a serializer that ties to the base model. Before upgrading from ruby 2.5 to 2.7,  ActiveModel::Serializer.serializer_for is able to return the base serializer with no issues. After upgrading, it doesnt seems to detect the serializer base on the base model. I cant seem to find anything related in the changelogs as well. 

Currently I'm on rails 5.2-stable with active_model_serializers 0.9.4.

Example:
```
Class Dog > Animal

Class AnimalSerializer < BaseSerializers

Ruby 2.5
ActiveModel::Serializer.serializer_for(Dog)
=> AnimalSerializer

Ruby 2.7
ActiveModel::Serializer.serializer_for(Dog)
=> nil
```

bf4:
If you can help write a failing test and are will to try and help by contributing a PR, I can facilitate.

Physium:
> If you can help write a failing test and are will to try and help by contributing a PR, I can facilitate.

I'm open to that. But before I dive deep, can I do a quick clarification if v0.9.x supports ruby 2.7? Also I understand that theres a difference between v0.9 and v0.10 and it seems like its recommended to upgrade from v0.8 to v0.10 but not v0.9. Are they both serving a different use-case?

bf4:
It should support it, but it's been years since it was actively developed.
0.9 was a rewrite of 0.8 and 0.10 was supposed to start out based on 0.8
but didn't have a good upgrade path.   0.9 is probably fine if it works for
you and you're already using it.   I got involved during the 0.10 rewrite.
 I'm just here now to facilitate.

Physium:
Apologies for the late response.

I did an initial round of digging and found out that main root cause was due to the usage of `_const_get` in v0.9.x [here](https://github.com/rails-api/active_model_serializers/blob/ff1a36ffbc82c6070905707b29978a7b897dea81/lib/active_model/serializer.rb#L71) 

Going from ruby 2.5 to 2.6 onwards will result in the example mentioned below to not work. 

```
Class Dog::Animal > Animal; end

Class AnimalSerializer < BaseSerializers; end

Ruby 2.5
# const_get is able to ignore Dog namespace and obtain the constant AnimalSerializer
ActiveModel::Serializer.serializer_for(Dog)
=> AnimalSerializer

[...]

PR #2446, opened 2023-05-07
Update 0.9-stable release to work with ruby 2.6 & 2.7

Physium:
#### Purpose
0.9-stable version as of now does not work with ruby 2.6 and above ( excluding ruby 3 ) due to how const_get has become stricter in terms searching for constants.

Examples mentioned in this issue: #2443 

#### Changes
- Update Gemfile to handle higher rails version if ruby is > 2.5
- Patch serializer_for method to search serializers based on a list of potential naming conventions

Note: this is my first rare attempt at making a PR, pardon my newbie-ness and appreciate any feedback.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
-version = ENV["RAILS_VERSION"] || "4.2"
+default_rails_version = RUBY_VERSION > '2.5' ? '5.2' : '4.2'
+version = ENV['RAILS_VERSION'] || default_rails_version
 
 if version == 'master'
   gem 'rack', github: 'rack/rack'
@@ -27,6 +28,7 @@ else
   gem 'activemodel', gem_version
   gem 'actionpack', gem_version
   gem 'activerecord', gem_version, group: :test
+  gem 'rails-controller-testing' if version > '4.2'
 end
 
 if RUBY_VERSION < '2'
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -66,10 +66,14 @@ end
             ArraySerializer
           end
         else
-          klass_name = build_serializer_class(resource, options)
-          Serializer.serializers_cache.fetch_or_store(klass_name) do
-            _const_get(klass_name)
-          end
+          search_list = build_serializer_class_list(resource, options)
+          result = search_list.map do |klass_name|
+                     Serializer.serializers_cache.fetch_or_store(klass_name) do
+                       _const_get(klass_name)
+                     end
+                   end
+
+          result.find { |serializer| !serializer.nil? }
         end
       end
 
@@ -116,6 +120,12 @@ end
         attr = attr.to_s.gsub(/\?\Z/, '')
         attr = attr.to_sym if symbolized
         attr
+      end
+
+      def build_serializer_class_list(resource, options)
+        list = []
+        list << build_serializer_class(resource, options)
+        list << build_serializer_class(resource.class.name.demodulize, {})
       end
 
       def build_serializer_class(resource, options)
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -122,7 +122,11 @@ end
         "".tap do |klass_name|
           klass_name << "#{options[:namespace]}::" if options[:namespace]
           klass_name << options[:prefix].to_s.classify if options[:prefix]
-          klass_name << "#{resource.class.name}Serializer"
+          if resource.is_a?(String)
+            klass_name << "#{resource}Serializer"
+          else
+            klass_name << "#{resource.class.name}Serializer"
+          end
         end
       end
 
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -125,6 +125,7 @@ end
       def build_serializer_class_list(resource, options)
         list = []
         list << build_serializer_class(resource, options)
+        list << build_serializer_class(resource, {})
         list << build_serializer_class(resource.class.name.demodulize, {})
       end
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -2,6 +2,8 @@ require 'bundler/setup'
 require 'minitest/autorun'
 require 'active_model_serializers'
 require 'fixtures/poro'
+require 'rails-controller-testing'
+Rails::Controller::Testing.install
 
 # Ensure backward compatibility with Minitest 4
 Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)
--- a/test/unit/active_model/array_serializer/serialization_test.rb
+++ b/test/unit/active_model/array_serializer/serialization_test.rb
@@ -49,6 +49,7 @@ module ActiveModel
         def object.serializer_class; CustomSerializer; end
 
         assert_equal CustomSerializer, Serializer.serializer_for(object)
+        assert_equal CustomSerializer, Serializer.serializer_for('Custom')
       end
     end
 
--- a/test/unit/active_model/array_serializer/serialization_test.rb
+++ b/test/unit/active_model/array_serializer/serialization_test.rb
@@ -53,6 +53,28 @@ module ActiveModel
       end
     end
 
+    class ModelSerializationNamespaceTest < Minitest::Test
+      def test_namespace
+        test1 = TestNamespace2::Test.new(name: 'Test 1', email: 'test1@test.com', gender: 'M')
+        test2 = TestNamespace2::Test.new(name: 'Test 2', email: 'test2@test.com', gender: 'M')
+        sub_test1 = TestNamespace2::SubTest.new(name: 'Name 1', description: 'Description 1', comments: 'Comments 1')
+        sub_test2 = TestNamespace2::SubTest.new(name: 'Name 2', description: 'Description 2', comments: 'Comments 2')
+        test1.sub_test = sub_test1
+        test2.sub_test = sub_test2
+
+        array = [test1, test2]
+        serializer = ArraySerializer.new(array)
+
+        expected = [
+          { name: 'Test 1', email: 'test1@test.com', sub_test: { name: 'Name 1', description: 'Description 1' }},
+          { name: 'Test 1', email: 'test2@test.com', sub_test: { name: 'Name 2', description: 'Description 2' }}
+        ]
+
+        assert_equal expected, serializer.serializable_array
+        assert_equal expected, serializer.as_json
+      end
+    end
+
     class ModelSerializationTest < Minitest::Test
       def test_array_serializer_serializes_models
         array = [Profile.new({ name: 'Name 1', description: 'Description 1', comments: 'Comments 1' }),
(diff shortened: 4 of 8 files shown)


test/unit/active_model/serializer/has_many_polymorphic_test.rb
```diff
@@ -2,7 +2,7 @@
 
 module ActiveModel
   class Serializer
-    class HasManyPolymorphicTest < ActiveModel::TestCase
+    class HasManyPolymorphicTest < Minitest::Test
```
bf4, line 5: what's the reason for this change in this PR? Is it required?

test/unit/active_model/array_serializer/serialization_test.rb
```diff
@@ -49,6 +49,7 @@ def test_serializer_for_array_returns_appropriate_type
         def object.serializer_class; CustomSerializer; end
 
         assert_equal CustomSerializer, Serializer.serializer_for(object)
+        assert_equal CustomSerializer, Serializer.serializer_for('Custom')
```
bf4, line 52: this is the failing test, right?

is accepting a String a new behavior?

Gemfile
```diff
@@ -27,6 +28,7 @@ else
   gem 'activemodel', gem_version
   gem 'actionpack', gem_version
   gem 'activerecord', gem_version, group: :test
+  gem 'rails-controller-testing' if version > '4.2'
```
bf4, line 31: do you mean >= 4.2?

lib/active_model/serializer.rb
```diff
@@ -118,11 +122,21 @@ def strip_attribute(attr)
         attr
       end
 
+      def build_serializer_class_list(resource, options)
+        list = []
+        list << build_serializer_class(resource, options)
+        list << build_serializer_class(resource.class.name.demodulize, {})
```
bf4, line 128: lookinng at 0-10-stable lib/active_model_serializers/lookup_chain.rb for how we solved this when it came up there looks like the same solution  more or less, so that's good :)

```ruby
    module_function

    def namespace_for(klass)
      klass.name.deconstantize
    end

    def resource_class_name(klass)
      klass.name.demodulize
    end

    def serializer_from_resource_name(name)
      "#{name}Serializer"
    end

    def serializer_from(klass)
      name = resource_class_name(klass)
      serializer_from_resource_name(name)
    end
```

and that in the serializer we called `serializer_class = lookup_chain.map(&:safe_constantize).find { |x| x && x < ActiveModel::Serializer }`

test/unit/active_model/array_serializer/serialization_test.rb
Physium: This test is fine. I added an extra test since i want to allow string instead of a whole class object. This somewhat relaxes the input for serializer_for. Not sure if its a good idea, but its one simple fix i can think off.

Upon upgrading to ruby 2.5 >, there are 17 tests that fails due to the const_get changes. I used thou failed test as a benchmark to validate if the fix work. Is that ok? Is there a need to include new test?

lib/active_model/serializer.rb
Physium: I got some inspiration from v0.10 for sure, thou i am still a little unclear on how the find condition works out. How does `x < ActiveModel::Serializer` work? 

Also, is there a need for me to make any changes to replicate how things are done in v0.10?

Physium:
@bf4 I added in a missing test case that I came across in my project. Found out that if you have namespace on your model and its association they do not really play well.

bf4:
Remind me if I don't look at this soon

Physium:
@bf4 apologies for bothering. is there a rough timeline for the review?

lib/active_model/serializer.rb
bf4: `x < ActiveModel::Serializer` is just one of those weird-looking Ruby things for asking if a thing is a subclass of the thing on the right

bf4: I'm guessing the 0.10 solution would work here as well. Up to you if you want to make another PR after this is merged

test/test_helper.rb
```diff
@@ -2,6 +2,8 @@
 require 'minitest/autorun'
 require 'active_model_serializers'
 require 'fixtures/poro'
+require 'rails-controller-testing'
```
bf4, line 5: this can always be required now?

test/unit/active_model/serializer/has_many_polymorphic_test.rb
bf4: I guess I misread this as removing ActiveSupport::TestCase (and not ActiveModel::TestCase)

I recall in 0.10x we went with ActiveSupport::TestCase over Minitest::Test since it provided better compatibility across minitest versions

Gemfile
Physium: not exactly sure and I cant seem to figure out where is this test metric at. Based on the gem file, rails-controller-testing is only added when version is set to 'master'.  

I dont think 0.9-stable has ever been configured to work with the offical release of rails 5 and above.

test/unit/active_model/serializer/has_many_polymorphic_test.rb
Physium: Personally have never worked with minitest and i notice most of these tests were written back in 2013/14. I made the switch simply by referencing what other model test files were referencing. I thought it made more sense for them to me the same across the board.

test/test_helper.rb
Physium: yes, previously this was automatically required pre rails 5. now its need to be required when needed for >= rails 5.

bf4: gotcha.  well, If these changes for you, I'm ok to merge them.

bf4: seems travis ci is dead and this branch wasn't running a CI. I ported some stuff over in https://github.com/rails-api/active_model_serializers/pull/2447 if you're interested in helping out (happy to let you take it over) and figure out why it's failing and if we care

Physium: I wouldn't mind taking a look when I have time. Would definitely love to rewrite the lookup to something similar to v0.10 as well. However it does feel like i'm the only one in the world right now still on v0.9 haha!

Just curious, will there still be a release once this gets merged? Or do i have to reference the stable branch for all latest patches?

Physium: appreciate your prompt response! looks good to me as well! I'll help to look into CI when possible.

bf4:
released 0.9.9

zavan:
FYI this 'patch' update wrecked havoc on my Rails v7.0.5/Ruby v3.2.2 app. A lot of namespaced serializer resolving problems.

Physium:
> FYI this 'patch' update wrecked havoc on my Rails v7.0.5/Ruby v3.2.2 app. A lot of namespaced serializer resolving problems.

Are you able to share examples of the problem? Perhaps open an issue and i can take a look. I am surprise that v9 even works with rails 7 and ruby 3.

bf4:
@zavan we can help with a detailed error report. Obviously nothing was supposed to break. If you're up to help with testing and or proposing a fix, that would be much appreciated.  I haven't used this gem myself in many years, so don't have a personal use case,  but am committed to helping maintain it.

Physium:
@zavan also did 0.9.8 work with ur existing setup?

zavan:
v0.9.8 does work, so I just downgraded to fix.

It's an old app that started on Rails v3 but I've been able to update everything except for a large API that uses AS.

I don't have time right now to go in depth and write some tests to replicate, but here's one example. Consider these serializers:

```ruby
class BusinessSerializer < ActiveModel::Serializer
  # This is the serializer for the `Business` model, not `Integration::MyIntegration::Business`.
  # The model has a `name` attribute.
  attributes :id, :name
end

class LocationSerializer < ActiveModel::Serializer
  attributes :id, :name

  # The model is an instance of `Integration::MyIntegration::Business`,
  # which doesn't have a serializer and does not have a `name` attribute.
  has_one :my_integration_business 
end
```

v0.9.8:
```ruby
irb(main):005:0> LocationSerializer.new(Location.first).as_json
=>
{
  "location" => {
    :id => 1,
    :name => "West Coast",
    :my_integration_business => {
      "id" => 2,
      "location_id" => 1,
      "business_id" => "465607d5-9cb1-4ffd-a796-b18e11c3decd", # This has nothing to do with the Business model. It's an external id.
      "created_at" => "2023-06-02T08:01:56.085-05:00",
      "updated_at" => "2023-06-02T09:03:52.946-05:00"
    },
  }
}
```

So as you can see, it didn't find a serializer for the `Integration::MyIntegration::Business` instance and instead just dumped all model attributes. The application relies on this behavior.

v0.9.9:
```ruby
irb(main):001:0> LocationSerializer.new(Location.first).as_json

=> /usr/local/lib/ruby/gems/3.2.0/gems/activemodel-7.0.5/lib/active_model/attribute_methods.rb:450:in `method_missing': undefined method `name' for #<Integration::MyIntegration::Business id: 2, location_id: 1, business_id: "465607d5-9cb1-4ffd-a796-b18e11c3decd", created_at: "2023-06-02 08:01:56.085837000 -0500", updated_at: "2023-06-02 09:03:52.946017000 -0500"> (NoMethodError)
```

AS tries to use the `BusinessSerializer` for the `Integration::MyIntegration::Business` model instance. Which breaks the application as the attributes don't match.

Physium:
Thank you @zavan! I'll look into it when I have time. The change should not affect existing way of doing things, all I did was to add additional look up chains if the initial lookup did not match. This is puzzling.

Physium:
I get the issue now. This is all boils down to how const_get became stricter in ruby 2.6 onwards and as a result the birth of this PR.

In 2.5 and below, the example you gave would have also return BusinessSerializer instead of nil. Hence, the purpose of this PR is to somewhat mimic that behaviour. It seems like you guys adhere to the stricter behaviour and thus this gem works with no issues even with ruby 3 (which is great of course).

Now, I personally have no idea what should be the ideal behaviour should be. Its only recently I've been tasked to work on a project that uses this gem and upgrading its ruby version has resulted in a ton of failures. 

I am fine with removing my release and leaving my "patch" within my personal fork. Eventually, given the status of this gem, is to also migrate out of AMS into something more actively maintained.

Thoughts @bf4 ?

bf4:
Maybe split the difference with a feature flag option for old behavior vs new?

Kartikvlog:
when i am upgrading rails 5 to rails 6.1then getting error ``` No serializer found for resource: #<Comments  _id``` we have association:  user has_many comments and comments belongs to user.

Physium:
> when i am upgrading rails 5 to rails 6.1then getting error ` No serializer found for resource: #<Comments _id` we have association: user has_many comments and comments belongs to user.

@Kartikvlog are you able to share more details? What version of the serializer are you on before you upgraded? If anything can you revert the serializer to 0.9.8 to see if it works? Upgrading rails shouldn’t cause any issue at all.

lib/active_model/serializer.rb
```diff
@@ -66,10 +66,14 @@ def serializer_for(resource, options = {})
             ArraySerializer
           end
         else
-          klass_name = build_serializer_class(resource, options)
-          Serializer.serializers_cache.fetch_or_store(klass_name) do
-            _const_get(klass_name)
-          end
+          search_list = build_serializer_class_list(resource, options)
```
bf4, line 69: Upon review,, I should have noticed these changes are out of scope for this pr and asked for them to be proposed separately.

PR #2370, opened 2020-01-06
Prevent creation of unnecessary fieldset

abcang:
#### Purpose
When I profiled a simple sample with stackprof, I found that `ActiveSupport::Inflector#apply_inflections` was taking a long time.

```rb
class Blog < ActiveModelSerializers::Model
  attributes :id, :name, :authors
end
class Author < ActiveModelSerializers::Model
  attributes :id, :name
end
class AuthorSerializer < ActiveModel::Serializer
  attributes :id, :name
end
class BlogSerializer < ActiveModel::Serializer
  attributes :id
  attribute :name, key: :title
  
  has_many :authors, serializer: ::AuthorSerializer
end

ActiveSupport::Notifications.unsubscribe(ActiveModelSerializers::Logging::RENDER_EVENT)
authors = [Author.new(id: 1, name: 'Blog Author')]
blog = Blog.new(id: 2, name: 'The Blog', authors: authors)

StackProf.run(mode: :cpu, out: Rails.root.join('tmp/stackprof.dump').to_s, raw: true) do
  10000.times do
    ActiveModelSerializers::SerializableResource.new(blog, serializer: BlogSerializer, adapter: :attributes).as_json
  end
end
```

```
$ bundle exec stackprof tmp/stackprof.dump --text
==================================
  Mode: cpu(1000)
  Samples: 1193 (0.00% miss rate)
  GC: 110 (9.22%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       318  (26.7%)         287  (24.1%)     ActiveSupport::Inflector#apply_inflections
        67   (5.6%)          67   (5.6%)     (sweeping)
        67   (5.6%)          67   (5.6%)     ActiveModel::Serializer::LazyAssociation#reflection_options
        69   (5.8%)          53   (4.4%)     Hash#deep_dup
        43   (3.6%)          43   (3.6%)     (marking)
       651  (54.6%)          41   (3.4%)     ActiveModel::Serializer#associations
        39   (3.3%)          37   (3.1%)     Hash#except
       910  (76.3%)          29   (2.4%)     ActiveModel::Serializer#serializable_hash
        26   (2.2%)          26   (2.2%)     ActiveSupport::Inflector::Inflections::Uncountables#uncountable?
        35   (2.9%)          25   (2.1%)     ActiveSupport::Inflector#underscore
        23   (1.9%)          23   (1.9%)     Set#include?
[...]

PR #2449, opened 2023-08-12
Prevent creation of unnecessary fieldset(mirror and revised)

liijunwei:
#### Purpose

see: https://github.com/rails-api/active_model_serializers/pull/2370

#### Changes

1. `fieldset` is not listed in `lib/active_model_serializers/serializable_resource.rb@ADAPTER_OPTION_KEYS`, remove the `||=` protection (simplification)
2. do not create `ActiveModel::Serializer::Fieldset` instance when `fieldset` is empty (optimization)
3. check whether is `fieldset` falsy before using it

#### Caveats

NA


liijunwei:
wrong target branch

PR #2452, opened 2023-10-05
Update active_model_serializers.gemspec

kalashnikovisme:
#### Purpose
To use this gem with Rails 7.1.0 project.

#### Changes
Upgrade `rails_versions` in dependencies

All tests (around 500) are passed in our projects.

### DIFF
--- a/active_model_serializers.gemspec
+++ b/active_model_serializers.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.1'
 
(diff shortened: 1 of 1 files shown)


wasifhossain:
Thank you @kalashnikovisme for your PR! Also great to hear that your test suite is happy with the change.

Still if you don't mind, can we close this in favor of #2453 ?

wasifhossain:
you are right, because the other one was created after yours, but with the updated test matrix :sweat_smile: 

thank you!

PR #2458, opened 2024-01-05
Update 0.9-stable tests to support Ruby 3 and Rails 6

RyanNielson:
I was seeing some failures when running tests against Ruby 3. It seems like all the problems were in the test code and not the library code so I updated a few things to make the tests pass in newer Ruby and Rails versions as well. As far as I could tell there wasn't an obvious reason that `0-9-stable` wouldn't support Ruby 3. The tests are passing so I assume I'm correct in that assumption. The upgrade from `0-9-stable` to `0-10-stable` doesn't seem trivial so I was hoping to be able to stick with this branch while being able to upgrade to Ruby 3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,12 @@ jobs:
       matrix:
         include:
           # Recent Rubies and Rails
-          # - ruby-version: '3.1'
-          #   rails-version: '7.0'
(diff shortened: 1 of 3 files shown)


bf4:
Addressed, I believe by https://github.com/rails-api/active_model_serializers/pull/2461

Thanks for your contribution!

PR #2460, opened 2024-02-21
Don't force ActionController::TestCase to load

eugeneius:
Instead of referencing `ActionController::TestCase` directly during boot, we can use a lazy load hook to apply our mixins if it is loaded later.

Aside from the direct benefit of not evaluating a [rather large Ruby file](https://github.com/rails/rails/blob/v7.1.3/actionpack/lib/action_controller/test_case.rb) until necessary, this will avoid forcing other lazy load hooks to run prematurely (e.g. https://github.com/amatsuda/routes_lazy_routes/pull/14).

The `action_controller_test_case` load hook was added in Rails 5.1, and the `run_once` option was added in Rails 5.2. I've left the behaviour unchanged for older Rails versions where they're not available.

### DIFF
--- a/lib/active_model_serializers/railtie.rb
+++ b/lib/active_model_serializers/railtie.rb
@@ -44,9 +44,19 @@ module ActiveModelSerializers
     end
     # :nocov:
 
+    def extend_action_controller_test_case(&block)
+      if Rails::VERSION::MAJOR >= 6 || Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 2
+        ActiveSupport.on_load(:action_controller_test_case, run_once: true, &block)
+      else
+        ActionController::TestCase.instance_eval(&block)
+      end
+    end
+
     if Rails.env.test?
-      ActionController::TestCase.send(:include, ActiveModelSerializers::Test::Schema)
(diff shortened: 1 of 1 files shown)


bf4:
awesome, thanks

lib/active_model_serializers/railtie.rb
```diff
@@ -44,9 +44,19 @@ class Railtie < Rails::Railtie
     end
     # :nocov:
 
+    def extend_action_controller_test_case(&block)
+      if Rails::VERSION::MAJOR >= 6 || Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 2
+        ActiveSupport.on_load(:action_controller_test_case, run_once: true, &block)
```
bf4, line 49: nice `run_once: true` on load tag

lib/active_model_serializers/railtie.rb
```diff
@@ -44,9 +44,19 @@ class Railtie < Rails::Railtie
...
+      else
+        ActionController::TestCase.instance_eval(&block)
```
bf4, line 51: nice checking version support for the on load hook

PR #2461, opened 2024-04-09
Shopify 0-9-stable-ruby-3.2

bf4:
Propose updates from https://github.com/Shopify/active_model_serializers/commits/0-9-stable-ruby-3.2

PR #2464, opened 2024-04-09
Avoid loading ActionController::TestCase in production

casperisfine:
Ref: https://github.com/Shopify/active_model_serializers/pull/3

> By referencing that object, it brings in all of the patches defined
> in `ActionController::TestCase` to the development and production
> environment, including behaviours like disabling the threading
> implementation of `ActionController::Live`

Extracted from: https://github.com/rails-api/active_model_serializers/pull/2461

cc @bf4


lib/active_model_serializers.rb
```diff
@@ -12,7 +12,9 @@
   ActiveSupport.on_load(:action_controller) do
     if ::ActionController::Serialization.enabled
       ActionController::Base.send(:include, ::ActionController::Serialization)
-      ActionController::TestCase.send(:include, ::ActionController::SerializationAssertions)
+      ActiveSupport.on_load(:action_controller_test_case) do
```
bf4, line 15: wow.. I'm kind of surprised that never came up...

palexvs:
Just to save time if somebody experience the same issue - ActiveRecord::ConnectionTimeoutError error after upgrading `active_model_serializers`:
```
ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use
```

due to this fix enables threading for ActiveStorage::Proxy

PR #2467, opened 2024-04-10
[0.9 stable] Lazily compute possible serializer class names

casperisfine:
There is no point computing them all and then later look them up.

cc @bf4

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -68,14 +68,13 @@ end
             ArraySerializer
           end
         else
-          search_list = build_serializer_class_list(resource, options)
-          result = search_list.map do |klass_name|
-                     Serializer.serializers_cache.fetch_or_store(klass_name) do
-                       _const_get(klass_name)
-                     end
-                   end
-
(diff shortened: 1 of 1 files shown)


lib/active_model/serializer.rb
```diff
@@ -68,14 +68,13 @@ def serializer_for(resource, options = {})
...
-                       _const_get(klass_name)
-                     end
-                   end
-
-          result.find { |serializer| !serializer.nil? }
+          each_possible_serializer(resource, options) do |klass_name|
+            serializer = Serializer.serializers_cache.fetch_or_store(klass_name) do
+              _const_get(klass_name)
+            end
+            return serializer unless serializer.nil?
```
bf4, line 75: Hmm,  I try to avoid return in a block. Break would work,  too, no?

casperisfine: Break could work yes. I find return from a block quite idomatic, but feel free to change the style if you feel strongly about it.

The goal here was first and foremost to reduce the overhead.

PR #2469, opened 2024-04-10
[0.9 stable] Fix various warnings in the test suite

casperisfine:
cc @bf4


lib/active_model/serializer/association.rb
```diff
@@ -9,6 +9,8 @@ class Serializer
     class Association
       def initialize(name, options={})
         if options.has_key?(:include)
+          puts '-' * 40
+          puts caller
```
bf4, line 13: ?

byroot: 🤦 forgotten debug, sorry.

PR #2471, opened 2024-05-31
Improve attribute methods generation

casperisfine:
I'm currently working through Ruby warnings in our app, and one of the top source is from ActiveModel Serializer sub classes.

Usually they look like:

```ruby
class SomethingSerializer < AMS
  attributes :title, :body

  def title # method redefinition warning
    titleize
  end
end
```

One solution would be to call `attributes` at the end of the method, so that `attributes` skips defining these, but it's quite unatural.

Instead we can use the self alias trick to mark these generated methods as OK to redefine. An alternative would be to define them in an included module like Active Model does, but not sure if it's worth it.

While I was at it, I improved the way these methods are generated.

cc @bf4

### DIFF
--- a/lib/active_model/serializer.rb
+++ b/lib/active_model/serializer.rb
@@ -90,15 +90,27 @@ end
       end
 
       def attributes(*attrs)
+        # Use `class_eval` rather than `define_method` for faster access
+        # and avoid retaining objects in the closure.
+        # Batch all methods in a single `class_eval` for efficiceny,
+        # and define all methods on the same line so the eventual backtrace
+        # properly maps to the `attributes` call.
+        source = ["# frozen_string_literal: true\n"]
         attrs.each do |attr|
           striped_attr = strip_attribute attr
 
           @_attributes << striped_attr
 
-          define_method striped_attr do
-            object.read_attribute_for_serialization attr
-          end unless method_defined?(attr)
+          unless method_defined?(attr)
+            source <<
+            "def #{striped_attr}" <<
+              "object.read_attribute_for_serialization(#{attr.inspect})" <<
+            "end" <<
+            "alias_method :#{striped_attr}, :#{striped_attr}" # suppress method redefinition warning
+          end
         end
+        caller = caller_locations(1, 1).first
+        class_eval(source.join(";"), caller.path, caller.lineno - 1)
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_model/serializer.rb
```diff
@@ -90,15 +90,27 @@ def root_name
       end
 
       def attributes(*attrs)
+        # Use `class_eval` rather than `define_method` for faster access
+        # and avoid retaining objects in the closure.
+        # Batch all methods in a single `class_eval` for efficiceny,
+        # and define all methods on the same line so the eventual backtrace
+        # properly maps to the `attributes` call.
+        source = ["# frozen_string_literal: true\n"]
```
bf4, line 98: @casperisfine Thanks for this.  I've never seen a class_eval with the magic comment header. On the one hand I'm surprised the context isn't inherited by the eval, on the other hand, it makes a certain amount of sense thought evaluated input would have a new context.

lib/active_model/serializer.rb
```diff
@@ -90,15 +90,27 @@ def root_name
...
 
-          define_method striped_attr do
-            object.read_attribute_for_serialization attr
-          end unless method_defined?(attr)
+          unless method_defined?(attr)
+            source <<
+            "def #{striped_attr}" <<
+              "object.read_attribute_for_serialization(#{attr.inspect})" <<
+            "end" <<
+            "alias_method :#{striped_attr}, :#{striped_attr}" # supress method redefinition warning
```
bf4, line 109:
```suggestion
            "alias_method :#{striped_attr}, :#{striped_attr}" # suppress method redefinition warning
```

bf4:
@casperisfine @byroot looks fine to merge with the comment typo fixed.  Can you add a comment to the https://github.com/rails-api/active_model_serializers/blob/0-9-stable/CHANGELOG.md while there? I can release a new patch after merge if you like.

casperisfine:
@bf4 done!

casperisfine:
@bf4 I just realized now that I forgot to ask for a release :P 

I'm trying to grind down our warnings again, a release would be super appreciated.

bf4:
@casperisfine Done. Release 0.9.13

byroot:
🙇 ❤️

PR #2492, opened 2025-10-26
Remove use of ActiveSupport::Configurable

t27duck:
#### Purpose

Rails 8.1 deprecated `ActiveSupport::Configurable` which this gem currently uses.

```
DEPRECATION WARNING: ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2.

You can emulate the previous behavior with `class_attribute`.
```

#### Changes

This replaces the use of `ActiveSupport::Configurable` with an `ActiveSupport::OrderedOptions` class attribute.

#### Related GitHub issues

N/A


wasifhossain:
Thanks @t27duck for the PR!

Could you please rebase your branch on https://github.com/rails-api/active_model_serializers/pull/2493 and then we check again, if CI is happy?

t27duck:
@wasifhossain Ok, I'm pretty sure I did it right.

serggl:
@wasifhossain any change this will be published to rubygems soon?

wasifhossain:
@bf4 could you please help us on this?

I am not so sure if the changes are worthy enough to get a new version at this moment.

Until then, you may continue with this:
```
gem 'active_model_serializers', github: 'rails-api/active_model_serializers', branch: '0-10-stable'
```

bjorntrondsen:
Thanks for fixing this :pray: 

> I am not so sure if the changes are worthy enough to get a new version at this moment

Every Rails 8.1 app gets a deprecation warning, and for Rails 8.2 v0.10.15 probably wont work at all, so I think that warrants a new release :slightly_smiling_face: 

Ref:
_ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2._

janko:
There is a bug in the current implementation, where the configuration object is shared across serializer subclasses. This is inherent behavior of `class_attribute`, which does copy if you use direct assignment (`#config=` in this case), but changing hashes or arrays directly doesn't have effect.

In https://github.com/rails/rails/pull/53970 that replaced `ActiveSupport::Configurable` usage, there was also a config copy in the `inherited` hook using `ActiveSupport::InheritableOptions`. I think `active_model_serializers` should do the same.

In our app, we currently have this regression on `0-10-stable`:

```rb
ActiveModelSerializers.config.adapter = :json_api

  config.adapter = :attributes
end

ActiveModelSerializers.config.adapter #=> :attributes
```

ebaranowski:
I vote that this is a perfect reason for a release. It provides a fix to a warning that pops up on any rails command anyone does when using this gem. Pefect reason to have a release

